PR #54, opened 2024-08-29
logging instead of prints

### LINKED ISSUE #10: [FEATURE]: Replace prints with proper logging facility

- [ ] I have searched the existing issues

### Problem statement

_No response_

### Proposed Solution

```
from databricks.labs.blueprint.logger import install_logger

install_logger()

import logging
logging.root.setLevel("DEBUG") # use only for development or demo purposes

logger = logging.getLogger("name.of.your.module")
logger.debug("This is a debug message")
logger.info("This is an table message")
logger.warning("This is a warning message")
logger.error("This is an error message", exc_info=KeyError(123))
logger.critical("This is a critical message")
```

See https://github.com/databrickslabs/blueprint?tab=readme-ov-file#nicer-logging-formatter

### Additional Context

_No response_

pratikk-databricks:
## Changes
Added logging to the project. Logging is imported from the labs blueprint project as that is standard used across other labs project. 
Replaced the prints with logger.info and added logger.debug in one of the method for test

### Linked issues
Resolves #10 

### Tests

- [ ] manually tested
- [x] added unit tests
- [ ] added integration tests

### DIFF
--- a/src/databricks/labs/dqx/__init__.py
+++ b/src/databricks/labs/dqx/__init__.py
@@ -1,0 +1,3 @@
+from databricks.labs.blueprint.logger import install_logger
+
+install_logger()
--- a/tests/unit/test_build_rules.py
+++ b/tests/unit/test_build_rules.py
@@ -1,5 +1,5 @@
 import pprint
-
+import logging
 import pytest
 from pyspark.sql import SparkSession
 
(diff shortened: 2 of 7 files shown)


tests/unit/test_build_rules.py
```diff
@@ -180,3 +180,17 @@ def test_build_checks_by_metadata_when_function_does_not_exist(spark_session: Sp
 
     with pytest.raises(Exception):
         build_checks_by_metadata(checks)
+
+
+def test_build_checks_by_metadata_logging_debug_calls(caplog, spark_session: SparkSession):
```
mwojtyczka, line 185: we don't need to test logging messages. It makes refactorization hard.

mwojtyczka (approved):

src/databricks/labs/dqx/__init__.py
```diff
@@ -0,0 +1,3 @@
+from databricks.labs.blueprint.logger import install_logger
+
+install_logger()
```
sbachstein, line 3: @mwojtyczka This leads to a complete teardown and reconfiguration of the root logger. Is this something that we want to happen at import time of a module? This is very much hidden from the user of the library and I would argue that this should probably happen only if explicitely asked for.

mwojtyczka: i think we should scope the formatting to DQX's own logger tree

sbachstein: Could also be an option, albeit I don't know how much logging is happening inside DQX?

sbachstein: Should we track this in a GitHub issue?

mwojtyczka: yes, @sbachstein would you mind creating an issue for this?

PR #68, opened 2024-11-22
Make profiler and generator dependent on UC and introduce run configurations

mwojtyczka:
## Changes
* Made profiler and generator dependent on UC/Databricks
* Introduced RunConfig to be able to specify multiple configurations in the config to be able to run the project (e.g. profiler) for different use cases
* Refactored tests
* Improved Readme
* Added new examples to demos

### Tests
- [x] manually tested the whole project on Databricks including demos
- [x] added unit tests
- [x] added integration tests


src/databricks/labs/dqx/cli.py
```diff
@@ -45,7 +45,6 @@ def installations(w: WorkspaceClient):
             config = installation.load(WorkspaceConfig)
             all_users.append(
                 {
-                    "input": config.input_location,
```
gergo-databricks, line 48: Please confirm that this was not an accidental delete!

src/databricks/labs/dqx/config.py
```diff
@@ -12,12 +26,23 @@ class WorkspaceConfig:
...
+            raise ValueError("No run configurations available")
+
+        if not run_config_name:
+            return self.run_configs[0]
+
+        for run in self.run_configs:
+            if run.name == run_config_name:
+                return run
+
+        return self.run_configs[0]
```
gergo-databricks: If I define run_config_name and there is no config with that name, then I would assume an error, not the default config. Please add an exception for that case!

src/databricks/labs/dqx/engine.py
```diff
@@ -355,24 +344,49 @@ def load_checks_from_local_file(filename: str) -> list[dict]:
             msg = f"Checks file {filename} missing"
             raise FileNotFoundError(msg) from None
 
-    def load_checks_from_file(self, install_folder: str | None = None) -> list[dict]:
+    def load_checks_from_workspace_file(self, workspace_path: str) -> list[dict]:
```
gergo-databricks, line 347: Kudos for the function renaming!

src/databricks/labs/dqx/config.py
```diff
@@ -2,7 +2,21 @@
...
+    """Configuration class for the data quality checks"""
+
+    name: str = "default"  # name of the run configuration
+    input_locations: str | None = None  # input data path or a table
+    input_format: str | None = "delta"  # input data format
+    output_table: str | None = None  # output data table
+    quarantine_table: str | None = None  # quarantined data table
+    curated_table: str | None = None  # curated data table
+    checks_file: str | None = "checks.yml"  # name of the file containing quality rules / checks
+    profile_summary_stats_file: str | None = "checks.yml"  # name of the file containing profile summary statistics
```
gergo-databricks: This default for "profile_summary_stats_file" is different then what you have in the installation script as a default value. Would it be better to have the same default value everyhwere and probably to not use the same "checks.yml" value as a default that we also use in "checks_file"?

src/databricks/labs/dqx/profiler/dlt_generator.py
```diff
@@ -2,174 +2,175 @@
...
+
+        lang = language.lower()
+
+        if lang == "sql":
+            return self._generate_dlt_rules_sql(rules, action)
+
+        if lang == "python":
+            return self._generate_dlt_rules_python(rules, action)
+
+        raise ValueError(f"Unsupported language '{language}'")
```
gergo-databricks: Nit picking/optional: add the supported formats to the error message.

tests/unit/test_config.py
```diff
@@ -0,0 +1,39 @@
...
+def test_get_run_config_when_name_as_none():
+    assert CONFIG.get_run_config(None) == DEFAULT_RUN_CONFIG
+
+
+def test_get_run_config_when_name_not_provided():
+    assert CONFIG.get_run_config() == DEFAULT_RUN_CONFIG
+
+
+def test_get_run_config_when_name_not_found():
+    assert CONFIG.get_run_config("not_found") == DEFAULT_RUN_CONFIG
```
gergo-databricks: Should be an error.

README.md
```diff
@@ -121,9 +125,29 @@ You'll be prompted to select a [configuration profile](https://docs.databricks.c
...
+  profile_summary_stats_file: profile_summary_stats.yml
+  quarantine_location: main.dqx.quarantine
+- name: another_run_config
+  ...
+```
+
+To select a specific run config when executing the dqx labs cli commands use `--run-config` parameter. 
+When not provided the "default" run config is used.
 
 By default, DQX is installed in the user home directory (under '/Users/<user>/.dqx'). You can also install DQX globally
```
gergo-databricks: Not in your change, so optional to fix: here and 2 lines below there is an accidental HTML tag "user", hidden in the text when seen on GitHub.

README.md
```diff
@@ -203,18 +240,16 @@ You must install DQX in the workspace before (see [installation](#installation-i
...
 ```
 
+If run config is not provided, the "default" run config will be used. The run config is used to select specific run configuration from 'config.yml'.
+
 The following DQX configuration from 'config.yml' will be used by default:
 - 'input_location': input data as a path or a table.
-- 'checks_file': location of the generated quality rule candidates (default: `checks.yml`). Can be json or yaml file.
-- 'profile_summary_stats_file':  location of the summary statistics (default: `profile_summary.yml`). Can be json or yaml file.
-
-You can overwrite the default parameters as follows:
```
gergo-databricks, line 214: Can you still override them? This part may not be deleted then.

README.md
```diff
@@ -233,21 +268,17 @@ valid = validate_rules(checks)  # returns boolean
...
 ```commandline
-databricks labs dqx validate
+databricks labs dqx validate --run-config "default"
 ```
-The following DQX configuration from 'config.yml' will be used by default:
-- 'checks_file': location of the quality rule (default: `checks.yml`).
 
-You can overwrite the default parameters as follows:
-```commandline
-databricks labs dqx validate --checks-file "checks.yml"
```
gergo-databricks, line 243: Can you still override them? This part may not be deleted then.

README.md
```diff
@@ -442,20 +503,17 @@ def silver():
 Once the application is run, the user should analyse and investigate data quality problems, 
 for example by looking at a summary, quality dashboard, and/or quarantined dataset etc.).
 
-Once the root cause analysis is performed, user can curate the quarantined data, 
-and then re-ingest it to the output by running:
+Once the root cause analysis is performed, user can curate the quarantined data and store it in the curated table.
```
gergo-databricks: This functionality does not exist in the codebase. The CLI command is not there, and the engine knows nothing about it. Only the config exists.
Is this under implementation? Can we delete this part or put here a warning that this feature is not yet works?

README.md
```diff
@@ -19,6 +19,7 @@ Simplified Data Quality checking at Scale for PySpark Workloads on streaming and
     * [Upgrade DQX in the Databricks workspace](#upgrade-dqx-in-the-databricks-workspace)
```
gergo-databricks, line 19: Kudos for improving the docs!

README.md
```diff
@@ -191,10 +222,16 @@ The generated rules are input for the quality checking (see [Adding quality chec
...
+from databricks.labs.dqx.profiler.generator import DQGenerator
+from databricks.sdk import WorkspaceClient
 
+ws = WorkspaceClient()
+profiler = DQProfiler(ws)
 df = spark.read.table("catalog1.schema1.table1")
-summary_stats, checks = profile(df)
+summary_stats, profiles = profiler.profile(df)
+generator = DQGenerator(ws)
+checks = generator.generate_dq_rules(profiles)  # with default level "error"
```
gergo-databricks, line 239: Add here an example for the DLT rule generation and a note that those are DLT expectations and not DQX rules!

src/databricks/labs/dqx/profiler/profiler.py
```diff
@@ -20,448 +22,448 @@ class DQProfile:
...
-        if isinstance(f.dataType, T.StructType):
-            fields.extend(get_fields(f.name, f.dataType))
-        else:
-            fields.append(f)
+            return int(value)
+        if typ == T.DoubleType() or typ == T.FloatType():
+            return float(value)
 
-    return [T.StructField(f"{col_name}.{f.name}", f.dataType, f.nullable) for f in fields]
+        # TODO: handle other types
```
gergo-databricks: I think there is no need for other types as this is only used to process "summary" output (https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.summary.html). I would rather raise here an error about unsupported type, just to be sure.

src/databricks/labs/dqx/profiler/profiler.py
```diff
@@ -20,448 +22,448 @@ class DQProfile:
...
+            dst2 = dst.dropDuplicates()
+            cnt = dst2.count()
+            if 0 < cnt < total_count * opts["distinct_ratio"] and cnt < opts["max_in_count"]:
+                dq_rules.append(
+                    DQProfile(name="is_in", column=field_name, parameters={"in": [row[0] for row in dst2.collect()]})
+                )
+        if typ == T.StringType():
+            dst2 = dst.filter(F.col(field_name) == "")
+            cnt = dst2.count()
+            if cnt <= (metrics["count"] * opts.get("max_empty_ratio", 0)):
```
gergo-databricks, line 166: We should also check if "is_not_null" was not added already and only add either of those. Preferably check this first and add it to the rules and not add the "is_not_null" rule if this is already added.

src/databricks/labs/dqx/profiler/profiler.py
```diff
@@ -20,448 +22,448 @@ class DQProfile:
...
+        outlier_cols = opts.get("outlier_columns", [])
+        column = dst.columns[0]
+        if opts.get("remove_outliers", True) and (
+            len(outlier_cols) == 0 or col_name in outlier_cols
+        ):  # detect outliers
+            if typ == T.DateType():
+                dst = dst.select(F.col(column).cast("timestamp").cast("bigint").alias(column))
+            elif typ == T.TimestampType():
+                dst = dst.select(F.col(column).cast("bigint").alias(column))
+            # TODO: do summary instead? to get percentiles, etc.?
```
gergo-databricks, line 287: It does "summary" before and those values should be in the "metrics" variable. Why are we calculating them again? There are some exceptions when we need this: in case of struct the inner fields are not in the metrics variable. But in other cases we should not recalculate the same values twice.

src/databricks/labs/dqx/profiler/profiler.py
```diff
@@ -20,448 +22,448 @@ class DQProfile:
...
+        :param typ: The data type of the column.
+        :return: A tuple containing the description, maximum limit, and minimum limit.
+        """
         if mn_mx and len(mn_mx) > 0:
             metrics["min"] = mn_mx[0][0]
             metrics["max"] = mn_mx[0][1]
-            min_limit = round_value(metrics.get("min"), "down", opts)
-            max_limit = round_value(metrics.get("max"), "up", opts)
-            descr = "Real min/max values were used"
+            sigmas = opts.get("sigmas", 3)
```
gergo-databricks, line 336: Kudos for adding sigma!

src/databricks/labs/dqx/cli.py
mwojtyczka: this is just for printing in the console. I removed it because, there could be multiple config runs now with multiple locations.

src/databricks/labs/dqx/config.py
mwojtyczka: corrected

src/databricks/labs/dqx/config.py
mwojtyczka: good catch, corrected

src/databricks/labs/dqx/profiler/dlt_generator.py
mwojtyczka: added

README.md
mwojtyczka: added

src/databricks/labs/dqx/profiler/profiler.py
mwojtyczka: added raising exception if unsupported type is encountered

src/databricks/labs/dqx/profiler/profiler.py
mwojtyczka: added

src/databricks/labs/dqx/profiler/profiler.py
mwojtyczka: not sure about that one. Will leave it for now. @gergo-databricks would be great if you can look at it and submit PR if needed

README.md
mwojtyczka: fixed

README.md
mwojtyczka: i think we want to generally reduce the amount of configuration. We had the same in ucx and with time we ended up with lot's of configuration options which are hard to manage. If the user want to use cli, then the tool must be installed in the workspace in which case, a config is created with these. There is also a run config now so it's very flexible as you can create multiple configurations. We can add the overwrite later.

README.md
mwojtyczka: same as above. To be added later. For now user should specify run config instead.

README.md
mwojtyczka: True. I removed this for now, also from the config. I think we agreed that automatic curation is not really feasible. And having a job to just copy data from one table to another is not very useful.

PR #95, opened 2025-01-16
Fix for image links in README on PyPi

alexott:
## Changes

Use of `fancy-pypi-readme` plugin helped with relative links, but links to images weren't working correctly because they led to GitHub page instead of actual link. This PR adds a separate pattern to handle image links (we need to use ``  for images.

### Linked issues

Resolves #..

### Tests

- [x] manually tested
- [ ] added unit tests
- [ ] added integration tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,7 @@ Databricks Labs DQX
 
 <img src="docs/logo.png" alt="DQX" width="50%">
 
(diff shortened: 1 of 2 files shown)


mwojtyczka (approved):
LGTM

Issue #55, opened 2024-09-17
[FEATURE]: Add context info to the output


mwojtyczka:
### Is there an existing issue for this?

- [ ] I have searched the existing issues

### Problem statement

Currently we only report warnings and errors. It is unknown when the check was performed and in which pipeline. 

### Proposed Solution

Add _context column to the output containing: timestamp, workflow it was run on.

pierre-monnet:
@mwojtyczka 
What do you think about replace the current structure of `_warning` and `_error`
FROM
```
{
  col_col1_is_null: "Column col1 is null",
  col_col2_is_null: "Column col2 is null"
}
```
TO
```
{
  "col_col1_is_null": {
    "rule": "is_not_null",
    "col": "col1",
    "filter": "col2<3",
    "message": "col1 is not null",
  },
  "col_col2_is_null": {
    "rule": "is_not_null",
    "col": "col2",
    "message": "col2 is not null",
  }
}
```

mwojtyczka:
thanks for the proposal. Reusing existing reporting columns is more elegant. The only disadvantage is that the execution time will be the same so a lot of repeated info. 
I would drop the workflow name or run_id as this info will be available in the Unity Catalog lineage when the dataframe is saved, along with info about the input datasets so we can skip. For the date ISO format would be better. 

```
{
  "col_col1_is_null": {
    "rule": "is_not_null",
    "col": "col1",
    "filter": "col2<3",
    "message": "col1 is not null",
    "execution_time": "2025-01-29T13:37:00+0000",
  },
  "col_col2_is_null": {
    "rule": "is_not_null",
    "col": "col2",
    "message": "col2 is not null",
    "execution_time": "2025-01-29T13:37:00+0000",
  }
}
```

gergo-databricks:
Question: How would we create this "filter" string? str(DQRule.check)?

I recommend using a timestamp for the time value. Most users of this library will store this data in a table either in this structure or even flatten it, so it's better to store it properly. This would mean converting the current Map to a Struct type.

I propose also giving some flexibility here and creating an option for users to add their metadata.

PR #141, opened 2025-01-28
feat: add filter to rule

### LINKED ISSUE #140: [FEATURE]: Add filter to rule

#### Problem
I need to apply rules to a part of a DataFrame.

#### Context
We have a table of tracking events, and depending on the type of event, we want to apply a rule or not.

#### Example
```
+---+------------+-----------+---------+--------+
| id|  event_type|button_name|page_from|page_url|
+---+------------+-----------+---------+--------+
|  1|   page_view|       NULL|     NULL|   /home|
|  2|button_click|      login|    /home|    NULL|
|  3|   page_view|       NULL|     NULL|  /login|
|  4|button_click|    sign_in|   /login|    NULL|
+---+------------+-----------+---------+--------+
```
##### In this case
- if the `event_type` is "page_view", then `page_url` must be non-null
- if the `event_type` is "button_click", then `button_name` and `page_from` must be non-null.

#### Purpose
For this simple example, a `sql_expression` with both conditions could solve the problem, but for readability, I find it more elegant to keep native rules (here `is_not_null`) to which we apply a filter. This is even more the case when we want to apply a rule to a list of columns, it's much more readable to do that.

### Proposed Solution

```
- criticality: "error"
  filter: "event_type='page_view'"
  check:
    function: "is_not_null"
    arguments:
      col_name: "page_url"
- criticality: "error"
  filter: "event_type='button_click'"
  check:
    function: "is_not_null"
    arguments:
      col_names: 
        - "button_name"
        - "page_from"
```

### Additional Context

I started a draft PR (#141) with a simple working purpose but there are still a lot of things to validate, for example, how to display filter in `_error` or `_warning` result.

pierre-monnet:
## Changes
Add filter to Rule

### Linked issues

Resolves #140

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests


mwojtyczka:
Thank you for your PR. It looks good. We currently have issue in our CI that they don't run on forks. Once this is solved, we should be able to merge.

tests/integration/test_apply_checks.py
```diff
@@ -381,6 +381,42 @@ def test_apply_checks_by_metadata(ws, spark):
     assert_df_equality(checked, expected, ignore_nullable=True)
 
 
+def test_apply_checks_with_filter(ws, spark):
+    dq_engine = DQEngine(ws)
+    test_df = spark.createDataFrame(
+        [[1, 3, 3], [2, None, 4], [3, 4, None], [4, None, None], [None, None, None]], SCHEMA
+    )
+
+    checks = [
```
mwojtyczka: please test DQRuleColSet as well

tests/integration/test_apply_checks.py
```diff
@@ -381,6 +381,42 @@ def test_apply_checks_by_metadata(ws, spark):
     assert_df_equality(checked, expected, ignore_nullable=True)
 
 
+def test_apply_checks_with_filter(ws, spark):
```
mwojtyczka, line 382: please add another test that uses apply_checks_by_metadata

src/databricks/labs/dqx/engine.py
```diff
@@ -87,6 +87,7 @@ class DQRule:
     check: Column
     name: str = ""
     criticality: str = Criticality.ERROR.value
+    filter: str | None = None
```
mwojtyczka: It would be useful if the filter info is also added to the reporting columns.

pierre-monnet:
I rebased from main, because there was lot of change on DQEngine

mwojtyczka (changes requested):

tests/integration/test_apply_checks.py
pierre-monnet: done

tests/integration/test_apply_checks.py
pierre-monnet: done -> `test_apply_checks_by_metadata_with_filter`

src/databricks/labs/dqx/engine.py
pierre-monnet: Do you mean in `_errors` and `_warnings` cols ? For new they are just dict[str, str] so I have no idea to add the information in them.
I can work on #55 to resolve this problem.

mwojtyczka: sounds good, we can keep them separate. Corret, it requires structure change.

mwojtyczka:
@pierre-monnet it looks good now. We still have to update the documentation.

Can you please extend the existing guide by adding a filter to one of the checks in the following places:
https://github.com/databrickslabs/dqx/blob/main/docs/dqx/docs/guide.mdx#quality-rules-defined-as-code
https://github.com/databrickslabs/dqx/blob/main/docs/dqx/docs/guide.mdx#method-2-using-yaml-config

Please also extend the demo in similar way:
https://github.com/databrickslabs/dqx/blob/main/demos/dqx_demo_library.py#L130
https://github.com/databrickslabs/dqx/blob/main/demos/dqx_demo_library.py#L184

There should also be a note about the filter functionality here at the bottom: https://github.com/databrickslabs/dqx/blob/main/docs/dqx/docs/reference.mdx#quality-rule-functions-checks

mwojtyczka (changes requested):
@pierre-monnet it looks good now. Thank you! The only missing part is the documentation. 

Can you please extend the existing guide by adding a filter to one of the checks in the following places:
https://github.com/databrickslabs/dqx/blob/main/docs/dqx/docs/guide.mdx#quality-rules-defined-as-code
https://github.com/databrickslabs/dqx/blob/main/docs/dqx/docs/guide.mdx#method-2-using-yaml-config

Please also extend the demo in similar way:
https://github.com/databrickslabs/dqx/blob/main/demos/dqx_demo_library.py#L130
https://github.com/databrickslabs/dqx/blob/main/demos/dqx_demo_library.py#L184

There should also be a note about the filter functionality here at the bottom: https://github.com/databrickslabs/dqx/blob/main/docs/dqx/docs/reference.mdx#quality-rule-functions-checks

pierre-monnet:
@mwojtyczka I just added missing documentation, don't hesitate to give me your feedback

mwojtyczka (approved):
LGTM

PR #142, opened 2025-01-28
Support datetime arguments for column range functions

### LINKED ISSUE #136: [FEATURE]: Add support for date and timestamp to range type of checks
### Problem statement

The current checks like not_less_than, not_greater_than, is_in_range work for numeric limits only. It's a common use case to check for date and datetime as well.

It's also common to have issue with date and time column if they are entered as strings. 

### Proposed Solution

Add equivalent check functions for working with date and datetime.

Add separate check to validate the dates and timestamps have valid format.

### Additional Context

_No response_

ghanse:
## Changes
Added support for both `datetime.date` and `datetime.datetime` arguments to the following column functions:
- `not_less_than`
- `not_greater_than`
- `is_in_range`
- `is_not_in_range`

I used type parameter lists for the `limit` parameter argument of each function. I created the limit as a `pyspark.sql.Column` using `lit(...)` which handles both `datetime.date` and `datetime.datetime` arguments.

### Linked issues
Resolves #136 

### Tests
Integration tests for each column function have been updated to check the condition for `int`, `datetime.date`, and `datetime.datetime` arguments against `IntegerType`, `DateType`, and `TimestampType` columns, respectively.

- [x] manually tested
- [ ] added unit tests
- [x] added integration tests

### DIFF
--- a/src/databricks/labs/dqx/col_functions.py
+++ b/src/databricks/labs/dqx/col_functions.py
@@ -1,3 +1,4 @@
+import datetime
 import re
 
 import pyspark.sql.functions as F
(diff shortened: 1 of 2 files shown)


mwojtyczka (approved):
LGTM

PR #151, opened 2025-02-05
Introducing to Spark Remote for Unit Tests

sundarshankar89:
## Changes

Introduced Managed Spark Remote to run tests effectively. In future, we can also introduce uc dependency to resolve other UC-related integrations if and when required.

Note: This makes the unit test more of an integration test and gets us away from mock sparksession. 

### Linked issues

Resolves #..

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests

### DIFF
--- a/.github/scripts/setup_spark_remote.sh
+++ b/.github/scripts/setup_spark_remote.sh
@@ -1,0 +1,60 @@
+#!/usr/bin/env bash
+
+set -xve
+
+mkdir -p "$HOME"/spark
+cd "$HOME"/spark || exit 1
+
+version=$(wget -O - https://dlcdn.apache.org/spark/ | grep 'href="spark' | grep -v 'preview' | sed 's:</a>:\n:g' | sed -n 's/.*>//p' | tr -d spark- | tr -d / | sort -r --version-sort | head -1)
+if [ -z "$version" ]; then
+  echo "Failed to extract Spark version"
+   exit 1
+fi
+
+spark=spark-${version}-bin-hadoop3
+spark_connect="spark-connect_2.12"
+
+mkdir -p "${spark}"
+
+
+SERVER_SCRIPT=$HOME/spark/${spark}/sbin/start-connect-server.sh
+
+## check the spark version already exist ,if not download the respective version
+if [ -f "${SERVER_SCRIPT}" ];then
+  echo "Spark Version already exists"
+else
+  if [ -f "${spark}.tgz" ];then
+    echo "${spark}.tgz already exists"
+  else
+    wget "https://dlcdn.apache.org/spark/spark-${version}/${spark}.tgz"
+  fi
+  tar -xvf "${spark}.tgz"
+fi
+
+cd "${spark}" || exit 1
+## check spark remote is running,if not start the spark remote
+result=$(${SERVER_SCRIPT} --packages org.apache.spark:${spark_connect}:"${version}" > "$HOME"/spark/log.out; echo $?)
+
+if [ "$result" -ne 0 ]; then
+    count=$(tail "${HOME}"/spark/log.out | grep -c "SparkConnectServer running as process")
+    if [ "${count}" == "0" ]; then
+            echo "Failed to start the server"
+        exit 1
+    fi
+    # Wait for the server to start by pinging localhost:4040
+    echo "Waiting for the server to start..."
+    for i in {1..30}; do
+        if nc -z localhost 4040; then
+            echo "Server is up and running"
+            break
+        fi
+        echo "Server not yet available, retrying in 5 seconds..."
+        sleep 5
+    done
+
+    if ! nc -z localhost 4040; then
+        echo "Failed to start the server within the expected time"
+        exit 1
+    fi
+fi
+echo "Started the Server"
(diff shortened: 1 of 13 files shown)


tests/unit/test_utils.py
```diff
@@ -27,6 +27,7 @@ def test_get_col_name_longer():
     assert actual == "local"
 
 
+@pytest.mark.skip(reason="Ignore")
```
sundarshankar89: We can move away from mock to actual results here since the initial work was to introduce the concept, staying away from introducing changes at this point. Once the approach is approved we can make changes to the test to reflect the expected behaviour.

tests/unit/conftest.py
```diff
@@ -1,13 +1,12 @@
 import os
 from pathlib import Path
-from unittest.mock import Mock
 from pyspark.sql import SparkSession
 import pytest
 
 
 @pytest.fixture
 def spark_session_mock():
```
mwojtyczka:
```suggestion
def spark_session():
```

.github/workflows/push.yml
```diff
@@ -35,6 +35,11 @@ jobs:
           cache-dependency-path: '**/pyproject.toml'
           python-version: ${{ matrix.pyVersion }}
 
+      - name: Setup Spark Remote
```
mwojtyczka, line 38: We can update the `test` command in the Makefile so this won't be needed here.

.github/scripts/setup_spark_remote.sh
```diff
@@ -0,0 +1,60 @@
...
+
+spark=spark-${version}-bin-hadoop3
+spark_connect="spark-connect_2.12"
+
+mkdir -p "${spark}"
+
+
+SERVER_SCRIPT=$HOME/spark/${spark}/sbin/start-connect-server.sh
+
+## check the spark version already exist ,if not download the respective version
```
mwojtyczka:
```suggestion
## check the spark version already exist, if not download the respective version
```

.github/scripts/setup_spark_remote.sh
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env bash
+
+set -xve
+
```
mwojtyczka, line 5:
```suggestion
echo "Setting up spark-connect"

```

tests/unit/test_utils.py
```diff
@@ -38,6 +39,7 @@ def test_read_input_data_unity_catalog_table(spark_session_mock):
     assert result == "dataframe"
 
 
+@pytest.mark.skip(reason="Ignore")
```
mwojtyczka: You can use a local file for this test.

Makefile
```diff
@@ -23,6 +23,9 @@ test:
 integration:
 	hatch run integration
 
+setup_spark_remote:
```
mwojtyczka, line 26: Add the command under `test` so that it is always run automatically when we run unit tests

```
test: setup_spark_remote
	hatch run test
```

tests/unit/test_utils.py
```diff
@@ -27,6 +27,7 @@ def test_get_col_name_longer():
     assert actual == "local"
 
 
+@pytest.mark.skip(reason="Ignore")
 def test_read_input_data_unity_catalog_table(spark_session_mock):
     input_location = "catalog.schema.table"
```
mwojtyczka: We can move this tests to under integration tests.

tests/unit/test_utils.py
```diff
@@ -50,6 +52,7 @@ def test_read_input_data_storage_path(spark_session_mock):
     assert result == "dataframe"
 
 
+@pytest.mark.skip(reason="Ignore")
```
mwojtyczka: We can move this tests to under integration tests.

mwojtyczka (changes requested):

tests/unit/test_utils.py
mwojtyczka: i like the idea

.github/workflows/push.yml
sundarshankar89: I would keep it separately for easier debugging

Makefile
sundarshankar89: I like to keep it separate and run `make setup_spark_remote test` or I will introduce `make local_test`

.github/workflows/push.yml
mwojtyczka: agree but we should also run it as part of the `test` command

sundarshankar89: I made some changes to Makefile I can do the other way name it as local test and keep the ci-test as test

Makefile
mwojtyczka: how will this work in the ci? won't we need to setup spark remote as well?

mwojtyczka (changes requested):

.github/workflows/push.yml
sundarshankar89: @mwojtyczka I still have 2 steps.

mwojtyczka: i see, sounds good

Makefile
mwojtyczka: sounds good

mwojtyczka (approved):
LGTM

PR #153, opened 2025-02-05
[Feature] Update is_(not)_in_range (#87) to support max/min limits from col

### LINKED ISSUE #87: [FEATURE]: is_(not)_in_range supports min/max_limit from the dataframe

- [X] I have searched the existing issues

### Problem statement

Customer GWSC wants to groupBy the entities in the data and check if a column’s values are within, say, 3 standard deviations of the mean for that entity group.

### Proposed Solution

Modify is_in_range() to accept dataframe columns for the min/max_limit. Then I can join a dataframe of pre-computed group-specific ranges and provide the column names to is_in_range(). Example ranges to support include multiples of standard deviation, interquartile range, or mean absolute deviation.

### Additional Context

This feature may be included in [FEATURE]: Data set level rules #43, but I want to make sure my customer's need is addressed.

karthik-ballullaya-db:
## Changes

Changes in the is_in_range and is_not_in_range column function to handle a column as the min/max limit along with the literal value.

### Linked issues

Resolves #87 

### Tests

- [X] manually tested
- [x] added unit tests
- [X] added integration tests


mwojtyczka (approved):
LGTM

src/databricks/labs/dqx/col_functions.py
```diff
@@ -315,18 +315,18 @@ def is_in_range(
 
 def is_not_in_range(
     col_name: str,
-    min_limit: int | datetime.date | datetime.datetime,
-    max_limit: int | datetime.date | datetime.datetime,
+    min_limit: int | datetime.date | datetime.datetime | str,
+    max_limit: int | datetime.date | datetime.datetime | str,
```
alexott: I would make it a bit different:

1. add support for a string as a valid range
2. add support for a column object as min/max arguments, because people may use not simple column names but more complex expressions (via `F.expr`).

alexott (changes requested):

src/databricks/labs/dqx/col_functions.py
karthik-ballullaya-db: agree, updated code to support string ranges and column expression

src/databricks/labs/dqx/col_functions.py
```diff
@@ -281,20 +281,50 @@ def not_greater_than(col_name: str, limit: int | datetime.date | datetime.dateti
     )
 
 
+def _get_min_max_column_expr(
+    min_limit: int | datetime.date | datetime.datetime | str = '',
+    max_limit: int | datetime.date | datetime.datetime | str = '',
+    min_limit_col_expr: str | Column = '',
+    max_limit_col_expr: str | Column = '',
```
alexott: I would use Optional type or `None` instead of default of empty string

src/databricks/labs/dqx/col_functions.py
```diff
@@ -281,20 +281,50 @@ def not_greater_than(col_name: str, limit: int | datetime.date | datetime.dateti
...
+
+
 def is_in_range(
     col_name: str,
-    min_limit: int | datetime.date | datetime.datetime,
-    max_limit: int | datetime.date | datetime.datetime,
+    min_limit: int | datetime.date | datetime.datetime | str = '',
+    max_limit: int | datetime.date | datetime.datetime | str = '',
+    min_limit_col_expr: str | Column = '',
+    max_limit_col_expr: str | Column = '',
```
alexott: Here as well

src/databricks/labs/dqx/col_functions.py
```diff
@@ -315,18 +345,23 @@ def is_in_range(
 
 def is_not_in_range(
     col_name: str,
-    min_limit: int | datetime.date | datetime.datetime,
-    max_limit: int | datetime.date | datetime.datetime,
+    min_limit: int | datetime.date | datetime.datetime | str = '',
+    max_limit: int | datetime.date | datetime.datetime | str = '',
+    min_limit_col_expr: str | Column = '',
+    max_limit_col_expr: str | Column = '',
```
alexott: and here

src/databricks/labs/dqx/col_functions.py
```diff
@@ -281,20 +281,50 @@ def not_greater_than(col_name: str, limit: int | datetime.date | datetime.dateti
...
+    :return: tuple containing min_limit_expr and max_limit_expr
+    """
+    if min_limit_col_expr is None:
+        min_limit_expr = F.lit(min_limit)
+    else:
+        min_limit_expr = F.col(min_limit_col_expr) if isinstance(min_limit_col_expr, str) else min_limit_col_expr
+    if max_limit_col_expr is None:
+        max_limit_expr = F.lit(max_limit)
+    else:
+        max_limit_expr = F.col(max_limit_col_expr) if isinstance(max_limit_col_expr, str) else max_limit_col_expr
```
alexott, line 308: we need to throw an error if both `min_limit` and `min_limit_col_expr` are `None` (same for `max_limit` and `max_limit_col_expr`).

Otherwise we'll get conditions like `(F.col(col_name) < F.lit(None)) | (F.col(col_name) > F.lit(None))`

alexott (changes requested):

PR #206, opened 2025-03-08
feat: add context info to output

### LINKED ISSUE #55: [FEATURE]: Add context info to the output

pierre-monnet:
## Changes

* Add context info to the output: name, message, col_name, filter, function, run_time, user_metadata.
* It enables the user to provide custom metadata that is stored in the reporting columns for every failed check.
* Updated docs and demos.

This PR introduces a breaking change for checkes defined using classes!!!

### Linked issues

Resolves #55 

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests


src/databricks/labs/dqx/rule.py
```diff
@@ -91,6 +106,7 @@ class DQRuleColSet:
 
     columns: list[str]
     check_func: Callable
+    name: str = ""
```
mwojtyczka, line 118: I agree having a name does make sense now since you can identify the actual test by other parameters in the output. 
Please add a test for this. All existing tests skip the name.

src/databricks/labs/dqx/schema/validation_result_schema.py
```diff
@@ -0,0 +1,14 @@
+from pyspark.sql.types import StructType, StructField, ArrayType, StringType, TimestampType
+
+validation_result_schema = ArrayType(
```
mwojtyczka: There should be a way to provide a custom key-value metadata by the users (as dict). 
I would propose an additional field, e.g. StructField("user_metadata", MapType(StringType(), StringType()), nullable=True) supplied using `extra_params`

tests/conftest.py
```diff
@@ -271,3 +274,11 @@ def delete(workspace_file_path: str) -> None:
         ws.workspace.delete(workspace_file_path)
 
     yield from factory("file", create, delete)
+
+
+@pytest.fixture
+def run_time_date():
+    run_datetime = dt.datetime(2025, 1, 1, 0, 0, 0, 0)
+    with patch("databricks.labs.dqx.engine.datetime") as mock_date:  # pylint: disable=explicit-dependency-required
```
mwojtyczka: Please avoid using a patch. It's a quick method to get things done but it's nightmare to maintain in the long term. There should be a way to provide this and other custom metadata by the user. You can use the `extra_params` which is part of the DQEngine constructor.
With that you wouldn't also have to disable the pylint error.

src/databricks/labs/dqx/engine.py
```diff
@@ -253,18 +263,24 @@ def _create_results_map(df: DataFrame, checks: list[DQRule], dest_col: str) -> D
...
         :param dest_col: name of the map column
         """
-        empty_type = F.lit(None).cast("map<string, string>").alias(dest_col)
+        empty_type = F.lit(None).cast(validation_result_schema).alias(dest_col)
+
         if len(checks) == 0:
             return df.select("*", empty_type)
-
-        name_cols = []
+        current_date = datetime.now()
```
mwojtyczka: This makes the code not testable. There should be a way to inject the time (use the provided time or datatime.now() if not given). You can use the `extra_params` which is already part of the DQEngine constructor.

src/databricks/labs/dqx/schema/validation_result_schema.py
```diff
@@ -0,0 +1,14 @@
...
+    StructType(
+        [
+            StructField("name", StringType(), nullable=True),
+            StructField("rule", StringType(), nullable=True),
```
mwojtyczka: I think `message` would be better than a `rule` here. This is really the message from the checks. Rule could mean a check as well which is ambiguous.

mwojtyczka (changes requested):

demos/dqx_demo_library.py
mwojtyczka, line 1: We should also add to the user guide, an example output of the check.

mwojtyczka (changes requested):

docs/dqx/docs/guide.mdx
mwojtyczka, line 1: Please also add to the user guide, an example output (how the output looks like in the error/warning column when e.g. 2 checks fail for a row) and explain this is a an array of struct that can be loaded in spark. There should also be an example on how to access individual fields in the output.

The motivation section has 2 images as well with the output - I can update that part.

src/databricks/labs/dqx/rule.py
```diff
@@ -39,20 +39,26 @@ class ExtraParams:
...
+    check_func: Callable
+    col_name: str = ""
     name: str = ""
     criticality: str = Criticality.ERROR.value
     filter: str | None = None
+    check_func_args: list[Any] = field(default_factory=list)
+    check_func_kwargs: dict[str, Any] = field(default_factory=dict)
 
     def __post_init__(self):
         # take the name from the alias of the column expression if not provided
```
mwojtyczka, line 66: We should also verify that the provided args and kwargs are actually correct. It should only be possible to create `DQRule` if correct correct args/kwargs are provided for a given `check_func`. Otherwise we risk that user code will break when applying checks. The validation should be done before checks are applied.

mwojtyczka (changes requested):

src/databricks/labs/dqx/engine.py
pierre-monnet: fixed in 
@pierre-monnet
[fix: add run_time to DQEngine extra_params](https://github.com/databrickslabs/dqx/pull/206/commits/fd7b1dceae5cb96eca78de0b49a28386bfcfa508)

tests/conftest.py
pierre-monnet: fixed in 
@pierre-monnet
[fix: add run_time to DQEngine extra_params](https://github.com/databrickslabs/dqx/pull/206/commits/fd7b1dceae5cb96eca78de0b49a28386bfcfa508)

mwojtyczka: Looks good, thank you!

src/databricks/labs/dqx/rule.py
```diff
@@ -34,25 +35,32 @@ class ExtraParams:
...
 
 
 @dataclass(frozen=True)
 class DQRule:
     """Class to represent a data quality rule consisting of following fields:
-    * `check` - Column expression to evaluate. This expression should return string value if it's evaluated to true -
-    it will be used as an error/warning message, or `null` if it's evaluated to `false`
+    * `check_func` - check function to be applied
+    * `col_name` - column name to which the given check function should be applied
     * `name` - optional name that will be given to a resulting column. Autogenerated if not provided
```
mwojtyczka: Docstring needs to be updated. Sth like "optional name that will be given to the resulting check name. "

mwojtyczka (changes requested):

tests/integration/test_apply_checks.py
```diff
@@ -826,266 +1633,8 @@ def test_apply_checks_with_is_unique(ws, spark, set_utc_timezone):
...
-      check:
-        function: sql_expression
-        arguments:
-          expression: col3 >= col2 and col3 <= 10
-          msg: ol3 is less than col2 and col3 is greater than 10
-          name: custom_output_name
-          negate: false
-    """
-    )
+    with open("tests/resources/all_checks.yaml", "r", encoding="utf-8") as f:
```
mwojtyczka: The path won't be resolved correctly. Need sth like this: `file_path = Path(__file__).parent.parent / "resources" / "all_checks.yaml"`

tests/integration/test_apply_checks.py
```diff
@@ -804,19 +1583,47 @@ def test_apply_checks_with_is_unique(ws, spark, set_utc_timezone):
             [
```
mwojtyczka, line 1705: The `test_apply_checks_with_is_unique` is missing EXTRA_PARAMS when creating the `DQEngine`. Therefore, the expected times won't match.

tests/integration/test_apply_checks.py
```diff
@@ -257,28 +470,83 @@ def test_apply_checks_and_split_by_metadata(ws, spark):
...
+                        "run_time": RUN_TIME,
+                    }
+                ],
+                [
+                    {
+                        "name": "col_a_is_not_in_the_list",
+                        "rule": "Value 2 is not in the allowed list: [1, 3, 4]",
+                        "col_name": "a",
+                        "filter": None,
+                        "function": "is_in_the_list",
```
mwojtyczka: needs to be `is_in_list`

tests/integration/test_apply_checks.py
```diff
@@ -317,28 +585,83 @@ def test_apply_checks_and_split_by_metadata_with_autogenerated_col_names(ws, spa
...
```
mwojtyczka: needs to be `is_in_list`

tests/integration/test_apply_checks.py
```diff
@@ -375,28 +698,83 @@ def test_apply_checks_by_metadata(ws, spark):
...
```
mwojtyczka: needs to be `is_in_list`

mwojtyczka (changes requested):

docs/dqx/docs/guide.mdx
mwojtyczka: Fixed

mwojtyczka:
@pierre-monnet I continue improving the PR. I hope that's fine for you. We have other PRs and would like to get this done as soon as possible to avoid conflicts.

src/databricks/labs/dqx/engine.py
```diff
@@ -240,31 +251,35 @@ def _append_empty_checks(self, df: DataFrame) -> DataFrame:
...
             return df.select("*", empty_type)
-
-        name_cols = []
         check_cols = []
         for check in checks:
-            check_cols.append(check.check_column())
-            name_cols.append(F.lit(check.name))
+            result = F.struct(
+                F.lit(check.name).alias("name"),
+                F.lit(check.check_column()).alias("rule"),
```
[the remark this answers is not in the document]
mwojtyczka: resolved

src/databricks/labs/dqx/rule.py
mwojtyczka: added tests

src/databricks/labs/dqx/schema/validation_result_schema.py
mwojtyczka: updated

src/databricks/labs/dqx/schema/validation_result_schema.py
mwojtyczka: added

src/databricks/labs/dqx/rule.py
mwojtyczka: Added

mwojtyczka (approved):
LGTM

pierre-monnet (commented):
LGTM

PR #212, opened 2025-03-11
Updated sql expression logic

### LINKED ISSUE #209: [BUG]: SQL Expression quality check quarantines rows that meet the passed SQL condition

### Current Behavior

The docstring for the check function `sql_expression()` says "Checks whether the condition provided as an SQL expression is met."

Assuming that quality checks are meant to identify valid rows as those that meet a condition, and invalid rows as those that do not, SQL expression checks do the opposite by default.

For example, when using the function `dq_engine.apply_checks_by_metadata_and_split(input_df, checks)`, when an expression such as `a_column > 10` is used in a check, rows where `a_column` values are greater than 10 are output in the quarantine DataFrame instead of the valid DataFrame.

https://github.com/databrickslabs/dqx/blob/main/src/databricks/labs/dqx/col_functions.py

### Expected Behavior

The expected behavior is that a check `a_column > 10` would output rows where `a_column` is greater than 10 in `valid_df`.

### Steps To Reproduce

Create a SQL expression-based check and pass a DataFrame where at least some rows satisfy the condition in the expression. Assuming there are no other checks, the row count of the rows that satisfy the condition in the SQL expression in `input_df` and the row count of rows in `quarantine_df` will be equal.

### Cloud

Azure

### Operating System

Ubuntu

### Relevant log output

```shell

```

mwojtyczka:
extended library demo

## Changes
* Update logic of the sql expression to follow convention of other checks, i.e. fail the check if condition is not met. This introduces a breaking change for the `sql_expression` check.
* Extended demo to include custom check using DQX classes
* Minor docs update

### Linked issues

Resolves #209 

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests


alexott (approved):
codewise is good, but we'll need to highlight this as a breaking change, maybe even bump a version to 0.3.0, or something like that

PR #301, opened 2025-03-17
Trim autogenerated check name to 255 chars

### LINKED ISSUE #216: [FEATURE]: Trim autogenerated check name to 255 characters

### Problem statement

DQX autogenerates name for checks if the name is not explicitly provided by the user. However, if long sql expressions are used the resulting check name can be very long and not readable. Also if a user wants to explode and pivot the results so that the a check name becomes a column in a table, this would be problematic as Unity Catalog and Hive metastore impose limit of 255 characters for column names.

### Proposed Solution

Trim the length of check name to 255 chars to avoid potential issues with analysing check results. 

### Additional Context

_No response_

mwojtyczka:
## Changes

* Trim autogenerated check name to 255 chars to avoid potential issues when using the results

### Linked issues

Resolves #216 

### Tests

- [ ] manually tested
- [x] added unit tests
- [ ] added integration tests

### DIFF
--- a/src/databricks/labs/dqx/utils.py
+++ b/src/databricks/labs/dqx/utils.py
@@ -14,11 +14,13 @@ def get_column_name(col: Column) -> str:
     It is necessary to parse this out from the string representation.
 
     This works on columns with one or more aliases as well as not aliased columns.
+    The resulting name is truncated to 255 characters to avoid potential issues with the column name length.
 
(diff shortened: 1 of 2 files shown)


ghanse:
LGTM

ghanse (approved):
LGTM

PR #302, opened 2025-03-17
Column expression inputs

ghanse:
## Changes

Add support for SQL expression inputs to column check functions. These can be passed as `str`.

### Linked issues

N/A

### Tests

Updated the integration tests.

- [ ] manually tested
- [ ] added unit tests
- [x] added integration tests


demos/dqx_demo_library.py
```diff
@@ -153,14 +153,24 @@
       allowed:
         - 1
         - 2
+- criticality: error
+  check:
+    function: is_not_null
+    arguments:
+      col_name: element_at(col5, 'key1')
```
mwojtyczka: What happens if the element is missing. Shouldn't we use `try_element_at` everywhere?

mwojtyczka (changes requested):

tests/integration/test_apply_checks.py
```diff
@@ -2107,3 +2107,65 @@ def test_apply_checks_with_sql_expression_for_map_and_array(ws, spark):
...
+            "check": {
+                "function": "is_not_greater_than",
+                "arguments": {"col_name": "try_element_at(try_element_at(col2, 2), 'key1')", "limit": 5},
+            },
+        },
+    ]
+
+    dq_engine = DQEngine(workspace_client=ws, extra_params=EXTRA_PARAMS)
+    checked = dq_engine.apply_checks_by_metadata(test_df, checks)
+    checked.show(10, False)
```
```suggestion

```

mwojtyczka (approved):
LGTM

src/databricks/labs/dqx/engine.py
```diff
@@ -278,7 +278,7 @@ def _create_results_map(self, df: DataFrame, checks: list[DQColRule], dest_col:
             result = F.struct(
                 F.lit(check.name).alias("name"),
                 check.check_column().alias("message"),
-                F.lit(check.col_name).alias("col_name"),
+                F.lit(check.col_name).cast("string").alias("col_name"),
```
mwojtyczka, line 281: Added as it was failing with strange col type in serverless.

PR #312, opened 2025-03-26
Added input spark options and schema for reading from the storage and example of uniqueness check for composite key

### LINKED ISSUE #161: [FEATURE]: Handle spark options while reading data from a file location

### Problem statement

We currently only provide format and input location. 
This works fine for delta but other formats like csv may require additional options.

### Proposed Solution

Extend config, profiler and utils to handle additional spark options when reading input data.

### Additional Context

_No response_

mwojtyczka:
## Changes

* Added input spark options when reading from the storage
* Added input schema when reading from the storage
* Allow usage of fully qualified Unity Catalog or Hive Metastore table name (e.g. catalog.schema.table, hive_metastore.schema.table) and default catalog (e.g. schema.table)
* Added example of uniqueness check for multiple columns (composite key) to the documentation
* Corrected check messages to display expressions correctly.
* Cast all expressions values to string when creating check message to avoid potential issues
* Updated check messages to have it consistent across all checks

### Linked issues

Resolves #161 

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests


tests/integration/test_utils.py
```diff
@@ -16,7 +16,22 @@ def test_read_input_data_unity_catalog_table(spark, make_schema, make_random):
...
+    input_location = f"{catalog_name}.{schema_name}.{make_random(6).lower()}"
+    input_format = None
+    input_read_options = {"versionAsOf": "0"}
+
+    schema = "a: int, b: int"
+    input_df = spark.createDataFrame([[1, 2]], schema)
+    input_df.write.format("delta").saveAsTable(input_location)
+
+    result_df = read_input_data(spark, input_location, input_format, input_read_options)
+    assert_df_equality(input_df, result_df)
```
ghanse: Should we do a 2nd insert to the table before reading? This would validate that `versionAsOf` is respected on-read.

tests/integration/test_utils.py
```diff
@@ -29,3 +44,19 @@ def test_read_input_data_workspace_file(spark, make_schema, make_volume):
...
+    input_location = info.full_name
+    input_format = "delta"
+    input_read_options = {"versionAsOf": "0"}
+
+    schema = "a: int, b: int"
+    input_df = spark.createDataFrame([[1, 2]], schema)
+    input_df.write.format("delta").saveAsTable(input_location)
+
+    result_df = read_input_data(spark, input_location, input_format, input_read_options)
+    assert_df_equality(input_df, result_df)
```
ghanse: Can we test this with a file type other than Delta (e.g. CSV with `header` option)?

ghanse (commented):
Added some ideas for testing.

tests/integration/test_utils.py
mwojtyczka: added

tests/integration/test_utils.py
mwojtyczka: Added. Also added input schema to the config.

PR #339, opened 2025-05-06
Load and save checks from a Delta table

### LINKED ISSUE #299: [FEATURE]: Provide option to define quality rules in a table

### Problem statement

Users can define DQX checks using yaml/json or DQX classes. To improve metadata driven approach, it should also be possible to define check in a table. 

### Proposed Solution

Provide option to save/load checks from a table.

### Additional Context

_No response_

ghanse:
## Changes
Added the following methods to the `DQEngine` class to save and load checks to a Delta table in a Databricks workspace:
- `save_checks_in_table`
- `load_checks_from_table`

### Linked issues
Resolves #299 

### Tests
Added integration tests.

- [x] manually tested
- [ ] added unit tests
- [x] added integration tests

### DIFF
--- a/src/databricks/labs/dqx/engine.py
+++ b/src/databricks/labs/dqx/engine.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import json
 import functools as ft
 import inspect
 import itertools
@@ -8,7 +9,8 @@ from collections.abc import Callable
 from typing import Any
 import yaml
 import pyspark.sql.functions as F
-from pyspark.sql import DataFrame
+from pyspark.sql import DataFrame, Row, SparkSession
+from pyspark.sql.types import StructType, StructField, StringType, MapType
 
 from databricks.labs.blueprint.installation import Installation
 from databricks.labs.dqx import row_checks
@@ -590,6 +592,18 @@ class DQEngine(DQEngineBase):
             raise ValueError(f"Invalid or no checks in workspace file: {installation.install_folder()}/{filename}")
         return parsed_checks
 
+    def load_checks_from_table(self, table_name: str, query: str | None = None) -> list[dict]:
+        """
+        Load checks (dq rules) from a Delta table in the workspace.
+        :param table_name: Unity catalog or Hive metastore table name
+        :param query: Spark SQL expression to filter checks loaded from the table
+        :return: List of dq rules or raise an error if checks file is missing or is invalid.
+        """
+        if not self.ws.tables.exists(table_name):
+            raise NotFound(f"Table {table_name} does not exist in the workspace")
+        logger.info(f"Loading quality rules (checks) from table {table_name}")
+        return DQEngine._load_checks_from_table(table_name, query)
+
     @staticmethod
     def save_checks_in_local_file(checks: list[dict], path: str):
         return DQEngineCore.save_checks_in_local_file(checks, path)
@@ -633,6 +647,17 @@ class DQEngine(DQEngineBase):
             workspace_path, yaml.safe_dump(checks).encode('utf-8'), format=ImportFormat.AUTO, overwrite=True
         )
 
+    def save_checks_in_table(self, checks: list[dict], table_name: str):
+        """
+        Save checks to a Delta table in the workspace.
+        :param checks: list of dq rules to save
+        :param table_name: Unity catalog or Hive metastore table name
+        """
+        if not self.ws.tables.exists(table_name):
+            logger.info(f"Creating table {table_name} for saving quality rules (checks)")
+        logger.info(f"Saving quality rules (checks) to table {table_name}")
+        DQEngine._save_checks_in_table(checks, table_name)
+
     def load_run_config(
         self, run_config_name: str | None = "default", assume_user: bool = True, product_name: str = "dqx"
     ) -> RunConfig:
@@ -670,3 +695,48 @@ class DQEngine(DQEngineBase):
         except NotFound:
             msg = f"Checks file {filename} missing"
             raise NotFound(msg) from None
+
+    @staticmethod
+    def _load_checks_from_table(
+        table_name: str, query: str | None = None, spark: SparkSession | None = None
+    ) -> list[dict]:
+        if spark is None:
+            spark = SparkSession.builder.getOrCreate()
+        rules_df = spark.read.table(table_name)
+        if query is not None:
+            rules_df = rules_df.where(query)
+        return [DQEngine._deserialize_check_table_rule(rule) for rule in rules_df.collect()]
+
+    @staticmethod
+    def _deserialize_check_table_rule(rule: Row) -> dict:
+        return {
+            "name": rule["name"],
+            "criticality": rule["criticality"],
+            "check": {
+                "function": rule["check"]["function"],
+                "arguments": {k: json.loads(v.replace("'", '"')) for k, v in rule["check"]["arguments"].items()},
+            },
+        }
+
+    @staticmethod
+    def _save_checks_in_table(checks: list[dict], table_name: str, spark: SparkSession | None = None):
+        if spark is None:
+            spark = SparkSession.builder.getOrCreate()
+        schema = StructType(
+            [
+                StructField("name", StringType()),
+                StructField("criticality", StringType()),
+                StructField(
+                    "check",
+                    StructType(
+                        [
+                            StructField("function", StringType()),
+                            StructField("arguments", MapType(StringType(), StringType())),
+                        ]
+                    ),
+                ),
+            ]
+        )
+        rows = [Row(**check) for check in checks]
+        rules_df = spark.createDataFrame(rows, schema)
+        rules_df.write.saveAsTable(table_name)
--- a/tests/integration/test_load_checks_from_table.py
+++ b/tests/integration/test_load_checks_from_table.py
@@ -1,0 +1,32 @@
+import pytest
+from databricks.labs.dqx.engine import DQEngine
+from databricks.sdk.errors import NotFound
+
+
+TEST_CHECKS = [
+    {"criticality": "error", "check": {"function": "is_not_null", "arguments": {"col_names": ["col1", "col2"]}}},
+    {"criticality": "warning", "check": {"function": "is_not_null_or_empty", "arguments": {"col_names": ["col_1"]}}},
+]
+
+
+def test_load_checks_when_checks_table_does_not_exist_in_workspace(ws, installation_ctx):
+    checks_table = installation_ctx.check_table
+    with pytest.raises(NotFound, match=f"Table {checks_table} does not exist in the workspace"):
+        engine = DQEngine(ws)
+        engine.load_checks_from_table(checks_table)
+
+
+def test_load_checks_from_table(ws, installation_ctx):
+    checks_table = installation_ctx.check_table
+    engine = DQEngine(ws)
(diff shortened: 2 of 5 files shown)


src/databricks/labs/dqx/engine.py
```diff
@@ -590,6 +592,18 @@ def load_checks_from_installation(
...
+    def load_checks_from_table(self, table_name: str, query: str | None = None) -> list[dict]:
+        """
+        Load checks (dq rules) from a Delta table in the workspace.
+        :param table_name: Unity catalog or Hive metastore table name
+        :param query: Spark SQL expression to filter checks loaded from the table
+        :return: List of dq rules or raise an error if checks file is missing or is invalid.
+        """
+        if not self.ws.tables.exists(table_name).table_exists:
+            raise NotFound(f"Table {table_name} does not exist in the workspace")
+        logger.info(f"Loading quality rules (checks) from table {table_name}")
```
mwojtyczka: You could log this before checking if the table exists so that the context is clear.

src/databricks/labs/dqx/engine.py
```diff
@@ -633,6 +647,17 @@ def save_checks_in_workspace_file(self, checks: list[dict], workspace_path: str)
...
         )
 
+    def save_checks_in_table(self, checks: list[dict], table_name: str):
+        """
+        Save checks to a Delta table in the workspace.
+        :param checks: list of dq rules to save
+        :param table_name: Unity catalog or Hive metastore table name
+        """
+        if not self.ws.tables.exists(table_name):
+            logger.info(f"Creating table {table_name} for saving quality rules (checks)")
```
mwojtyczka: i think it's enough to log that quality rules are saved to a table.

src/databricks/labs/dqx/engine.py
```diff
@@ -670,3 +695,48 @@ def _load_checks_from_file(installation: Installation, filename: str) -> list[di
...
+            "name": rule["name"],
+            "criticality": rule["criticality"],
+            "check": {
+                "function": rule["check"]["function"],
+                "arguments": {k: json.loads(v.replace("'", '"')) for k, v in rule["check"]["arguments"].items()},
+            },
+        }
+
+    @staticmethod
+    def _save_checks_in_table(checks: list[dict], table_name: str, spark: SparkSession | None = None):
```
mwojtyczka: should also have `mode`, e.g. to overwrite existing checks in the table.

src/databricks/labs/dqx/engine.py
```diff
@@ -670,3 +695,48 @@ def _load_checks_from_file(installation: Installation, filename: str) -> list[di
...
+            [
+                StructField("name", StringType()),
+                StructField("criticality", StringType()),
+                StructField(
+                    "check",
+                    StructType(
+                        [
+                            StructField("function", StringType()),
+                            StructField("arguments", MapType(StringType(), StringType())),
+                        ]
```
mwojtyczka: filter is missing

tests/integration/test_load_checks_from_table.py
```diff
@@ -0,0 +1,52 @@
...
+    client.schemas.create(name="dqx", catalog_name="labs")
+
+    checks_table = installation_ctx.check_table
+    with pytest.raises(NotFound, match=f"Table {checks_table} does not exist in the workspace"):
+        engine = DQEngine(ws)
+        engine.load_checks_from_table(checks_table)
+
+
+def test_load_checks_from_table(ws, installation_ctx):
+    client = installation_ctx.workspace_client
```
mwojtyczka, line 32: take catalog and schema from nstallation_ctx.check_table

src/databricks/labs/dqx/engine.py
```diff
@@ -670,3 +695,48 @@ def _load_checks_from_file(installation: Installation, filename: str) -> list[di
         except NotFound:
             msg = f"Checks file {filename} missing"
             raise NotFound(msg) from None
+
+    @staticmethod
+    def _load_checks_from_table(
+        table_name: str, query: str | None = None, spark: SparkSession | None = None
```
mwojtyczka: why we need the query?

mwojtyczka (changes requested):

src/databricks/labs/dqx/engine.py
ghanse: In case users want to filter checks on-load (e.g. `where criticality <> "warning"`).

src/databricks/labs/dqx/engine.py
```diff
@@ -142,6 +144,36 @@ def save_checks_in_local_file(checks: list[dict], filepath: str):
             msg = f"Checks file {filepath} missing"
             raise FileNotFoundError(msg) from None
 
+    @staticmethod
+    def load_checks_from_dataframe(df: DataFrame, query: str | None = None) -> list[dict]:
```
mwojtyczka: Pls add docstring.
I would rename this method, e.g. `build_quality_rules_from_df`. When we say load or save we typically mean an external system (file, table etc). It's a bit unusual to use it in the context of a dataframe.

src/databricks/labs/dqx/engine.py
```diff
@@ -142,6 +144,36 @@ def save_checks_in_local_file(checks: list[dict], filepath: str):
...
+            "filter",
+        )
+        if query is not None:
+            checks_df = checks_df.where(query)
+        if checks_df.count() > COLLECT_LIMIT_WARNING:
+            warnings.warn(f"Collecting large number of rows from Spark DataFrame: {checks_df.count()}")
+        return [row.asDict() for row in checks_df.collect()]
+
+    @staticmethod
+    def save_checks_in_dataframe(checks: list[dict], spark: SparkSession | None = None) -> DataFrame:
```
mwojtyczka: Pls add docstring.
I would rename this method, e.g. `build_df_from_quality_rules`.

src/databricks/labs/dqx/engine.py
```diff
@@ -142,6 +144,36 @@ def save_checks_in_local_file(checks: list[dict], filepath: str):
...
+        if spark is None:
+            spark = SparkSession.builder.getOrCreate()
+        schema = "name STRING, criticality STRING, check STRUCT<function STRING, arguments MAP<STRING, STRING>>, filter STRING"
+        dq_rule_checks = []
```
mwojtyczka: Please reuse the logic already implemented in `build_checks_by_metadata` as much as possible.

src/databricks/labs/dqx/engine.py
mwojtyczka: IMO it would be more useful to be able to load a set of checks for particular use case / run config. We already have the concept of run config. I would use `run_config_name` instead of `query` when saving and loading. By default use `default` run config.
I would not make it too flexible because it increases usage complexity.

tests/integration/test_load_checks_from_table.py
```diff
@@ -0,0 +1,52 @@
...
+    },
+    {
+        "name": "column_not_less_than",
+        "criticality": "warning",
+        "check": {"function": "is_not_less_than", "arguments": {"col_name": "col_1", "min_limit": "0"}},
+    },
+]
+
+
+def test_load_checks_when_checks_table_does_not_exist_in_workspace(installation_ctx, make_schema, make_random):
```
mwojtyczka: removed the "in_workspace" from the name. We actually use UC here.

mwojtyczka (changes requested):

src/databricks/labs/dqx/engine.py
```diff
@@ -633,6 +677,16 @@ def save_checks_in_workspace_file(self, checks: list[dict], workspace_path: str)
             workspace_path, yaml.safe_dump(checks).encode('utf-8'), format=ImportFormat.AUTO, overwrite=True
         )
 
+    def save_checks_in_table(self, checks: list[dict], table_name: str, mode: str = "append"):
```
mwojtyczka: Please update documentation and library demo accordingly.

src/databricks/labs/dqx/engine.py
ghanse: I removed `query`. We can add `run_config_name` if needed.

src/databricks/labs/dqx/engine.py
ghanse: This is done

src/databricks/labs/dqx/engine.py
ghanse: This is done

tests/integration/test_load_checks_from_table.py
ghanse: This is done

src/databricks/labs/dqx/engine.py
mwojtyczka: we should have `run_config_name` here with default as `default` to be able to store configuration for different run configs without a need to create new table

demos/dqx_demo_library.py
```diff
@@ -62,12 +62,17 @@
 user_name = spark.sql("select current_user() as user").collect()[0]["user"]
 checks_file = f"/Workspace/Users/{user_name}/dqx_demo_checks.yml"
 dq_engine = DQEngine(ws)
-dq_engine.save_checks_in_workspace_file(checks, workspace_path=checks_file)
+dq_engine.save_checks_in_workspace_file(checks=checks, workspace_path=checks_file)
+
+# save generated checks in a Delta table
+spark.sql("create catalog if not exists demo")
```
mwojtyczka: we could also use `main.default` as it exists in every workspace by default. Not all users will have access to create new catalog. Or check if the `main.default` exists, and if not create new catalog and schema.

src/databricks/labs/dqx/engine.py
```diff
@@ -590,6 +655,20 @@ def load_checks_from_installation(
             raise ValueError(f"Invalid or no checks in workspace file: {installation.install_folder()}/{filename}")
         return parsed_checks
 
+    def load_checks_from_table(self, table_name: str, spark: SparkSession | None = None) -> list[dict]:
```
mwojtyczka: Please add `run_config_name` as well (default as `default`). This will make it possible to reuse the same table for multiple run configs.

src/databricks/labs/dqx/engine.py
```diff
@@ -633,6 +712,17 @@ def save_checks_in_workspace_file(self, checks: list[dict], workspace_path: str)
             workspace_path, yaml.safe_dump(checks).encode('utf-8'), format=ImportFormat.AUTO, overwrite=True
         )
 
+    @staticmethod
+    def save_checks_in_table(checks: list[dict], table_name: str, mode: str = "append"):
```
mwojtyczka: Please add `run_config_name` as well (default as `default`). This will make it possible to reuse the same table for multiple run configs.

src/databricks/labs/dqx/engine.py
ghanse: Added ✅

src/databricks/labs/dqx/engine.py
ghanse: added ✅

demos/dqx_demo_library.py
ghanse: I used `main.default.dqx_checks_table`. I think we can keep it simple. Users can update with their own catalog and schema where they have access.

mwojtyczka (approved):
LGTM

Issue #344, opened 2025-05-21
[BUG]: is_unique checks with composite unique columns marks all-NULL values as not unique


ovidiu-eremia:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

def test_col_is_unique_handle_nulls_in_composite_cols(spark: SparkSession) -> None:
    """Test shows that is_unique with composite struct doesn't ignore null inner fields.

    The SQL standard states that:
    'A unique constraint on T is satisfied if and only if there do not exist two rows R1
    and R2 of T such that R1 and R2 have the same non-NULL values in the unique cols.'

    Applying this statement to the particular case of two rows with NULL values means:

    # given the table
    T = [Row(a=None, b=None, c=None),
            Row(a=None, b=None, c=None)]

    # when we run is unique with the composite struct of the unique columns
    is_unique(struct('b', 'c'))

    # then we get that the unique constraint is not satisfied
    Value '{null, null}' in Column 'struct(b, c)' is not unique
    Value '{null, null}' in Column 'struct(b, c)' is not unique

    Whereas this test shows that we can't have two rows with only NULL values.

    Args:
        spark (SparkSession): The spark session needed for context

    """
    test_df = spark.createDataFrame(
        data=[
            [1, "", None],
            [2, "", None],
            [3, "str1", 1],
            [4, "str1", 1],
            [None, None, None],
            [None, None, None],
        ],
        schema="a: int, b: string, c:int",
    )

    actual = test_df.select(is_unique(f.col("a")), is_unique(f.struct("b", "c")))

    checked_schema = "a_is_not_unique: string, struct_b_c_is_not_unique: string"
    expected = spark.createDataFrame(
        [
            [None, "Value '{, null}' in Column 'struct(b, c)' is not unique"],
            [None, "Value '{, null}' in Column 'struct(b, c)' is not unique"],
            [None, "Value '{str1, 1}' in Column 'struct(b, c)' is not unique"],
            [None, "Value '{str1, 1}' in Column 'struct(b, c)' is not unique"],
            [None, "Value '{null, null}' in Column 'struct(b, c)' is not unique"],
            [None, "Value '{null, null}' in Column 'struct(b, c)' is not unique"],
        ],
        checked_schema,
    )
    assertDataFrameEqual(actual, expected)

### Expected Behavior

def test_col_is_unique_handle_nulls_in_composite_cols(spark: SparkSession) -> None:

    test_df = spark.createDataFrame(
        data=[
[...]

PR #345, opened 2025-05-23
Handle nulls in uniqueness check for composite keys

mwojtyczka:
## Changes

* Corrected `is_unique` check to consistently handle nulls.  To conform with the SQL ANSI Standard), null values are treated now as unknown, thus not duplicates, e.g. "(NULL, NULL) not equals (NULL, NULL); (1, NULL) not equals (1, NULL). A new parameter called `nulls_distinct` has been added to the check which is set to True by default. If set to False, null values are treated as duplicates, e.g. eg. (1, NULL) equals (1, NULL) and (NULL, NULL) equals (NULL, NULL)
* Added `columns` parameter to the `is_unique` check to be able to pass list of columns to handle composite keys.
* Renamed `col_name` field in the checks to `column` to better reflect the fact that it can be a column name or column expression. This is a breaking change.
* The `for_each_column` field was moved to the check level from the check function arguments for clarity. This is a breaking change.
* Updated documentation to provide guidance on applying quality checks to multiple data sets.
* Removed usage of spark remote for testing to avoid dependency issues and test all spark related tests on Databricks.
* Bufixed saving and loading checks from a table to handle data types correctly.
* Bugixed overwrite mode when saving checks to a table to only overwrite the specific run config name.
* Renamed `DQColRule` rules to `DQRowRule` for clarity, and introduced `DQRowSingleColRule` and `DQRowMultiColRule` classes to better support different type of rules internally. Updated documentation and tests to use `DQRowRule` for row-level checks. This is a breaking change.
* Renamed `DQColSetRule` to `DQRowRuleForEachCol` for clarity. This is a breaking change.
* Minor documentation refactor in various places.

### Linked issues

Resolves [#344](https://github.com/databrickslabs/dqx/issues/344)

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests


docs/dqx/docs/reference/quality_rules.mdx
```diff
@@ -748,6 +754,50 @@ checks = [
...
+      expression: col3 >= col2 and col3 <= 10
+      msg: col3 is less than col2 and col3 is greater than 10
+      name: custom_output_name
+      negate: false
+""")
+
+good_df, quarantine_df = dq_engine.apply_checks_by_metadata_and_split(input_df, checks)
+```
+</details>
+
```
ghanse: We should add an example in the demos.

ghanse (changes requested):

src/databricks/labs/dqx/row_checks.py
```diff
@@ -474,6 +474,8 @@ def is_unique(col_name: str | Column, window_spec: str | Column | None = None) -
     :return: Column object for condition
     """
     col_name_str_norm, col_expr_str, col_expr = _get_norm_col_name_and_expr(col_name)
+    col_expr = _ignore_if_all_struct_fields_null(col_expr, col_expr_str)
```
ovidiu-eremia: This solution will follow the SQL Standard for the situation when all columns of a composite key are NULL:
`(NULL, NULL) not equals (NULL, NULL)`
because all-NULLs rows will be ignored.

But it will not conform to the SQL Standard if at least one column has the same non-value in two or more rows:
`(1, NULL) equals (1, NULL)`
because two struct rows with the non-NULL values in the same position will conflict.

As stated in the bug report, the SQL standard requires that a composite unique constraint is not violated if two or more rows have identical values in the non-null columns but contain NULL in at least one of the other columns of the unique key:
https://github.com/databrickslabs/dqx/issues/344#issuecomment-2906456191

The SQL Standard behavior (expected from the Data Quality Framework):

```
(1, NULL) not equals (1, NULL)
(NULL, NULL) not equals (NULL, NULL)
```

src/databricks/labs/dqx/row_checks.py
```diff
@@ -492,6 +494,31 @@ def is_unique(col_name: str | Column, window_spec: str | Column | None = None) -
...
+    If any of the struct fields are not null, the col expression is returned as is.
+    In pySpark, calling .isNotNull() on a struct column only checks if the entire struct
+    itself is null, not whether all the nested fields are non-null. To check that each field within the struct
+    is non-null, we need to access them individually.
+
+    :param col_expr: Column to check as column expression
+    :param col_expr_str: Column name as string
+    :return: Column expression with a not null condition for struct fields
+    """
+    struct_fields = extract_struct_fields(col_expr_str)
```
ovidiu-eremia: As stated in the [bug report](https://github.com/databrickslabs/dqx/issues/344#issue-3080899919), most probably the solution of using struct columns to package the components of a composite unique constraint, was proposed by DQX as a convention, to workaround a limitation of it not being able to pass variadic positional parameters through the YAML configuration, alongside the variadic named parameters. As in YAML you can't have something like this:

```
arguments:
  - col1
  - col2
  window_spec: window(coalesce(col6, '1970-01-01'), '10 minutes')
```

But if we really want to implement a generic Data Quality Framework, then I think we shouldn't build on this convention and start decomposing every struct field that we are passed to the `is_unique` check, as it might not be struct as part of the convention that the DQX needs because of its limitation, but it could be actually a struct column from the original dataframe that needs to be checked for uniqueness as is. 

Otherwise we'll introduce unexpected behavior, like:

`struct(null, null) NOT equals struct(null, null)`

So a dataframe with these duplicates in its original struct field would pass the `is_unique` check, whereas it shouldn't have.

ovidiu-eremia (changes requested):
I left some change requests as I think that: 

- the current fix doesn't fit the purpose of the #344 bug report, as per the last comment thanks to @mwojtyczka 
- it introduces other incorrect behavior for unique checks if the passed column is a struct field

src/databricks/labs/dqx/row_checks.py
mwojtyczka: Thank you @ovidiu-eremia for your comments. This PR is not finilized yet. The implementation has to be updated to follow standard sql as we discussed earlier + introduce `null_not_distinct` option

src/databricks/labs/dqx/row_checks.py
mwojtyczka: I don't like decomposing the struct either. It would be better to provide the list of columns explicitly. We could either change the interface to accept list of Columns or think of an extra argument for this check.

mwojtyczka: Updated the `is_unique` to use `columns` as argument.

ovidiu-eremia (approved):
LGTM 👍

tests/integration/test_apply_checks.py
```diff
@@ -1907,27 +1939,244 @@ def test_apply_checks_with_is_unique(ws, spark, set_utc_timezone):
...
+    schema = "col1: int, col2: timestamp, col3: string"
+    test_df = spark.createDataFrame(
+        [
+            [1, datetime(2025, 1, 1), "a"],
+            [1, datetime(2025, 1, 2), "a"],
+            [None, None, ""],
+            [None, None, ""],
+        ],
+        schema,
+    )
```
ghanse: We should add a row that has a partial duplicate with a null value like `[1, datetime(2025, 1, 1), None]` for testing `is_unique`.

docs/dqx/docs/guide.mdx
```diff
@@ -263,7 +340,38 @@ In addition, you can also perform a standalone syntax validation of the checks a
...
+# +------------------+-----------------------+---------------------------------------+----------------------+--------------------+
+```
+Fields:
+- `criticality`: either "error" (data going only into "bad/quarantine" dataframe) or "warn" (data going into both "good" and "bad" dataframes). If not provided, the default is "error".
+- `check`: a `StructType` value with the following fields:
+    - `function`: Name of the DQX check function to apply
+    - `arguments`: A `MapType` value with the function's keyword arguments stored as json
+- (optional) `name`: Name to use for the check
+- (optional) `filter`: Spark expression to filter the rows for which the check is applied (e.g. `"business_unit = 'Finance'"`)
+- `run_config_name`: A run or workflow name. Can be used to load and apply a subset of checks to specific workflows. Default value is `"default"`.
```
ghanse: We should include `for_each_column` here in the `check` column.

src/databricks/labs/dqx/rule.py
```diff
@@ -117,86 +121,146 @@ def check_condition(self) -> Column:
...
+        if self.columns is not None:
+            return F.array(*[F.lit(get_column_as_string(column)) for column in self.columns])
+        return F.lit(None).cast("array<string>")
+
+    @ft.cached_property
+    def _check(self) -> Column:
+        """Spark Column expression representing the check condition.
+        :return: A Spark Column object representing the check condition.
+        """
+        args = [self.columns] if self.columns is not None else []
```
ovidiu-eremia, line 201: `self.columns` is already a list. Does it makes sense to nest it in another list in the `args` variable? There is no test for this new type of rule, for me to understand its proper usage, that's why I'm asking...

mwojtyczka: There are tests for the `is_unique` check for this type of rule. 
This type of check takes list[str | Column] as input not str | Column.
I hope that helps.

docs/dqx/docs/reference/engine.mdx
```diff
@@ -38,8 +38,8 @@ The following table outlines the available methods of the `DQEngine` and their f
 <summary>**Available DQX engine methods**</summary>
 | Check                              | Description                                                                                                                 | Arguments                                                                                                                                                                                | Supports local execution |
```
ghanse: We should change _"Check"_ to _"Method"_

src/databricks/labs/dqx/rule.py
```diff
@@ -201,8 +201,21 @@ def _check(self) -> Column:
...
+    *, column: str | Column | None = None, columns: list[str | Column] | None = None, **kwargs
+) -> DQRule:
+    """Factory function to create a row-level data quality rule.
+    :param column: A single column to which the check function is applied.
+    :param columns: A list of columns to which the check function is applied.
+    :param kwargs: Additional keyword arguments to pass to the rule.
+    """
+    if columns is not None:
+        return DQRowMultiColRule(columns=columns, **kwargs)
+    return DQRowSingleColRule(column=column, **kwargs)
```
ghanse: We should raise a warning or error if both `column` and `columns` are specified.

```
if column is not None and columns is not None:
  raise ValueError("Cannot specify both 'column' and 'columns' when creating a DQRowRule.")
```

src/databricks/labs/dqx/rule.py
```diff
@@ -201,8 +201,21 @@ def _check(self) -> Column:
...
+
+
 @dataclass(frozen=True)
-class DQColSetRule:
+class DQRowRuleForEachCol:
     """Represents a row-level data quality rule set that applies a quality check function to multiple columns.
```
ghanse: Maybe say _"that repeatedly applies a quality check function to multiple columns"_? To differentiate from `DQRowMultiColRule`.

src/databricks/labs/dqx/engine.py
```diff
@@ -180,43 +194,56 @@ def build_dataframe_from_quality_rules(
...
         :param spark: SparkSession to use for DataFrame operations
         :return: Spark DataFrame with data quality check rules
         """
         if spark is None:
             spark = SparkSession.builder.getOrCreate()
-        schema = "name STRING, criticality STRING, check STRUCT<function STRING, arguments MAP<STRING, STRING>>, filter STRING, run_config_name STRING"
+        schema = (
+            "name STRING, criticality STRING, check STRUCT<function STRING, arguments MAP<STRING, STRING>>, "
+            "filter STRING, run_config_name STRING"
+        )
```
ghanse: I think we need to include `for_each_column` in the `check` fields.

src/databricks/labs/dqx/engine.py
```diff
@@ -167,10 +170,21 @@ def build_quality_rules_from_dataframe(df: DataFrame, run_config_name: str = "de
...
+                "name": row.name,
+                "criticality": row.criticality,
+                "check": {
+                    "function": row.check["function"],
+                    "arguments": (
+                        {k: json.loads(v) for k, v in row.check["arguments"].items()}
+                        if row.check["arguments"] is not None
+                        else {}
+                    ),
+                },
```
ghanse, line 183: I think we need to include `for_each_column` from the `row.check["for_each_column"]`.

tests/unit/test_build_rules.py
ghanse, line 1: Do we need a unit test for the `@register_rule` decorator?

tests/integration/test_apply_checks.py
mwojtyczka: Added, however, It's already covered by composite key check on col1 and col3

docs/dqx/docs/guide.mdx
mwojtyczka: added

src/databricks/labs/dqx/engine.py
mwojtyczka: Added

src/databricks/labs/dqx/engine.py
mwojtyczka: Added

tests/unit/test_build_rules.py
mwojtyczka: added

docs/dqx/docs/reference/engine.mdx
mwojtyczka: Corrected

src/databricks/labs/dqx/rule.py
mwojtyczka: added

src/databricks/labs/dqx/rule.py
mwojtyczka: corrected

docs/dqx/docs/guide.mdx
```diff
@@ -360,8 +363,9 @@ Fields:
 - `check`: a `StructType` value with the following fields:
     - `function`: Name of the DQX check function to apply
     - `arguments`: A `MapType` value with the function's keyword arguments stored as json
-- (optional) `name`: Name to use for the check
-- (optional) `filter`: Spark expression to filter the rows for which the check is applied (e.g. `"business_unit = 'Finance'"`)
+- (optional) `name`: Name to use for the check.
+- (optional) `filter`: Spark expression to filter the rows for which the check is applied (e.g. `"business_unit = 'Finance'"`).
+- (optional) `for_each_column`: List of column names or expressions to which the check will be applied individually.
```
ghanse: `for_each_column` should be a field in the `check` column:

```
Fields:
- `criticality`: either "error" (data going only into "bad/quarantine" dataframe) or "warn" (data going into both "good" and "bad" dataframes). If not provided, the default is "error".
- `check`: a `StructType` value with the following fields:
    - `function`: Name of the DQX check function to apply
    - `arguments`: A `MapType` value with the function's keyword arguments stored as json
    - (optional) `for_each_column`: List of column names or expressions to which the check will be applied individually.
- (optional) `name`: Name to use for the check
- (optional) `filter`: Spark expression to filter the rows for which the check is applied (e.g. `"business_unit = 'Finance'"`)
```

mwojtyczka: corrected

ghanse (approved):
LGTM

PR #347, opened 2025-05-25
Add variadic positional arguments (check_func_args) to the YAML interface to implement (is_unique) checks on composite keys

### LINKED ISSUE #344: [BUG]: is_unique checks with composite unique columns marks all-NULL values as not unique

ovidiu-eremia:
## Background context
This PR tries to come-up with a proper solution for the checks that need to be applied to composite keys from multiple columns in a dataframe. This issue was raised as a follow-up to #345 which tried to apply a workaround for this limitation of DQX using `pyspark.sql.functions.struct`, because currently it can't define a check on multiple columns from a YAML configuration.

As a first application of this solution for a composite key check, it tries to implement a fix for `is_unique` on multiple columns, by passing them explicitly as a list in the YAML configuration. Currently the workaround proposed by DQX for this use-case was to pack those multiple columns in a `pyspark.sql.functions.struct`, but this introduced a subtle bug in respect to handling the nulls comparison as part of the key to be checked for uniqueness: #344 

## Changes
Currently there is no way of supplying the `check_func_args` attribute for the DQColRule class, from the YAML configuration.
The `arguments` key in a check's block populates the `check_func_kwargs` of the DQColRule class, hence why maybe it should have been `named_arguments` to follow the same approach as the Databricks Workflow's YAML config for a python task which has `parameters` for the positional arguments and `named_parameters` for the keyword arguments.

Considering that renaming the `arguments` to `named_arguments` in the DQX YAML involved backwards compatibility issues with the already existing configurations, I propose `unnamed_args_list` as a key in the YAML config for the variadic positional arguments to be passed to the `check_func_args` attribute of the DQColRule class.

### Linked issues

Resolves #344 

### Tests

- [x] manually tested
- [x] added unit tests
- [ ] added integration tests

### DIFF
--- a/src/databricks/labs/dqx/row_checks.py
+++ b/src/databricks/labs/dqx/row_checks.py
@@ -472,7 +472,7 @@ def is_unique(*columns: str | Column, window_spec: str | Column | None = None, n
     :param window_spec: window specification for the partition by clause. Default value for NULL in the time column
     of the window spec must be provided using coalesce() to prevent rows exclusion!
     e.g. "window(coalesce(b, '1970-01-01'), '2 hours')"
-    :nulls_distinct: flag to have the check consider each null value as an unknown, distinct value, thus not duplicated. 
+    :param nulls_distinct: flag to have the check consider each null value as an unknown, distinct value, thus not duplicated.
     This is by default True to conform with the SQL ANSI Standard.
     e.g. "(NULL, NULL) not equals (NULL, NULL); (1, NULL) not equals (1, NULL); (1, NULL, 'str1') not equals (1, NULL, 'str1')"
     :return: Column object for condition
@@ -488,7 +488,7 @@ def is_unique(*columns: str | Column, window_spec: str | Column | None = None, n
         partition_by_spec = Window.partitionBy(window_spec)
 
     condition = F.count(F.expr("*")).over(partition_by_spec) == 1
-    
+
     if nulls_distinct:
         cols_is_not_null = ft.reduce(
             lambda prev_cols_is_not_null, next_col: prev_cols_is_not_null & next_col.isNotNull(),
@@ -497,15 +497,16 @@ def is_unique(*columns: str | Column, window_spec: str | Column | None = None, n
         )
         condition = F.when(cols_is_not_null, condition)
 
+    cols_values = [
+        F.when(col_expr.isNull(), F.lit("null")).otherwise(col_expr.cast("string")) for col_expr in cols_expr
+    ]
+
     return make_condition(
         ~condition,
         F.concat_ws(
             "",
             F.lit("Value '"),
-            F.concat_ws(
-                ", ",
-                *[col_expr.cast("string") for col_expr in cols_expr],
-            ),
+            F.concat_ws(", ", *cols_values),
             F.lit(f"' in Column '{', '.join(cols_expr_str)}' is not unique"),
         ),
         f"{'_'.join(cols_name_str_norm)}_is_not_unique",


ovidiu-eremia:
@mwojtyczka I would really appreciate your feedback on the proposed solution. 
Thank you!

coverage-unit.xml
```diff
@@ -0,0 +1,2543 @@
+<?xml version="1.0" ?>
```
mwojtyczka, line 1: Unused file

docs/dqx/docs/reference/quality_rules.mdx
```diff
@@ -281,14 +281,29 @@ For brevity, the `name` field in the examples is omitted and it will be auto-gen
     arguments:
       col_name: col1
 
+# is_unique with NULLS NOT DISTINCT option
+- criticality: error
+  check:
+    function: is_unique
+    arguments:
+      col_name: col1
+      nulls_distinct: false
```
mwojtyczka, line 290: We should follow convention used in databases: `nulls_not_distinct` instead of `nulls_distinct`

docs/dqx/docs/reference/quality_rules.mdx
```diff
@@ -281,14 +281,29 @@ For brevity, the `name` field in the examples is omitted and it will be auto-gen
...
+
 # is_unique for multiple columns (composite key)
 - criticality: error
   name: col1_and_col2_is_not_unique
   check:
     function: is_unique
+    unnamed_args_list: [col1, col2]
```
mwojtyczka, line 297: If we really have a compelling reason to provide unnamed args then we should probably rename the `arguments` to `named_arguments` and use `arguments` for the `unnamed_args_list `

docs/dqx/docs/reference/quality_rules.mdx
```diff
@@ -281,14 +281,29 @@ For brevity, the `name` field in the examples is omitted and it will be auto-gen
...
+
+# is_unique for composite key with NULLS NOT DISTINCT
+- criticality: error
+  name: col1_and_col2_is_not_unique
+  check:
+    function: is_unique
+    unnamed_args_list: [col1, col2]
```
mwojtyczka, line 304: why not simply a new argument `columns: [col1, col2]`?
I really don't want to complicate the syntax.

src/databricks/labs/dqx/row_checks.py
```diff
@@ -460,35 +461,55 @@ def is_valid_timestamp(col_name: str | Column, timestamp_format: str | None = No
     )
 
 
-def is_unique(col_name: str | Column, window_spec: str | Column | None = None) -> Column:
+def is_unique(*columns: str | Column, window_spec: str | Column | None = None, nulls_distinct: bool = True) -> Column:
```
mwojtyczka, line 464: Same as before, we could also just use a single argument `columns: list[str | Column]`

docs/dqx/docs/reference/quality_rules.mdx
ovidiu-eremia: I think currently the YAML interface does not match the DQColRule classes interface and it will become a problem in the future run. Yes, sure you could implement a workaround and have a named arg in the is_unique function, but we'll have similar issues in the future because what's possible using the DQColRule classes won't be possible for the YAML config.

If we'll following the already established by Databricks interface of having `arguments` for positional arguments list and `named_arguments` for the keyword arguments, then this won't be complicated for the users as it's idiomatic for YAML, Python but as well for Databricks users.

src/databricks/labs/dqx/row_checks.py
ovidiu-eremia: If we remove the `col_name: str | Column` argument and replace it with the `columns: list[str | Column]` arguments then the previous is_unique checks will fail, as they will send not a list but a str or Column. I tried to have this solution backwards compatible.

docs/dqx/docs/reference/quality_rules.mdx
ovidiu-eremia: The NULL DISTINCT concept is already a medium complexity concept to grasp. If we include NOT it will add a layer of indirection to the logic, thus adding to that complexity. So the user will first need to understand what NULLs DISTINCT means only to negate it afterward. And this makes also coding it more complicated, as we'll need to check for `not nulls_not_distinct` which adds a third layer of indirection to the logic in the `is_unique` function. And we need to bear in mind that the `is_unique` is part of the also user facing module `row_checks` that also needs to be well understood by the users in order for them to enhance or add upon it.

ovidiu-eremia:
Thank you @mwojtyczka for your review! Could you kindly address also my follow-up responses?

docs/dqx/docs/reference/quality_rules.mdx
mwojtyczka: let's just look at an example, even after renaming:
```
- criticality: error
  check:
    function: is_unique
    arguments: [col1, col2]
    named_arguments:
      null_not_distinct: True
```
What does arguments mean. Why I need to define 2 types of arguments to implement my check in this case. Very confusing for the users IMO.

ovidiu-eremia: I understand your point of view. Thank you for your honest feedback and I'm sorry for wasting your time with this...

mwojtyczka: @ovidiu-eremia it's not wasting time, it's good thinking. The implementation would definitely be better. We just have to see how and if we can keep this simple for users. Maybe there is no simpler approach.
At the moment we don't really have checks that require multiple columns except `is_unique`, but with introduction of data set level checks this may change so worth to consider improvements.

docs/dqx/docs/reference/quality_rules.mdx
mwojtyczka: Agree, `nulls_distinct` is easier to understand.

src/databricks/labs/dqx/row_checks.py
mwojtyczka: yes, but any of the proposed changes so far will be breaking anyway. We will bump the version so it's not a big problem. It should work with all args passed as named args or positional args.

docs/dqx/docs/reference/quality_rules.mdx
ovidiu-eremia: In the SQL Standard (and also in the industry standards - PostgreSQL) the setting is not a Boolean flag, but a distinct setting like `NULLS DISTINCT` for true:
https://modern-sql.com/caniuse/unique-nulls-distinct

... and `NULLS NOT DISTINCT` for false:
https://modern-sql.com/caniuse/unique-nulls-not-distinct

So, we get to choose which setting to use...

mwojtyczka: as you pointed out `nulls_distinct` (default to True) is easier to understand and implement so we can go with this.

src/databricks/labs/dqx/row_checks.py
ovidiu-eremia: Actually, I've tested and the current implementation will work with `is_unique` with a single `col_name` (passed in the YAML as the previous usage was) but also with multiple columns passed as positional arguments.

PR #352, opened 2025-06-02
Allow user metadata for individual checks

### LINKED ISSUE #350: [FEATURE]: Add metadata to checks

### Problem statement

We currently add `user_metadata` passed into the `DQEngine` to the row-level metadata generated by DQX checks.

Customers have asked if we can add custom metadata for individual checks.

### Proposed Solution

Allow customers to add `user_metadata` as an argument to each quality check.

### Additional Context

_No response_

ghanse:
## Changes

These change allow users to append custom metadata to individual checks:
- Added `user_metadata` fields to the `DQRule` class
- Updated `DQEngine` methods to serialize/deserialize user metadata from configuration (e.g. files or tables)

### Linked issues

Resolves #350 

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests


src/databricks/labs/dqx/engine.py
```diff
@@ -384,16 +395,19 @@ def _create_results_map(self, df: DataFrame, checks: list[DQRule], dest_col: str
 
         check_cols = []
         for check in checks:
+            user_metadata = self.user_metadata
+            if check.user_metadata is not None:
+                user_metadata |= check.user_metadata
```
mwojtyczka: Add a comment pls that if a the same user metadata exists is already defined at the engine level it will get overwritten by the user metadata from the check.
Please add this info the documentation as well.

tests/integration/test_save_and_load_checks_from_table.py
```diff
@@ -21,11 +22,13 @@
         "name": "col_col1_is_null",
         "criticality": "error",
         "check": {"function": "is_not_null", "arguments": {"column": "col1"}},
+        "user_metadata": {"check_type": "completeness", "check_owner": "someone@email.com"},
     },
     {
         "name": "col_col2_is_null",
         "criticality": "error",
         "check": {"function": "is_not_null", "arguments": {"column": "col2"}},
+        "user_metadata": {"check_type": "completeness", "check_owner": "someone@email.com"},
```
mwojtyczka, line 32: define different user metadata for each check in the test

mwojtyczka (changes requested):
Implementation LGTM.
Request to update tests. Please test apply checks functionality to make sure it actually works and is saved correctly in the results, not only that the checks can be build.
Please also include scenario where the same user metadata is defined at the engine and check level at the same time.

tests/integration/test_save_and_load_checks_from_table.py
ghanse: This is done ✅

tests/integration/test_apply_checks.py
```diff
@@ -3122,3 +3190,367 @@ def test_apply_checks_complex_types_using_classes(ws, spark):
...
+            ],
+        ],
+        expected_schema,
+    )
+
+    actual_df = dq_engine.apply_checks_by_metadata(test_df, checks)
+    assert_df_equality(actual_df, expected_df)
+
+
+def test_apply_checks_with_check_and_engine_metadata_from_classes(ws, spark):
```
mwojtyczka, line 3313: is this the same test as test_apply_checks_with_check_metadata_from_classes? 

It seems it's the same test with just a second rule different.

mwojtyczka (approved):
LGTM. 

Just double check test_apply_checks_with_check_metadata_from_classes is not a duplicate.

demos/dqx_demo_library.py
```diff
@@ -263,6 +274,14 @@
         check_func=row_checks.is_not_null_and_not_empty,
         column="col4"
      ),
+    DQRowRule(
+        criticality="warn",
```
mwojtyczka, line 278: Missing column

mwojtyczka (changes requested):

src/databricks/labs/dqx/engine.py
```diff
@@ -384,16 +395,21 @@ def _create_results_map(self, df: DataFrame, checks: list[DQRule], dest_col: str
 
         check_cols = []
         for check in checks:
+            if check.user_metadata is not None:
+                # Checks defined in the user metadata will override checks defined in the engine
+                user_metadata = self.user_metadata | check.user_metadata
+            else:
+                user_metadata = self.user_metadata
```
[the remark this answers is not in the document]

PR #354, opened 2025-06-03
Fix older than

### LINKED ISSUE #353: [BUG]: is_older_than_col2_for_n_days gives inverted result

### Current Behavior

When evaluating whether a column is n days older than another column, when the result is true,  based on the description given in the documentation 
> Checks whether the values in one input column are at least N days older than the values in another column.

If column 1 is indeed at least n days older, should be treated as a good result, for example when evaluating if 1992-10-18 is at least 6570 days than 2024-10-28, it actually is at least 6570 days older, but it is giving this message in the _errors column
_Value '1992-10-18' in Column 'A' is less than Value '2024-10-28' in Column 'B' for more than 6570 days_ So it is confusing since I'm checking if in fact the col A is at least 6570 days older, It should not be an error, but a good result, since the goal is to check if col 'a' is older than col 'b' for at least n days

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Cloud

AWS

### Operating System

macOS

### Relevant log output

```shell

```

ghanse:
## Changes

Updated the check functions:
- `is_older_than_col2_for_n_days` fails when `col1` is not n days older than `col2`
- `is_older_than_n_days` fails when `col_name` is not n days older than the current/supplied date

### Linked issues

Resolves #353 

### Tests

- [x] manually tested
- [ ] added unit tests
- [x] added integration tests


mwojtyczka (changes requested):
The original intent of these checks was as currently implemented (just the name is wrong, it should be with `_not_`). We wanted to check that a column is not older than another column or current timestamp by N days. Can we add negate flag so that we can choose the behaviour?

Issue #43, opened 2024-07-23
[FEATURE]: DataFame level rules


mwojtyczka:
### Is there an existing issue for this?

- [ ] I have searched the existing issues

### Proposed Solution

Whole data set level checks and window based checks, constraint checks like primary key.

Example checks:
* row count not less than value or expression
* row count not greater than value or expression
* row count equal to value or expression
* foreign key check
* column is in another dataframe
* column match ordered list from table/df
* pk check (duplicate)
* outliers - can be implemented as row level check
* column /mean/median/avg not less/greater than
* ability to apply the checks on groups of rows

### Additional Context

Separate issues should be created for each of the checks once the framework is updated to accept this type of checks.

Palkers76:
could detect FK to PK , i.e. relationship between fact and dim/dict tables

PR #357, opened 2025-06-05
Add aggregation type of checks

mwojtyczka:
## Changes
* Added 2 new checks: `is_aggr_greater_than` and `is_aggr_less_than` to check whether the aggregated values over group of rows are not greater/less than the provided limit
* Updated `is_unique` check to handle check filter properly (filter out records in the windows using the check filter)
* The aggregation checks `is_aggr_greater_than`, `is_aggr_less_than`, `is_unique` only work with spark batch. Support for Spark Structured Streaming will be added in a subsequent PR. 

### Linked issues

Resolves partially: https://github.com/databrickslabs/dqx/issues/43

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests

### DIFF
--- a/src/databricks/labs/dqx/check_funcs.py
+++ b/src/databricks/labs/dqx/check_funcs.py
@@ -603,6 +603,7 @@ def is_aggr_not_greater_than(
     )
 
 
+@register_rule("single_column")
 def is_aggr_not_less_than(
     column: str | Column,
     limit: int | float | str | Column,


src/databricks/labs/dqx/check_funcs.py
```diff
@@ -560,6 +560,92 @@ def is_unique(
     )
 
 
+@register_rule("no_column")
+def is_aggr_less_than(
+    row_filter: str | None,
+    limit: int | str | Column,
+    aggr_type: str = "count",
+    partition_by: list[str | Column] | None = None,
+) -> Column:
```
ghanse: I think this should accept an optional column as `str | Column`. When the aggregation is `avg` for example we can do `avg(col_to_aggregate).over(window_spec)`.

src/databricks/labs/dqx/engine.py
```diff
@@ -557,14 +559,37 @@ def _validate_func_args(arguments: dict, func: Callable, check: dict, func_param
...
+        if origin is list:
+            if not isinstance(value, list):
+                return False
+            if not args:
+                return True  # no inner type to check
+            return all(DQEngineCore._check_type(item, args[0]) for item in value)
+
+        if origin:
+            return isinstance(value, origin)
+        return isinstance(value, expected_type)
```
ghanse: Do we need to check the inner types of any `dict` or `tuple` variables?

tests/integration/test_apply_checks.py
```diff
@@ -3454,3 +3459,367 @@ def test_apply_checks_with_check_and_engine_metadata_from_classes(ws, spark):
...
+            filter="b is not null",
+            check_func_kwargs={"partition_by": ["b"], "limit": 1},
+        ),
+        DQRowRule(
+            name="row_count_partition_by_a_b_less_than_limit",
+            criticality="error",
+            check_func=check_funcs.is_aggr_less_than,
+            check_func_kwargs={"partition_by": ["a", "b"], "limit": 1},
+        ),
+    ]
```
ghanse: We should test the different aggregation types.

tests/unit/test_build_rules.py
ghanse, line 1: We should add a test with aggregate checks with an invalid aggregate type specified.

src/databricks/labs/dqx/engine.py
ghanse, line 1: Can we validate that the aggregate type is supported in `validate_checks`?

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: added, for count aggr one can define column as "col1" or "*". With "*" all rows are counted regardless of nulls in any column

src/databricks/labs/dqx/engine.py
mwojtyczka: added

tests/unit/test_build_rules.py
mwojtyczka: added

src/databricks/labs/dqx/engine.py
mwojtyczka: added

tests/integration/test_apply_checks.py
mwojtyczka: added

ghanse (approved):
LGTM

PR #387, opened 2025-06-24
Add equality checks for aggregate values

### LINKED ISSUE #369: [FEATURE]: `is_aggr_equal` function

### Problem statement

I have a use case where a specific column is expected to sum to `n` at a specific level of `group_by` column.

The simplest example is likely something like a column that represents a `percent`, where we know that at a column / set of columns the sum should be `100`. 

Example:
- We expect the sum of `percent_sales` by `date` to always be 100

| date       | name       | percent_sales |
|------------|------------|---------------|
| 2023-10-01 | Product A  | 75            |
| 2023-10-01 | Product B  | 15            |
| 2023-10-01 | Product C  | 10            |
| 2023-10-02| Product A  | 30            |
| 2023-10-02 | Product B  | 15            |
| 2023-10-02| Product C  | 55            |

Right now to ensure that the sum is `100` I can add two rules, but it seems pretty repetitive - perhaps there is a better way I am omitting? Maybe I need to consider adding a custom class?

```python
[
  DQRowRule(
    name="percent_agg_less_than",
    check_func=check_funcs.is_aggr_not_less_than,
    column = "percent_sales",
    check_func_kwargs={
      "group_by": ['date'],
      "aggr_type": "sum",
      "limit": 100}
  ),
  DQRowRule(
    name="percent_agg_greater_than",
    check_func=check_funcs.is_aggr_not_greater_than,
    column = "percent_sales",
    check_func_kwargs={
      "group_by": ['date'],
      "aggr_type": "sum",
      "limit": 100}
  )
]
```

### Proposed Solution

It would be nice to have a `check_func` that could handle an equivalent statement. 

### Additional Context

`percent` columns in general are where I see this being most useful, but I could see some other use cases - such as ensuring something is properly distributed (e.g. we have a starting value distributed across a series of columns where the sum across those series of columns should be equivalent to the starting value)

ghanse:
## Changes
This PR introduces built-in equality checks for aggregate values:
- `is_aggr_equal_to` checks that the aggregate value is equal to a provided limit
- `is_aggr_not_equal_to` checks that the aggregate value is not equal to a provided limit

### Linked issues
Resolves #369 

### Tests

- [x] manually tested
- [ ] added unit tests
- [x] added integration tests


docs/dqx/docs/reference/quality_rules.mdx
```diff
@@ -1112,6 +1172,74 @@ checks = [
       },
     ),
 
+    # is_aggr_equal check with count aggregation over all rows
+    DQDatasetRule(
+      criticality="error",
+      check_func=check_funcs.is_aggr_equal,
+      check_func_kwargs={
+        "column": "*",
```
mwojtyczka: Please define `column` param at the rule level instead of inside check func kwargs. In that case the column is logged inside column field and it is consistent with other examples.

```
    DQDatasetRule(
      criticality="error",
      column: "*",
      check_func=check_funcs.is_aggr_equal,
      check_func_kwargs={
        "aggr_type": "count",
        "limit": 3
      },
    ),
```

demos/dqx_demo_library.py
```diff
@@ -233,6 +233,20 @@
       column: col1
       aggr_type: count
       limit: 1.2
+- criticality: warn
```
mwojtyczka: We don't necessarily need to update the demo. We have an example of aggr check already in there. We don't want to grow the demo too much.

mwojtyczka (changes requested):

docs/dqx/docs/reference/quality_rules.mdx
ghanse: Refactored ✅

demos/dqx_demo_library.py
ghanse: I removed these changes from the demo. Maybe we can create a separate, quick-demo for the dataset-level checks.

mwojtyczka (approved):
LGTM

PR #415, opened 2025-07-02
dataframe compare

grusin-db:
### Linked issues

Resolves #..

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests


src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1229,3 +1229,111 @@ def _validate_with_ref_params(
 
     if len(columns) != len(ref_columns):
         raise ValueError("The number of columns to check against the reference columns must be equal.")
+
+
+@register_rule("dataset")
+def compare_datasets(
+    columns: list[str],
```
mwojtyczka, line 1248: this should be list[str | Column]
We support providing columns as expressions in all other checks.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1229,3 +1229,111 @@ def _validate_with_ref_params(
...
+    for changed rows, with details on column-level differences.
+    Only columns common across both datasets will be compered.
+    Mismatched columns will be ingored when doing comparison. Consider using schema
+    compare functionality to ensure schema is the matching if desire is to do full dataset comparison.
+
+    :param columns: List of primary key column names used for join.
+    :param ref_df_name: Optional name of reference DataFrame.
+    :param ref_table: Optional name of reference table.
+    :param row_filter: Optional SQL expression for pre-filtering.
+    :param check_missing_records: Performed FULL OUTER JOIN between dataframe, and refernace table to find also records that could be missing from the dataframe
```
mwojtyczka, line 1265: a warning would also be good here, to inform that this may increase the number of records in the output

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1229,3 +1229,111 @@ def _validate_with_ref_params(
...
+
+        ref_df = ref_dfs.get(ref_df_name) if ref_df_name else spark.read.table(ref_table or "")
+        if not ref_df:
+            raise ValueError('Reference DataFrame not found')
+
+        df = df.alias("df1")
+        ref_df = ref_df.alias("df2")
+
+        if row_filter:
+            df = df.filter(F.expr(row_filter))
```
mwojtyczka, line 1288: it must not filter on the input dataframe as it will change the number of records in the output. Instead, the comparison should only be only for records satisfying the filter

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1229,3 +1229,111 @@ def _validate_with_ref_params(
...
+
+        # rows added in the ref df are not considered changed
+        joined = df.join(ref_df, on=columns, how=("full_outer" if check_missing_records else 'left_outer'))
```
mwojtyczka, line 1291: should there be protection against situation that the columns are not unique? 
Generally all checks are expected to return the original data + issues. The join can potentially return more records if the keys are not unique. 
Exception would be with `check_missing_records` where more records may be returned.

PR #416, opened 2025-07-03
Column kwargs

### LINKED ISSUE #398: [FEATURE]: Ability to pass column names for columns in the struct when using sql_expression

### Problem statement

DQX allows to use arbitrary SQL expression in the function `sql_expression`, this ability allows the users to write complex rules but also makes it difficult to determine columns used. Therefore, in the dq result, the `columns` is left empty and unknown.

Image

### Proposed Solution

I want an ability to pass the values for the columns myself manually in the config so that it gets populated in the result.

### Additional Context

_No response_

mwojtyczka:
## Changes
* Populate the `columns` field in the results if `column` or `columns` are specified as kwargs of the check.
* Improved docs for foreign key and is_in_list checks.
* Added optional `columns` argument to `sql_expression` for reporting.
* Added info on using DAB with DQX.

### Linked issues

Resolves #398 

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests


docs/dqx/docs/installation.mdx
```diff
@@ -6,26 +6,41 @@ import Admonition from '@theme/Admonition';
...
+resources:
+  jobs:
+    my_job:
+      # ...
+      tasks:
+        - task_key: my_task
+          # ...
+          libraries:
+            - pypi:
+                package: databricks-labs-dqx==0.6.0
```
ghanse, line 43: This is good. Creating a follow-up item to add a more complete demo with asset bundles.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -148,7 +155,10 @@ def sql_expression(expression: str, msg: str | None = None, name: str | None = N
         expr_col = ~expr_col
         message = F.concat_ws("", F.lit(f"Value is not matching expression: {expr_msg}"))
 
-    name = name if name else get_column_as_string(expression, normalize=True)
+    if not name:
+        name = get_column_as_string(expression, normalize=True)
+        if columns:
+            name = "_".join([get_column_as_string(col, normalize=True) for col in columns]) + "_" + name
```
ghanse: `get_column_as_string` applies character limits. Do we need to check and limit the length of the string after joining the columns?

For example if we have many column inputs or a few column inputs with very large names.

src/databricks/labs/dqx/rule.py
```diff
@@ -161,26 +162,55 @@ class DQRule(abc.ABC, DQRuleTypeMixin):
...
+
+    def _initialize_columns_if_missing(self):
+        """Handle scenarios where 'columns' is provided in check_func_kwargs but not as an attribute."""
+        if "columns" in self.check_func_kwargs:
+            if self.columns is None:
+                object.__setattr__(
+                    self,
+                    "columns",
+                    self.check_func_kwargs.get("columns"),
+                )
```
ghanse: Should we raise a warning if the user provides column arguments using both methods? We will be ignoring the column(s) provided in `check_func_kwargs`.

docs/dqx/docs/installation.mdx
mwojtyczka: yes, that would be great, parked here: https://github.com/databrickslabs/dqx/issues/417

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: absolutely, I think there are few other places where we need to do this

src/databricks/labs/dqx/rule.py
mwojtyczka: it's evaluated already in _validate_attributes. Also tested in test_validate_column_and_columns_provided_as_args.

Issue #448, opened 2025-07-13
[BUG]: criticality attribute in YAML checks throws error in v0.7.0


febiyan:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

In version 0.7.0 YAML checks throw `AttributeError: 'dict' object has no attribute 'check_criticality'`.

### Expected Behavior

No error thrown like its Python API.

### Steps To Reproduce

1. Login Databricks Free edition
2. Create a notebook
3. Create a cell to install the library `%pip install databricks-labs-dqx@0.9.3`
4. Create a cell to restart python `dbutils.library.restartPython()`
5. Create a cell with this code to create checks:
```
import yaml
checks_yaml_text = """
- criticality: error
  name: NoAMEX
  check:
    function: sql_expression
    arguments:
      expression: "paymentMethod NOT IN ('amex')"
"""
checks_yaml = yaml.safe_load(checks_yaml_text)
```
6. Create a cell with this code to run checks:
```
from databricks.labs.dqx.engine import DQEngine
from databricks.sdk import WorkspaceClient

data_source_name = "samples.bakehouse.sales_transactions"
dq_engine2 = DQEngine(WorkspaceClient())
data_df2 = spark.read.table(data_source_name)

valid_df2, quarantine_df2 = dq_engine2.apply_checks_and_split(df=data_df2, checks=checks_yaml)
```

You should get an `AttributeError: 'dict' object has no attribute 'check_criticality'` message

### Cloud

Azure

### Operating System

macOS

### Relevant log output

```shell
File /local_disk0/.ephemeral_nfs/envs/pythonEnv-a541ffe0-66c3-40c2-98f2-8a7e8a2b748b/lib/python3.11/site-packages/databricks/labs/dqx/engine.py:736, in DQEngine.apply_checks_and_split(self, df, checks, ref_dfs)
    724 def apply_checks_and_split(
    725     self, df: DataFrame, checks: list[DQRule], ref_dfs: dict[str, DataFrame] | None = None
    726 ) -> tuple[DataFrame, DataFrame]:
    727     """Applies data quality checks to a given dataframe and split it into two ("good" and "bad"),
    728     according to the data quality checks.
    729 
   (...)
    734     error and warning rows and corresponding result columns
    735     """
--> 736     return self._engine.apply_checks_and_split(df, checks, ref_dfs)
[...]

PR #449, opened 2025-07-13
Fix missing check_criticality attribute

febiyan:
## Changes
Replace the `check_criticality` attribute in the `_get_check_columns` function to `criticality`. There is no other parts of code that use / create that `check_criticality` attribute -- so this is still a guess of the fix.

Let me know if you want me to make tests and how to structure them.

### Linked issues
Resolves https://github.com/databrickslabs/dqx/issues/448

### Tests
Not tested yet. I need help in configuring codespace environment.

### DIFF
--- a/src/databricks/labs/dqx/engine.py
+++ b/src/databricks/labs/dqx/engine.py
@@ -400,7 +400,7 @@ class DQEngineCore(DQEngineCoreBase):
         :param criticality: criticality
         :return: list of check columns
         """
-        return [check for check in checks if check.check_criticality == criticality]
+        return [check for check in checks if check.criticality == criticality]
 
     def _append_empty_checks(self, df: DataFrame) -> DataFrame:
         """Append empty checks at the end of dataframe.


src/databricks/labs/dqx/engine.py
```diff
@@ -400,7 +400,7 @@ def _get_check_columns(checks: list[DQRule], criticality: str) -> list[DQRule]:
         :param criticality: criticality
         :return: list of check columns
         """
-        return [check for check in checks if check.check_criticality == criticality]
+        return [check for check in checks if check.criticality == criticality]
```
mwojtyczka, line 403: yes it is better to use `criticality` field but the existing logic of `check_criticality` should be moved to `_validate_attributes` so that we validate the value passed by the user in this field. After that the `check_criticality` property can be removed.

The unit tests have to be adjusted accordingly. Pls sign your commits as well. See https://databrickslabs.github.io/dqx/docs/dev/contributing/

mwojtyczka (changes requested):
Thx for the PR. Pls address the comments.

PR #463, opened 2025-07-16
Check to compare datasets

grusin-db:
## Changes

* Added dataset-level check that compares two datasets. Only columns that are common across both datasets are compared.Mismatched columns are ignored. Detailed information about the differences is provided. The reference data can be provided as a dataframe or table name. If `check_missing_records` is enabled FULL OUTER JOIN between the DataFrames is performed to find also records that could be missing from the DataFrame. Use this with caution as it may produce output with more rows than in the original DataFrame. List of columns to exclude from the comparison can be provided. Map type columns are excluded from the comparison automatically as they cannot be compared directly.

### Linked issues

Resolves #..

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests


mwojtyczka (approved):
LGTM

PR #464, opened 2025-07-17
Feat add ip address validation

### LINKED ISSUE #452: [FEATURE]: Add IP Address Validation

### Problem statement

IP address validation is a common task

### Proposed Solution

Add specialized check.

Should validate if IPV4 and IPV6 addresses are valid.
Should validate if IP address is a subset of provided CIDR block.

### Additional Context

_No response_

cornzyblack:
## Changes

- Added a new row check called `is_valid_in_ipv4_address`. I used an IPv4 address, as I also have an IPv6 address, but IPv4 is more common. Also, another check can be created for IPv6, if the need arises

image

### Linked issues

Resolves #452 

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests

### DIFF
--- a/src/databricks/labs/dqx/check_funcs.py
+++ b/src/databricks/labs/dqx/check_funcs.py
@@ -528,6 +528,21 @@ def is_valid_timestamp(column: str | Column, timestamp_format: str | None = None
         f"{col_str_norm}_is_not_valid_timestamp",
     )
 
+@register_rule("row")
+def is_valid_ipv4_address(column: str | Column) -> Column:
+    """Checks whether the values in the input column have valid IPv4 address formats.
+
+    :param column: column to check; can be a string column name or a column expression
+    :return: Column object for condition
+    """
+    col_str_norm, col_expr_str, col_expr = _get_norm_column_and_expr(column)
+    condition = ~col_expr.rlike("^(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$")
+
+    return make_condition(
+        condition,
+        f"Column '{col_expr_str}' is not a valid IPv4 address",
+        f"{col_str_norm}_is_not_valid_ipv4_address",
+    )
 
 @register_rule("dataset")
 def is_unique(
--- a/src/databricks/labs/dqx/check_funcs.py
+++ b/src/databricks/labs/dqx/check_funcs.py
@@ -528,6 +528,7 @@ def is_valid_timestamp(column: str | Column, timestamp_format: str | None = None
         f"{col_str_norm}_is_not_valid_timestamp",
     )
 
+
 @register_rule("row")
 def is_valid_ipv4_address(column: str | Column) -> Column:
     """Checks whether the values in the input column have valid IPv4 address formats.
@@ -536,13 +537,14 @@ def is_valid_ipv4_address(column: str | Column) -> Column:
     :return: Column object for condition
     """
     col_str_norm, col_expr_str, col_expr = _get_norm_column_and_expr(column)
-    condition = ~col_expr.rlike("^(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$")
+    condition = ~col_expr.rlike(r"^((25[0-5]|(2[0-4]|1\d|[1-9]|)\d)\.?\b){4}$")
 
     return make_condition(
         condition,
         f"Column '{col_expr_str}' is not a valid IPv4 address",
         f"{col_str_norm}_is_not_valid_ipv4_address",
     )
+
 
 @register_rule("dataset")
 def is_unique(
--- a/tests/integration/test_apply_checks.py
+++ b/tests/integration/test_apply_checks.py
@@ -382,19 +382,19 @@ def test_foreign_key_check_on_composite_keys(ws, spark):
           check:
             function: foreign_key
             arguments:
-              columns: 
+              columns:
               - a
-              ref_columns: 
+              ref_columns:
               - ref_a
               ref_df_name: ref_df
         - criticality: error
           check:
             function: foreign_key
             arguments:
-              columns: 
+              columns:
               - a
               - b
-              ref_columns: 
+              ref_columns:
               - ref_a
               - ref_b
               ref_df_name: ref_df2
@@ -494,10 +494,10 @@ def test_foreign_key_check_on_composite_keys_negate(ws, spark):
           check:
             function: foreign_key
             arguments:
-              columns: 
+              columns:
               - a
               - b
-              ref_columns: 
+              ref_columns:
               - ref_a
               - ref_b
               ref_df_name: ref_df2
@@ -601,9 +601,9 @@ def test_foreign_key_check_yaml(ws, spark):
           check:
             function: foreign_key
             arguments:
-              columns: 
+              columns:
               - a
-              ref_columns: 
+              ref_columns:
               - {ref_column}
               ref_df_name: ref_df
           user_metadata:
@@ -614,9 +614,9 @@ def test_foreign_key_check_yaml(ws, spark):
           check:
             function: foreign_key
             arguments:
-              columns: 
+              columns:
               - a
-              ref_columns: 
+              ref_columns:
               - {ref_column}
               ref_df_name: ref_df
         """
@@ -2540,7 +2540,7 @@ def test_apply_checks_with_sql_query_and_ref_df(ws, spark):
                     sensor.*,
                     COALESCE(specs.min_threshold, 100) AS effective_threshold
                 FROM {{ sensor }} sensor
-                LEFT JOIN {{ sensor_specs }} specs 
+                LEFT JOIN {{ sensor_specs }} specs
                     ON sensor.sensor_id = specs.sensor_id
             )
             SELECT
@@ -2646,7 +2646,7 @@ def test_apply_checks_with_sql_query_and_ref_table(ws, spark):
                         sensor.*,
                         COALESCE(specs.min_threshold, 100) AS effective_threshold
                     FROM {{ sensor }} sensor
-                    LEFT JOIN {{ sensor_specs }} specs 
+                    LEFT JOIN {{ sensor_specs }} specs
                         ON sensor.sensor_id = specs.sensor_id
                 )
                 SELECT
@@ -4201,7 +4201,7 @@ def test_apply_checks_all_row_checks_as_yaml_with_streaming(ws, make_schema, mak
 
     schema = (
         "col1: string, col2: int, col3: int, col4 array<int>, col5: date, col6: timestamp, "
-        "col7: map<string, int>, col8: struct<field1: int>"
+        "col7: map<string, int>, col8: struct<field1: int>, col9: string"
     )
     test_df = spark.createDataFrame(
         [
@@ -4214,6 +4214,7 @@ def test_apply_checks_all_row_checks_as_yaml_with_streaming(ws, make_schema, mak
                 datetime(2025, 1, 12, 1, 0, 0),
                 {"key1": 1},
                 {"field1": 1},
+                "192.168.1.1",
             ],
             [
                 "val2",
@@ -4224,6 +4225,7 @@ def test_apply_checks_all_row_checks_as_yaml_with_streaming(ws, make_schema, mak
                 datetime(2025, 1, 12, 2, 0, 0),
                 {"key1": 1},
                 {"field1": 1},
+                "0.0.0.0",
             ],
             [
                 "val3",
@@ -4234,6 +4236,7 @@ def test_apply_checks_all_row_checks_as_yaml_with_streaming(ws, make_schema, mak
                 datetime(2025, 1, 12, 3, 0, 0),
                 {"key1": 1},
                 {"field1": 1},
+                "255.255.255.255",
             ],
         ],
         schema,
--- a/tests/integration/test_apply_checks.py
+++ b/tests/integration/test_apply_checks.py
@@ -4325,7 +4325,7 @@ def test_apply_checks_all_checks_as_yaml(ws, spark):
 
     schema = (
         "col1: string, col2: int, col3: int, col4 array<int>, col5: date, col6: timestamp, "
-        "col7: map<string, int>, col8: struct<field1: int>"
+        "col7: map<string, int>, col8: struct<field1: int>, col9: string"
     )
     test_df = spark.createDataFrame(
         [
@@ -4338,6 +4338,7 @@ def test_apply_checks_all_checks_as_yaml(ws, spark):
                 datetime(2025, 1, 12, 1, 0, 0),
                 {"key1": 1},
                 {"field1": 1},
+                "255.255.255.255",
             ],
             [
                 "val2",
@@ -4348,6 +4349,7 @@ def test_apply_checks_all_checks_as_yaml(ws, spark):
                 datetime(2025, 1, 12, 2, 0, 0),
                 {"key1": 1},
                 {"field1": 1},
+                "127.0.0.1",
             ],
             [
                 "val3",
@@ -4358,6 +4360,7 @@ def test_apply_checks_all_checks_as_yaml(ws, spark):
                 datetime(2025, 1, 12, 3, 0, 0),
                 {"key1": 1},
                 {"field1": 1},
+                "0.0.0.0",
             ],
         ],
         schema,
@@ -4380,6 +4383,7 @@ def test_apply_checks_all_checks_as_yaml(ws, spark):
                 datetime(2025, 1, 12, 1, 0, 0),
                 {"key1": 1},
                 {"field1": 1},
+                "255.255.255.255",
                 None,
                 None,
             ],
@@ -4392,6 +4396,7 @@ def test_apply_checks_all_checks_as_yaml(ws, spark):
                 datetime(2025, 1, 12, 2, 0, 0),
                 {"key1": 1},
                 {"field1": 1},
+                "127.0.0.1",
                 None,
                 None,
             ],
@@ -4404,6 +4409,7 @@ def test_apply_checks_all_checks_as_yaml(ws, spark):
                 datetime(2025, 1, 12, 3, 0, 0),
                 {"key1": 1},
                 {"field1": 1},
+                "0.0.0.0",
                 None,
                 None,
             ],
@@ -4889,13 +4895,20 @@ def test_apply_checks_all_checks_using_classes(ws, spark):
                 F.try_element_at("col4", F.lit(1)),  # array col
             ],
         ).get_rules(),
+        # is_valid_ipv4_address check
+        DQRowRule(
+            criticality="error",
+            check_func=check_funcs.is_valid_ipv4_address,
+            column="col9",
+            user_metadata={"tag1": "value4", "tag2": "030"},
+        ),
     ]
 
     dq_engine = DQEngine(ws)
 
     schema = (
         "col1: string, col2: int, col3: int, col4 array<int>, col5: date, col6: timestamp, "
-        "col7: map<string, int>, col8: struct<field1: int>"
+        "col7: map<string, int>, col8: struct<field1: int>, col9: string"
     )
     test_df = spark.createDataFrame(
         [
@@ -4908,6 +4921,7 @@ def test_apply_checks_all_checks_using_classes(ws, spark):
                 datetime(2025, 1, 12, 1, 0, 0),
                 {"key1": 1},
                 {"field1": 1},
+                "255.255.255.255",
             ],
             [
                 "val2",
@@ -4918,6 +4932,7 @@ def test_apply_checks_all_checks_using_classes(ws, spark):
                 datetime(2025, 1, 12, 2, 0, 0),
                 {"key1": 1},
                 {"field1": 1},
+                "0.0.0.0",
             ],
             [
                 "val3",
@@ -4928,6 +4943,7 @@ def test_apply_checks_all_checks_using_classes(ws, spark):
                 datetime(2025, 1, 12, 3, 0, 0),
                 {"key1": 1},
                 {"field1": 1},
+                "127.0.0.1",
             ],
         ],
         schema,
@@ -4947,6 +4963,7 @@ def test_apply_checks_all_checks_using_classes(ws, spark):
                 datetime(2025, 1, 12, 1, 0, 0),
                 {"key1": 1},
                 {"field1": 1},
+                "255.255.255.255",
                 None,
                 None,
             ],
@@ -4959,6 +4976,7 @@ def test_apply_checks_all_checks_using_classes(ws, spark):
                 datetime(2025, 1, 12, 2, 0, 0),
                 {"key1": 1},
                 {"field1": 1},
+                "0.0.0.0",
                 None,
                 None,
             ],
@@ -4971,6 +4989,7 @@ def test_apply_checks_all_checks_using_classes(ws, spark):
                 datetime(2025, 1, 12, 3, 0, 0),
                 {"key1": 1},
                 {"field1": 1},
+                "127.0.0.1",
                 None,
                 None,
             ],
--- a/tests/integration/test_row_checks.py
+++ b/tests/integration/test_row_checks.py
@@ -21,6 +21,7 @@ from databricks.labs.dqx.check_funcs import (
     is_not_null_and_not_empty_array,
     is_valid_date,
     is_valid_timestamp,
+    is_valid_ipv4_address,
 )
 
 SCHEMA = "a: string, b: int"
@@ -248,7 +249,7 @@ def test_col_sql_expression(spark):
         sql_expression(
             """
             CASE
-                WHEN TRY_CAST(a AS BIGINT) IS NOT NULL AND TRY_CAST(b AS BIGINT) IS NOT NULL 
+                WHEN TRY_CAST(a AS BIGINT) IS NOT NULL AND TRY_CAST(b AS BIGINT) IS NOT NULL
                     THEN SUBSTRING(a, 1, LENGTH(b)) = b
                 ELSE FALSE
             END""",
@@ -939,9 +940,9 @@ def test_col_is_valid_date(spark, set_utc_timezone):
     )
 
     checked_schema = """
-        a_is_not_valid_date: string, 
-        b_is_not_valid_date: string, 
-        c_is_not_valid_date: string, 
+        a_is_not_valid_date: string,
+        b_is_not_valid_date: string,
+        c_is_not_valid_date: string,
         d_is_not_valid_date: string,
         unresolvedextractvalue_e_dt_is_not_valid_date: string
         """
@@ -1008,9 +1009,9 @@ def test_col_is_valid_timestamp(spark, set_utc_timezone):
     )
 
     checked_schema = """
-        a_is_not_valid_timestamp: string, 
-        b_is_not_valid_timestamp: string, 
-        c_is_not_valid_timestamp: string, 
+        a_is_not_valid_timestamp: string,
+        b_is_not_valid_timestamp: string,
+        c_is_not_valid_timestamp: string,
         d_is_not_valid_timestamp: string,
         e_is_not_valid_timestamp: string,
         unresolvedextractvalue_f_dt_is_not_valid_timestamp: string
@@ -1044,3 +1045,27 @@ def test_col_is_valid_timestamp(spark, set_utc_timezone):
     expected = spark.createDataFrame(checked_data, checked_schema)
 
     assert_df_equality(actual, expected, ignore_nullable=True)
+
+
+def test_col_is_valid_in_ipv4_address(spark):
+    schema_ipv4 = "a: string"
+
+    test_df = spark.createDataFrame(
+        [["255.255.255.255"], ["192.168.01.1"], ["0.0.0.0"], ["192.168.1"], ["abc.def.ghi.jkl"]], schema_ipv4
+    )
+
+    actual = test_df.select(is_valid_ipv4_address("a"))
+
+    checked_schema = "a_is_valid_ipv4_address: string"
+
+    expected = spark.createDataFrame(
+        [
+            [None],
+            ["Value '192.168.01.1' in Column 'a' is not a valid IPv4 address"],
+            [None],
+            ["Value '192.168.1' in Column 'a' is not a valid IPv4 address"],
+            ["Value 'abc.def.ghi.jkl' in Column 'a' is not a valid IPv4 address"],
+        ],
+        checked_schema,
+    )
+    assert actual.select("a_is_valid_ipv4_address") != expected.select("a_is_valid_ipv4_address")
--- a/tests/integration/test_row_checks.py
+++ b/tests/integration/test_row_checks.py
@@ -1056,16 +1056,16 @@ def test_col_is_valid_in_ipv4_address(spark):
 
     actual = test_df.select(is_valid_ipv4_address("a"))
 
-    checked_schema = "a_is_valid_ipv4_address: string"
+    checked_schema = "a_is_not_valid_ipv4_address: string"
 
     expected = spark.createDataFrame(
         [
             [None],
-            ["Value '192.168.01.1' in Column 'a' is not a valid IPv4 address"],
+            ["Column 'a' is not a valid IPv4 address"],
             [None],
-            ["Value '192.168.1' in Column 'a' is not a valid IPv4 address"],
-            ["Value 'abc.def.ghi.jkl' in Column 'a' is not a valid IPv4 address"],
+            ["Column 'a' is not a valid IPv4 address"],
+            ["Column 'a' is not a valid IPv4 address"],
         ],
         checked_schema,
     )
-    assert actual.select("a_is_valid_ipv4_address") != expected.select("a_is_valid_ipv4_address")
+    assert_df_equality(actual, expected, ignore_nullable=True)
--- a/tests/resources/all_row_checks.yaml
+++ b/tests/resources/all_row_checks.yaml
@@ -330,3 +330,9 @@
     - try_element_at(col7, 'key1') # map col
     - try_element_at(col4, 1) # array col
 
+# is_valid_ipv4_address check
+- criticality: error
+  check:
+    function: is_valid_ipv4_address
+    arguments:
+      column: col9
--- a/tests/unit/test_build_rules.py
+++ b/tests/unit/test_build_rules.py
@@ -12,6 +12,7 @@ from databricks.labs.dqx.check_funcs import (
     is_aggr_not_greater_than,
     is_aggr_not_less_than,
     foreign_key,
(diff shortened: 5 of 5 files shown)


src/databricks/labs/dqx/check_funcs.py
```diff
@@ -529,6 +529,23 @@ def is_valid_timestamp(column: str | Column, timestamp_format: str | None = None
...
 
+@register_rule("row")
+def is_valid_ipv4_address(column: str | Column) -> Column:
+    """Checks whether the values in the input column have valid IPv4 address formats.
+
+    :param column: column to check; can be a string column name or a column expression
+    :return: Column object for condition
+    """
+    col_str_norm, col_expr_str, col_expr = _get_norm_column_and_expr(column)
+    condition = ~col_expr.rlike(r"^((25[0-5]|(2[0-4]|1\d|[1-9]|)\d)\.?\b){4}$")
```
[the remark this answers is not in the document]
cornzyblack: Changed the condition to follow copilot's suggestion
Tested
image

Tested regex in Regexr
image

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -529,6 +529,23 @@ def is_valid_timestamp(column: str | Column, timestamp_format: str | None = None
...
 
+@register_rule("row")
+def is_valid_ipv4_address(column: str | Column) -> Column:
+    """Checks whether the values in the input column have valid IPv4 address formats.
+
+    :param column: column to check; can be a string column name or a column expression
+    :return: Column object for condition
+    """
+    col_str_norm, col_expr_str, col_expr = _get_norm_column_and_expr(column)
+    condition = ~col_expr.rlike(r"^(?:25[0-5]|2[0-4]\d|1\d{2}|[1-9]?\d)(?:\.(?:25[0-5]|2[0-4]\d|1\d{2}|[1-9]?\d)){3}$")
```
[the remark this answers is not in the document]
alexott: Couldn't be `[1-9]?\d` simplified to `\d?\d`?

alexott (commented):
In general looks good, I'm not sure (can't check) if we should additionally escape `\` to pass them to the JVM?

ghanse (changes requested):
This check falls under a broader category of pattern matching. To better onboard new patterns in the future, we should implement a `DQPattern` class:

```
class DQPattern(Enum):
  IPV4_ADDRESS = "..."
```

We can then define a standard pattern matching check:
```
def matches_pattern(column: str | Column, pattern: DQPattern) -> Column:
  col_str_norm, col_expr_str, col_expr = _get_norm_column_and_expr(column)
  condition = ~col_expr.rlike(pattern.value)
  return make_condition(
    condition,
    ...
  )
```

And then reference `matches_pattern` for individual checks:
```
@register_rule("row")
def is_valid_ipv4_address(column: str | Column) -> Column:
  return matches_pattern(column, DQPattern.IPV4_ADDRESS)
```

**NOTE:** We will need to introduce method(s) for handling `DQPattern` when loading checks from YAML/JSON.

mwojtyczka (changes requested):
IPV6 should also be covered.

In the original [idea](https://github.com/databrickslabs/dqx/issues/452) we also, wanted to check if IP address is a subset of provided CIDR block. This can also be done in a separate ticket/PR. If not done here, pls create a follow up ticket.

cornzyblack:
Thanks, all, for the feedback. I will make the changes. @mwojtyczka I have created a new ticket for IPV6 and work on the PR #466

cornzyblack:
@ghanse,

Regarding your note:

> NOTE: We will need to introduce method(s) for handling DQPattern when loading checks from YAML/JSON.

Could you clarify if the idea is for users to define and load `DQPattern` objects directly within the YAML/JSON configuration files?

For example, the `regex_match` rule already allows a user to pass a `regex` string as an argument, and this should be fine since the user is defining their pattern (not pre-defined in DQX):

```python
def regex_match(column: str | Column, regex: str, negate: bool = False) -> Column:
    # ... implementation ...
```

```yaml
# Regex Based Check
- criticality: warn
  check:
    function: regex_match
    arguments:
      column: machine_id
      regex: '^MCH-\\d{3}$' # A direct regex string
```

Given that DQPattern will be more of a pre-defined, DQX-level abstraction (rather than user-defined logic), I wanted to get your thoughts on why it should be handled in loading YAML/JSON checks. Are you thinking of a way for users to reference shared patterns inside a check?

mwojtyczka:
> Thanks, all, for the feedback. I will make the changes. @mwojtyczka I have created a new ticket for IPV6 and work on the PR #466

How about the check `Should validate if IP address is a subset of provided CIDR block.`
It is also not covered in this PR. This would need a separate ticket if not covered here.

cornzyblack:
> > Thanks, all, for the feedback. I will make the changes. @mwojtyczka I have created a new ticket for IPV6 and work on the PR #466
> 
> Thanks! How about the requirement `Should validate if IP address is a subset of provided CIDR block.`? It is also not covered in this PR. Either it's covered here or we need a separate ticket for this.

Hi @mwojtyczka I'll cover the `Should validate if IP address is a subset of provided CIDR block.` in this PR.

ghanse:
> @ghanse,
> 
> Regarding your note:
> 
> > NOTE: We will need to introduce method(s) for handling DQPattern when loading checks from YAML/JSON.
> 
> Could you kindly clarify if the idea is for users to define and load `DQPattern` objects directly within the YAML/JSON configuration files?
> 
> For example, the `regex_match` rule already allows a user to pass a `regex` string as an argument, and this should be fine since the user is defining their pattern (not pre-defined in DQX):
> 
> ```python
> def regex_match(column: str | Column, regex: str, negate: bool = False) -> Column:
>     # ... implementation ...
> ```
> 
> ```yaml
> # Regex Based Check
> - criticality: warn
>   check:
>     function: regex_match
>     arguments:
>       column: machine_id
>       regex: '^MCH-\\d{3}$' # A direct regex string
> ```
> 
> Given that DQPattern will be more of a pre-defined, DQX-level abstraction (rather than user-defined logic), I wanted to get your thoughts on why it should be handled in loading YAML/JSON checks. Are you thinking of a way for users to reference shared patterns inside a check?

I think you're right. `DQPattern` can be used internally for now. If we want to expose `DQPattern` as an argument for `regex_match`, we can do that in a later PR.

mwojtyczka:
> **Updates** The PR is ready for review. I'd appreciate some feedback on the implemented rule. Based on my understanding, the intended behaviour is for the user to optionally pass in a CIDR block to check against. Could you please confirm if this aligns with the expected solution, or was the intention to perform the check against a column instead of a parameter?
> 
> ```python
>         DQRowRule(
>             criticality="error",
>             check_func=check_funcs.is_ipv4_in_cidr,
>             column="hostmin",
>             check_func_kwargs={"cidr_block" :"1.178.4.0/24"},
>         ),
> ```

This is correct, the user should pass `cidr_block` as param.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -529,6 +555,80 @@ def is_valid_timestamp(column: str | Column, timestamp_format: str | None = None
     )
 
 
+def _extract_octets_to_bits(column: Column, pattern: str) -> Column:
```
mwojtyczka: please move the private methods to the end of the module. Keep the public methods on top.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -529,6 +555,80 @@ def is_valid_timestamp(column: str | Column, timestamp_format: str | None = None
...
+
+    col_str_norm, col_expr_str, col_expr = _get_norm_column_and_expr(column)
+    cidr_col_expr = F.lit(cidr_block)
+    is_valid_ip = col_expr.rlike(DQPattern.IPV4_ADDRESS.value)
+    ip_bits_col = _convert_ipv4_to_bits(col_expr)
+    cidr_ip_bits_col, cidr_prefix_length_col = _convert_cidr_to_bits_and_prefix(cidr_col_expr)
+
+    ip_net = _get_network_address(ip_bits_col, cidr_prefix_length_col)
+    cidr_net = _get_network_address(cidr_ip_bits_col, cidr_prefix_length_col)
+    condition = F.when(col_expr.isNull() | ~is_valid_ip | (ip_net != cidr_net), F.lit(True)).otherwise(F.lit(False))
```
mwojtyczka: You can reuse `is_valid_ipv4_address` function to validate the ip address, instead of writing the logic again, and display accurate message that ip is not valid if that happens instead of generic message that col is not within CIDR. 

Example:
```
@register_rule("row")
def is_ipv4_in_cidr(column: str | Column, cidr_block: str) -> Column:
    if not re.match(DQPattern.IPV4_CIDR_BLOCK.value, cidr_block):
        raise ValueError(f"CIDR block '{cidr_block}' is not a valid IPv4 CIDR block.")

    col_str_norm, col_expr_str, col_expr = _get_norm_column_and_expr(column)
    cidr_col_expr = F.lit(cidr_block)

    ipv4_msg_col = is_valid_ipv4_address(column)

    ip_bits_col = _convert_ipv4_to_bits(col_expr)
    cidr_ip_bits_col, cidr_prefix_length_col = _convert_cidr_to_bits_and_prefix(cidr_col_expr)
    ip_net = _get_network_address(ip_bits_col, cidr_prefix_length_col)
    cidr_net = _get_network_address(cidr_ip_bits_col, cidr_prefix_length_col)

    cidr_msg = F.concat_ws(
        "",
        F.lit("Value '"),
        col_expr.cast("string"),
        F.lit(f"' in Column '{col_expr_str}' is not in the CIDR block '{cidr_block}'"),
    )

    return F.when(
        ipv4_msg_col.isNotNull(), ipv4_msg_col
    ).when(
        ip_net != cidr_net, cidr_msg
    ).otherwise(F.lit(None).cast("string")).alias(f"{col_str_norm}_is_not_ipv4_in_cidr")
```

tests/integration/test_row_checks.py
```diff
@@ -1044,3 +1046,51 @@ def test_col_is_valid_timestamp(spark, set_utc_timezone):
     expected = spark.createDataFrame(checked_data, checked_schema)
 
     assert_df_equality(actual, expected, ignore_nullable=True)
+
+
+def test_col_is_valid_ipv4_address(spark):
+    schema_ipv4 = "a: string"
+
+    test_df = spark.createDataFrame(
+        [["255.255.255.255"], ["192.168.01.1"], ["0.0.0.0"], ["192.168.1"], ["abc.def.ghi.jkl"], [None]], schema_ipv4
```
mwojtyczka: Pls add more edge cases:
* string without dot
* empty string
* Out-of-range octets e.g 256.168.1.1
* Mix of numeric and non-numeric values: 192.abc.1.1
* Missing numbers, wrong format and extra dots, e.g. "192.168.1", "192.168.1.1.1", "192.168..1"

tests/integration/test_row_checks.py
```diff
@@ -1044,3 +1046,51 @@ def test_col_is_valid_timestamp(spark, set_utc_timezone):
...
+        checked_schema,
+    )
+    assert_df_equality(actual, expected, ignore_nullable=True)
+
+
+def test_is_ipv4_in_cidr(spark):
+    schema_ipv4 = "a: string"
+
+    test_df = spark.createDataFrame(
+        [["255.255.255.255"], ["10.0.0.5"], ["172.16.1.1"], ["192.168.1"], ["abc.def.ghi.jkl"], [None]], schema_ipv4
```
mwojtyczka: Pls add more edge cases:
* string without dot
* empty string
* Out-of-range octets e.g 256.168.1.1
* Mix of numeric and non-numeric values: 192.abc.1.1
* Missing numbers, wrong format and extra dots, e.g. "192.168.1", "192.168.1.1.1", "192.168..1"
* CIDR boundary values (first and last IP in block)
* Invalid CIDR notation, eg. "192.168.1.0/33"

Another test where cidr_block is provided as None or empty.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -30,6 +39,23 @@ def make_condition(condition: Column, message: Column | str, alias: str) -> Colu
     return (F.when(condition, msg_col).otherwise(F.lit(None).cast("string"))).alias(_cleanup_alias_name(alias))
 
 
+def matches_pattern(column: str | Column, pattern: DQPattern) -> Column:
+    """Helper function to match a pattern in a column.
```
ghanse: Maybe update the docstring to something like: `"Checks whether the values in the input column match a given pattern"`

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -529,6 +555,57 @@ def is_valid_timestamp(column: str | Column, timestamp_format: str | None = None
...
+    ip_bits_col = _convert_ipv4_to_bits(col_expr)
+    cidr_ip_bits_col, cidr_prefix_length_col = _convert_cidr_to_bits_and_prefix(cidr_col_expr)
+    ip_net = _get_network_address(ip_bits_col, cidr_prefix_length_col)
+    cidr_net = _get_network_address(cidr_ip_bits_col, cidr_prefix_length_col)
+
+    cidr_msg = F.concat_ws(
+        "",
+        F.lit("Value '"),
+        col_expr.cast("string"),
+        F.lit(f"' in Column '{col_expr_str}' is not in the CIDR block '{cidr_block}'"),
```
ghanse, line 605: Can we add the column value to the message? Similar to `is_valid_timestamp` or other checks.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -30,6 +39,23 @@ def make_condition(condition: Column, message: Column | str, alias: str) -> Colu
...
+    :param column: column to check; can be a string column name or a column expression
+    :param pattern: pattern to match against
+    :return: Column object for condition
+    """
+    col_str_norm, col_expr_str, col_expr = _get_norm_column_and_expr(column)
+    condition = ~col_expr.rlike(pattern.value)
+    final_condition = F.when(col_expr.isNotNull(), condition).otherwise(F.lit(True))
+    return make_condition(
+        final_condition,
+        f"Column '{col_expr_str}' value does not match pattern '{pattern.name}'",
```
ghanse: Can we add the column value to the message? Similar to `is_valid_timestamp` or other checks.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -30,6 +39,23 @@ def make_condition(condition: Column, message: Column | str, alias: str) -> Colu
...
+def matches_pattern(column: str | Column, pattern: DQPattern) -> Column:
+    """Helper function to match a pattern in a column.
+
+    :param column: column to check; can be a string column name or a column expression
+    :param pattern: pattern to match against
+    :return: Column object for condition
+    """
+    col_str_norm, col_expr_str, col_expr = _get_norm_column_and_expr(column)
+    condition = ~col_expr.rlike(pattern.value)
+    final_condition = F.when(col_expr.isNotNull(), condition).otherwise(F.lit(True))
```
ghanse: We typically return `None` when the expression resolves to `NULL`.

docs/dqx/docs/reference/quality_rules.mdx
```diff
@@ -301,7 +303,25 @@ For brevity, the `name` field in the examples is omitted and it will be auto-gen
       regex: '[0-9]+'
       negate: false
 
+# is_valid_ipv4_address check
+
```
```suggestion
```

docs/dqx/docs/reference/quality_rules.mdx
```diff
@@ -301,7 +303,25 @@ For brevity, the `name` field in the examples is omitted and it will be auto-gen
...
+- criticality: error
+  check:
+  function: is_valid_ipv4_address
+  arguments:
+  column: col2
+
+# is_ipv4_in_cidr check
+
```
mwojtyczka, line 312:
```suggestion
```

docs/dqx/docs/reference/quality_rules.mdx
```diff
@@ -301,7 +303,25 @@ For brevity, the `name` field in the examples is omitted and it will be auto-gen
...
+- criticality: error
+  check:
+  function: is_ipv4_in_cidr
+  arguments:
+  column: col2
+  cidr_block: '192.168.1.0/24'
+
 # sql_expression check
+
```
```suggestion
```

tests/integration/test_row_checks.py
```diff
@@ -1044,3 +1046,178 @@ def test_col_is_valid_timestamp(spark, set_utc_timezone):
...
+        ],
+        checked_schema,
+    )
+    assert_df_equality(actual, expected, ignore_nullable=True)
+
+
+def test_is_ipv4_in_cidr(spark):
+    schema_ipv4 = "a: string, b: string"
+
+    test_df = spark.createDataFrame(
```
mwojtyczka, line 1165: Can we also have a few positive tests? these are are failing examples

tests/integration/test_apply_checks.py
```diff
@@ -4879,13 +4888,28 @@ def test_apply_checks_all_checks_using_classes(ws, spark):
...
+            criticality="error",
+            check_func=check_funcs.is_valid_ipv4_address,
+            column="col9",
+            user_metadata={"tag1": "value4", "tag2": "030"},
+        ),
+        # is_ipv4_in_cidr check
+        DQRowRule(
+            criticality="error",
+            check_func=check_funcs.is_ipv4_in_cidr,
+            column="col9",
```
```suggestion
            column=F.col("col9"),
```
Please make at least one test where the column is provided as column expression and not string.

tests/integration/test_row_checks.py
```diff
@@ -1044,3 +1046,178 @@ def test_col_is_valid_timestamp(spark, set_utc_timezone):
...
+            ["255.255.255.255", "192.168.01.1"],
+            ["0.0.0.0", "192.168.1"],
+            ["abc.def.ghi.jkl", None],
+            ["255255155255", "127.0.0.1"],
+            ["192.168.1.1", "10.0.0.1"],
+            ["172.16.0.1", "224.0.0.1"],
+            ["240.0.0.1", "192.168.1"],
+            ["192.168.1.1.1", ""],
+            [" ", "abc.def.ghi.jkl"],
+            ["1.178.7.255", "1.178.4.0"],
```
cornzyblack, line 1176: `1.178.4.0` Column 2, and column 2 `1.178.4.255` on lines 1176 & 1177 are positive checks. Would you like for me to add more to it? @mwojtyczka

mwojtyczka: ah ok, sorry, missed that, it's fine

tests/integration/test_apply_checks.py
cornzyblack: I have now made the changes by adding two extra checks using `F.col` for 
- `is_valid_ipv4_address` on line 4980 - 4985
- `is_ipv4_in_cidr` on line 4994 - 5000

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -529,6 +558,57 @@ def is_valid_timestamp(column: str | Column, timestamp_format: str | None = None
...
+    :param cidr_block: CIDR block string (e.g., '192.168.1.0/24')
+    :raises ValueError: If cidr_block is not a valid string in CIDR notation.
+
+    :return: Column object for condition
+    """
+
+    if not isinstance(cidr_block, str) or not cidr_block.strip():
+        raise ValueError("'cidr_block' must be a non-empty string.")
+
+    if not re.match(DQPattern.IPV4_CIDR_BLOCK.value, cidr_block):
```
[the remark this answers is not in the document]
mwojtyczka: there are multiple places where IPV4_CIDR_BLOCK is used so compiling it once would make sense.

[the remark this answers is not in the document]
cornzyblack: @mwojtyczka  I can make this change, but it will only apply to the lines Copilot suggested. The other references to `DQPattern.IPV4_CIDR_BLOCK.value` use PySpark’s `rlike` or `regexp_extract`, which expect a string and not a compiled regex.

For example, in `_convert_cidr_to_bits_and_prefix` and `_extract_octets_to_bits`, using a compiled pattern would fail in Spark:

```python
compiled_pattern = re.compile('[0-9]')
df.ip_address.rlike(compiled_pattern)  # This will fail
```
re.compile() returns a Pattern object, but Spark requires a string.

image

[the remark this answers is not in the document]
mwojtyczka: True, thx for the answer

docs/dqx/docs/reference/quality_rules.mdx
```diff
@@ -301,6 +303,21 @@ For brevity, the `name` field in the examples is omitted and it will be auto-gen
       regex: '[0-9]+'
       negate: false
 
+# is_valid_ipv4_address check
+- criticality: error
+  check:
+  function: is_valid_ipv4_address
```
mwojtyczka: Please correct the indentations

docs/dqx/docs/reference/quality_rules.mdx
```diff
@@ -301,6 +303,21 @@ For brevity, the `name` field in the examples is omitted and it will be auto-gen
...
+  arguments:
+  column: col2
+
+# is_ipv4_in_cidr check
+- criticality: error
+  check:
+  function: is_ipv4_in_cidr
```
mwojtyczka: Please correct the indentations

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -13,6 +14,13 @@
 from databricks.labs.dqx.utils import get_column_as_string, is_sql_query_safe, normalize_col_str
 
 
+class DQPattern(Enum):
+    """Enum class to represent DQ patterns used to match data in columns."""
+
+    IPV4_ADDRESS = r"^(25[0-5]|2[0-4]\d|1\d{2}|[1-9]?\d)\.(25[0-5]|2[0-4]\d|1\d{2}|[1-9]?\d)\.(25[0-5]|2[0-4]\d|1\d{2}|[1-9]?\d)\.(25[0-5]|2[0-4]\d|1\d{2}|[1-9]?\d)$"
```
mwojtyczka: For simplicity pls declare and reuse the octet:
```
_IPV4_OCTET = r"(25[0-5]|2[0-4]\d|1\d{2}|[1-9]?\d)"

class DQPattern(Enum):
    """Enum class to represent DQ patterns used to match data in columns."""

    IPV4_ADDRESS = rf"^{_IPV4_OCTET}\.{_IPV4_OCTET}\.{_IPV4_OCTET}\.{_IPV4_OCTET}$"
    IPV4_CIDR_BLOCK = rf"^{_IPV4_OCTET}\.{_IPV4_OCTET}\.{_IPV4_OCTET}\.{_IPV4_OCTET}\/(3[0-2]|[12]?\d)$"
```

mwojtyczka (changes requested):

src/databricks/labs/dqx/check_funcs.py
cornzyblack: Okay, I have implemented that change now

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -12,6 +13,15 @@
 from databricks.labs.dqx.rule import register_rule
 from databricks.labs.dqx.utils import get_column_as_string, is_sql_query_safe, normalize_col_str
 
+_IPV4_OCTET = r"(25[0-5]|2[0-4]\d|1\d{2}|[1-9]?\d)"
```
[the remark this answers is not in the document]
cornzyblack: True that the regex pattern `_IPV4_OCTET` alone picks single digits, but when we use it in 

```python
    IPV4_ADDRESS = rf"^{_IPV4_OCTET}\.{_IPV4_OCTET}\.{_IPV4_OCTET}\.{_IPV4_OCTET}$"
    IPV4_CIDR_BLOCK = rf"^{_IPV4_OCTET}\.{_IPV4_OCTET}\.{_IPV4_OCTET}\.{_IPV4_OCTET}\/(3[0-2]|[12]?\d)$"
```

We specify `^` and `$`; a fullmatch is done on the DQPattern regex, so this is fine

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -531,6 +561,57 @@ def is_valid_timestamp(column: str | Column, timestamp_format: str | None = None
...
+
+    if not isinstance(cidr_block, str) or not cidr_block.strip():
+        raise ValueError("'cidr_block' must be a non-empty string.")
+
+    if not re.match(DQPattern.IPV4_CIDR_BLOCK.value, cidr_block):
+        raise ValueError(f"CIDR block '{cidr_block}' is not a valid IPv4 CIDR block.")
+
+    col_str_norm, col_expr_str, col_expr = _get_norm_column_and_expr(column)
+    cidr_col_expr = F.lit(cidr_block)
+    ipv4_msg_col = is_valid_ipv4_address(column)
```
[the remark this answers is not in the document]
cornzyblack: This is because the `is_valid_ipv4_address` function behind the scenes uses `matches_pattern` function to perform the regex validation. While it's a good suggestion, the only reason why we use the  `is_valid_ipv4_address` function instead of `matches_pattern` directly is because we will be duplicating code again.

The solution this seems to be proposing is instead of

```python
ipv4_msg_col = is_valid_ipv4_address(column)
```

 use 

```
ipv4_msg_col = matches_pattern(column, DQPattern.IPV4_ADDRESS)
```

But then that is still the same definition in `is_valid_ipv4_address`, so we will be duplicating code.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1473,3 +1554,32 @@ def _validate_with_ref_params(
...
+
+def _convert_ipv4_to_bits(ip_col: Column) -> Column:
+    """Returns 32-bit binary string from IPv4 address (no CIDR). (e.g., '11000000101010000000000100000001')."""
+    return _extract_octets_to_bits(ip_col, DQPattern.IPV4_ADDRESS.value)
+
+
+def _convert_cidr_to_bits_and_prefix(cidr_col: Column) -> tuple[Column, Column]:
+    """Returns binary IP and prefix length from CIDR (e.g., '192.168.1.0/24')."""
+    ip_bits = _extract_octets_to_bits(cidr_col, DQPattern.IPV4_CIDR_BLOCK.value)
+    prefix_length = F.regexp_extract(cidr_col, DQPattern.IPV4_CIDR_BLOCK.value, 5).cast("int").alias("prefix_length")
```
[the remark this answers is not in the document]
cornzyblack: I have implemented this.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -531,6 +559,57 @@ def is_valid_timestamp(column: str | Column, timestamp_format: str | None = None
...
+    :param column: column to check; can be a string column name or a column expression
+    :param cidr_block: CIDR block string (e.g., '192.168.1.0/24')
+    :raises ValueError: If cidr_block is not a valid string in CIDR notation.
+
+    :return: Column object for condition
+    """
+
+    if not isinstance(cidr_block, str) or not cidr_block.strip():
+        raise ValueError("'cidr_block' must be a non-empty string.")
+
```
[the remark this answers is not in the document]
cornzyblack: I removed the stripping functionality, as the user is meant to be responsible for providing a correct cidr block, e.g:
`191.1.1.1/10 ` is not the same as `191.1.1.1/10` 
The user is responsible for passing in the correct CIDR block, and logic for removing spaces should not be handled by this

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -12,6 +13,16 @@
 from databricks.labs.dqx.rule import register_rule
 from databricks.labs.dqx.utils import get_column_as_string, is_sql_query_safe, normalize_col_str
 
+_IPV4_OCTET = r"(25[0-5]|2[0-4]\d|1\d{2}|[1-9]?\d)"
+CIDR_PREFIX_LENGTH_GROUP = 5
```
mwojtyczka: `CIDR_PREFIX_LENGTH_GROUP ` is only used in the `_convert_cidr_to_bits_and_prefix`. I would kept it local in the func, or do we plan to use it somewhere else?

mwojtyczka (changes requested):
LGTM

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -531,6 +562,57 @@ def is_valid_timestamp(column: str | Column, timestamp_format: str | None = None
...
+    Checks if an IP column value falls within the given CIDR block.
+
+    :param column: column to check; can be a string column name or a column expression
+    :param cidr_block: CIDR block string (e.g., '192.168.1.0/24')
+    :raises ValueError: If cidr_block is not a valid string in CIDR notation.
+
+    :return: Column object for condition
+    """
+
+    if not isinstance(cidr_block, str):
```
```suggestion
    if not cidr_block:
```
should just check if value exists and is not empty

mwojtyczka (changes requested):

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -531,6 +560,57 @@ def is_valid_timestamp(column: str | Column, timestamp_format: str | None = None
...
+    ip_net = _get_network_address(ip_bits_col, cidr_prefix_length_col)
+    cidr_net = _get_network_address(cidr_ip_bits_col, cidr_prefix_length_col)
+
+    cidr_msg = F.concat_ws(
+        "",
+        F.lit("Value '"),
+        col_expr.cast("string"),
+        F.lit(f"' in Column '{col_expr_str}' is not in the CIDR block '{cidr_block}'"),
+    )
+    return (
```
mwojtyczka: please return `make_condition`

docs/dqx/docs/reference/quality_rules.mdx
```diff
@@ -34,6 +34,8 @@ You can also define your own custom checks (see [Creating custom checks](#creati
...
+| `is_ipv4_in_cidr`                  | Checks if an IP column value falls within the given CIDR block.                                                                                                                                                                                                                                                                                                                                                                   | `column`: column to check (can be a string column name or a column expression); `cidr_block`: CIDR block string                                                                                                                                                                                                                                                                                                                                                                        |
```
```suggestion
| `is_ipv4_in_cidr`                  | Checks whether the values in the input column have valid IPv4 address format and fall within the given CIDR block.                                                                                                                                                                                                                                                                                                                | `column`: column to check (can be a string column name or a column expression); `cidr_block`: CIDR block string                                                                                                                                                                                                                                                                                                                                                                        |
```

docs/dqx/docs/reference/quality_rules.mdx
```diff
@@ -34,6 +34,8 @@ You can also define your own custom checks (see [Creating custom checks](#creati
...
+| `is_valid_ipv4_address`            | Checks whether the values in the input column have valid IPv4 address formats.                                                                                                                                                                                                                                                                                                                                                    | `column` to check (can be a string column name or a column expression)                                                                                                                                                                                                                                                                                                                                                                                                                 |
```
```suggestion
| `is_valid_ipv4_address`            | Checks whether the values in the input column have valid IPv4 address format.                                                                                                                                                                                                                                                                                                                                                     | `column` to check (can be a string column name or a column expression)                                                                                                                                                                                                                                                                                                                                                                                                                 |
```

tests/integration/test_apply_checks.py
```diff
@@ -4309,6 +4311,7 @@ def test_apply_checks_all_row_checks_as_yaml_with_streaming(ws, make_schema, mak
                 datetime(2025, 1, 12, 3, 0, 0),
                 {"key1": 1},
                 {"field1": 1},
+                "255.255.255.255",
```
mwojtyczka: This makes `test_apply_checks_all_row_checks_as_yaml_with_streaming` to fail. Update the value so that it falls within the CIDR: '192.168.1.0/24', e.g. "192.168.1.3",

Also values from `col9` must also be added to the `expected`.

tests/integration/test_apply_checks.py
```diff
@@ -4299,6 +4300,7 @@ def test_apply_checks_all_row_checks_as_yaml_with_streaming(ws, make_schema, mak
                 datetime(2025, 1, 12, 2, 0, 0),
                 {"key1": 1},
                 {"field1": 1},
+                "0.0.0.0",
```
mwojtyczka: This makes `test_apply_checks_all_row_checks_as_yaml_with_streaming` to fail. Update the value so that it falls within the CIDR: '192.168.1.0/24', e.g. "192.168.1.2",

mwojtyczka (changes requested):

docs/dqx/docs/reference/quality_rules.mdx
```diff
@@ -34,6 +34,8 @@ You can also define your own custom checks (see [Creating custom checks](#creati
...
+| `is_ipv4_in_cidr`                  | Checks whether the values in the input column have valid IPv4 address format and fall within the given CIDR block.                                                                                                                                                                                                                                                                                                                | `column`: column to check (can be a string column name or a column expression); `cidr_block`: CIDR block string                                                                                                                                                                                                                                                                                                                                                                        |
```
mwojtyczka: For consistency with the other check, can we rename it to `is_ipv4_address_in_cidr ` pls

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -531,6 +561,56 @@ def is_valid_timestamp(column: str | Column, timestamp_format: str | None = None
...
+    """Checks whether the values in the input column have valid IPv4 address formats.
+
+    :param column: column to check; can be a string column name or a column expression
+    :return: Column object for condition
+    """
+    return matches_pattern(column, DQPattern.IPV4_ADDRESS)
+
+
+@register_rule("row")
+def is_ipv4_in_cidr(column: str | Column, cidr_block: str) -> Column:
```
mwojtyczka: For consistency with the other check, can we rename it to `is_ipv4_address_in_cidr` pls

mwojtyczka (changes requested):
Tests passing now. Just last request is to rename `is_ipv4_in_cidr` to `is_ipv4_address_in_cidr` to have consistent naming convention.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1473,3 +1553,33 @@ def _validate_with_ref_params(
 
     if len(columns) != len(ref_columns):
         raise ValueError("The number of columns to check against the reference columns must be equal.")
+
+
+def _extract_octets_to_bits(column: Column, pattern: str) -> Column:
```
mwojtyczka, line 1558: we are currently applying regexp_extract for each of the 4 octets — that’s 4 regex operations per row. 
For performance reason it would be better to extract all 4 octets in one shot and split them, for example: 
```
def _extract_octets_to_bits(column: Column, pattern: str) -> Column:
    """Extracts 4 octets from an IP column and returns the binary string."""
    match = F.regexp_extract(column, pattern, 0)
    octets = F.split(match, r"\.")
    octet_bin = [F.lpad(F.conv(octets[i], 10, 2), 8, "0") for i in range(4)]
    return F.concat(*octet_bin).alias("ip_bits")
```

mwojtyczka (changes requested):

src/databricks/labs/dqx/check_funcs.py
mwojtyczka, line 1558: Sorry, just one more. We are currently applying `regexp_extract` for each of the 4 octets — that’s 4 regex operations per row. 
For performance reason it would be better to extract all 4 octets in one shot and split them, for example:
```
def _extract_octets_to_bits(column: Column, pattern: str) -> Column:
    """Extracts 4 octets from an IP column and returns the binary string."""
    match = F.regexp_extract(column, pattern, 0)
    octets = F.split(match, r"\.")
    octet_bin = [F.lpad(F.conv(octets[i], 10, 2), 8, "0") for i in range(4)]
    return F.concat(*octet_bin).alias("ip_bits")
```

mwojtyczka:
@cornzyblack for the future, to run integration tests pls follow this (you can use any dbx workspace):
https://databrickslabs.github.io/dqx/docs/dev/contributing/#running-integration-tests-and-code-coverage

ghanse (approved):
LGTM

Issue #475, opened 2025-07-25
[BUG]: Datetime profiling fails on `datetime.max` and ignores `round=False` option


bsr-the-mngrm:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When calling `DQProfiler.profile_table()` with `options={"round": False}` on a Delta table that contains `datetime.datetime.max` values, the profiler crashes with:

```bash
OverflowError: date value out of range
```

This happens in `_round_datetime` during profiling. The `round=False` option is also ignored due to a logic bug in the code.

Relevant traceback:
```bash
  File .../profiler.py, line 680, in DQProfiler._round_datetime
    return value.replace(hour=0, minute=0, second=0, microsecond=0) + datetime.timedelta(days=1)
OverflowError: date value out of range
```

Root causes:

1. `_get_min_max()` receives the `opts` dictionary, but does not pass it to `_adjust_min_max_limits()`. As a result, `_round_value()` always uses `opts={"round": True}`, even when the user specifies `round=False`.

1. `_round_datetime()` does not safely handle values near `datetime.datetime.max`, causing overflow when attempting to round “up”.

### Expected Behavior

- `options={"round": False}` should prevent all rounding, including datetime rounding, throughout the profiling logic.
- If a value is `datetime.max`, rounding logic should detect this and avoid performing `+ timedelta(days=1)` that results in overflow.
- Ideally, the function should cap values at `datetime.max` or log a warning, rather than fail.

###  Suggested Fixes
- Propagate `opts` to `_adjust_min_max_limits()` to respect user config.
- Update `_round_datetime()` to detect near-overflow conditions and skip rounding or cap safely.

### Steps To Reproduce

1. Create a Delta table with a timestamp column containing a value equal to or close to datetime.datetime.max:
    ```python
    import datetime
    from pyspark.sql import SparkSession

    spark = SparkSession.builder.getOrCreate()

    df = spark.createDataFrame([
        (datetime.datetime.max.replace(microsecond=0),),
    ], ["_end_date"])

    df.write.format("delta").mode("overwrite").saveAsTable("your_catalog.your_schema.your_table")
    ```
1. Use `databricks.labs.dqx` to profile this table, specifying `options={"round": False}`:
    ```python
    from databricks.labs.dqx import DQProfiler
    from databricks.sdk import WorkspaceClient

    ws = WorkspaceClient()
    profiler = DQProfiler(ws)

[...]

PR #483, opened 2025-07-25
Fix datetime rounding logic and ensure round=False is respected

bsr-the-mngrm:
## TL;DR

Fixes two issues in datetime profiling:
- `round=False` option was not respected due to missing `opts` propagation.
- `datetime.max` caused an `OverflowError` when rounded up — now safely handled.

## Changes

- Fixes a bug where `opts={"round": False}` was ignored during min/max adjustment due to `opts` not being passed to `_adjust_min_max_limits`.
- Prevents `OverflowError` when attempting to round up `datetime.datetime.max` by capping the result to `datetime.datetime.max`.
- Adds a warning log when overflow is caught during datetime rounding.

### Linked issues

Resolves [#475](https://github.com/databrickslabs/dqx/issues/475)

### Tests

- [x] manually tested
- [ ] added unit tests
- [ ] added integration tests

> A unit test for `_round_datetime` was considered but intentionally omitted to avoid accessing a protected method directly. The logic is indirectly covered by existing profiling workflows.

### DIFF
--- a/src/databricks/labs/dqx/profiler/profiler.py
+++ b/src/databricks/labs/dqx/profiler/profiler.py
@@ -675,10 +675,21 @@ class DQProfiler(DQEngineBase):
 
     @staticmethod
     def _round_datetime(value: datetime.datetime, direction: str) -> datetime.datetime:
+        """
+        Rounds a datetime value to midnight based on the specified direction.
+
+        :param value: The datetime value to round.
+        :param direction: The rounding direction ("up" or "down").
+        :return: The rounded datetime value.
+        """
         if direction == "down":
             return value.replace(hour=0, minute=0, second=0, microsecond=0)
         if direction == "up":
-            return value.replace(hour=0, minute=0, second=0, microsecond=0) + datetime.timedelta(days=1)
+            try:
+                return value.replace(hour=0, minute=0, second=0, microsecond=0) + datetime.timedelta(days=1)
+            except OverflowError:
+                logger.warning("Rounding datetime up caused overflow; returning datetime.max instead.")
+                return datetime.datetime.max
         return value
 
     @staticmethod
--- a/src/databricks/labs/dqx/profiler/profiler.py
+++ b/src/databricks/labs/dqx/profiler/profiler.py
@@ -576,13 +576,13 @@ class DQProfiler(DQEngineBase):
                     f"stddev={stddev}, min={metrics.get('min')}"
                 )
             # we need to preserve type at the end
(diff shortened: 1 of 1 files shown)


src/databricks/labs/dqx/profiler/profiler.py
```diff
@@ -674,10 +675,21 @@ def _type_supports_min_max(typ: T.DataType) -> bool:
...
+        :param direction: The rounding direction ("up" or "down").
+        :return: The rounded datetime value.
+        """
         if direction == "down":
             return value.replace(hour=0, minute=0, second=0, microsecond=0)
         if direction == "up":
-            return value.replace(hour=0, minute=0, second=0, microsecond=0) + datetime.timedelta(days=1)
+            try:
+                return value.replace(hour=0, minute=0, second=0, microsecond=0) + datetime.timedelta(days=1)
+            except OverflowError:
```
mwojtyczka, line 693: The change is LGTM but can you please add test for this. Either create new or extend existing (there a couple of tests that use timestamp), e.g. 
https://github.com/databrickslabs/dqx/blob/main/tests/integration/test_profiler.py#L104

Here you can find instruction on how to run integration tests: https://databrickslabs.github.io/dqx/docs/dev/contributing/#running-integration-tests-and-code-coverage

mwojtyczka (changes requested):

mwojtyczka (approved):
LGTM

tests/integration/test_profiler.py
```diff
@@ -195,7 +198,198 @@ def test_profiler_non_default_profile_options(spark, ws):
...
+            parameters={"min": date(2023, 1, 6), "max": date(9999, 12, 31)},
+        ),
+    ]
+    print(stats)
+    assert len(stats.keys()) > 0
+    assert rules == expected_rules
+
+
 def test_profiler_empty_df(spark, ws):
+
```
mwojtyczka:
```suggestion
```

tests/integration/test_profiler.py
```diff
@@ -195,6 +198,193 @@ def test_profiler_non_default_profile_options(spark, ws):
...
+        DQProfile(
+            name="min_max",
+            column="s1.s2.ns3",
+            description="Real min/max values were used",
+            parameters={"min": date(2023, 1, 6), "max": date(9999, 12, 31)},
+        ),
+    ]
+    assert len(stats.keys()) > 0
+    assert rules == expected_rules
+
```
mwojtyczka, line 387:
```suggestion


```

mwojtyczka, line 387:
```suggestion


```

PR #486, opened 2025-07-27
Add built-in methods for PII detection

### LINKED ISSUE #419: [FEATURE]: Add PII detection check func to the built-in checks using extras

### Problem statement

The PII detection check is currently showcased in the demo: https://github.com/databrickslabs/dqx/blob/v0.6.0/demos/dqx_demo_pii_detection.py

However, it's the user responsibility to implement it as a custom check. 

### Proposed Solution

Add PII detection to the standard / built-in checks. The required dependencies should be installed using extras.: `pip install databricks-labs-dqx[pii]`

The module should be registered only if the pii extras is installed. 
The check should be stored in a separate module, e.g. `check_funcs_pii`

### Additional Context

_No response_

ghanse:
## Changes

This PR introduces the `pii` module which users can install to access built-in checks for Personally Identifiable Information (PII). These built-in checks use the [Presidio framework](https://github.com/microsoft/presidio).

#### Installation
Users install PII detection capabilities as an extras module with additional dependencies:

```
pip install databricks-labs-dqx[pii]
```

#### Usage
Users can define checks using a new built-in function, `contains_pii`.

```
from databricks.labs.dqx.rule import DQRowRule
from databricks.labs.dqx.pii.pii_detection_funcs import contains_pii

checks = [
    # PII detection check using stanford-deidentifier-base as a named entity recognizer
    DQRowRule(
        name="col_description_contains_pii",
        criticality="error",
        column="description",
        check_func=contains_pii
    ),    
]
```

#### Configuration
Users can configure the underlying model by passing parameters to `contains_pii`. This includes NLP entity recognition configuration which can be passed as a Python dictionary or as an `NLPEngineConfig`. In the future, we can add `NLPEngineConfig` values to support different named entity recognizers.

**TO-DO:**
- [x] Introduce methods to configure the underlying named entity recognizer
- [x] Add tests
- [x] Add documentation
- [x] Update demos to use the built-in check

### Linked issues

Resolves #419 

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests


docs/dqx/docs/reference/quality_rules.mdx
```diff
@@ -2234,64 +2287,167 @@ Using DQX classes:
 
 When using dataset-level checks, the filter condition is applied before aggregation, ensuring that the check operates only on the relevant subset of rows rather than on the aggregated results.
 
-## Detecting Personally Identifiable Information (PII) using a Python function
+## Detecting Personally Identifiable Information (PII)
+
+DQX provides both built-in PII detection and support for custom PII detection implementations. PII detection checks can be installed as extras:
+
+```bash
+   pip install 'databricks-labs-dqx[pii-detection]'
```
```suggestion
   pip install 'databricks-labs-dqx[pii]'
```

I would simply use `pii`

docs/dqx/docs/reference/quality_rules.mdx
```diff
@@ -37,7 +37,7 @@ You can also define your own custom checks (see [Creating custom checks](#creati
...
+| `contains_pii`                     | Checks whether the values in the input column contain Personally Identifiable Information (PII). Uses Microsoft Presidio to detect various named entities (e.g. PERSON, ADDRESS, EMAIL_ADDRESS). Requires installation of PII detection extras: `pip install 'databricks-labs-dqx[pii-detection]'`.                                                                                                                               | `column`: column to check (can be a string column name or a column expression); `threshold`: confidence threshold for PII detection (0.0 to 1.0, default: 0.7); `language`: optional language of the text (default: 'en'); `entities`: optional list of entities to detect; `nlp_engine_config`: optional dictionary configuring the NLP engine used for PII detection                                                                                                                 |
```
mwojtyczka: would be good to provide official link to presidio configuration options

docs/dqx/docs/reference/quality_rules.mdx
```diff
@@ -354,6 +380,7 @@ For brevity, the `name` field in the examples is omitted and it will be auto-gen
 ```python
 from databricks.labs.dqx.rule import DQRowRule, DQForEachColRule
 from databricks.labs.dqx import check_funcs
+from databricks.labs.dqx.pii import check_funcs as pii_check_funcs
```
mwojtyczka: why we need alias? why not to use `checks_funcs` module directly similar to other examples?

src/databricks/labs/dqx/pii/check_funcs.py
```diff
@@ -0,0 +1,130 @@
...
+from pyspark.sql import Column
+from pyspark.sql.functions import call_udf, concat_ws, lit, to_json, udf
+from pyspark.sql.types import ArrayType, FloatType, IntegerType, StringType, StructType, StructField
+
+from databricks.labs.dqx.rule import register_rule
+from databricks.labs.dqx.check_funcs import make_condition, _get_norm_column_and_expr
+from databricks.labs.dqx.pii.config import NLPEngineConfig
+
+try:
+    from presidio_analyzer import AnalyzerEngine
```
mwojtyczka: This looks quite hacky. I would move this to a separate module, e.g. `checks_funcs_pii`?
Trying to import this module will fail with a clear ImportError without additional checks. This would need to be handled in the `resolve_check_function`

src/databricks/labs/dqx/pii/check_funcs.py
```diff
@@ -0,0 +1,130 @@
...
+                     Lower values = more sensitive, more potential false positives
+    :param language: Optional language of the text (default: 'en')
+    :param entities: Optional list of entities to detect
+    :param nlp_engine_config: Optional NLP engine configuration used for PII detection; Can be `dict` or `NLPEngineConfiguration`
+    :return: Column object for condition that fails when PII is detected
+    """
+    analyzer = _get_analyzer(nlp_engine_config)
+
+    @udf(returnType=_detection_result_type)
+    def _detect_named_entities_udf(value):
```
mwojtyczka: as this is using udf a comment regarding possible performance issue would be good here

src/databricks/labs/dqx/pii/config.py
```diff
@@ -0,0 +1,55 @@
+from enum import Enum
```
mwojtyczka: having the same module under one project may be confusing even though it leaves in different folder. Maybe rename it to `pii_config`?

src/databricks/labs/dqx/pii/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+except ImportError:
+    import warnings
+
+    warnings.warn(
+        "PII detection dependencies not found. Install with: pip install dqx[pii-detection]",
+        ImportWarning,
+        stacklevel=2,
+    )
+
+    def contains_pii(*args, **kwargs):
```
mwojtyczka: Maybe create a helper for this, e.g.

```
    def _missing_dependency(*args, **kwargs):
        raise ImportError(
            "PII detection dependencies not installed. "
            "Install with: pip install 'databricks-labs-dqx[pii-detection]'"
        )

    contains_pii = _missing_dependency
```

Reusable function in utils for this would also be good as we will need this pattern for other things:
```
import importlib

def require_module(module_name: str, extra_name: str):
    try:
        return importlib.import_module(module_name)
    except ImportError as e:
        raise ImportError(
            f"Optional dependency not installed. "
            f"Install with: pip install 'databricks-labs-dqx[{extra_name}]'"
        ) from e
```

Usage:
```
contains_pii = require_module(
    "databricks.labs.dqx.pii.pii_detection_funcs", "pii-detection"
).contains_pii
```

pyproject.toml
```diff
@@ -758,3 +770,6 @@ replacement = '[\1](https://github.com/databrickslabs/dqx/tree/main/\g<2>)'
 # replace relative image links with absolute links
 pattern = '<img (.*?)src="((?!https?://)\S+?)"(.*?)>'
 replacement = '<img \1src="https://raw.githubusercontent.com/databrickslabs/dqx/refs/heads/main/\g<2>"\g<3>>'
+
+[tool.hatch.metadata]
+allow-direct-references = true
```
mwojtyczka, line 759: why is this needed? i suppose because of the github packages, comment would be good

.github/workflows/acceptance.yml
```diff
@@ -38,7 +38,7 @@ jobs:
         with:
           cache: 'pip'
           cache-dependency-path: '**/pyproject.toml'
-          python-version: '3.10'
+          python-version: '3.12'
```
mwojtyczka: i think it's ok to upgrade to 3.11. Non severless cluster that we use for testing has 15.4 which has python 3.11. I would probably no go with 3.12 at this point.

.github/workflows/acceptance.yml
```diff
@@ -90,7 +90,7 @@ jobs:
...
```
mwojtyczka: same as above

docs/dqx/docs/reference/quality_rules.mdx
```diff
@@ -15,30 +15,30 @@ You can also define your own custom checks (see [Creating custom checks](#creati
...
+| `contains_pii`                     | Checks whether the values in the input column contain Personally Identifiable Information (PII). Uses Microsoft Presidio to detect various named entities (e.g. PERSON, ADDRESS, EMAIL_ADDRESS). Requires installation of PII detection extras: `pip install 'databricks-labs-dqx[pii-detection]'`.                                                                                                                               | `column`: column to check (can be a string column name or a column expression); `threshold`: confidence threshold for PII detection (0.0 to 1.0, default: 0.7); `language`: optional language of the text (default: 'en'); `entities`: optional list of entities to detect; `nlp_engine_config`: optional dictionary configuring the NLP engine used for PII detection, see the [Presidio documentation](https://microsoft.github.io/presidio/analyzer/customizing_nlp_models/) for more information |
```
```suggestion
| `contains_pii`                     | Checks whether the values in the input column contain Personally Identifiable Information (PII). Uses Microsoft Presidio to detect various named entities (e.g. PERSON, ADDRESS, EMAIL_ADDRESS). Requires installation of PII detection extras: `pip install 'databricks-labs-dqx[pii]'`.                                                                                                                                         | `column`: column to check (can be a string column name or a column expression); `threshold`: confidence threshold for PII detection (0.0 to 1.0, default: 0.7); `language`: optional language of the text (default: 'en'); `entities`: optional list of entities to detect; `nlp_engine_config`: optional dictionary configuring the NLP engine used for PII detection, see the [Presidio documentation](https://microsoft.github.io/presidio/analyzer/customizing_nlp_models/) for more information |
```

mwojtyczka: I would remove the PII detection demo. It won't be needed anymore.

pyproject.toml
```diff
@@ -34,6 +34,14 @@ dependencies = ["databricks-labs-blueprint>=0.9.1,<0.10",
 
 [project.optional-dependencies]
 cli = ["pyspark~=3.5.0"]
+pii = [
+    "databricks-connect>=16.4.1; python_version >= '3.12'",
```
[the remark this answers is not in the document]
mwojtyczka: agree, this should be aligned. I would also prefer 3.11

pyproject.toml
```diff
@@ -71,11 +80,11 @@ dependencies = [
     "ruff~=0.3.4",
     "types-PyYAML~=6.0.12",
     "types-requests~=2.31.0",
-    "databricks-connect~=15.4",
-    "pyspark~=3.5.0",
+    "databricks-connect~=16.4.3",
```
mwojtyczka: i think it's not possible to use newer version of databricks connect on older runtimes. So this won't work on our tests cluster that has 15.4. We require runtime 15.4 or higher for dqx so we should still support the older one.

pyproject.toml
```diff
@@ -71,11 +80,11 @@ dependencies = [
...
+    "pylint-per-file-ignores",
 ]
 
-python="3.10"
+python="3.12"
```
mwojtyczka: i would probably forget about 3.10, but use 3.11 by default.

src/databricks/labs/dqx/pii/pii_detection_funcs.py
```diff
@@ -0,0 +1,159 @@
...
+
+logging.getLogger("presidio_analyzer").setLevel(logging.ERROR)
+logger = logging.getLogger(__name__)
+
+_default_nlp_engine_config = NLPEngineConfig.SPACY_SMALL
+_required_modules = ["presidio_analyzer", "spacy"]
+
+
+@register_rule("row")
+def contains_pii(
```
mwojtyczka: to be aligned with the existing naming convention, should we call this `is_pii`?

docs/dqx/docs/reference/quality_rules.mdx
mwojtyczka: Please improve the following:
* It should be 'databricks-labs-dqx[pii]' and not 'databricks-labs-dqx[pii-detection]'. 
* A link to the PII section on this page would also be good
* Info that this function is not supported when running checks with Databricks Connect

pyproject.toml
```diff
@@ -72,10 +81,10 @@ dependencies = [
     "types-PyYAML~=6.0.12",
     "types-requests~=2.31.0",
     "databricks-connect~=15.4",
-    "pyspark~=3.5.0",
```
mwojtyczka, line 75: Why are removing this, we need this dev dependency. We have databricks connect as well but to be on safe side we should have pyspark as well.

src/databricks/labs/dqx/pii/pii_detection_funcs.py
```diff
@@ -0,0 +1,159 @@
...
+
+            results = analyzer.analyze(
+                text=text,
+                entities=entities,
+                language=language,
+                score_threshold=threshold,
+            )
+
+            qualified_results = [result for result in results if result.score >= threshold]
+            if not qualified_results or len(qualified_results) == 0:
```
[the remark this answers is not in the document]
mwojtyczka: agree

src/databricks/labs/dqx/pii/pii_detection_funcs.py
```diff
@@ -0,0 +1,159 @@
...
+    :param nlp_engine_config: Optional NLP engine configuration used for PII detection; Can be `dict` or `NLPEngineConfiguration`
+    :return: Column object for condition that fails when PII is detected
+    """
+    warnings.warn(
+        "PII detection uses user-defined functions which may degrade performance. "
+        "Sample or limit large datasets when running PII detection.",
+    )
+
+    if threshold < 0.0 or threshold > 1.0:
+        raise ValueError(f"Provided threshold {threshold} must be between 0.0 and 1.0")
```
mwojtyczka, line 53: pls add unit test for this

src/databricks/labs/dqx/pii/pii_detection_funcs.py
```diff
@@ -0,0 +1,159 @@
...
+
+    if not nlp_engine_config:
+        nlp_engine_config = _default_nlp_engine_config
+
+    config_dict = nlp_engine_config if isinstance(nlp_engine_config, dict) else nlp_engine_config.value
+    if not isinstance(config_dict, dict):
+        raise ValueError(f"Invalid type provided for 'nlp_engine_config': {type(nlp_engine_config)}")
```
mwojtyczka, line 59: pls add unit test for this

src/databricks/labs/dqx/pii/pii_detection_funcs.py
```diff
@@ -0,0 +1,159 @@
...
+def _validate_environment() -> None:
+    """
+    Validates that the environment can run PII detection checks which use Python dependencies.
+    """
+    connect_session_pattern = re.compile(r"127.0.0.1|.*grpc.sock")
+    session = pyspark.sql.SparkSession.builder.getOrCreate()
+    if isinstance(session, pyspark.sql.connect.session.SparkSession) and not connect_session_pattern.search(
+        session.client.host
+    ):
+        raise TypeError("'contains_pii' is not supported when running checks with Databricks Connect")
```
mwojtyczka: pls add comment why is this needed. As much info as possible based on the research.

src/databricks/labs/dqx/pii/pii_detection_funcs.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+def _get_model_names(nlp_engine_config: dict) -> list[str]:
```
mwojtyczka: it does not seem this function is used anywhere

tests/e2e/test_run_demos.py
```diff
@@ -101,8 +101,11 @@ def test_run_dqx_demo_pii_detection(make_notebook, make_cluster, make_job):
         notebook = make_notebook(content=f, format=ImportFormat.SOURCE)
 
     notebook_path = notebook.as_fuse().as_posix()
-    notebook_task = NotebookTask(notebook_path=notebook_path, base_parameters={"test_library_ref": TEST_LIBRARY_REF})
-    cluster = make_cluster(single_node=True, spark_version="15.4.x-scala2.12")
+    notebook_task = NotebookTask(
+        notebook_path=notebook_path,
+        base_parameters={"test_library_ref": f"databricks-labs-dqx[pii] @ {TEST_LIBRARY_REF}"},
+    )
+    cluster = make_cluster(single_node=True, spark_version="16.4.x-scala2.12")
```
mwojtyczka: A comment here why we need to use a new cluster would be useful.

tests/integration/test_row_checks.py
```diff
@@ -1224,6 +1226,22 @@ def test_is_ipv4_address_in_cidr(spark):
     assert_df_equality(actual, expected, ignore_nullable=True)
 
 
+def test_contains_pii_fails_session_validation(spark):
+    schema_pii = "col1: string"
+    test_df = spark.createDataFrame(
+        [
+            ["John Smith works at Acme Corp"],
+            ["Contact us at info@company.com"],
```
mwojtyczka: could also reduce the amount of data. Just one row would be enough

tests/integration/test_row_checks.py
```diff
@@ -1224,6 +1226,22 @@ def test_is_ipv4_address_in_cidr(spark):
...
+            ["No PII here"],
+            [None],
+        ],
+        schema_pii,
+    )
+
+    with pytest.raises(TypeError, match="'contains_pii' is not supported when running checks with Databricks Connect"):
```
mwojtyczka: A comment would be useful here that due to this limitation functional testing is done in the e2e tests.

demos/dqx_demo_pii_detection.py
```diff
@@ -22,124 +12,164 @@
...
+    name='contains_pii',
+  )
+]
+
+# Initialize the DQX engine:
+dq_engine = DQEngine(WorkspaceClient())
+
+# Create some sample data:
+data = [
+  ['My name is John Smith'],
```
mwojtyczka: Please add row with Null/None as well

tests/e2e/test_run_demos.py
ghanse: I removed the cluster here. This demo should now run on serverless to align with the others.

src/databricks/labs/dqx/pii/pii_detection_funcs.py
ghanse: Removed

src/databricks/labs/dqx/pii/pii_detection_funcs.py
```diff
@@ -0,0 +1,154 @@
...
+
+from databricks.labs.dqx.rule import register_rule
+from databricks.labs.dqx.check_funcs import make_condition, _get_normalized_column_and_expr
+from databricks.labs.dqx.pii.nlp_engine_config import NLPEngineConfig
+
+logging.getLogger("presidio_analyzer").setLevel(logging.ERROR)
+logger = logging.getLogger(__name__)
+
+_default_nlp_engine_config = NLPEngineConfig.SPACY_SMALL
+_required_modules = ["presidio_analyzer", "spacy"]
```
mwojtyczka: this is actually not used

ghanse: Removed

tests/integration/test_row_checks.py
```diff
@@ -1224,6 +1226,23 @@ def test_is_ipv4_address_in_cidr(spark):
     assert_df_equality(actual, expected, ignore_nullable=True)
 
 
+def test_contains_pii_fails_session_validation(spark):
+    """
+    We restrict running `contains_pii` using Databricks Connect. Because tests are run
```
mwojtyczka: would be useful to have info why we cannot use databricks connect

.github/workflows/nightly.yml
```diff
@@ -78,7 +78,7 @@ jobs:
         with:
           cache: 'pip'
           cache-dependency-path: '**/pyproject.toml'
-          python-version: '3.10'
+          python-version: '3.11'
```
mwojtyczka, line 81: there are more places in .github where this has to be updated, e.g. puhs.yml, downstream.yml, release.yml

src/databricks/labs/dqx/pii/pii_detection_funcs.py
ghanse: I was thinking `contains_pii` because this will identify the PII anywhere in the input string.

mwojtyczka: makes sense

tests/integration/test_row_checks.py
```diff
@@ -1224,6 +1226,23 @@ def test_is_ipv4_address_in_cidr(spark):
...
+    we can only validate that the correct error is raised.
+
+    Complete testing of `contains_pii` and its options has been added to e2e tests.
+    We run many scenarios in `dqx_demo_pii_detection` to validate `contains_pii` from
+    a Databricks workspace.
+    """
+    schema_pii = "col1: string"
+    test_df = spark.createDataFrame([["Contact us at info@company.com"]], schema_pii)
+
+    with pytest.raises(TypeError, match="'contains_pii' is not supported when running checks with Databricks Connect"):
```
mwojtyczka: Shouldn't that be `ImportError`?

A unit tests would also be nice. Simply create instance of DQRowRule with this check func.

mwojtyczka (approved):
LGTM - left minor comment. Will merge after testing.

tests/integration/test_row_checks.py
ghanse: Sounds good. Do you think `ImportError` or `NotImplementedError` is better?

mwojtyczka: `ImportError` is more clear IMO

src/databricks/labs/dqx/pii/pii_detection_funcs.py
```diff
@@ -0,0 +1,153 @@
...
+from databricks.labs.dqx.pii.nlp_engine_config import NLPEngineConfig
+
+logging.getLogger("presidio_analyzer").setLevel(logging.ERROR)
+logger = logging.getLogger(__name__)
+
+_default_nlp_engine_config = NLPEngineConfig.SPACY_SMALL
+
+
+@register_rule("row")
+def contains_pii(
```
mwojtyczka: the logic is reverted. The quality check should fail if the condition is violated. The function name should be updated to `not_contains_pii`. I would personally go with `is_not_pii`.
For example: `is_not_null` check will fail if the value is null.

mwojtyczka (changes requested):
func name needs to be updated

src/databricks/labs/dqx/pii/pii_detection_funcs.py
```diff
@@ -0,0 +1,153 @@
...
+    Builds a UDF with the provided threshold, entities, language, and analyzer.
+
+    :param nlp_engine_config: Dictionary configuring the NLP engine used for PII detection
+    :param language: Language of the text
+    :param threshold: Confidence threshold for named entity detection (0.0 to 1.0)
+    :param entities: List of entities to detect
+    :return: PySpark UDF which can be called to detect PII with the given configuration
+    """
+
+    @pandas_udf("string", PandasUDFType.SCALAR)
```
mwojtyczka: This way of defining pandas udfs will be depreciated in the future. Since Spark 3.0+, the preferred way is to use type hints and omit PandasUDFType.SCALAR entirely:
```
@pandas_udf("string")
def handler(batch: pd.Series) -> pd.Series:
```

mwojtyczka: updated

PR #487, opened 2025-07-27
Refactored methods for loading and saving checks

mwojtyczka:
## Changes

* Moved all load and save checks methods from `DQEngine` to a separate module. The previous approach of continuously expanding DQEngine was not scalable or maintainable. This change will make it easier to add more ways of saving and loading checks.
* Refactored code in various places to improve modularity and adhere to the separation of concerns. The original `DQEngine` had grown beyond its intended scope, so responsibilities were redistributed. The change should provide a good basis for extensions without overloading the engine.
* All conversion methods for checks (serialization and deserialization) from/to dict and `DQRule` has been centralized in `checks_serializer` module. It is now possible to convert checks defined in dict to `DQRule` and vice versa. Detailed documentation is available [here](https://databrickslabs.github.io/dqx/docs/reference/quality_rules/).
* Added support for saving checks to json file/workspace file
* Added validation of location when saving and loading checks
* Added validation of columns passed to `compare_datasets` check to make sure only simple expressions are allowed
* Added parameter to `validate-checks` cli command to enable/disable validating custom check functions (`validate-custom-check-functions`)

BREAKING CHANGES!

If you are loading or saving checks from a storage (file, workspace file, table, installation), you are affected. We are deprecating the the below methods. We are keeping the methods in the `DQEngine` but you should update your code as these methods will be removed in future versions.

* Loading checks to storage has been unified under `load_checks` method. The following methods have been removed from the `DQEngine`:
`load_checks_from_local_file`, `load_checks_from_workspace_file`, `load_checks_from_installation`, `load_checks_from_table`.

* Saving checks in storage has been unified under `load_checks` method. The following methods have been removed from the `DQEngine`:
`save_checks_in_local_file`, `save_checks_in_workspace_file`, `save_checks_in_installation`, `save_checks_in_table`.

The `save_checks` and  `load_checks` take `config` as a parameter, which determines the storage types used. The following storage configs are currently supported: 
* `FileChecksStorageConfig`: file in the local filesystem (YAML or JSON)
* `WorkspaceFileChecksStorageConfig`: file in the workspace (YAML or JSON)
* `TableChecksStorageConfig`: a table
* `InstallationChecksStorageConfig`: storage defined in the installation context, using either the `checks_table` or `checks_file` field from the run configuration.

* The `load_run_config` method has been moved to `config_loader.RunConfigLoader`, as it is not intended for direct use and falls outside the `DQEngine` core responsibilities.

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests


tests/integration/test_cli.py
```diff
@@ -121,8 +121,13 @@ def test_profiler(ws, setup_workflows, caplog):
 
     profile(installation_ctx.workspace_client, run_config=run_config.name, ctx=installation_ctx.workspace_installer)
 
-    checks = DQEngine(ws).load_checks_from_installation(
-        run_config_name=run_config.name, assume_user=True, product_name=installation_ctx.installation.product()
+    checks = DQEngine(ws).load_checks(
+        config=InstallationChecksStorageConfig(
+            location="installation",
```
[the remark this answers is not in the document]
mwojtyczka: fixed

docs/dqx/docs/reference/engine.mdx
```diff
@@ -45,25 +45,20 @@ The following table outlines the available methods of the `DQEngine` and their f
...
+| `load_checks`                                  | Loads quality rules from a storage. Multiple storage types are supported including tables, files or workspace files, installation-specific sources where the location is inferred automatically from run config  | `config`: Storage configuration for loading checks from a storage, i.e. `FileChecksStorageConfig`: file in the local filesystem (YAML or JSON); `WorkspaceFileChecksStorageConfig`: file in the workspace (YAML or JSON); `TableChecksStorageConfig`: a table; `InstallationChecksStorageConfig`: storage defined in the installation context, using either the `checks_table` or `checks_file` field from the run configuration. See more details below.                                                                                                                                                                                         | Yes (only with `FileChecksStorageConfig`) |
```
ghanse: Maybe just `Loads quality rules from storage.` or `Loads stored quality rules.`

Are the installation-specific sources only inferred when DQX is installed as a tool in the workspace?

docs/dqx/docs/reference/engine.mdx
```diff
@@ -45,25 +45,20 @@ The following table outlines the available methods of the `DQEngine` and their f
...
+| `save_checks`                                  | Saves quality rules to a storage. Multiple storage types are supported including tables, files or workspace files, installation-specific targets where the location is inferred automatically from run config    | `checks`: List of checks defined as dictionary; `config`: Storage configuration for saving checks in a storage, i.e. `FileChecksStorageConfig`: file in the local filesystem (YAML); `WorkspaceFileChecksStorageConfig`: file in the workspace (YAML); `TableChecksStorageConfig`: a table; `InstallationChecksStorageConfig`: storage defined in the installation context, using either the `checks_table` or `checks_file` field from the run configuration. See more details below.                                                                                                                                                            | Yes (only with `FileChecksStorageConfig`) |
```
ghanse: Maybe `Saves quality rules to storage.`

src/databricks/labs/dqx/config.py
ghanse, line 1: We have several implementations of config with a similar underlying attribute (`location`) and handler methods (`load` and `save`).

Can we instead create a single `ChecksStorageConfig` builder for users to instantiate the config?

mwojtyczka: The problem is that when we use a single `ChecksStorageConfig` the user has no idea which parameters to use and there is a high chance of misconfiguration. It's not a good design. 

I prefer dedicated configs for each storage type:
* It improves developer experience
* It avoids ambiguous or unused fields
* It makes the configuration self-describing and IDE-friendly
* It is easier to add new storage types because each type defines its own dedicated configuration. This makes the system more flexible and avoids the risk of future storage types having incompatible or conflicting parameters within a shared, generic config.

Separate configs requires slightly more code.

With a single config:
* Users don't have to know which config class to use. But also:
* Poor discoverability: Users don’t know which parameters apply unless they read the docs or guess
* Extra friction: They must specify a storage type and then figure out compatible fields
* Higher chance of misconfiguration: Users might mix parameters from different storage backends

Maybe we can introduce a generic config in addition. So user either use the generic config `ChecksStorageConfig` or a specific one?

docs/dqx/docs/reference/engine.mdx
mwojtyczka: updated

docs/dqx/docs/reference/engine.mdx
mwojtyczka: Updated. 

Yes the installation specific source will only work if the tool is installed in the workspace. There are tests for this: https://github.com/databrickslabs/dqx/pull/487/files#diff-332a010367eb7a5e6ce63b06e0fdc1da63a7cb5772bd3bff2898a41f8d1558d2L107

docs/dqx/docs/reference/engine.mdx
```diff
@@ -83,4 +76,18 @@ The 'Supports local execution' in the above table indicates which methods can be
...
+    * `location`: workspace file path (JSON or YAML)
+* `TableChecksStorageConfig` with fields:
+    * `location`: table fully qualified name
+    * `run_config_name`: (optional) run configuration name to load (use "default" if not provided)
+* `InstallationChecksStorageConfig` with fields:
+    * `location` (optional): automatically set based on the `checks_table` or `checks_file` field from the run configuration, depending on which is provided. If both are specified, `checks_table` takes precedence
+    * `run_config_name` (optional) - run configuration name to load (use "default" if not provided)
+    * `product_name`: name of the product/installation directory
+    * `assume_user`: if True, assume user installation
+
```
ghanse: It may be helpful to explicitly define each type:

```
* `FileChecksStorageConfig` can be used to save or load checks from the local filesystem
...
* `WorkspaceFileChecksStorageConfig` can be used to save or load checks from a workspace file
...
```

src/databricks/labs/dqx/storage.py
```diff
@@ -0,0 +1,307 @@
...
+        The returned checks can be used as input for `apply_checks_by_metadata` or
+        `apply_checks_by_metadata_and_split` functions.
+
+        :param config: configuration for loading checks, including the table location and run configuration name.
+        :return: list of dq rules or raise an error if checks file is missing or is invalid.
+        """
+
+    @abstractmethod
+    def save(self, checks: list[dict], config: T) -> None:
+        """Save quality rules to the target."""
```
ghanse: Do we want to enforce creation of these methods for any new handlers by raising a `NotImplementedError` from the abstract methods?

src/databricks/labs/dqx/storage.py
```diff
@@ -0,0 +1,307 @@
...
+        :param checks: list of dq rules to save
+        :param config: configuration for saving checks, including the file location and storage type.
+        """
+        workspace_dir = os.path.dirname(config.location)
+
+        logger.info(f"Saving quality rules (checks) to {config.location} in the workspace.")
+        self.ws.workspace.mkdirs(workspace_dir)
+        self.ws.workspace.upload(
+            config.location, yaml.safe_dump(checks).encode('utf-8'), format=ImportFormat.AUTO, overwrite=True
+        )
```
ghanse: We support reading the checks from JSON. Should we also support writing to JSON?

src/databricks/labs/dqx/storage.py
```diff
@@ -0,0 +1,307 @@
...
+        """
+        if not config.location:
+            raise ValueError("filepath must be provided")
+
+        try:
+            with open(config.location, 'w', encoding="utf-8") as file:
+                yaml.safe_dump(checks, file)
+        except FileNotFoundError:
+            msg = f"Checks file {config.location} missing"
+            raise FileNotFoundError(msg) from None
```
ghanse: Same comment about JSON support for writing checks applies here.

docs/dqx/docs/reference/engine.mdx
mwojtyczka: updated

src/databricks/labs/dqx/storage.py
mwojtyczka: I normally don't do it because python enforces that all subclasses must implement abstract methods. If a subclass does not implement them, python raises TypeError, preventing object creation. The only reason to add it would be for documentation and dev clarity but given this is internal class I don't think this is needed.

src/databricks/labs/dqx/storage.py
mwojtyczka: added

src/databricks/labs/dqx/storage.py
mwojtyczka: added

PR #495, opened 2025-07-30
Add new data quality checks for stale data checks

### LINKED ISSUE #373: [FEATURE]: Add freshness check

### Problem statement

Freshness check is useful to identify issues with the upstream systems.
It can detect stale data due to delayed pipelines, helping teams catch upstream issues early.

### Proposed Solution

Supported scenarios can include:
* Per-row reporting: Validate that timestamp column is not older than X minutes from now on a per-row basis
* Summary reporting only: [Completness] Validate that records arrive at least every X minutes, with a threshold for the expected number of rows per interval. Support configurable time window (e.g., past 12, 24, or 48 hours). Example: validate that, within each X-minute window, at least Y records are received.
* Summary reporting only: Validate the number of late events is not greater than threshold

### Additional Context

_No response_

karthik-ballullaya-db:
## Changes
Add new data quality checks for stale data checks

### Linked issues
Resolves #373 

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests

### DIFF
--- a/src/databricks/labs/dqx/check_funcs.py
+++ b/src/databricks/labs/dqx/check_funcs.py
@@ -611,6 +611,39 @@ def is_ipv4_address_in_cidr(column: str | Column, cidr_block: str) -> Column:
     )
 
 
+@register_rule("row")
+def is_stale_data(column: str | Column, max_age_minutes: int, base_timestamp: Column = F.current_timestamp()) -> Column:
+    """Checks whether the values in the timestamp column are not older than the specified number of minutes from the base timestamp column.
+
+    This is useful for identifying stale data due to delayed pipelines and helps catch upstream issues early.
+
+    :param column: column to check; can be a string column name or a column expression containing timestamp values
+    :param max_age_minutes: maximum age in minutes before data is considered stale
+    :param base_timestamp: (optional) set base timestamp column from which the stale check is calculated, if not provided uses current_timestamp()
+    :return: Column object for condition
+    """
+    col_str_norm, col_expr_str, col_expr = _get_norm_column_and_expr(column)
+
+    # Calculate the threshold timestamp (current time - max_age_minutes)
+    threshold_timestamp = F.from_unixtime(base_timestamp - F.expr("INTERVAL {max_age_minutes} MINUTES"))
+
+    # Check if the timestamp is older than the threshold (stale)
+    condition = col_expr < threshold_timestamp
+
+    return make_condition(
+        condition,
+        F.concat_ws(
+            "",
+            F.lit("Value '"),
+            col_expr.cast("string"),
+            F.lit(f"' in Column '{col_expr_str}' is older than {max_age_minutes} minutes from current time '"),
+            base_timestamp.cast("string"),
+            F.lit("'"),
+        ),
+        f"{col_str_norm}_is_stale",
+    )
+
+
 @register_rule("dataset")
 def is_unique(
     columns: list[str | Column],


src/databricks/labs/dqx/check_funcs.py
```diff
@@ -611,6 +611,39 @@ def is_ipv4_address_in_cidr(column: str | Column, cidr_block: str) -> Column:
     )
 
 
+@register_rule("row")
+def is_stale_data(column: str | Column, max_age_minutes: int, base_timestamp: Column = F.current_timestamp()) -> Column:
```
mwojtyczka: Should we call this `is_data_fresh` instead to indicate it's type of freshness checks?

docs/dqx/docs/reference/quality_rules.mdx
```diff
@@ -38,6 +38,7 @@ You can also define your own custom checks (see [Creating custom checks](#creati
...
+| `is_stale_data`                    | Checks whether the values in the input timestamp column are not older than the specified number of minutes from the base timestamp column. This is useful for identifying stale data due to delayed pipelines and helps catch upstream issues early.                                                                                                                                                                              | `column`: column to check (can be a string column name or a column expression); `max_age_minutes`: maximum age in minutes before data is considered stale; `base_timestamp`: optional base timestamp column from which the stale check is calculated, if not provided current_timestamp() function is used                                                                                                                                                                             |
```
mwojtyczka: Please add info that the column provided should be a timestamp col

mwojtyczka (changes requested):
Generally LGTM. But it only covers the first requirement from the issue: https://github.com/databrickslabs/dqx/issues/373

This should also cover the second requirement to validate that records arrive at least every X minutes, with a threshold for the expected number of rows per interval. Support configurable time window (e.g., past 12, 24, or 48 hours).

For the second requirement, dataset level check would be needed, similar to below:
```
import uuid
from typing import Callable
from pyspark.sql import DataFrame, Column, functions as F, Window

@register_rule("dataset"
def is_data_fresh_and_sufficient(
    column: str | Column,
    max_age_minutes: int,
    records_limit: int,
    window_minutes: int = 60,
    row_filter: str | None = None,
) -> tuple[Column, Callable[[DataFrame], DataFrame]]:
    """
    Dataset-level check for data freshness.

    Validates that for *every* sliding time window of size `window_minutes` within the
    last `max_age_minutes`, the count of records is at least `records_limit`.

    :param column: Timestamp column to check (str or Column).
    :param max_age_minutes: Time window length in minutes to look back from current time.
    :param records_limit: Minimum number of records expected per time window.
    :param window_minutes: Size of each sliding window in minutes (default 60).
    :param row_filter: Optional SQL filter expression applied before the check.
    :return: Tuple of (condition Column, apply callable) where:
        - condition Column signals data freshness violations,
        - apply callable applies the necessary columns to the DataFrame.
    """

    col_str_norm, col_expr_str, col_expr = _get_norm_column_and_expr(column)

    unique_str = uuid.uuid4().hex
    condition_col = f"__condition_{col_str_norm}_{unique_str}"
    count_col = f"__count_{col_str_norm}_{unique_str}"

    def apply(df: DataFrame) -> DataFrame:
        filter_condition = F.lit(True)
        if row_filter:
            filter_condition = filter_condition & F.expr(row_filter)

        cutoff = F.current_timestamp() - F.expr(f"INTERVAL {max_age_minutes} MINUTES")

        # Include cutoff condition in filter
        filter_condition = filter_condition & (col_expr >= cutoff)

        # Window specification for grouping records into time buckets
        # Aggregate counts per window
        windowed_counts = (
            df.where(filter_condition)
            .groupBy(F.window(col_expr, f"{window_minutes} minutes", f"{window_minutes} minutes").alias("time_window"))
            .agg(F.count("*").alias(count_col))
        )

        # Add condition column for windows where record count is below threshold
        windowed_counts = windowed_counts.withColumn(condition_col, F.col(count_col) < records_limit)

        return df

    condition = F.lit(True)
    return condition, apply
```

src/databricks/labs/dqx/check_funcs.py
karthik-ballullaya-db: done

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -611,6 +611,39 @@ def is_ipv4_address_in_cidr(column: str | Column, cidr_block: str) -> Column:
...
+    # Check if the timestamp is older than the threshold (stale)
+    condition = col_expr < threshold_timestamp
+
+    return make_condition(
+        condition,
+        F.concat_ws(
+            "",
+            F.lit("Value '"),
+            col_expr.cast("string"),
+            F.lit(f"' in Column '{col_expr_str}' is older than {max_age_minutes} minutes from base timestamp '"),
```
mwojtyczka, line 639: as part of this we can also report what is the value of base timestamp

tests/integration/test_row_checks.py
```diff
@@ -1221,3 +1222,44 @@ def test_is_ipv4_address_in_cidr(spark):
...
+        [
+            ["row1", datetime(2023, 1, 2, 0, 0, 0), datetime(2023, 1, 1).date()],
+            ["row2", datetime(2025, 1, 1, 0, 0, 0), datetime(2025, 1, 1).date()],
+            ["row3", None, None],
+            ["row4", datetime(2023, 12, 31, 23, 59, 59), datetime(2022, 12, 31).date()],
+        ],
+        input_schema,
+    )
+
+    reference_date = datetime(2024, 1, 1)
```
mwojtyczka, line 1244: Please also add negative test, where the date is None

tests/integration/test_row_checks.py
```diff
@@ -1221,3 +1222,44 @@ def test_is_ipv4_address_in_cidr(spark):
         checked_schema,
     )
     assert_df_equality(actual, expected, ignore_nullable=True)
+
+
+def test_is_data_fresh(spark):
```
mwojtyczka: Please also add this check here: https://github.com/databrickslabs/dqx/blob/main/tests/resources/all_row_checks.yaml

docs/dqx/docs/reference/quality_rules.mdx
```diff
@@ -38,6 +38,7 @@ You can also define your own custom checks (see [Creating custom checks](#creati
...
+| `is_data_fresh`                    | Checks whether the values in the input timestamp column are not older than the specified number of minutes from the base timestamp column. This is useful for identifying stale data due to delayed pipelines and helps catch upstream issues early.                                                                                                                                                                              | `column`: column of type timestamp/date to check (can be a string column name or a column expression); `max_age_minutes`: maximum age in minutes before data is considered stale; `base_timestamp`: optional base timestamp column from which the stale check is calculated, if not provided current_timestamp() function is used                                                                                                                                                      |
```
mwojtyczka: Please also add this to the examples using yaml and dqx classes
https://github.com/databrickslabs/dqx/blob/main/docs/dqx/docs/reference/quality_rules.mdx?plain=1#L57
https://github.com/databrickslabs/dqx/blob/main/docs/dqx/docs/reference/quality_rules.mdx?plain=1#L353

tests/integration/test_row_checks.py
karthik-ballullaya-db: The test might be transient as the current_timestamp might different due to small delay in function execution, or can I mock to set the value of current_timestamp?

mwojtyczka: true for the test, you need to provide the current_timestamp.

tests/integration/test_row_checks.py
```diff
@@ -1221,3 +1222,62 @@ def test_is_ipv4_address_in_cidr(spark):
...
+        input_schema,
+    )
+
+    reference_date = datetime(2024, 1, 1)
+    mins_threshold_b = 120
+    mins_threshold_c = 3600
+
+    actual = test_df.select(
+        is_data_fresh("b", mins_threshold_b, F.lit(reference_date)),
+        is_data_fresh("c", mins_threshold_c, F.lit(reference_date)),
```
mwojtyczka: can we also test with time as string?

docs/dqx/docs/reference/quality_rules.mdx
```diff
@@ -346,6 +347,23 @@ For brevity, the `name` field in the examples is omitted and it will be auto-gen
...
     - col5
+
+# is_data_fresh check with base_timestamp
+- criticality: error
+  check:
+    function: is_data_fresh
+    arguments:
+      column: col5
+      max_age_minutes: 15
+      base_timestamp: col6
```
mwojtyczka, line 358: example with timestamp provided as string would also be nice

docs/dqx/docs/reference/quality_rules.mdx
mwojtyczka: add pls for the `base_timestamp` that it can be a string column name or a column expression, or literal value F.lit(datetime(2024, 1, 1))

docs/dqx/docs/reference/quality_rules.mdx
mwojtyczka, line 358: would also be good to have an example with literal, where we provide the timestamp explicitly and not use a column

tests/integration/test_row_checks.py
karthik-ballullaya-db: time will be a datetime literal; column will be a string; Added a test case with column name

tests/integration/test_row_checks.py
karthik-ballullaya-db: added a seperate test case

docs/dqx/docs/reference/quality_rules.mdx
```diff
@@ -346,6 +347,32 @@ For brevity, the `name` field in the examples is omitted and it will be auto-gen
...
+# is_data_fresh check with base_timestamp as string
+- criticality: error
+  check:
+    function: is_data_fresh
+    arguments:
+      column: col5
+      max_age_minutes: 15
+      base_timestamp: col6
+
+# is_data_fresh check with base_timestamp as datetime object
```
mwojtyczka: this is not date time object, it's date time string

docs/dqx/docs/reference/quality_rules.mdx
```diff
@@ -673,6 +700,22 @@ checks = [
...
     ),
 
+    # is_data_fresh check with base_timestamp
+    DQRowRule(
+      criticality="error",
+      check_func=check_funcs.is_data_fresh,
+      column="col5",
+      check_func_kwargs={"column": "col5", "max_age_minutes": 15, "base_timestamp": "col6"}
+    ),
+
```
mwojtyczka, line 718: please add the same examples as in yaml, the literal example is missing

tests/integration/test_apply_checks.py
```diff
@@ -4276,7 +4276,7 @@ def test_apply_checks_all_row_checks_as_yaml_with_streaming(ws, make_schema, mak
 
     schema = (
         "col1: string, col2: int, col3: int, col4 array<int>, col5: date, col6: timestamp, "
-        "col7: map<string, int>, col8: struct<field1: int>, col9: string"
+        "col7: map<string, int>, col8: struct<field1: int>, col9: string, col10: timestamp"
```
mwojtyczka: do we need another column? there are already 2 timestamp cols

tests/integration/test_apply_checks.py
```diff
@@ -4401,9 +4407,10 @@ def test_apply_checks_all_checks_as_yaml(ws, spark):
     status = dq_engine.validate_checks(checks)
     assert not status.has_errors
 
+    curr_tmsp = datetime.now()
     schema = (
         "col1: string, col2: int, col3: int, col4 array<int>, col5: date, col6: timestamp, "
-        "col7: map<string, int>, col8: struct<field1: int>, col9: string"
+        "col7: map<string, int>, col8: struct<field1: int>, col9: string, col10: timestamp"
```
mwojtyczka: Same as above, do we need another column?

tests/integration/test_apply_checks.py
```diff
@@ -4290,6 +4290,7 @@ def test_apply_checks_all_row_checks_as_yaml_with_streaming(ws, make_schema, mak
                 {"key1": 1},
                 {"field1": 1},
                 "192.168.1.1",
+                datetime.date(),
```
mwojtyczka: it makes the test non-deterministic, pls reuse existing cols for testing.

docs/dqx/docs/reference/quality_rules.mdx
karthik-ballullaya-db: updated the comment

docs/dqx/docs/reference/quality_rules.mdx
karthik-ballullaya-db: done

tests/integration/test_apply_checks.py
karthik-ballullaya-db: This column is used to check the current_timestamp check of the is_data_fresh function, if not required I can remove it

tests/integration/test_apply_checks.py
karthik-ballullaya-db: This column is used to check the current_timestamp check of the is_data_fresh function, if not required I can remove it

tests/integration/test_apply_checks.py
mwojtyczka: yes, this is not required

tests/resources/all_row_checks.yaml
```diff
@@ -344,3 +344,29 @@
...
+- criticality: error
+  check:
+    function: is_data_fresh
+    arguments:
+      column: col6
+      max_age_minutes: 1440
+      base_timestamp: 2025-01-02 10:00:00
+
+# is_data_fresh check with base_timestamp as current_timestamp
+- criticality: error
```
mwojtyczka: not required, there are already 2 tests for this check

mwojtyczka (approved):
LGTM

PR #498, opened 2025-07-31
Feat add ipv6 address

### LINKED ISSUE #466: [FEATURE]: Add IPV6 Address Validation

### Problem statement

IPv6 address validation is a common task

### Proposed Solution

Add specialized check.

Should validate an IPV6 address

### Additional Context

_No response_

cornzyblack:
## Changes

Adds checks for the following:
   - Validate IPv6 addresses: `is_valid_ipv6_address`
   - Check if IPV6 is within a CIDR block: `is_ipv6_address_in_cidr`

### Linked issues

Resolves #466 

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests
- [x] added end-to-end tests


src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1623,20 +1753,83 @@ def _convert_ipv4_to_bits(ip_col: Column) -> Column:
...
+
+    unpadded_array = F.when(is_compressed, F.concat(left_hextets, zeros, right_hextets)).otherwise(F.split(ip_col, ":"))
+
+    # Option 1
+    return F.array_join(unpadded_array, ":").alias("ip_bits")
+
+    # Option 2
+    # return F.transform(
+    #     unpadded_array,
+    #     lambda hextet: F.lpad(hextet, 4, '0')
```
cornzyblack: Hi @mwojtyczka, so wanted to get your thoughts on whether it is okay to use higher-order functions in Spark (option 2). Option 1 still achieves the same purpose but I know UDFs are a no-go, but what of higher-order functions?

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1616,11 +1737,20 @@ def _validate_ref_params(
         raise ValueError("The number of columns to check against the reference columns must be equal.")
 
 
+def _does_not_match_pattern(column: Column, pattern: DQPattern) -> Column:
+    """
+    Internal function that returns a Boolean Column indicating if values
+    in the column do NOT match the given pattern.
+    """
+    _, _, col_expr = _get_normalized_column_and_expr(column)
```
mwojtyczka: Better to create a separate function for getting the col_expr, instead of ignoring fields. I think it could also be used in some other checks.

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: Higher order function should be ok. The computation should remain within the Catalyst optimizer and Spark's JVM execution context. Spark should use native logical plan. There is probably slight serialization overhead due to lambda.

It would be interesting to see what is the performance diff between the option 1 and 2.

tests/integration/test_row_checks.py
```diff
@@ -1221,3 +1222,128 @@ def test_is_ipv4_address_in_cidr(spark):
...
     )
     assert_df_equality(actual, expected, ignore_nullable=True)
+
+
+def test_col_is_valid_ipv6_address(spark):
+    schema_ipv6 = "a: string"
+
+    test_df = spark.createDataFrame(
+        [
+            ["192.170.01.1"],
```
mwojtyczka, line 1784: would be nice to test in addition edge cases:
```
    ["::1"],                        # Loopback address (compressed)
    ["::"],                         # Unspecified address (all zeros)
    ["1::"],                        # Trailing zero compression
    ["::1:2:3:4:5:6"],              # Leading zero compression
    ["2001:db8::1"],                 # Compressed zeros in middle
    ["2001:0db8:85a3:0000:0000:8a2e:0370:7334"],  # Full 8 segments with leading zeros
    ["ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff"],  # Max possible value
    ["2001:DB8::1"],                 # Uppercase letters are allowed
    ["fe80::1%eth0"],                # Link-local with interface zone
    ["fe80::%12"],                   # Link-local with numeric zone
    ["1:2:3:4:5:6:7:8"],             # Exactly 8 segments, no compression
    ["2001:db8::192.168.0.1"],       # IPv4-mapped IPv6 address
    ["::ffff:192.0.2.128"],          # IPv4-embedded IPv6 (special use)
```

tests/integration/test_apply_checks.py
```diff
@@ -5001,13 +5001,20 @@ def test_apply_checks_all_checks_using_classes(ws, spark):
             user_metadata={"tag1": "value7", "tag2": "033"},
```
mwojtyczka, line 5084: Please also add this check `all_row_checks.yaml` so that execution of it is checks in streaming.

mwojtyczka (changes requested):
Generally looking good, minor comments

docs/dqx/docs/reference/quality_rules.mdx
```diff
@@ -319,6 +321,21 @@ For brevity, the `name` field in the examples is omitted and it will be auto-gen
       column: col2
       cidr_block: '192.168.1.0/24'
 
+# is_valid_ipv6_address check
+- criticality: error
+  check:
+    function: is_valid_ipv6_address
+    arguments:
+      column: col2
```
```suggestion
      column: col7
```

docs/dqx/docs/reference/quality_rules.mdx
```diff
@@ -319,6 +321,21 @@ For brevity, the `name` field in the examples is omitted and it will be auto-gen
       column: col2
       cidr_block: '192.168.1.0/24'
 
+# is_valid_ipv6_address check
+- criticality: error
+  check:
+    function: is_valid_ipv6_address
+    arguments:
+      column: col7
```
```suggestion
      column: col10
```

docs/dqx/docs/reference/quality_rules.mdx
```diff
@@ -319,6 +321,21 @@ For brevity, the `name` field in the examples is omitted and it will be auto-gen
...
+    function: is_valid_ipv6_address
+    arguments:
+      column: col10
+
+# is_ipv6_address_in_cidr check
+- criticality: error
+  check:
+    function: is_ipv6_address_in_cidr
+    arguments:
+      column: col7
```
```suggestion
      column: col10
```

docs/dqx/docs/reference/quality_rules.mdx
```diff
@@ -681,6 +698,23 @@ checks = [
       }
     ),
 
+    # is_valid_ipv6_address check
+    DQRowRule(
+      criticality="error",
+      check_func=check_funcs.is_valid_ipv6_address,
+      column="col7"
```
```suggestion
      column="col10"
```

docs/dqx/docs/reference/quality_rules.mdx
```diff
@@ -681,6 +698,23 @@ checks = [
...
+      criticality="error",
+      check_func=check_funcs.is_valid_ipv6_address,
+      column="col10"
+    ),
+
+    # is_ipv6_address_in_cidr check
+    DQRowRule(
+      criticality="error",
+      check_func=check_funcs.is_ipv6_address_in_cidr,
+      column="col7",
```
```suggestion
      column="col10",
```

src/databricks/labs/dqx/llm/resources/yaml_checks_examples.yml
```diff
@@ -208,6 +208,17 @@
...
+- criticality: error
+  check:
+    function: is_valid_ipv6_address
+    arguments:
+      column: col2
+- criticality: error
+  check:
+    function: is_ipv6_address_in_cidr
+    arguments:
+      column: col7
```
mwojtyczka:
```suggestion
      column: col10
```

src/databricks/labs/dqx/llm/resources/yaml_checks_examples.yml
```diff
@@ -208,6 +208,17 @@
     arguments:
       column: col2
       cidr_block: 192.168.1.0/24
+- criticality: error
+  check:
+    function: is_valid_ipv6_address
+    arguments:
+      column: col2
```
mwojtyczka:
```suggestion
      column: col10
```

mwojtyczka (changes requested):
The peformance of the ipv6 within cidr seems to be pretty bad. I think it's because the higher order funciton forces row-by-row processing instead of vectorized operations.
Checking now.

cornzyblack:
> > > The peformance of the ipv6 within cidr seems to be pretty bad. I think it's because the higher order funciton forces row-by-row processing instead of vectorized operations. Checking now.
> > 
> > 
> > Just out of curiosity, is there a way you measured the performance of the change?
> 
> I run the integration tests and checked execution time, and also did some manual testing. Even for a few rows it takes more a couple of minutes for `is_ipv6_address_in_cidr` to finish. The `is_valid_ipv6_address` looks good. We need to improve this. I removed the higher order function but it did not really help.

Hmm, alright, I will look at any possible improvements I can make as well.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -19,13 +19,101 @@
...
+    + r":){0,4}"
+    + _IPV6_HEXTET
+    + r"?|"
+    + r"("
+    + _IPV6_HEXTET
+    + r":){3}:("
+    + _IPV6_HEXTET
+    + r":){0,3}"
+    + _IPV6_HEXTET
+    + r"?|"
```
mwojtyczka: @cornzyblack i extended the pattern as it was not catching all cases. I also added more test cases

mwojtyczka:
> > > > The peformance of the ipv6 within cidr seems to be pretty bad. I think it's because the higher order funciton forces row-by-row processing instead of vectorized operations. Checking now.
> > > 
> > > 
> > > Just out of curiosity, is there a way you measured the performance of the change?
> > 
> > 
> > I run the integration tests and checked execution time, and also did some manual testing. Even for a few rows it takes more a couple of minutes for `is_ipv6_address_in_cidr` to finish. The `is_valid_ipv6_address` looks good. We need to improve this. I removed the higher order function but it did not really help.
> 
> Hmm, alright, I will look at any possible improvements I can make as well.

thank you! if i would guess it's because of the array operations. I expected slow performance for this as it is complex validation and the more accurate we want to be the more expensive. But we should check how we can improve the perf.

mwojtyczka:
@cornzyblack were you able to investigate the issue. Do you need support?

cornzyblack:
> @cornzyblack were you able to investigate the issue. Do you need support? We plan a release tomorrow and was wandering if we will be able to include this.

Ah, sorry, not yet, but I will investigate this now

src/databricks/labs/dqx/llm/resources/yaml_checks_examples.yml
```diff
@@ -244,6 +244,17 @@
...
+- criticality: error
+  check:
+    function: is_valid_ipv6_address
+    arguments:
+      column: col10
+- criticality: error
+  check:
+    function: is_ipv6_address_in_cidr
+    arguments:
+      column: col10
```
mwojtyczka:
```suggestion
      column: col_ipv6
```

src/databricks/labs/dqx/llm/resources/yaml_checks_examples.yml
```diff
@@ -244,6 +244,17 @@
     arguments:
       column: col2
       cidr_block: 192.168.1.0/24
+- criticality: error
+  check:
+    function: is_valid_ipv6_address
+    arguments:
+      column: col10
```
mwojtyczka:
```suggestion
      column: col_ipv6
```

tests/integration/test_apply_checks.py
```diff
@@ -5081,6 +5094,34 @@ def test_apply_checks_all_checks_using_classes(ws, spark):
...
             check_func_kwargs={"cidr_block": "255.255.255.255/16"},
         ),
+        # is_valid_ipv6_address check
+        DQRowRule(
+            criticality="error",
+            check_func=check_funcs.is_valid_ipv6_address,
+            column="col10",
```
```suggestion
            column="col_ipv6",
```

tests/resources/all_row_checks.yaml
```diff
@@ -417,3 +417,18 @@
       column: col5
       max_age_minutes: 18000
       base_timestamp: col6
+
+# is_valid_ipv6_address check
+- criticality: error
+  check:
+    function: is_valid_ipv6_address
+    arguments:
+      column: col10
```
mwojtyczka:
```suggestion
      column: col_ipv6
```

tests/resources/all_row_checks.yaml
```diff
@@ -417,3 +417,18 @@
...
+    function: is_valid_ipv6_address
+    arguments:
+      column: col_ipv6
+
+# is_ipv6_address_in_cidr check
+- criticality: error
+  check:
+    function: is_ipv6_address_in_cidr
+    arguments:
+      column: col10
```
mwojtyczka:
```suggestion
      column: col_ipv6
```

tests/integration/test_apply_checks.py
```diff
@@ -5081,6 +5094,34 @@ def test_apply_checks_all_checks_using_classes(ws, spark):
...
+        DQRowRule(
+            criticality="error",
+            check_func=check_funcs.is_valid_ipv6_address,
+            column="col_ipv6",
+            user_metadata={"tag1": "value8", "tag2": "034"},
+        ),
+        DQRowRule(
+            criticality="error",
+            check_func=check_funcs.is_valid_ipv6_address,
+            column="col10",
```
```suggestion
            column="col_ipv6",
```

docs/dqx/docs/reference/quality_rules.mdx
```diff
@@ -782,6 +799,23 @@ checks = [
...
+      criticality="error",
+      check_func=check_funcs.is_valid_ipv6_address,
+      column="col10"
+    ),
+
+    # is_ipv6_address_in_cidr check
+    DQRowRule(
+      criticality="error",
+      check_func=check_funcs.is_ipv6_address_in_cidr,
+      column="col10",
```
```suggestion
      column="col_ipv6",
```

docs/dqx/docs/reference/quality_rules.mdx
```diff
@@ -373,6 +375,21 @@ For brevity, the `name` field in the examples is omitted and it will be auto-gen
       column: col2
       cidr_block: '192.168.1.0/24'
 
+# is_valid_ipv6_address check
+- criticality: error
+  check:
+    function: is_valid_ipv6_address
+    arguments:
+      column: col10
```
```suggestion
      column: col_ipv6
```

docs/dqx/docs/reference/quality_rules.mdx
```diff
@@ -373,6 +375,21 @@ For brevity, the `name` field in the examples is omitted and it will be auto-gen
...
+    function: is_valid_ipv6_address
+    arguments:
+      column: col_ipv6
+
+# is_ipv6_address_in_cidr check
+- criticality: error
+  check:
+    function: is_ipv6_address_in_cidr
+    arguments:
+      column: col10
```
```suggestion
      column: col_ipv6
```

docs/dqx/docs/reference/quality_rules.mdx
```diff
@@ -782,6 +799,23 @@ checks = [
       }
     ),
 
+    # is_valid_ipv6_address check
+    DQRowRule(
+      criticality="error",
+      check_func=check_funcs.is_valid_ipv6_address,
+      column="col10"
```
```suggestion
      column="col_ipv6"
```

tests/integration/test_apply_checks.py
```diff
@@ -5081,6 +5094,34 @@ def test_apply_checks_all_checks_using_classes(ws, spark):
...
+            criticality="error",
+            check_func=check_funcs.is_valid_ipv6_address,
+            column="col_ipv6",
+            user_metadata={"tag1": "value8", "tag2": "034"},
+        ),
+        # is_ipv6_address_in_cidr check
+        DQRowRule(
+            criticality="error",
+            check_func=check_funcs.is_ipv6_address_in_cidr,
+            column=F.col("col10"),
```
```suggestion
            column=F.col("col_ipv6"),
```

tests/integration/test_apply_checks.py
```diff
@@ -5081,6 +5094,34 @@ def test_apply_checks_all_checks_using_classes(ws, spark):
...
+            criticality="error",
+            check_func=check_funcs.is_ipv6_address_in_cidr,
+            column=F.col("col_ipv6"),
+            user_metadata={"tag1": "value9", "tag2": "035"},
+            check_func_kwargs={"cidr_block": "2001:db8:85a3:8d3:1319:8a2e:3.112.115.68/64"},
+        ),
+        DQRowRule(
+            criticality="error",
+            check_func=check_funcs.is_ipv6_address_in_cidr,
+            column=F.col("col10"),
```
```suggestion
            column=F.col("col_ipv6"),
```

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -746,6 +834,97 @@ def is_ipv4_address_in_cidr(column: str | Column, cidr_block: str) -> Column:
...
+
+    This rule checks for four accepted IPv6 formats:
+      - Fully uncompressed (e.g. '2001:0db8:0000:0000:0000:0000:0000:0001')
+      - Compressed (e.g. '2001:db8::1')
+      - Loopback ('::1')
+      - Unspecified ('::')
+
+    A value fails the check if it does not match **any** of these valid IPv6 patterns.
+
+    :param column: column to check; can be a string column name or a column expression
```
mwojtyczka: please convert all docstrings to google style docs. We recently converted all docstrings, as there is better support for api generation from these

mwojtyczka:
@cornzyblack  For ipv6 within cidr could also verify the performance of python ipaddress executed as pandas udf. We have an example [here](https://github.com/databrickslabs/dqx/blob/main/src/databricks/labs/dqx/pii/pii_detection_funcs.py).

cornzyblack:
> @cornzyblack For ipv6 within cidr you could also verify the performance of python ipaddress executed as pandas udf. We have an example [here](https://github.com/databrickslabs/dqx/blob/main/src/databricks/labs/dqx/pii/pii_detection_funcs.py).

Thanks I will check the performance with `pandas_udf`

cornzyblack:
Hi @mwojtyczka I have created an e2e flow using `udfs` and the `ipaddress `module, for IPV6 address validation and CIDR, similar to `pii_detection_notebook`. 

If it looks good to you, I will proceed to remove my old implementation from `check_funcs.py`and refactor the unit tests to import the new functions, and then add it as a demo as well.

cornzyblack:
updates:
- Created demos and removed the previous IPv6 implementations
- Refactored the e2e test to reduce code duplication

demos/dqx_demo_ipv6address_validation.py
```diff
@@ -0,0 +1,106 @@
+# Databricks notebook source
```
mwojtyczka: why do we need dedicated demo for ipv6 validation?
we don't want to have demo for every check type

src/databricks/labs/dqx/ipaddress/ipaddress_funcs.py
```diff
@@ -0,0 +1,205 @@
+import re
```
mwojtyczka: Move this to the existing `check_funcs` module. We don't need a separate module.

src/databricks/labs/dqx/ipaddress/ipaddress_funcs.py
```diff
@@ -0,0 +1,205 @@
...
+        True if the string is a valid CIDR block, False otherwise.
+    """
+    try:
+        ipaddress.IPv6Network(cidr, strict=False)
+        return True
+    except (ipaddress.AddressValueError, ipaddress.NetmaskValueError):
+        return False
+
+
+def _validate_environment(func_name: str) -> None:
```
mwojtyczka: We don't need this validation for ipv6 validation since we are using standard python library here.

src/databricks/labs/dqx/ipaddress/ipaddress_funcs.py
```diff
@@ -0,0 +1,205 @@
...
+
+    Returns:
+        pd.Series: A Series indicating whether each value is a valid IPv6 address.
+    """
+    warnings.warn(
+        "IPv6 Address validation uses pandas user-defined functions which may degrade performance. "
+        "Sample or limit large datasets when running IPV6 address validation.",
+    )
+
+    _validate_environment("is_valid_ipv6_address")
```
mwojtyczka: Remove, this is not needed

src/databricks/labs/dqx/ipaddress/ipaddress_funcs.py
```diff
@@ -0,0 +1,205 @@
...
+        UserWarning,
+    )
+
+    if not cidr_block:
+        raise ValueError("'cidr_block' must be a non-empty string.")
+
+    if not _is_valid_cidr_block(cidr_block):
+        raise ValueError(f"CIDR block '{cidr_block}' is not a valid IPv6 CIDR block.")
+
+    _validate_environment("is_ipv6_address_in_cidr")
```
mwojtyczka: Remove, this is not needed

src/databricks/labs/dqx/ipaddress/ipaddress_funcs.py
```diff
@@ -0,0 +1,205 @@
...
+def _is_valid_ipv6(ip_address: str) -> bool:
+    """Validate if the string is a valid IPv6 address."""
+    try:
+        ipaddress.IPv6Address(ip_address)
+        return True
+    except ipaddress.AddressValueError:
+        return False
+
+
+def _is_ipv6_check(ip_address: str) -> bool:
```
mwojtyczka:
```suggestion
def _is_valid_ipv6(ip_address: str) -> bool:
```

src/databricks/labs/dqx/ipaddress/ipaddress_funcs.py
```diff
@@ -0,0 +1,205 @@
...
+    """
+    Check if a string is a valid IPv6 address.
+
+    Args
+        ip_address: The string to check.
+
+    Returns
+        True if the string is a valid IPv6 address, False otherwise.
+    """
+    return _is_valid_ipv6(ip_address)
```
mwojtyczka: move the logic of `_is_valid_ipv6` here

tests/e2e/notebooks/ipaddress_validation_notebook.py
```diff
@@ -0,0 +1,761 @@
...
+
+# COMMAND ----------
+
+from databricks.labs.dqx.ipaddress.ipaddress_funcs import is_ipv6_address_in_cidr, is_valid_ipv6_address
+from chispa import assert_df_equality
+
+# COMMAND ----------
+# DBTITLE 1,test_is_valid_ipv6_address
+
+def test_is_valid_ipv6_address():
```
mwojtyczka: can this be implemented as standard integration test?

mwojtyczka (changes requested):

mwojtyczka:
> o remove m

Thank you! but why do we need e2e test for ipv6 validation?

demos/dqx_demo_ipv6address_validation.py
cornzyblack: I followed the pattern I saw with `does_not_contain_pii`, which has its own module, demo, and e2e. That's why I did the same for IPv6. But no problem, I'll switch it over to check_funcs and skip the separate `demo/e2e`

mwojtyczka: we needed this for pii detection because we install nlp models that are not available in the executor. For ipv6 we use standard python library that is available without additional installations.

mwojtyczka: How is this new ipv6 implementation perform compared to previous implementation?

cornzyblack: so just by timing it and plotting it on a serverless cluster the previous implementation seems faster, however I am also going ro run the same with pytest-benchmark.

I will add the notebook i used the next commit as well (will remove it later). 

This was just by timing the separate runs using `timeit` and not with `pytest-benchmark` (in the picture `ipmodule` is the new implementation)

benchmark-ipv6address-timeit

mwojtyczka: @cornzyblack  I run it in apply checks and it finishes within limited time, previously it was hanging for many minutes even for a few rows. It looks much better now. Maybe running this function in isolation now is slower but overall it looks better.

Please remove the demo for this, we don't need it.

tests/e2e/notebooks/ipaddress_validation_notebook.py
mwojtyczka: @cornzyblack i tried to run it as standard integration tests and it failed, complaining about python version mismatch, so let's keep this e2e test as you have it

`pyspark.errors.exceptions.connect.SparkConnectGrpcException: Python versions in the Spark Connect client and server are different. To execute user-defined functions, client and server should have the same minor Python version. Please update the Python version in the client. You can check the current Python version on the server by consulting Databricks release notes. Release notes for cluster compute can be found at https://docs.databricks.com/en/release-notes/runtime/index.html and release notes for serverless can be found at https://docs.databricks`

tests/e2e/test_run_demos.py
```diff
@@ -112,6 +112,28 @@ def test_run_dqx_demo_pii_detection(make_notebook, make_job, library_ref):
     logging.info(f"Job run {run.run_id} completed successfully for dqx_demo_pii_detection")
 
 
+def test_run_dqx_demo_ipv6address_validation(make_notebook, make_job, library_ref):
```
mwojtyczka: remove, we don't need separate demo for ipv6 validation

tests/e2e/notebooks/ipaddress_validation_notebook.py
cornzyblack: Yes i faced a smilar error, and that was why I put the `_validate_environment` as I thought that was the reason why `pii` had it. I noticed in the integration tests for `pii` it was checking for `pytest error` and not really doing the main checks.

I will live them in the e2e tests then

image

mwojtyczka: yes it is there in pii detection because we have models as dependencies and there is limit to the size of dependencies. We don't have this issue in ipv6 validation

tests/integration/test_row_checks.py
```diff
@@ -1222,12 +1435,346 @@ def test_is_ipv4_address_in_cidr(spark):
...
+        is_ipv4_address_in_cidr("a", "224.0.0.0/4"),  # Multicast
+    )
+
+    checked_schema = (
+        "a_is_not_ipv4_in_cidr: string, "
+        "a_is_not_ipv4_in_cidr: string, "
+        "a_is_not_ipv4_in_cidr: string, "
+        "a_is_not_ipv4_in_cidr: string, "
+        "a_is_not_ipv4_in_cidr: string, "
+        "a_is_not_ipv4_in_cidr: string"
```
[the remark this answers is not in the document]
mwojtyczka: yes, please make the columns unique

docs/dqx/docs/reference/quality_checks.mdx
```diff
@@ -43,6 +43,8 @@ You can also define your own custom checks (see [Creating custom checks](#creati
...
+| `is_valid_ipv6_address`            | Checks whether the values in the input column have valid IPv6 address format.                                                                                                                                                                                                                                                                                                                                                     | `column` to check (can be a string column name or a column expression)                                                                                                                                                                                                                                                                                                                                                                                                                               |
```
mwojtyczka, line 46: we would need to add performance benchmark for these. But this must be done from a branch so that the benchmark is added automatically. I will create a follow up for this

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -2004,23 +2108,98 @@ def _convert_ipv4_to_bits(ip_col: Column) -> Column:
...
+    """
+
+    @F.pandas_udf("boolean")  # type: ignore[call-overload]
+    def handler(ipv6_column: pd.Series, cidr_column: pd.Series) -> pd.Series:
+        return ipv6_column.combine(cidr_column, _ipv6_in_cidr)
+
+    return handler
+
+
+def _is_valid_cidr_block(cidr: str) -> bool:
```
```suggestion
def _is_valid_ipv6_cidr_block(cidr: str) -> bool:
```

docs/dqx/docs/reference/quality_checks.mdx
cornzyblack: On this, I finally re-ran it wth pytest-benchmark with a different dataset on serverless, and the new one does seem to be faster for is_valid_ipv6.

`tests/perf/conftest.py `

```python
@pytest.fixture
def generated_ipv6_df(spark, rows=DEFAULT_ROWS):
    IPV6_SCHEMA_STR = (
        "col1_ipv6_u_upper: string, "
        "col2_ipv6_u_lower: string, "
        "col3_ipv6_c_min1: string, "
        "col4_ipv6_c_r3: string, "
        "col5_ipv6_c_l3: string, "
        "col6_ipv6_c_mid1: string, "
        "col7_ipv6_c_mid4: string, "
        "col8_ipv6_u_prefix: string"
    )
    schema = _parse_datatype_string(IPV6_SCHEMA_STR)
    spec = (
        dg.DataGenerator(spark, rows=DEFAULT_ROWS, partitions=DEFAULT_PARTITIONS)
        .withSchema(schema)
        .withColumnSpec("col1_ipv6_u_upper", template=r'XXXX:XXXX:XXXX:XXXX:XXXX:XXXX:XXXX:XXXX')
        .withColumnSpec("col2_ipv6_u_lower", template=r'xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx')
        .withColumnSpec("col3_ipv6_c_min1", template='::')
        .withColumnSpec("col4_ipv6_c_r3", template='::XXXX:XXXX:XXXX')
        .withColumnSpec("col5_ipv6_c_l3", template='XXXX:XXXX:XXXX::')
        .withColumnSpec("col6_ipv6_c_mid1", template='XXXX::XXXX')
        .withColumnSpec("col7_ipv6_c_mid4", template='XXXX:XXXX::XXXX')
        .withColumnSpec("col8_ipv6_u_prefix", template='2001:0DB8:XXXX:XXXX:XXXX:XXXX:XXXX:XXXX')
    )
    return spec.build()
```

`tests/perf/test_apply_checks.py `

```python
@pytest.mark.parametrize(
    "column",
    [
        "col1_ipv6_u_upper",
        "col2_ipv6_u_lower",
        "col3_ipv6_c_min1",
        "col4_ipv6_c_r3",
        "col5_ipv6_c_l3",
        "col6_ipv6_c_mid1",
        "col7_ipv6_c_mid4",
        "col8_ipv6_u_prefix",
    ],
)
@pytest.mark.benchmark(group="test_benchmark_is_valid_ipv6_address.previous")
def test_benchmark_prev_is_valid_ipv6_address(benchmark, ws, generated_ipv6_df, column):
    dq_engine = DQEngine(workspace_client=ws, extra_params=EXTRA_PARAMS)
    checks = [
        DQRowRule(
            name=f"{column}_prev_is_valid_ipv6_address",
            criticality="warn",
            check_func=check_funcs.is_valid_ipv6_address,
            column=column,
        ),
    ]
    benchmark.group += f" {column}"
    checked = dq_engine.apply_checks(generated_ipv6_df, checks)


@pytest.mark.parametrize(
    "column",
    [
        "col1_ipv6_u_upper",
        "col2_ipv6_u_lower",
        "col3_ipv6_c_min1",
        "col4_ipv6_c_r3",
        "col5_ipv6_c_l3",
        "col6_ipv6_c_mid1",
        "col7_ipv6_c_mid4",
        "col8_ipv6_u_prefix",
    ],
)
@pytest.mark.benchmark(group="test_benchmark_is_valid_ipv6_address.current")
def test_benchmark_current_is_valid_ipv6_address(benchmark, ws, generated_ipv6_df, column):
    dq_engine = DQEngine(workspace_client=ws, extra_params=EXTRA_PARAMS)
    checks = [
        DQRowRule(
            name=f"{column}_current_is_valid_ipv6_address",
            criticality="warn",
            check_func=ipmodule.is_valid_ipv6_address,
            column=column,
        ),
    ]
    benchmark.group += f" {column}"
    checked = dq_engine.apply_checks(generated_ipv6_df, checks)
```

running in cli 

```bash
 pytest -q \                  
  tests/perf/test_apply_checks.py::test_benchmark_prev_is_valid_ipv6_address \
  tests/perf/test_apply_checks.py::test_benchmark_current_is_valid_ipv6_address \
  --benchmark-warmup=on --benchmark-min-time=1.0 --benchmark-min-rounds=5 \
  --benchmark-json=out_ipv6_benchmark.json
```
generates

[out_ipv6_benchmark.json](https://github.com/user-attachments/files/22367573/out_ipv6_benchmark.json)

And plotting:

image

mwojtyczka: @cornzyblack awesome. I found the issue with the databricks connect python version mismatch. I updated the python version to 3.11. The `databricks-connect~=15.4` that we use for testing expects python 3.11, and we used 3.10. I updated the `pyproject.toml`. I also updated the udf to be self contained. The pandas UDFs were referencing external functions (_is_ipv6_check, _ipv6_in_cidr) that caused Spark to try to serialize the entire module context, including databricks.labs imports.

Please remove the e2e tests and notebook and revert tests back to the version just before the switch to udfs. We can execute the check as part of normal integration tests now.

mwojtyczka: @cornzyblack I will create a branch for this change so that we can also add performance tests. Can you please do a PR to this Branch please?

cornzyblack: yes sure!

> @cornzyblack I will create a branch for this change so that we can also add performance tests. Can you please do a PR to this Branch please?

cornzyblack: > @cornzyblack awesome. I found the issue with the databricks connect python version mismatch. I updated the python version to 3.11. The `databricks-connect~=15.4` that we use for testing expects python 3.11, and we used 3.10. I updated the `pyproject.toml`. I also updated the udf to be self contained. The pandas UDFs were referencing external functions (_is_ipv6_check, _ipv6_in_cidr) that caused Spark to try to serialize the entire module context, including databricks.labs imports.
> 
> Please remove the e2e tests and notebook and revert tests back to the version just before the switch to udfs. We can execute the check as part of normal integration tests now.

ahh okay, thanks will revert now!

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -2004,23 +2107,100 @@ def _convert_ipv4_to_bits(ip_col: Column) -> Column:
...
     """
-    return F.rpad(F.substring(ip_bits, 1, prefix_length), 32, "0")
+
+    @F.pandas_udf("boolean")  # type: ignore[call-overload]
+    def _is_valid_ipv6_address_udf(column: pd.Series) -> pd.Series:
+        # Self-contained validation logic to avoid serialization issues
+        def is_valid_ipv6_local(ip_str):
+            if pd.isna(ip_str) or ip_str is None:
+                return False
+            import ipaddress  # pylint: disable=import-outside-toplevel
```
cornzyblack, line 2147: Had to add `disable=import-outside-toplevel` if not `make lint` complains about the import not being on outer level

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -2004,23 +2107,100 @@ def _convert_ipv4_to_bits(ip_col: Column) -> Column:
...
+
+def _is_valid_ipv6_cidr_block(cidr: str) -> bool:
+    """Validate if the string is a valid IPv6 CIDR block.
+
+    Args:
+        cidr: The CIDR block string to validate.
+    Returns:
+        True if the string is a valid CIDR block, False otherwise.
+    """
+    import ipaddress  # pylint: disable=import-outside-toplevel
```
cornzyblack, line 2198: Had to add disable=import-outside-toplevel if not make lint complains about the import not being on outer level.

PR #525, opened 2025-08-18
Feat add performance benchmark

### LINKED ISSUE #56: [FEATURE]: Add Performance benchmark

### Problem statement

There are no performance benchmarks available for profiler and quality checks. 
DQX is using native spark apis for quality checking so the impact is minimal. Nevertheless, it would be useful to have performance benchmarks.

### Proposed Solution

Create performance benchmarks on how the performance is impacted by running profiler and applying quality rules on bigger data sets. 
This should be written as performance tests that can be run weekly or monthly.

### Additional Context

_No response_

cornzyblack:
### Linked issues

Resolves #56 

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests
- [ ] added end-to-end tests


pyproject.toml
```diff
@@ -95,6 +99,7 @@ test        = "pytest tests/unit/ -n 10 --cov --cov-report=xml:coverage-unit.xml
 coverage    = "pytest tests/ -n 10 --cov --cov-report=html --timeout 600 --durations 20"
 integration = "pytest tests/integration/ -n 10 --cov --cov-report=xml --timeout 600 --durations 20"
 e2e         = "pytest tests/e2e/ -n 10 --cov --cov-report=xml --timeout 600 --durations 20"
+perf         = "pytest tests/perf/ -n 10 --cov --cov-report=xml --timeout 600 --durations 20"
```
mwojtyczka: Please add the execution to the `.github/workflows/nightly.yml`. Ideally, we should also have it in the `.github/workflows/acceptance.yml` so that it runs on every PR, but only we it is not too heavy.

mwojtyczka (commented):
The approach looks good. Yes we should execute all checks all at once from the yaml definitions and each individual check separately.

We will need to compare the execution time against the baseline. The idea is to create a baseline that we can compare against the current run.

As part of ci/cd in github action we would need to run: 
pytest --benchmark-compare=baseline

The baseline (json file) should be stored in the repo. It can be generated with: 
pytest --benchmark-save=baseline

If you need help generating it, pls let me know.

.github/workflows/nightly.yml
```diff
@@ -162,3 +162,52 @@ jobs:
           ARM_CLIENT_ID: ${{ secrets.ARM_CLIENT_ID }}
           ARM_TENANT_ID: ${{ secrets.ARM_TENANT_ID }}
           DATABRICKS_SERVERLESS_COMPUTE_ID: ${{ env.DATABRICKS_SERVERLESS_COMPUTE_ID }}
+
+  benchmark:
+    environment: tool
+    runs-on: larger
+    env:
+      DATABRICKS_SERVERLESS_COMPUTE_ID: auto
```
mwojtyczka: it's probably better to use a standard cluster for more deterministic and repeatable performance results. For this just remove DATABRICKS_SERVERLESS_COMPUTE_ID

.github/workflows/nightly.yml
```diff
@@ -162,3 +162,52 @@ jobs:
...
+            echo "missing=true"  >> $GITHUB_OUTPUT
+          else
+            echo "missing=false" >> $GITHUB_OUTPUT
+            echo "file=${files[0]}" >> $GITHUB_OUTPUT
+          fi
+
+      # Create baseline if missing (first run)
+      - name: Generate initial baseline
+        if: steps.detect.outputs.missing == 'true'
+        run: hatch run pytest tests/perf/test_apply_checks.py --benchmark-save=benchmark
```
mwojtyczka: this will need to be run similar to other tests to setup the auth correctly:
`uses: databrickslabs/sandbox/acceptance@acceptance/v0.4.4`

You will need to create and provide` codegen_path: tests/perf/.codegen_baseline.json`
```
{
  "version": {
    "src/databricks/labs/dqx/__about__.py": "__version__ = \"$VERSION\""
  },
  "toolchain": {
    "required": ["python3", "hatch"],
    "pre_setup": ["hatch env create"],
    "prepend_path": ".venv/bin",
    "acceptance_path": "tests/perf/baseline.py"
  }
}
```
The problem is that this only points to a path containing tests, and it does not provide more customizations. Maybe wrapping this inside another test would work as below.

baseline.py

```
def test_create_apply_checks_benchmark_baseline():
    # Equivalent to:
    # hatch run pytest tests/perf/test_apply_checks.py --benchmark-save=benchmark
    result = pytest.main([
        "tests/perf/test_apply_checks.py",
        "--benchmark-save=benchmark"
    ])
```

.github/workflows/nightly.yml
```diff
@@ -162,3 +162,52 @@ jobs:
...
+          DATABRICKS_SERVERLESS_COMPUTE_ID: ${{ env.DATABRICKS_SERVERLESS_COMPUTE_ID }}
+
+      - name: Commit baseline to repo
+        if: steps.detect.outputs.missing == 'true'
+        run: |
+          git config user.name  "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git add .benchmarks/*/0001_*.json
+          git commit -m "Add pytest-benchmark baseline from Nightly build" || true
+          git push
```
mwojtyczka: this will not work as we cannot push to the main directly. Nightly works on the main branch. 

We could also just print the baseline. But I think we just need to generate the baseline once, we can do this locally. I can help. If we need to bump any timing we could do this directly in the baseline file.

This nightly job will also need to compare the perf test results against the baseline.

.github/workflows/nightly.yml
cornzyblack: Sure, will remove that now

.github/workflows/nightly.yml
cornzyblack: What about if we use [PYTEST_ADDOPTS](https://docs.pytest.org/en/7.1.x/reference/reference.html#envvar-PYTEST_ADDOPTS) `PYTEST_ADDOPTS: "--benchmark-save=benchmark"` and pass it as an environmental variable. It can allow other command-line flags to be passed to pytest.

image

.github/workflows/nightly.yml
cornzyblack: Ah, okay, I think generating the first baseline locally could work, but I'm concerned about where the baseline is stored. The path is machine-dependent, so it varies depending on the system it's run on (I'm still researching how to customize this).

For example, on my machine, it's stored under: `dqx/.benchmarks/Darwin-CPython-3.10-64bit`

If someone else runs it on a different system, the baseline folder will be named differently. That means when comparing, there’s a chance the runner won’t find the baseline because the directory structure won’t match. 

I tried modifying the baseline path locally and I tried to compare again and I got this 
image

cornzyblack: We can do something like `pytest tests/perf/test_apply_checks.py --benchmark-storage='.benchmarkstest' --benchmark-autosave`. But it still appends the `machine_id` (a combination of the system, Python implementation, python version, and machine architecture) to the storage path.

Running this
`pytest tests/perf/test_apply_checks.py --benchmark-storage='.benchmarkstest' --benchmark-autosave`

still creates a folder `.benchmarkstest/Darwin-CPython-3.10-64bit` on my machine

.github/workflows/nightly.yml
mwojtyczka: let me test this

.github/workflows/nightly.yml
mwojtyczka: let me create dummy PR to generate the file then so that we get the right path

ghanse: Have we tried using `--benchmark-json`?

```
pytest --benchmark-json=./my_benchmarks/my_file.json
```

tests/perf/conftest.py
```diff
@@ -0,0 +1,96 @@
...
+
+@pytest.fixture
+def table_name(make_schema, make_random):
+    catalog = "main"
+    schema = make_schema(catalog_name=catalog).name
+    return f"{catalog}.{schema}.{make_random(6).lower()}"
+
+
+@pytest.fixture
+def generated_df(spark):
```
ghanse, line 68: Can we update the fixture to accept a number of input rows and columns, then parametrize the input dataset size and number of checks in the`test_apply_checks.py` methods?

tests/perf/test_apply_checks.py
ghanse, line 1: We currently benchmark checks on individual columns and on the entire dataset.

Should we also benchmark the performance when applying a check to multiple columns using `DQForEachColRule`? There may be a performance difference when applying 1 rule, 10 rules, or 100 rules.

The number of checks could be parametrized if we modify `generated_df` to create dummy columns and use `DQForEachColRule`.

.github/workflows/nightly.yml
cornzyblack: `pytest --benchmark-json=./my_benchmarks/my_file.json` does save the results to ./my_benchmarks/my_file.json.

But when I try to use this file as a baseline for comparison, I can't find a flag in the docs that explains how to run pytest and compare new results against the new baseline JSON file `./my_benchmarks/my_file.json`.

tests/perf/conftest.py
cornzyblack: yeah sure, I can update that

tests/perf/test_apply_checks.py
```diff
@@ -0,0 +1,493 @@
+from databricks.labs.dqx.engine import DQEngine
+from datetime import datetime
+from databricks.labs.dqx.rule import ExtraParams, DQRowRule, DQDatasetRule
```
mwojtyczka:
```suggestion
from databricks.labs.dqx.rule import DQRowRule, DQDatasetRule
from databricks.labs.dqx.config import ExtraParams
```

tests/perf/conftest.py
```diff
@@ -0,0 +1,97 @@
+import logging
+from datetime import datetime
+from pathlib import Path
+import pytest
+import yaml
+import dbldatagen as dg  # type: ignore[import-untyped]
+from pyspark.sql.types import _parse_datatype_string
+
+from databricks.labs.dqx.engine import DQEngine
+from databricks.labs.dqx.rule import ExtraParams
```
mwojtyczka:
```suggestion
from databricks.labs.dqx.config import ExtraParams
```

.github/workflows/nightly.yml
mwojtyczka: yes it seems using json would require manual comparison.

tests/perf/conftest.py
```diff
@@ -0,0 +1,97 @@
...
+
+SCHEMA_STR = (
+    "col1: int, col2: int, col3: int, col4: array<int>, "
+    "col5: date, col6: timestamp, col7: map<string, int>, "
+    'col8: struct<field1: int>, col9: string'
+)
+
+REF_SCHEMA_STR = "ref_col1: int, ref_col2: int, ref_col3: int"
+
+RUN_TIME = datetime(2025, 1, 1, 0, 0, 0)
```
mwojtyczka:
```suggestion
RUN_TIME = datetime(2025, 1, 1, 0, 0, 0, 0, tzinfo=timezone.utc)
```

tests/perf/conftest.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+@pytest.fixture
+def extra_params():
+    return ExtraParams(run_time=RUN_TIME)
```
mwojtyczka:
```suggestion
    return ExtraParams(run_time=RUN_TIME.isoformat())
```

tests/perf/test_apply_checks.py
```diff
@@ -0,0 +1,494 @@
+from databricks.labs.dqx.engine import DQEngine
+from datetime import datetime
+from databricks.labs.dqx.rule import DQRowRule, DQDatasetRule
+from databricks.labs.dqx.config import ExtraParams
+import pytest
+from databricks.labs.dqx import check_funcs
+from tests.perf.conftest import ROWS
+
+RUN_TIME = datetime(2025, 1, 1, 0, 0, 0, 0)
```
mwojtyczka:
```suggestion
RUN_TIME = datetime(2025, 1, 1, 0, 0, 0, 0, tzinfo=timezone.utc)
```

tests/perf/test_apply_checks.py
```diff
@@ -0,0 +1,494 @@
...
+EXTRA_PARAMS = ExtraParams(run_time=RUN_TIME)
```
mwojtyczka:
```suggestion
EXTRA_PARAMS = ExtraParams(run_time=RUN_TIME.isoformat())
```

tests/perf/conftest.py
```diff
@@ -0,0 +1,97 @@
+import logging
+from datetime import datetime
```
mwojtyczka:
```suggestion
from datetime import datetime, timezone
```

tests/perf/test_apply_checks.py
```diff
@@ -0,0 +1,494 @@
+from databricks.labs.dqx.engine import DQEngine
+from datetime import datetime
```
mwojtyczka:
```suggestion
from datetime import datetime, timezone
```

tests/perf/conftest.py
```diff
@@ -0,0 +1,97 @@
...
+
+logger = logging.getLogger(__name__)
+
+ROWS = 10_000_000
+PARTITIONS = 8
+
+SCHEMA_STR = (
+    "col1: int, col2: int, col3: int, col4: array<int>, "
+    "col5: date, col6: timestamp, col7: map<string, int>, "
+    'col8: struct<field1: int>, col9: string'
```
mwojtyczka:
```suggestion
    'col8: struct<field1: int>, col9: string, col10: string'
```

tests/perf/test_apply_checks.py
```diff
@@ -0,0 +1,494 @@
...
+            columns=["col1", "col2"],
+            check_func_kwargs={
+                "ref_columns": ["ref_col1", "ref_col2"],
+                "ref_df_name": "ref_df",
+            },
+        ),
+    ]
+    refs_df = {"ref_df": make_ref_df}
+
+    checked = benchmark(dq_engine.apply_checks, generated_df, checks, refs_df)
```
mwojtyczka:
```suggestion
    checked = dq_engine.apply_checks(generated_df, checks, refs_df)
```

tests/perf/test_apply_checks.py
```diff
@@ -0,0 +1,494 @@
...
+            check_func=check_funcs.foreign_key,
+            columns=["col1", "col2"],
+            check_func_kwargs={
+                "ref_columns": ["ref_col1", "ref_col2"],
+                "ref_df_name": "ref_df",
+            },
+        ),
+    ]
+    refs_df = {"ref_df": make_ref_df}
+    checked = benchmark(dq_engine.apply_checks, generated_df, checks, refs_df)
```
mwojtyczka:
```suggestion
    checked = dq_engine.apply_checks(generated_df, checks, refs_df)
```

.github/workflows/nightly.yml
mwojtyczka: @cornzyblack  It works. Here the complete output: https://github.com/databrickslabs/dqx/actions/runs/17238539667/job/48909182834#logs

As you suggested, this does the trick: `PYTEST_ADDOPTS: "--benchmark-save=benchmark"`

I updated some tests, but generally all went thru: https://github.com/databrickslabs/dqx/actions/runs/17238812294/workflow

There are 10 benchmarks files in the output: tests/perf/.benchmarks/Linux-CPython-3.10-64bit/0001_benchmark.json .... 0010_benchmark.json

cornzyblack:
Yeah sure of course, I can continue from the new PR

mwojtyczka:
> n continue from th

thank you! I am testing the best way to execute this, you can focus on adding more tests (make sure we test all check funcs)

tests/perf/test_apply_checks.py
mwojtyczka: yes, let's add this

mwojtyczka:
Closing this PR. The change has been moved to this PR to be able to properly test it in github actions: https://github.com/databrickslabs/dqx/pull/548

PR #533, opened 2025-08-20
Add an end-to-end test for the asset bundles demo

### LINKED ISSUE #497: [FEATURE]: Add DAB demo to e2e tests

### Problem statement

We run all demos as part of e2e tests which are triggered for every PR and in nightly ci/cd job. The DAB demo is currently missing there. 

### Proposed Solution

Add DAB demo to the e2e tests.

### Additional Context

_No response_

ghanse:
## Changes

This PR adds an end-to-end test for the asset bundles demo. Bundle commands are run via Python's subprocess module.

### Linked issues

Resolves #497 

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests
- [x] added end-to-end tests


demos/dqx_demo_asset_bundle/databricks.yml
```diff
@@ -29,12 +35,14 @@ resources:
               maintenance_rules_file: "${var.maintenance_rules_file}"
               sensor_rules_file: "${var.sensor_rules_file}"
             source: WORKSPACE
-          environment_key: Default
+      job_clusters:
+        - job_cluster_key: dqx_demo_cluster
```
mwojtyczka, line 52: can we use serverless cluster so that we don't have to maintain the cluster config?

ghanse: I tested a few options. Setting the serverless environment with Python dependencies is not currently supported for notebook tasks deployed via DABs.

mwojtyczka: but this was working before, isn't it? serverless is supported by DAB: https://docs.databricks.com/aws/en/dev-tools/bundles/examples#job-that-uses-serverless-compute

PR #556, opened 2025-08-31
Perf benchmark improvement

cornzyblack:
## Changes
- Added foreach Performance tests for row checks
- Added a custom way to generate dataset types in conftest 

### Linked issues

Resolves #..

### Tests

- [x] manually tested
- [ ] added unit tests
- [ ] added integration tests
- [ ] added end-to-end tests
- [x] added performance tests

### DIFF
--- a/tests/perf/conftest.py
+++ b/tests/perf/conftest.py
@@ -152,8 +152,9 @@ def generated_array_string_df(request, spark):
     opts = params.get("opts", {})
     col_names, data_gen = make_data_gen(spark, n_rows=n_rows, n_columns=n_columns)
     for col in col_names:
-        data_gen = data_gen.withColumn(col, "string", template=r'\\w.\\w@\\w.com',
-                 numFeatures=(1, array_length), structType="array", **opts)
+        data_gen = data_gen.withColumn(
+            col, "string", template=r'\\w.\\w@\\w.com', numFeatures=(1, array_length), structType="array", **opts
+        )
     return col_names, data_gen.build(), n_rows
 
 


tests/perf/conftest.py
```diff
@@ -15,19 +15,35 @@
...
 
 
+def make_data_gen(
+    spark,
+    n_rows: int = DEFAULT_ROWS,
+    n_columns: int = DEFAULT_COLUMNS,
+    partitions: int = DEFAULT_PARTITIONS,
+):
+    col_names = [f"col{i+1}" for i in range(n_columns)]
+    gen = dg.DataGenerator(spark, rows=n_rows, partitions=partitions)
```
mwojtyczka, line 43: it does not make sense to parametrize columns if they are statically defined in generated_df

mwojtyczka (changes requested):
please remove changes to workflows
Please remove the baselines. The baseline is generated/updated automatically.

tests/perf/test_apply_checks.py
```diff
@@ -93,6 +120,28 @@ def test_benchmark_is_not_empty(benchmark, ws, generated_df, column):
...
 
 
+@pytest.mark.parametrize(
+    "generated_integer_df",
+    [{"n_rows": DEFAULT_ROWS, "n_columns": 5}],
+    indirect=True,
+    ids=lambda param: f"n_rows_{param['n_rows']}_n_columns_{param['n_columns']}",
+)
+@pytest.mark.benchmark(group="test_benchmark_foreach_is_not_empty")
+def test_benchmark_foreach_is_not_empty(benchmark, ws, generated_integer_df):
```
mwojtyczka, line 130: use string columns for this, not integer

tests/perf/test_apply_checks.py
```diff
@@ -149,6 +243,29 @@ def test_benchmark_is_in_list(benchmark, ws, generated_df, column):
...
 
 
+@pytest.mark.parametrize(
+    "generated_string_df",
+    [{"n_rows": DEFAULT_ROWS, "n_columns": 5}],
+    indirect=True,
+    ids=lambda param: f"n_rows_{param['n_rows']}_n_columns_{param['n_columns']}",
+)
+@pytest.mark.benchmark(group="test_benchmark_foreach_is_in_list")
+def test_benchmark_foreach_is_in_list(benchmark, ws, generated_string_df):
```
mwojtyczka, line 253: this should use int columns, not string

tests/perf/test_apply_checks.py
```diff
@@ -272,6 +554,29 @@ def test_benchmark_is_not_less_than(benchmark, ws, generated_df):
     assert actual_count == EXPECTED_ROWS
 
 
+@pytest.mark.parametrize(
+    "generated_timestamp_df",
```
mwojtyczka, line 558: i would use int for this test

tests/perf/test_apply_checks.py
```diff
@@ -287,6 +592,29 @@ def test_benchmark_is_not_greater_than(benchmark, ws, generated_df):
     assert actual_count == EXPECTED_ROWS
 
 
+@pytest.mark.parametrize(
+    "generated_date_df",
```
mwojtyczka, line 596: same as above, int would be better for this

tests/perf/test_apply_checks.py
```diff
@@ -376,6 +819,30 @@ def test_benchmark_is_valid_timestamp(benchmark, ws, generated_df):
...
```
mwojtyczka, line 823: shouldn't that be timestamp?

Issue #480, opened 2025-07-25
[Task]: Method for LLM based rules generation


mwojtyczka:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Problem statement

We need to expose a method to perform llm-based rules generation

### Proposed Solution

Method to perform llm-based rules generation based on params, e.g.
* (option) User questions/input, if not provided generate rule candidates
* Usage examples of rules (from docs)
* Rules definitions, including custom checks
* (option) Schema of the input table from UC

Issue #478, opened 2025-07-25
[Task]: LLM deps setup


mwojtyczka:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Problem statement

Required dependencies and endpoints must be setup.

### Proposed Solution

We will initially only use foundational models so not additional setup is required on DBX. However, we need to install required build and runtime dependencies (e.g. dspy, langchain).

PR #577, opened 2025-09-15
LLM assisted rules generation

### LINKED ISSUE #370: [FEATURE]: LLM to generate DQX rules

### Problem statement

A user can currently profile the data and generate quality rule candidates, or define rules manually.
Users cannot use natural language to generate the checks.

### Proposed Solution

Add natural language support with LLM to allow rule definition via plain English

### Additional Context

_No response_

souravg-db2:
## Changes

* Rules generation based on natural language and optionally schema from input table. Providing support for AI-assisted rules generation using programmatic and no-code (workflows) approach. 
* Using Databricks foundational model by default and support providing different LLM models.
* Generated rules are validated for correctness.
* Added demo for AI-assisted rules generation.
* Added new method to the `DQGenerator`: `generate_dq_rules_ai_assisted`

### Linked issues

Resolves #370 #478 #480 

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests
- [x] added end-to-end tests
- [ ] added performance tests

### DIFF
--- a/src/databricks/labs/dqx/profiler/generator.py
+++ b/src/databricks/labs/dqx/profiler/generator.py
@@ -89,7 +89,7 @@ class DQGenerator(DQEngineBase):
 
         return dq_rules
 
-    @telemetry_logger("generator", "generate_dq_rules_with_llm")
+    @telemetry_logger("generator", "generate_dq_rules_ai_assisted")
     def generate_dq_rules_ai_assisted(self, user_input: str, table_name: str = "") -> list[dict]:
         """
         Generates data quality rules using LLM based on natural language input.


src/databricks/labs/dqx/llm/llm_core.py
```diff
@@ -0,0 +1,272 @@
...
+        result = self.forward(schema_info, business_description, available_functions)
+        return result.quality_rules
+
+
+def _configure_dspy_lm(
+    model: str = "databricks/databricks-meta-llama-3-3-70b-instruct", api_key: str = "", api_base: str = ""
+):
+    """Configure the Dspy language model.
+
+    :param model: The model to use for the Dspy language model.
```
mwojtyczka: update all docstrings to use google style

src/databricks/labs/dqx/llm/llm_core.py
```diff
@@ -0,0 +1,272 @@
...
+    :param api_key: API key for authenticating with the language model service.
```
mwojtyczka:
```suggestion
    :param api_key: API key for authenticating with the language model service. Not required for Databricks foundational models.
```

src/databricks/labs/dqx/llm/llm_core.py
```diff
@@ -0,0 +1,272 @@
...
+    :param api_base: Base URL for the language model API endpoint.
```
mwojtyczka:
```suggestion
    :param api_base: Base URL for the language model API endpoint. Not required for Databricks foundational models.
```

src/databricks/labs/dqx/llm/resources/training_examples.yml
```diff
@@ -0,0 +1,93 @@
+- name: "product_code_not_null_or_empty"
```
mwojtyczka, line 1: This needs to be extended with more examples, or create a follow up issue. Even for the first release there should be a better coverage.

src/databricks/labs/dqx/llm/llm_core.py
```diff
@@ -0,0 +1,304 @@
...
+    # Rules validation score (50%)
+    validation_status = DQEngineCore.validate_checks(actual_rules)
+    if not validation_status.has_errors:
+        total_score += rule_weight
+        logger.info(f"✓ Rules validation passed (+{rule_weight:.1f})")
+    else:
+        logger.warning(f"✗ Rules validation errors: {validation_status.errors}")
+
+    # Content similarity score (30%)
+    similarity_score = _calculate_rule_similarity(expected_rules, actual_rules)
```
mwojtyczka: this is not needed: `DQEngineCore.validate_checks(actual_rules)` validates the generated rule is correct. There should be no need to compare against expected. It would also be problematic to maintain because a lot of arguments can be optional and providing examples of all possible combinations won't be possible

src/databricks/labs/dqx/llm/llm_core.py
```diff
@@ -0,0 +1,304 @@
...
+    # Score weights
+    json_weight = 0.2
+    rule_weight = 0.5
+    similarity_weight = 0.3
+
+    expected_rules = json.loads(expected)
+
+    # Json parsing score (20%)
+    try:
+        actual_rules = json.loads(actual)
```
mwojtyczka, line 141: this does not seem to be necessary since the `forward` method in `DQRuleGeneration` is making sure a valid json is outputed

mwojtyczka (changes requested):

src/databricks/labs/dqx/llm/llm_core.py
```diff
@@ -0,0 +1,315 @@
...
+        Args:
+            business_description (str): Natural language description of the dataset and its use case.
+
+        Returns:
+            dspy.primitives.prediction.Prediction: A Prediction object containing the guessed schema
+            and assumptions made during the inference process.
+        """
+        return self.guess(business_description=business_description)
+
+    def get_schema_dict(self, business_description: str) -> dict:
```
mwojtyczka: not used anywhere

src/databricks/labs/dqx/llm/llm_core.py
```diff
@@ -0,0 +1,315 @@
...
+        Initialize the DQ rule generation module.
+
+        Args:
+            enable_schema_inference (bool): If True, will use SchemaGuesser when schema_info is empty.
+        """
+        super().__init__()
+        # Use Predict for reliable output
+        self.generator = dspy.Predict(RuleSignature)
+        self.enable_schema_inference = enable_schema_inference
+        self.schema_guesser = SchemaGuesser() if enable_schema_inference else None
```
mwojtyczka: make `schema_guesser` available in the constructor so that it can be mocked/supplied if needed. It's a bad practice to hardcode creation of depencencies. There should be a way to inject dependencies, eg.
```
def __init__(self, enable_schema_inference: bool = True, schema_guesser: dspy.Module | None = None):
   self.schema_guesser = schema_guesser if enable_schema_inference and schema_guesser is not None else (SchemaGuesser() if enable_schema_inference else None)
```

src/databricks/labs/dqx/llm/llm_core.py
```diff
@@ -0,0 +1,315 @@
...
+        if not schema_info or not schema_info.strip():
+            if self.enable_schema_inference and self.schema_guesser:
+                logger.info("Schema not provided, inferring from business description...")
+                schema_result = self.schema_guesser(business_description=business_description)
+                schema_info = schema_result.guessed_schema_json
+                guessed_schema_json = schema_result.guessed_schema_json
+                assumptions_bullets = schema_result.assumptions_bullets
+                logger.info(f"Inferred schema: {schema_info}")
+            else:
+                raise ValueError("schema_info is required when enable_schema_inference=False")
```
mwojtyczka:
```suggestion
                raise MissingParameterError("schema_info is required when enable_schema_inference=False")
```
from databricks.labs.dqx.errors import MissingParameterError

src/databricks/labs/dqx/llm/llm_core.py
```diff
@@ -0,0 +1,315 @@
...
+
+        Args:
+            schema_info (str): JSON string containing table schema with column names, types, and sample data.
+                              If empty and enable_schema_inference=True, schema will be inferred.
+            business_description (str): Natural language description of data quality requirements.
+            available_functions (str): JSON string of available DQX check functions.
+
+        Returns:
+            dspy.primitives.prediction.Prediction: A Prediction object containing the generated data quality rules,
+            reasoning, and optionally guessed_schema_json and assumptions_bullets if schema was inferred.
```
mwojtyczka:
```suggestion
            reasoning, and optionally guessed_schema_json and assumptions_bullets if schema was inferred.
        
        Raises: 
            MissingParameterError: if schema_info is empty and enable_schema_inference is False.
```

src/databricks/labs/dqx/llm/llm_core.py
```diff
@@ -0,0 +1,315 @@
...
+class DQRuleGeneration(dspy.Module):
+    """
+    Generate data quality rules with improved JSON output reliability.
+
+    This class provides functionality to generate data quality rules based on schema information,
+    business descriptions, and available functions. It can optionally infer the schema from the
+    business description if schema_info is not provided or is empty.
+    """
+
+    def __init__(self, enable_schema_inference: bool = True):
```
mwojtyczka: Does that need to be even a parameter? we should always try to guess the schema if not provided.

src/databricks/labs/dqx/llm/llm_core.py
```diff
@@ -0,0 +1,315 @@
...
+        Returns:
+            dspy.primitives.prediction.Prediction: A Prediction object containing the generated data quality rules,
+            reasoning, and optionally guessed_schema_json and assumptions_bullets if schema was inferred.
+        """
+        # Step 1: Infer schema if needed
+        guessed_schema_json = None
+        assumptions_bullets = None
+
+        if not schema_info or not schema_info.strip():
+            if self.enable_schema_inference and self.schema_guesser:
```
mwojtyczka: I think we should always guess the schema if not provided and remove the `enable_schema_inference` param

src/databricks/labs/dqx/llm/llm_core.py
```diff
@@ -0,0 +1,315 @@
...
+            else:
+                raise ValueError("schema_info is required when enable_schema_inference=False")
+
+        # Step 2: Generate rules using the schema (provided or inferred)
+        result = self.generator(
+            schema_info=schema_info, business_description=business_description, available_functions=available_functions
+        )
+
+        # Validate and clean the JSON output
+        if hasattr(result, 'quality_rules'):
```
mwojtyczka: shouldn't that always exist since it's in the signature?

src/databricks/labs/dqx/llm/llm_core.py
```diff
@@ -0,0 +1,315 @@
...
+                # Re-serialize to ensure consistent formatting
+                result.quality_rules = json.dumps(parsed_rules, separators=(',', ':'))
+            except json.JSONDecodeError as e:
+                logger.warning(f"Generated invalid JSON: {e}. Raw output: {result.quality_rules}")
+                # Return a fallback empty array if JSON is invalid
+                result.quality_rules = "[]"
+
+        # Add schema inference results to the prediction if they exist
+        if guessed_schema_json:
+            result.guessed_schema_json = guessed_schema_json
```
mwojtyczka, line 192: we need need this field? we assigned `guessed_schema_json` to `schema_info` before so they are the same

src/databricks/labs/dqx/llm/llm_core.py
```diff
@@ -0,0 +1,315 @@
...
+            result.reasoning = (
+                f"[Schema Inference] The schema was automatically inferred from the question:\n"
+                f"{guessed_schema_json}\n\n"
+                f"Assumptions made:\n{assumptions_bullets}\n\n"
+                f"[Rule Generation] {original_reasoning}"
+            )
+
+        return result
+
+    def get_json_rules(self, schema_info: str, business_description: str, available_functions: str) -> str:
```
mwojtyczka: this function is not used

src/databricks/labs/dqx/llm/llm_core.py
```diff
@@ -0,0 +1,315 @@
...
+    Args:
+        actual (str): JSON string of the actual generated rules.
+
+    Returns:
+        float: A score between 0.0 and 1.0 representing the quality of the generated rules.
+    """
+    total_score = 0.0
+
+    # Score weights
+    json_weight = 0.4
```
mwojtyczka: Let's remove the json weight completely since the generator only output valid json.

src/databricks/labs/dqx/llm/llm_core.py
mwojtyczka:
```suggestion
        if result.quality_rules:
```

src/databricks/labs/dqx/llm/llm_core.py
```diff
@@ -0,0 +1,315 @@
...
+    """
+    _configure_dspy_model(api_key=api_key, api_base=api_base, model=model)
+
+    # Use standard DSPy approach with improved prompting
+    dq_model = DQRuleGeneration(enable_schema_inference=enable_schema_inference)
+    train_set = create_optimizer_training_set()
+
+    # Standard metric for JSON output validation
+    def json_metric(_example, pred, _trace=None):
+        if hasattr(pred, 'quality_rules'):
```
mwojtyczka:
```suggestion
        if pred.quality_rules:
```
don't use reflection if you can access the field directly

src/databricks/labs/dqx/profiler/generator.py
```diff
@@ -45,6 +81,54 @@ def generate_dq_rules(self, profiles: list[DQProfile] | None = None, level: str
...
+
+        Returns:
+            A list of dictionaries representing the generated data quality rules.
+
+        Raises:
+            RuntimeError: If DSPy compiler is not available.
+        """
+        # Check if DSPy compiler is available
+        if self.dspy_compiler is None:
+            raise RuntimeError(
```
mwojtyczka:
```suggestion
            raise MissingParameterError(
```
from databricks.labs.dqx.errors import MissingParameterError

let's use custom exception  here

src/databricks/labs/dqx/profiler/generator.py
```diff
@@ -45,6 +81,54 @@ def generate_dq_rules(self, profiles: list[DQProfile] | None = None, level: str
...
+
+        # Generate rules using pre-initialized LLM compiler
+        prediction = get_business_rules_with_llm(
+            user_input=user_input, dspy_compiler=self.dspy_compiler, schema_info=schema_info
+        )
+
+        # Parse the generated rules
+        try:
+            dq_rules = json.loads(prediction.quality_rules)
+        except json.JSONDecodeError as e:
```
mwojtyczka: the compiler will only return valid json, this validation is redundant

src/databricks/labs/dqx/profiler/generator.py
```diff
@@ -45,6 +81,54 @@ def generate_dq_rules(self, profiles: list[DQProfile] | None = None, level: str
...
+            logger.error(f"Failed to parse LLM generated rules: {e}")
+
+        # Validate the generated rules using DQEngine
+        status = DQEngine.validate_checks(dq_rules)
```
mwojtyczka: the compiler should only return valid rules, this validation is redundant

src/databricks/labs/dqx/llm/llm_core.py
```diff
@@ -0,0 +1,315 @@
...
+            return validate_generated_rules(pred.quality_rules)
```
mwojtyczka: it would be great to have a few passes if the score is 0. A simple loop where we try to generate rules with the validation errors as input would be enough to start with, without going into teacher-student patterns.

src/databricks/labs/dqx/llm/llm_utils.py
```diff
@@ -0,0 +1,146 @@
...
+                "doc": doc or "",
+                "signature": str(sig),
+                "parameters": str(sig.parameters),
+                "implementation": inspect.getsource(func),
+            }
+        )
+    return function_docs
+
+
+def load_yaml_checks_examples() -> str:
```
mwojtyczka: if we don't plan to use examples from docs we can remove this and related code

mwojtyczka (changes requested):
Missing: docs, integration tests, integration into profiler workflow

src/databricks/labs/dqx/llm/llm_core.py
souravg-db2: Not needed

src/databricks/labs/dqx/llm/llm_core.py
souravg-db2: Removed this

src/databricks/labs/dqx/profiler/generator.py
souravg-db2: Done

src/databricks/labs/dqx/profiler/generator.py
souravg-db2: This is run time check and rules are evaluated before sharing it to user

src/databricks/labs/dqx/llm/llm_core.py
souravg-db2: DqEngine validates the list[dict] checks. We need to score both json and valid rule while during compile time. I have changed the score though to give higher weightage to valid rule

src/databricks/labs/dqx/llm/llm_utils.py
souravg-db2: Yes we can remove this.

src/databricks/labs/dqx/llm/llm_utils.py
```diff
@@ -0,0 +1,128 @@
...
+    """
+    Extract only required function information (name and doc).
+
+    Returns:
+        list[dict[str, str]]: A list of dictionaries containing the name, doc, type, signature, and parameters of each function.
+    """
+    required_function_docs: list[dict[str, str]] = []
+    for func in get_check_function_definition():
+        required_func_info = {
+            "check_function_name": func.get("name", ""),
```
mwojtyczka, line 66: why do we skip things like signature and docstring? is it really enough to have just name and parameters? we need at least a comment here

souravg-db2: I got better results with name and Parameters . It looks like if context length increases then we don't get good results.

mwojtyczka: thank you! i put this as a comment in the code

PR #584, opened 2025-09-23
Profile subset dataframe

### LINKED ISSUE #569: [FEATURE]: Run data Profile in a subset of the input data

### Problem statement

DQX allows you to perform data profiling and generate quality rule candidates only on an entire dataset, but sometimes you need to do this on a specific subset of your input data. For example, instead of monitoring the minimum and maximum selling price across the entire product catalog, you might want to do this individually, as the range of prices varies significantly across products.

### Proposed Solution

Add support for applying a filter on our dataset/table in the profile method options by replicating the underlying filter to the yaml check file (Ex: filter: Product_name = 'cell_phone')

### Additional Context

_No response_

STEFANOVIVAS:
### Linked issues

Resolves #569

### Tests

- [ ] manually tested
- [ ] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests


databricks.yml
```diff
@@ -0,0 +1,19 @@
+# This is a Databricks asset bundle definition for dqx.
```
mwojtyczka: pls remove your local files

STEFANOVIVAS: I'll close the PR and recreate it after I do more manual testing. By the way, the code you showed for installing DQX from the development branch worked! Thanks!

PR #589, opened 2025-09-26
Profile subset dataframe

### LINKED ISSUE #569: [FEATURE]: Run data Profile in a subset of the input data

STEFANOVIVAS:
## Changes

* Added ability to profile and generate rules on subset of the input data by introducing a sql expression filter.

### Linked issues

Resolves #569

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests


tests/integration/test_load_checks_from_uc_volume.py
```diff
@@ -9,15 +9,33 @@
 TEST_CHECKS = [
     {
         "criticality": "error",
-        "check": {"function": "is_not_null", "for_each_column": ["col1", "col2"], "arguments": {}},
-    }
+        "check": {"function": "is_not_null", "for_each_column": ["col1", "col2"], "arguments": {}, "filter": None},
```
mwojtyczka: Create new test instead of modifying existing.
```suggestion
        "check": {"function": "is_not_null", "for_each_column": ["col1", "col2"], "arguments": {}},
```

tests/integration/test_profiler.py
```diff
@@ -67,18 +67,25 @@ def test_profiler(spark, ws):
     stats, rules = profiler.profile(inp_df, options={"sample_fraction": None})
 
     expected_rules = [
-        DQProfile(name="is_not_null", column="t1", description=None, parameters=None),
+        DQProfile(name="is_not_null", column="t1", description=None, parameters=None, filter=None),
```
mwojtyczka: Don't modify existing tests. Filter is by default None.
```suggestion
        DQProfile(name="is_not_null", column="t1", description=None, parameters=None),
```

src/databricks/labs/dqx/profiler/profiler.py
```diff
@@ -109,6 +112,10 @@ def profile(
             return summary_stats, dq_rules
 
         self._profile(df, df_columns, dq_rules, options, summary_stats, total_count)
+        filter = options.get("filter", None)
```
mwojtyczka: The filter should filter the input DataFrame

`df = df.filter(filter)`

STEFANOVIVAS: The filter itself was being applied in the self._sample method, because i think that the filter must call before sample and limit operations. If you prefer i can put it in the profile method, before sample it. In fact, the code that i have included on line 115 and below makes no sense at all.

tests/integration/test_profiler.py
```diff
@@ -1255,3 +1264,323 @@ def test_profile_tables_no_tables_or_patterns(ws):
...
+
+    custom_options = {
+        "sample_fraction": None,
+        "round": False,
+        "limit": None,
+        "filter": "machine_id IN ('MCH-002', 'MCH-003') AND maintenance_type = 'preventive'",
+    }
+
+    profiler = DQProfiler(ws)
+    filtered_df = profiler._sample(input_df, opts=custom_options)
```
mwojtyczka: don't access private methods in tests

tests/integration/test_profiler.py
```diff
@@ -1255,3 +1264,323 @@ def test_profile_tables_no_tables_or_patterns(ws):
...
```
mwojtyczka: pls don't use private methods in tests. It makes the code and tests hard to maintain.

src/databricks/labs/dqx/profiler/profiler.py
mwojtyczka: ah ok, thx for the clarfication, it's better to have it in _sample

tests/integration/test_load_checks_from_uc_volume.py
```diff
@@ -216,3 +216,50 @@ def test_save_checks_in_volume_file_as_json(ws, make_schema, make_volume, instal
 
     checks = dq_engine.load_checks(config=VolumeFileChecksStorageConfig(location=checks_path))
     assert TEST_CHECKS == checks, "Checks were not saved correctly"
+
+
+TEST_CHECKS_FILTER = [
+    {
+        "criticality": "error",
+        "check": {"function": "is_not_null", "for_each_column": ["col1", "col2"], "arguments": {}, "filter": None},
```
mwojtyczka: Filter must be at the base level, not inside a check. We don't want to change the existing syntax.

```
{
        "criticality": "error",
        "filter": None,
        "check": {"function": "is_not_null", "for_each_column": ["col1", "col2"], "arguments": {}},
}

```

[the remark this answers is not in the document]
mwojtyczka: pls correct

tests/integration/test_load_checks_from_uc_volume.py
```diff
@@ -216,3 +216,50 @@ def test_save_checks_in_volume_file_as_json(ws, make_schema, make_volume, instal
...
+    {
+        "check": {
+            "function": "is_not_null",
+            "arguments": {"column": "next_scheduled_date"},
+            "filter": "machine_id IN ('MCH-002', 'MCH-003') AND maintenance_type = 'preventive'",
+        },
+        "name": "next_scheduled_date_is_null",
+        "criticality": "Error",
+    },
+]
```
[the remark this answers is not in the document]
mwojtyczka: good suggestion

tests/integration/test_load_checks_from_uc_volume.py
```diff
@@ -216,3 +216,50 @@ def test_save_checks_in_volume_file_as_json(ws, make_schema, make_volume, instal
...
+        "check": {"function": "is_not_null", "for_each_column": ["col1", "col2"], "arguments": {}, "filter": None},
+    },
+    {
+        "check": {
+            "function": "is_not_null",
+            "arguments": {"column": "next_scheduled_date"},
+            "filter": "machine_id IN ('MCH-002', 'MCH-003') AND maintenance_type = 'preventive'",
+        },
+        "name": "next_scheduled_date_is_null",
+        "criticality": "Error",
```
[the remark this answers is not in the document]
mwojtyczka: good suggestion. Please ensure to use small letters for criticality

mwojtyczka (changes requested):
Please address remaining comments

tests/integration/test_save_checks_to_workspace_file.py
```diff
@@ -18,7 +18,21 @@
...
-    }
+    },
+    {
+        "check": {
+            "function": "is_not_null",
+            "arguments": {"column": "next_scheduled_date"},
+            "filter": "machine_id IN ('MCH-002', 'MCH-003') AND maintenance_type = 'preventive'",
+        },
+        "name": "next_scheduled_date_is_null",
+        "criticality": "Error",
```
mwojtyczka:
```suggestion
        "criticality": "error",
```

mwojtyczka (changes requested):
formatting is still wrong

mwojtyczka (changes requested):
Documentation has to be updated:
* https://github.com/STEFANOVIVAS/dqx/blob/profile_subset_dataframe/docs/dqx/docs/guide/data_profiling.mdx
* https://github.com/STEFANOVIVAS/dqx/blob/profile_subset_dataframe/docs/dqx/docs/reference/profiler.mdx
* https://github.com/STEFANOVIVAS/dqx/blob/profile_subset_dataframe/docs/dqx/docs/installation.mdx#configuration-file

The filter parameter should also be added to the `ProfilerConfig` for workflows

PR #609, opened 2025-10-11
Skip check evaluation if columns or filter cannot be resolved in the input DataFrame

### LINKED ISSUE #608: [FEATURE]: Skips checks if columns are not present in the input dataset

### Problem statement

In the current implementation of the data quality framework, checks are defined based on metadata that may reference specific columns in the input dataset. However, if a check references a column that is not present in the dataset at runtime, the entire process fails immediately.

This behavior introduces several challenges:

**- **Brittle Execution:**** A single missing column—whether due to schema drift, upstream changes, or optional fields—can cause the entire data quality run to terminate, even if all other checks are valid.

**- Reduced Observability**: The failure does not provide granular insight into which specific checks failed due to missing columns, making it harder to debug or trace issues.

**- Limited Flexibility:** In real-world scenarios, datasets often evolve over time. Some columns may be conditionally present or vary across environments. The current approach lacks the flexibility to handle such variability gracefully.

**- Developer Friction:** During development or testing, incomplete datasets may be used. The inability to run partial validations without full schema compliance slows down iteration and increases overhead.

### Proposed Solution

Enhance the the apply_checks series of methods to gracefully handle missing columns by skipping the check and logging a warning or error, instead of failing the entire execution process.

### Additional Context

_No response_

mwojtyczka:
## Changes

Currently, applying a check to a column or list of columns that are invalid (not present in the input DataFrame or containing an unresolvable expression) would cause the check to fail. With this change, the `column`, `columns`, and `filter` fields are validated against the input DataFrame. If any of these fields are invalid, the check is skipped, and the results include a failure message identifying the invalid fields. This ensures that the data quality process can continue gracefully, without failing due to missing or invalid fields.

### Linked issues

Resolves #608 

### Tests

- [x] manually tested
- [ ] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests


src/databricks/labs/dqx/manager.py
```diff
@@ -87,3 +128,34 @@ def _build_result_struct(self, condition: Column) -> Column:
...
+
+        return invalid_cols_message
+
+    def _is_invalid_column(self, column: str | Column) -> bool:
+        """
+        Returns name of a missing column, if any.
+        """
+        try:
+            # if column is not accessible or column expression cannot be evaluated AnalysisException is thrown
+            col_expr = F.expr(column) if isinstance(column, str) else column
```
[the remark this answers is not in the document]
mwojtyczka: would not validate string expressions correctly

src/databricks/labs/dqx/manager.py
```diff
@@ -53,26 +58,63 @@ def user_metadata(self) -> dict[str, str]:
...
+        Returns list of invalid check columns in the input DataFrame.
+        """
+        invalid_cols = []
+
+        # Validate a single column
+        if self.check.column is not None and self._is_invalid_column(self.check.column):
+            invalid_cols.append(get_column_name_or_alias(self.check.column))
+
+        # Validate multiple columns
+        if self.check.columns is not None:
```
alexott: Should we have

```
if
...
elif
....
```
?

We shouldn't have both check.column and check.columns at the same time?

src/databricks/labs/dqx/manager.py
```diff
@@ -53,26 +58,63 @@ def user_metadata(self) -> dict[str, str]:
...
+    @cached_property
+    def has_invalid_filter(self) -> bool:
+        return self._is_invalid_column(self.filter_condition)
+
+    @cached_property
+    def has_invalid_columns(self) -> bool:
+        """
+        Returns a boolean indicating whether any of the specified check columns are invalid in the input DataFrame.
+        """
+        return len(self.invalid_columns) > 0
```
alexott: could be expressed as `return bool(self.invalid_columns)`...

mwojtyczka: updated

src/databricks/labs/dqx/manager.py
```diff
@@ -87,3 +129,34 @@ def _build_result_struct(self, condition: Column) -> Column:
...
+        invalid_cols_message = "; ".join(invalid_cols_message_parts)
+
+        return invalid_cols_message
+
+    def _is_invalid_column(self, column: str | Column) -> bool:
+        """
+        Returns True if the specified column is invalid (i.e., cannot be resolved in the input DataFrame), otherwise False.
+        """
+        try:
+            # if column is not accessible or column expression cannot be evaluated AnalysisException is thrown
```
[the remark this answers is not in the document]
mwojtyczka: updated

src/databricks/labs/dqx/manager.py
```diff
@@ -87,3 +129,34 @@ def _build_result_struct(self, condition: Column) -> Column:
...
+            col_expr = F.expr(column) if isinstance(column, str) else column
+            _ = self.df.select(col_expr).schema
```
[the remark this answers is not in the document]
mwojtyczka: added comments, accessing schema does not trigger computation

src/databricks/labs/dqx/manager.py
mwojtyczka: updated

docs/dqx/docs/guide/quality_checks_storage.mdx
```diff
@@ -72,18 +72,18 @@ If you create checks as a list of `DQRule` objects, you can convert them to meta
...
     dq_engine.save_checks(checks, config=WorkspaceFileChecksStorageConfig(location="/Shared/App1/checks.yml"))
 
-    # save checks in a Delta table to "default" run config (append checks)
+    # save checks in a Delta table to "default" run config, append checks
     dq_engine.save_checks(checks, config=TableChecksStorageConfig(location="catalog.schema.checks_table", mode="append"))
 
-    # save checks in a Delta table with specific run config for filtering (overwrite checks)
-    dq_engine.save_checks(checks, config=TableChecksStorageConfig(location="catalog.schema.checks_table", run_config_name="workflow_001", mode="overwrite"))
+    # save checks in a Delta table with specific run config for filtering (e.g. input table name), overwrite checks
+    dq_engine.save_checks(checks, config=TableChecksStorageConfig(location="catalog.schema.checks_table", run_config_name="main.default.input_table", mode="overwrite"))
```
ghanse: I see why we provide this example but I worry that this may be misleading. Can we potentially update the docstring to something like `(e.g. input table or job name)`?

ghanse (approved):
LGTM! Left a comment on the docs.

docs/dqx/docs/guide/quality_checks_storage.mdx
mwojtyczka: yes, i will update this, i also worry that it may be misleading, it's just whenever people use it they put a table name there.

mwojtyczka: updated

alexott (approved):

PR #613, opened 2025-10-14
add null island check

### LINKED ISSUE #603: [FEATURE]: Add `is_not_null_island` check

### Problem statement

Geospatial data columns often contain `POINT(0 0)` to indicate that no known position exists. We essentially have the `is_null` problem for point geometries/geographies.

### Proposed Solution

Write a new (geo)check that tests if a column of points does/doesn't contain the null island.

### Additional Context

[NULL island](https://en.wikipedia.org/wiki/Null_Island)

tdikland:
## Changes

Add new geo check: `is_not_null_island` to check whether the values in the input column are NULL island geometries (e.g. POINT(0 0), POINTZ(0 0 0), or POINTZM(0 0 0 0)).

### Linked issues

Resolves #603

### Tests

- [ ] manually tested
- [ ] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [x] added performance tests


src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -357,6 +357,35 @@ def is_non_empty_geometry(column: str | Column) -> Column:
...
+        This function requires Databricks serverless compute or runtime 17.1 or above.
+    """
+    col_str_norm, col_expr_str, col_expr = _get_normalized_column_and_expr(column)
+    # NOTE: This function is currently only available in Databricks runtime 17.1 or above or in
+    #   Databricks SQL, due to the use of the `try_to_geometry`, `st_geometrytype`, `st_x`, and `st_y` functions.
+    geom_cond = F.expr(f"try_to_geometry({col_str_norm}) IS NULL")
+    is_point_cond = F.expr(f"st_geometrytype(try_to_geometry({col_str_norm})) = '{POINT_TYPE}'")
+    is_zero_zero = F.expr(f"st_x(try_to_geometry({col_str_norm})) = 0.0 AND st_y(try_to_geometry({col_str_norm})) = 0.0")
+    condition = F.when(col_expr.isNull(), F.lit(None)).otherwise(~geom_cond & is_point_cond & is_zero_zero)
+    condition_str = f"column `{col_expr_str}` contains a null island"
```
ghanse: What about POINTZ and POINTZM types?

tests/integration/test_row_checks_geo.py
```diff
@@ -333,6 +334,29 @@ def test_is_non_empty_geometry(skip_if_runtime_not_geo_compatible, spark):
     assert_df_equality(actual, expected, ignore_nullable=True)
 
 
+def test_is_not_null_island(skip_if_runtime_not_geo_compatible, spark):
+    input_schema = "geom: string"
+    test_df = spark.createDataFrame(
+        [["POINT(1 1)"], ["POINT(0 0)"], ["LINESTRING(0 0, 1 1)"], ["nonsense"], [None]],
```
ghanse: Should we test GeoJSON or WKB as well?

src/databricks/labs/dqx/geo/check_funcs.py
tdikland: What would you expect? When people talk about the null island, they mean the point at 0 lat, 0 long. I don't know if POINTZ/POINTZM makes sense in this setting, and was therefore a bit "liberal" in my definition.

tests/integration/test_row_checks_geo.py
tdikland: Ideally we would be able to construct geometries directly in this constructor instead of starting with strings and converting them. The other formats are already checked in adjacent tests, so I think it is a bit overkill here.

src/databricks/labs/dqx/geo/check_funcs.py
ghanse: I think something like `POINT Z (0 0 0)` and `POINT ZM (0 0 0 0)` would be equivalent? Just to validate that the data is not a placeholder geometry or generated by some process that defaults values to 0. We might be able to use `ST_Z` and `ST_M`.

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -357,6 +357,35 @@ def is_non_empty_geometry(column: str | Column) -> Column:
...
+
+    Note:
+        This function requires Databricks serverless compute or runtime 17.1 or above.
+    """
+    col_str_norm, col_expr_str, col_expr = _get_normalized_column_and_expr(column)
+    # NOTE: This function is currently only available in Databricks runtime 17.1 or above or in
+    #   Databricks SQL, due to the use of the `try_to_geometry`, `st_geometrytype`, `st_x`, and `st_y` functions.
+    geom_cond = F.expr(f"try_to_geometry({col_str_norm}) IS NULL")
+    is_point_cond = F.expr(f"st_geometrytype(try_to_geometry({col_str_norm})) = '{POINT_TYPE}'")
+    is_zero_zero = F.expr(f"st_x(try_to_geometry({col_str_norm})) = 0.0 AND st_y(try_to_geometry({col_str_norm})) = 0.0")
```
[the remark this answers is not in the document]
ghanse: Fixed

src/databricks/labs/dqx/geo/check_funcs.py
ghanse: Fixed

Issue #71, opened 2024-12-27
[FEATURE]: Handle timestamps and dates in the checks generator


mwojtyczka:
### Is there an existing issue for this?

- [ ] I have searched the existing issues

### Problem statement

profiler provides metrics for timestamp and date as min_limit and max_limit. These are not handled yet properly in the generator, as range, greater, less than check functions accept only int. 

### Proposed Solution

Extend generator.

louis-vines:
Has this been implemented? I'm trying to find a good small issue as my first contribution to the codebase, but from what I can gather from the codebase, this looks already done

mwojtyczka:
@bsr-the-mngrm since you were working on profiler recently, maybe you can have a look at this one?

bsr-the-mngrm:
> [@bsr-the-mngrm](https://github.com/bsr-the-mngrm) since you were working on profiler recently, maybe you can have a look at this one?

@mwojtyczka :

I’m planning a minimal first PR to address the core of this feature:

- Generator: Update DQGenerator.generate_dq_rules to emit temporal checks when profiles provide min_limit/max_limit as datetime.date or datetime.datetime:

  - Both bounds → is_in_range(col, min, max, inclusive=True)
  - Only min → is_not_less_than(col, min)
  - Only max → is_not_greater_than(col, max)
  - Pass Python date/datetime objects through without stringification.

- Tests: Add unit tests covering DateType and TimestampType:

  - date min/max → is_in_range with date args
  - timestamp only min → is_not_less_than with datetime arg
  - timestamp only max → is_not_greater_than with datetime arg
  - Validate that generated rules are accepted by existing execution paths.

- Scope boundaries / TODOs for follow-ups:

  - Serializer: robust JSON/YAML round-trip for temporal types (typed JSON encoding, YAML native datetime).
  - Validator: explicit acceptance and normalization of datetime/date (incl. timezone handling).
  - DLT generator: proper SQL/Python literal rendering for DATE/TIMESTAMP.
  - Docs: examples for temporal generation and execution.

Please let me know if you’d prefer a broader initial scope; otherwise I’ll proceed with this minimal slice and layer the serializer/validator/DLT/doc updates in subsequent PRs.

mwojtyczka:
@bsr-the-mngrm this is fine. Feel free to proceed.

STEFANOVIVAS:
[...]

PR #624, opened 2025-10-18
Improve Generator to emit temporal checks for min/max date & datetime

bsr-the-mngrm:
## Changes
- Update profiler and generator to emit temporal checks when profiles provide datetime.date or datetime.datetime:
  - both bounds → is_in_range(column, min_limit, max_limit)
  - only min → is_not_less_than(column, limit)
  - only max → is_not_greater_than(column, limit)
- Update integration test expectations to include a temporal and decimal check.
- Add unit tests for date/datetime pass-through in the generator.
- Preserve numeric behavior; no changes to DLT generator or check funcs.

Notes:
- Serializer/round-trip for JSON and checks table remains string-only for temporal values; follow-ups planned to add typed encoding.

### Linked issues
Resolves [databrickslabs/dqx#71](https://github.com/databrickslabs/dqx/issues/71)

### Tests
- [ ] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

Additional details:
- Unit tests verify generator pass-through of datetime.date and datetime.datetime.
- Integration test updated to expect is_not_less_than for product_launch_date.

### DIFF
--- a/tests/unit/test_generator_temporal.py
+++ b/tests/unit/test_generator_temporal.py
@@ -3,30 +3,30 @@ import datetime
 from databricks.labs.dqx.profiler.generator import DQGenerator
 
 
-def test_date_both_bounds_is_in_range():
+def test_date_both_bounds_is_in_range(set_utc_timezone):
     result = DQGenerator.dq_generate_min_max(
         "dcol", **{"min": datetime.date(2020, 1, 1), "max": datetime.date(2020, 12, 31)}
     )
     assert result["check"]["function"] == "is_in_range"
     args = result["check"]["arguments"]
     assert args["column"] == "dcol"
-    assert args["min_limit"] == datetime.date(2020, 1, 1)
-    assert args["max_limit"] == datetime.date(2020, 12, 31)
+    assert args["min_limit"] == "2020-01-01"
+    assert args["max_limit"] == "2020-12-31"
 
 
-def test_timestamp_only_min_is_not_less_than():
+def test_timestamp_only_min_is_not_less_than(set_utc_timezone):
     timestamp = datetime.datetime(2024, 6, 1, 12, 0, 0)
     result = DQGenerator.dq_generate_min_max("tscol", **{"min": timestamp, "max": None})
     assert result["check"]["function"] == "is_not_less_than"
     args = result["check"]["arguments"]
     assert args["column"] == "tscol"
-    assert args["limit"] == timestamp
+    assert args["limit"] == "2024-06-01T12:00:00.000000"
 
 
-def test_timestamp_only_max_is_not_greater_than():
+def test_timestamp_only_max_is_not_greater_than(set_utc_timezone):
     timestamp = datetime.datetime(2024, 6, 30, 23, 59, 59)
     result = DQGenerator.dq_generate_min_max("tscol", **{"min": None, "max": timestamp})
     assert result["check"]["function"] == "is_not_greater_than"
     args = result["check"]["arguments"]
     assert args["column"] == "tscol"
-    assert args["limit"] == timestamp
+    assert args["limit"] == "2024-06-30T23:59:59.000000"


bsr-the-mngrm:
I closed it because I want to be sure all integration tests are running successfully.

src/databricks/labs/dqx/profiler/generator.py
mwojtyczka:
```suggestion
    def dq_generate_min_max(column: str, level: str = "error", **params: dict) -> list[dict]:
```
please add return types to other generate methods as well

src/databricks/labs/dqx/profiler/generator.py
```diff
@@ -70,49 +70,59 @@ def dq_generate_min_max(column: str, level: str = "error", **params: dict):
...
+            return isinstance(value, int)
+
+        def _is_temporal(value):
+            return isinstance(value, (datetime.date, datetime.datetime))
+
+        def _same_family(value_a, value_b):
+            # numeric with numeric OR temporal with temporal
+            if value_a is None or value_b is None:
+                return True
+            return (_is_num(value_a) and _is_num(value_b)) or (_is_temporal(value_a) and _is_temporal(value_b))
```
mwojtyczka:
```suggestion
            return any([
                _is_num(value_a) and _is_num(value_b),
                _is_temporal(value_a) and _is_temporal(value_b),
            ])
```
to simplify and make it easier to extend

tests/integration/test_generator.py
```diff
@@ -22,19 +22,6 @@
         parameters={"min": datetime.date(2020, 1, 1), "max": None},
         description="Real min/max values were used",
     ),
-    DQProfile(
```
mwojtyczka, line 25: why are these cases removed?

src/databricks/labs/dqx/profiler/generator.py
```diff
@@ -70,49 +70,59 @@ def dq_generate_min_max(column: str, level: str = "error", **params: dict):
...
+        if min_limit is not None and max_limit is not None and _same_family(min_limit, max_limit):
             return {
                 "check": {
                     "function": "is_in_range",
                     "arguments": {
                         "column": column,
-                        "min_limit": val_maybe_to_str(min_limit, include_sql_quotes=False),
-                        "max_limit": val_maybe_to_str(max_limit, include_sql_quotes=False),
+                        # pass through Python ints or datetime/date without stringification
+                        "min_limit": min_limit,
```
mwojtyczka: wouldn't that cause issues down the line?  shouldn't the `val_maybe_to_str` still be used?

src/databricks/labs/dqx/profiler/generator.py
```diff
@@ -70,49 +70,59 @@ def dq_generate_min_max(column: str, level: str = "error", **params: dict):
...
         min_limit = params.get("min")
         max_limit = params.get("max")
 
-        if not isinstance(min_limit, int) or not isinstance(max_limit, int):
-            return None  # TODO handle timestamp and dates: https://github.com/databrickslabs/dqx/issues/71
+        if min_limit is None and max_limit is None:
+            return None
+
+        def _is_num(value):
+            return isinstance(value, int)
```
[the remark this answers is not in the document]
mwojtyczka: yeah, i guess we should also support float and decimal here

src/databricks/labs/dqx/profiler/generator.py
mwojtyczka: updated

src/databricks/labs/dqx/profiler/generator.py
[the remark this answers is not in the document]
mwojtyczka: updated

mwojtyczka (changes requested):
@bsr-the-mngrm I fixed the current issues and tests. But yes, the Serializer/round-trip for JSON and checks table remains string-only for temporal values, also it does not handle decimal

src/databricks/labs/dqx/profiler/generator.py
```diff
@@ -239,49 +238,64 @@ def dq_generate_min_max(column: str, criticality: str = "error", **params: dict)
...
-        if not isinstance(min_limit, int) or not isinstance(max_limit, int):
-            return None  # TODO handle timestamp and dates: https://github.com/databrickslabs/dqx/issues/71
+        if min_limit is None and max_limit is None:
+            return None
+
+        def _is_num(value):
+            return isinstance(value, (int, float))
+
+        def _is_temporal(value):
+            return isinstance(value, (datetime.date, datetime.datetime))
```
ghanse, line 259: Thoughts on using some union types?

```
Numeric = int | float
Temporal = datetime.date | datetime.datetime

if isinstance(var, Numeric):
  ...

if isinstance(var, Temporal):
  ...
```

src/databricks/labs/dqx/profiler/generator.py
```diff
@@ -239,49 +238,64 @@ def dq_generate_min_max(column: str, criticality: str = "error", **params: dict)
...
+                return True
+            return any(
+                [
+                    _is_num(value_a) and _is_num(value_b),
+                    _is_temporal(value_a) and _is_temporal(value_b),
+                ]
+            )
+
+        # Both bounds
+        if min_limit is not None and max_limit is not None and _same_family(min_limit, max_limit):
```
ghanse, line 271: Given we will only enter this block if both `min_limit` and `max_limit` exist, do we need to handle `None` on lines 262-263? It seems counter-intuitive that we would return `True` if one of the arguments is `None`.

mwojtyczka: updated

src/databricks/labs/dqx/profiler/generator.py
mwojtyczka: it's not bringing much here

mwojtyczka (approved):
LGTM

PR #676, opened 2025-11-01
Improved Config Serialization

mwojtyczka:
## Changes

- Added saving of configuration and run configuration, required for the App.
- Improved config serialization to ensure boolean resources are properly serialized and deserialized from config. Previously if a field has been assigned False, it did not appear in the config at all. When reading it and the default field value in the model was True, it was incorrectly read as True.
- Added telemetry for DLT.
- Refactored get installation functionality to enforce single responsibility.
- Updated acceptance GitHub Action to not cancel the workflow if already in progress, ensuring fixture resources are properly released and cleaned. Previously, interrupted pipelines (e.g., when pushing new changes) could skip fixture teardown, leaving orphaned resources in the workspace. This is especially important for limited resources like Lakebase which has strict quota limits.
- Removed test for cleaning up lakbase instances (not needed anymore, as per above).
- Minor documentation updates.
- Updated tests to use "dqx" UC catalog for testing.

### Tests

- [x] manually tested
- [ ] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

### DIFF
--- a/.github/workflows/acceptance.yml
+++ b/.github/workflows/acceptance.yml
@@ -16,7 +16,7 @@ permissions:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: false  # don't cancel ongoing runs to ensure fixtures are properly terminated
+  cancel-in-progress: false  # don't cancel ongoing runs to ensure fixtures are completed and resources terminated
 
 jobs:
 
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -13,7 +13,7 @@ permissions:
 
 concurrency:
   group: single-acceptance-job-per-repo
-  cancel-in-progress: false  # don't cancel ongoing runs to ensure fixtures are properly terminated
+  cancel-in-progress: false  # don't cancel ongoing runs to ensure fixtures are completed and resources terminated
 
 jobs:
   integration:


docs/dqx/docs/guide/quality_checks_apply.mdx
```diff
@@ -197,8 +197,8 @@ In the example below, checks are defined in YAML syntax for convenience and then
 
 <Admonition type="note" title="Note">
 Checks are applied to a `column` specified in the `arguments` of the `check`. This can be either a column name as `string` or column expression.
-Some checks require `columns` (e.g. `is_unique`) instead of `column`, which is a list of column names or expressions. There are also checks that require no columns at all (e.g. `sql_expression`).
-Alternatively, `for_each_column` can be used to define a list of columns that the check should be applied for one by one.
+Some checks require `columns` (e.g. `is_unique`) instead of `column`, which is a list of column names or expressions. There are also checks that require no columns at all (e.g. `sql_expression` and `sql_query`).
+Alternatively, `for_each_column` can be used to define a list of columns that the check should be applied for one by one. The `for_each_column` is supported for quality rules that take `column` or `columns` as arguments (e.g. not supported by `sql_expression` and `sql_query`).
```
ghanse: Maybe we should change this wording?

> Alternatively, `for_each_column` can be used to define a list of columns that the check should be applied for one by one.

This might mislead users into thinking we will trigger multiple actions for each checked column. Maybe something like this is better?

> Alternatively, `for_each_column` can be used to define a list of columns for which the same check should be applied.

src/databricks/labs/dqx/config.py
```diff
@@ -135,6 +135,13 @@ class WorkspaceConfig:
     quality_checker_max_parallelism: int = 4  # max parallelism for quality checking multiple tables
     custom_metrics: list[str] | None = None  # custom summary metrics tracked by the observer when applying checks
 
+    def as_dict(self) -> dict:
+        """
+        Convert the WorkspaceConfig to a dictionary for serialization.
+        This method ensures that all fields, including boolean False values, are properly serialized.
+        """
```
ghanse, line 146: We might want to document that this is called by the installer during installation. Docstring needs a `Returns` section.

src/databricks/labs/dqx/config_serializer.py
```diff
@@ -0,0 +1,85 @@
...
+        self, assume_user: bool = True, product_name: str = "dqx", install_folder: str | None = None
+    ) -> WorkspaceConfig:
+        """
+        Load workspace configuration from the installation. The workspace config contains all run configs.
+
+        Args:
+            product_name: name of the product
+            assume_user: if True, assume user installation
+            install_folder: Custom workspace installation folder. Required if DQX is installed in a custom folder.
+        """
```
ghanse, line 23: We should reorder the docstring arguments to match the method signature and add a `Returns` section.

Same comment applies to the other methods.

src/databricks/labs/dqx/engine.py
ghanse, line 461: Should we add `Args` for instantiating the engine?

docs/dqx/docs/reference/quality_checks.mdx
```diff
@@ -19,7 +19,7 @@ You can explore the implementation details of the check functions [here](https:/
 ## Row-level checks reference
 
 Row-level checks are applied to each row in a PySpark DataFrame. The quality check results are reported for individual rows in the result columns.
-You can also define your own custom checks (see [Creating custom checks](#creating-custom-row-level-checks)).
+You can also define your own custom checks in python (see [Creating custom checks](#creating-custom-row-level-checks)).
```
alexott: `python` -> `Python`?

docs/dqx/docs/guide/quality_checks_apply.mdx
mwojtyczka: updated, also other places that describe for each column

src/databricks/labs/dqx/config.py
mwojtyczka: added

src/databricks/labs/dqx/config_serializer.py
mwojtyczka: reordered in all methods and added Returns section

docs/dqx/docs/reference/quality_checks.mdx
mwojtyczka: updated

src/databricks/labs/dqx/engine.py
mwojtyczka: added

src/databricks/labs/dqx/engine.py
```diff
@@ -1222,3 +1232,11 @@ def _wait_for_one_time_trigger_streaming_queries(
             output_query.awaitTermination()
         if quarantine_query and quarantine_config and is_one_time_trigger(quarantine_config.trigger):
             quarantine_query.awaitTermination()
+
+    def _log_telemetry_for_streaming(self, df):
+        log_telemetry(self.ws, "streaming", str(df.isStreaming).lower())
+
+        try:
+            log_telemetry(self.ws, "dlt", "true" if self.spark.conf.get('pipelines.id', None) else "false")
```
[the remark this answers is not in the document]
mwojtyczka: this is intentional since the telemetry should never fail the process. The exception type seems to be different for different cluster versions, therefore, using the safe option to catch all exception here

alexott: I would move `self.spark.conf.get('pipelines.id', None)` into a separate function where I'll catch exception separately, instead of skipping telemetry completely as right now

mwojtyczka: very good point, updated

ghanse (approved):
LGTM

PR #931, opened 2025-11-13
Fix pickling of DQEngine

### LINKED ISSUE #922: [BUG]: Using instantiated DQengine inside foreachbatch in serverless/shared access mode is throwing up error

### Current Behavior

 In serverless/shared access mode dqengine instance  cant be shipped to executors , it needs to be instantiated within foreachbatch using manual credentials, it would make cumbersome from efficiency POV as well as user needs to provide credentials.

[github_issue_dqx_streaming_demo_diy.html](https://github.com/user-attachments/files/23424754/github_issue_dqx_streaming_demo_diy.html)

### Expected Behavior

_No response_

### Steps To Reproduce

i have attached the notebook ,you can just run that to reproduce the error

### Cloud

Azure

### Operating System

Linux

### Relevant log output

```shell

```

ghanse:
## Changes

This PR fixes issues with pickling and unpickling of the `DQEngine` and associated objects which prevent injection of `DQEngine` into methods like `foreachBatch`.

I have introduced a `WorkspaceClientSerDeMixin` to implement `__getstate__` and `__setstate__` methods to reset the `WorkspaceClient`.

I have also introduced a `get_workspace_client()` utility method which is called when `DQEngine` needs to be used without an internal `WorkspaceClient` attribute.

### Linked issues

Resolves #922 

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

### DIFF
--- a/src/databricks/labs/dqx/base.py
+++ b/src/databricks/labs/dqx/base.py
@@ -3,15 +3,18 @@ from collections.abc import Callable
 from functools import cached_property
 from typing import final
 from pyspark.sql import DataFrame, Observation
-from databricks.labs.dqx.checks_validator import ChecksValidationStatus
-from databricks.labs.dqx.rule import DQRule
-from databricks.sdk import WorkspaceClient
 from databricks.labs.dqx.__about__ import __version__
+from databricks.labs.dqx.checks_validator import ChecksValidationStatus
+from databricks.labs.dqx.mixins import PickleableMixin
+from databricks.labs.dqx.rule import DQRule
+from databricks.labs.dqx.utils import get_workspace_client
+from databricks.sdk import WorkspaceClient
 
 
-class DQEngineBase(abc.ABC):
-    def __init__(self, workspace_client: WorkspaceClient):
-        self._workspace_client = self._verify_workspace_client(workspace_client)
+class DQEngineBase(PickleableMixin, abc.ABC):
+    def __init__(self, workspace_client: WorkspaceClient | None = None):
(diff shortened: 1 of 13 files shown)


src/databricks/labs/dqx/base.py
```diff
@@ -20,7 +23,7 @@ def ws(self) -> WorkspaceClient:
         Ensures workspace connectivity and sets the product info used for
         telemetry so that requests are attributed to *dqx*.
         """
-        return self._workspace_client
+        return self._ws or get_workspace_client()
```
mwojtyczka, line 26: can we simplify this, and have this logic in constructor only to avoid duplications

src/databricks/labs/dqx/mixins.py
```diff
@@ -0,0 +1,23 @@
...
+    Mixin to make classes with WorkspaceClient pickleable. Used to broadcast objects to Spark executors.
+
+    This mixin removes WorkspaceClient instances during pickling to avoid serialization of thread locks and network
+    connections. The WorkspaceClient is recreated lazily on first access after unpickling.
+    """
+
+    def __getstate__(self):
+        """Removes the WorkspaceClient when the object is pickled."""
+        state = self.__dict__.copy()
+        if '_ws' in state:
```
mwojtyczka, line 12: maybe we should standarize the naming and always use either ws or workspace client? we do the same in telemetry, we could avoid all these if statements.

PR #954, opened 2025-12-01
Added telemetry to track count of input tables

### LINKED ISSUE #940: [BUG]: is_not_equal_to function treats string values as SQL expressions instead of literals, causing "column not found" errors

### Current Behavior

When using the **is_not_equal_to** function with a string value in filtered rules, the function treats the string as a SQL expression (column name) instead of a string literal, causing Spark to fail with "UNRESOLVED_COLUMN" errors.

**Example Rule (from LLM-generated text expectations):**
{
  "check": {
    "function": "is_not_equal_to",
    "arguments": {
      "column": "order_status",
      "value": "cancelled"
    }
  },
  "filter": "discount_percentage > 50"
}
{  "check": {    "function": "is_not_equal_to",    "arguments": {      "column": "order_status",      "value": "cancelled"    }  },  "filter": "discount_percentage > 50"}

**Error Message:**
[UNRESOLVED_COLUMN.WITH_SUGGESTION] A column, variable, or function parameter with name `cancelled` cannot be resolved.

Spark is interpreting "cancelled" as a column name instead of a string literal 'cancelled'.

### Expected Behavior

The is_not_equal_to function should treat string values as literals, not SQL expressions. 

### Steps To Reproduce

1. Create a rule using is_not_equal_to with a string value:
 rule = {

    "check": {
        "function": "is_not_equal_to",
        "arguments": {
            "column": "status_column",
            "value": "pending"  # String value
        }
    }
}
rule = {    "check": {        "function": "is_not_equal_to",        "arguments": {            "column": "status_column",            "value": "pending"  # String value        }    }}
2. Apply the rule to a DataFrame
3. "UNRESOLVED_COLUMN" error for the string value

### Cloud

AWS

### Operating System

macOS

### Relevant log output

```shell

```

mwojtyczka:
## Changes
* Added telemetry to track count of input tables and non-table inputs
* Updated AI-assisted rules generation to create correct conditions with string literals
* Updated docs

### Linked issues

Resolves #940 

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

### DIFF
--- a/tests/unit/test_telemetry.py
+++ b/tests/unit/test_telemetry.py
@@ -25,7 +25,7 @@ class MockDataFrameWithPlan:
     def __init__(self, plan_output):
         self.plan_output = plan_output
 
-    def explain(self, extended=False):
+    def explain(self, _extended=False):
         print(self.plan_output, end='')
 
 


tests/integration/test_telemetry.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+def test_count_tables_from_streaming_table_based_dataframe(ws, spark, make_schema, make_random):
+    catalog_name = TEST_CATALOG
+    schema_name = make_schema(catalog_name=catalog_name).name
+    table_name = f"{catalog_name}.{schema_name}.{make_random(10).lower()}"
+
+    input_schema = T.StructType([T.StructField("id", T.IntegerType())])
+    df = spark.createDataFrame([[1]], schema=input_schema)
+    df.write.format("delta").saveAsTable(table_name)
```
grusin-db, line 128: joins, unions, etc. should be tested to make sure it parses correctly number of tables used

tests/integration/test_telemetry.py
```diff
@@ -0,0 +1,99 @@
...
+    assert count == 2, f"Expected 2 tables, but found {count}"
+
+
+def test_count_tables_from_streaming_table_based_dataframe(ws, spark, make_schema, make_random):
+    catalog_name = TEST_CATALOG
+    schema_name = make_schema(catalog_name=catalog_name).name
+    table_name = f"{catalog_name}.{schema_name}.{make_random(10).lower()}"
+
+    input_schema = T.StructType([T.StructField("id", T.IntegerType())])
+    df = spark.createDataFrame([[1]], schema=input_schema)
```
grusin-db, line 127: what will happen to telemetry when source is not a createDataFrame, but a real table?

mwojtyczka: This is tested, e.g. `test_count_tables_from_table_based_dataframe`

tests/integration/test_telemetry.py
mwojtyczka: join is there, added union and aggr

PR #969, opened 2025-12-15
Implementation of is_not_in_list check function

### LINKED ISSUE #964: [FEATURE]: is_not_in_list check function

### Problem statement

For our team, the opposite of is_in_list is the functionality that is wanted. Why is there not a is_not_in_list or a negate parameter for this check function?

### Proposed Solution

Add a simple is_not_in_list check function

### Additional Context

_No response_

larsmoan:
## Changes

Added a new check function called `is_not_in_list`, which works opposite to that of `is_in_list`.

### Linked issues

Resolves #964 

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] ~~added end-to-end tests~~
- [x] added performance tests

Integration tests written but not run locally, requires Databricks workspace


docs/dqx/docs/reference/quality_checks.mdx
```diff
@@ -33,6 +33,7 @@ You can also define your own custom checks in Python (see [Creating custom check
...
+| `is_not_in_list`                   | Checks whether the values in the input column are NOT present in the list of forbidden values (null values are allowed). This check is not suited for large lists of forbidden values. This check is not suited for `MapType` or `StructType` columns.                                                                                                                                                                            | `column`: column to check (can be a string column name or a column expression); `forbidden`: list of forbidden values; `case_sensitive`: optional boolean flag for case-sensitive comparison (default: True)                                                                                                                                                                                                                                                                                                                                         |
```
```suggestion
| `is_not_in_list`                   | Checks whether the values in the input column are NOT present in the list of forbidden values (null values are allowed). This check is not suited for large lists of forbidden values. In such cases, it’s recommended to use the `foreign_key` dataset-level check instead with negate parameter enabled. This check is not suited for `MapType` or `StructType` columns.                                                                                                                                                                            | `column`: column to check (can be a string column name or a column expression); `forbidden`: list of forbidden values; `case_sensitive`: optional boolean flag for case-sensitive comparison (default: True)                                                                                                                                                                                                                                                                                                                                         |
```
please add info to use `foreign_key` with negate instead for larger lists

mwojtyczka (changes requested):
Please also add the following:
* example to the `tests/resources/all_row_checks.yaml` - this will make sure the func is tested in apply checks scenario.
* examples in the docs (quality_checks.mdx). There are examples for every check function, below the table definition.

PR #992, opened 2026-01-10
Fix loading checks from a delta table with special characters

### LINKED ISSUE #991: [BUG]: `load_checks` fails when fully qualified table name includes back quote

### Current Behavior

When the fully qualified path to a table needs to include a backtick due to the inclusion of a character such as `-`, the table is unable to be loaded for the list of checks.

This happens because the workspace client checks for the existence of the table [here](https://github.com/databrickslabs/dqx/blob/0b146a69c1233ebcb4613591eab59bfd909195f9/src/databricks/labs/dqx/checks_storage.py#L110-L111). The Databricks SDK for `tables.exists` enforces NOT using backticks. However, if you omit the backticks such that the workspace client can confirm the existence of the table, then when the table is loaded into a spark df [here](https://github.com/databrickslabs/dqx/blob/0b146a69c1233ebcb4613591eab59bfd909195f9/src/databricks/labs/dqx/checks_storage.py#L112), the table cannot be loaded because spark DOES enforce the usage of backticks.

### Expected Behavior

Backticks can be included in the `location` argument of calls to the `load_checks` function when using a `TableChecksStorageConfig` without throwing any errors.

### Steps To Reproduce

1. Create a table with the fully qualified name of `example-catalog.example-schema.example-table` (notice the inclusion of the hyphens)
2. Attempt to load checks from this table:
```python
checks = dq_engine.load_checks(config=TableChecksStorageConfig(location="`example-catalog`.`example-schema`.`example-table`"))
```
3. Confirm that DQX throws an exception about the table not existing in the workspace (even though it was created in step 1 above).

### Cloud

Azure

### Operating System

Windows

### Relevant log output

```shell
When including the backticks:

NotFound: Checks table `example-catalog`.`example-schema`.`example-table` does not exist in the workspace


When excluding the backticks:

ParseException: [INVALID_IDENTIFIER] ...
```

cornzyblack:
## Changes

* Fixed loading checks from a delta table, when the fully qualified table name includes backticks

### Linked issues

Resolves #991 

### Tests

- [x] manually tested
- [ ] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

### DIFF
--- a/tests/integration/test_ai_rules_generator.py
+++ b/tests/integration/test_ai_rules_generator.py
@@ -94,7 +94,10 @@ def test_generate_dq_rules_ai_assisted_with_custom_functions(ws, spark):
 
     custom_check_functions = {"not_ends_with_suffix": not_ends_with_suffix}
 
-    user_input = USER_INPUT + "\nEmail address must not end with '@gmail.com'. Use not_ends_with_suffix function."
+    user_input = USER_INPUT + (
+        "\nEmail address must not end with '@gmail.com'. "
+        "Use not_ends_with_suffix function and don't add any extra quotes for suffix."
+    )
 
     generator = DQGenerator(ws, spark, custom_check_functions=custom_check_functions)
     actual_checks = generator.generate_dq_rules_ai_assisted(user_input=user_input)


src/databricks/labs/dqx/checks_storage.py
```diff
@@ -107,7 +107,8 @@ def load(self, config: TableChecksStorageConfig) -> list[dict]:
             NotFound: if the table does not exist in the workspace
         """
         logger.info(f"Loading quality rules (checks) from table '{config.location}'")
-        if not self.ws.tables.exists(config.location).table_exists:
+        api_ready_location = quote(config.location.replace("`", ""))
+        if not self.ws.tables.exists(api_ready_location).table_exists:
```
taylorsweetman: Are there any other locations in the codebase that use a workspace client to do an `exists` right before using spark to read the same table? I just so happened to find this location when attempting to load checks from a table, but it's also possible the same bug exists in other locations too (I haven't done a full analysis). Worth checking!

cornzyblack: Hi @taylorsweetman, that's true. Looking at the codebase, I can currently see two files that utilise this.

`TableChecksStorageHandler.load` in [src/databricks/labs/dqx/checks_storage.py](https://github.com/databrickslabs/dqx/blob/main/src/databricks/labs/dqx/checks_storage.py), which is the one with this issue, and the `test_engine_without_observer_no_metrics_saved` in [src/databricks/labs/dqx/tests/integration/test_summary_metrics.py](https://github.com/databrickslabs/dqx/blob/main/tests/integration/test_summary_metrics.py)

image

taylorsweetman: Since it happens in two different places, it might make sense to extract your backtick removal logic into a utility function that is imported in the two places where it is needed. Then this same function can be used in the future for the same purpose.

cornzyblack: Sure, that makes sense, added that. I've extracted that logic into a helper function called `to_api_identifier`

```python
def to_api_identifier(location: str) -> str:
    """
    Transform a Spark SQL table identifier into a Databricks SDK-compatible string.

    This function bridges the gap between Spark's SQL requirements (backticks for
    special characters) and the Databricks Workspace Client's REST requirements
    (literal strings in URL paths). It removes SQL-style backticks and applies
    URL encoding to ensure safety for characters like '$' or '#'.

    Args:
        location (str): The fully qualified table location (e.g., "`cat`.`schema`.`tab`").

    Returns:
        str: A URL-encoded string compatible with Databricks SDK methods like tables.exists().
    """
    return quote(location.replace("`", ""))
```

I also wanted to highlight an alternative approach that doesn't rely on the SDK directly: using Spark's native catalog API. It's a cleaner 'Spark-native' way to check the existence of a table and allows for backticks.

```python
spark.catalog.tableExists("default.tbl1")
```

image

I've stuck with the SDK-based fix for now to minimize changes to the existing architecture, but I'm curious if @mwojtyczka has a preference between the two!"

taylorsweetman (approved):

src/databricks/labs/dqx/checks_storage.py
mwojtyczka: yes, it is better to use `spark.catalog.tableExists`. It handles special characters and backticks correctly so we can avoid custom code for handling this. It is compatible with DBR 14+. I pushed a change for this.

tests/integration/test_config_serializer.py
```diff
@@ -185,3 +185,18 @@ def test_save_run_config_in_custom_folder_installation(ws, installation_ctx_cust
...
+    run_config = RunConfig(name="fake")
+    config = WorkspaceConfig(run_configs=[run_config])
+    # extra configs are skipped
+    config.extra_field = "extra_value"
+
+    product_name = installation_ctx.product_info.product_name()
+    ConfigSerializer(ws).save_config(config=config, assume_user=True, product_name=product_name)
+
+    actual_config = ConfigSerializer(ws).load_config(assume_user=True, product_name=product_name)
+    assert actual_config == config
```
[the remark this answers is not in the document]
mwojtyczka: correct, this is increasing test coverage

PR #1002, opened 2026-01-19
Update handling of metadata columns during schema validation

### LINKED ISSUE #989: [BUG]: has_valid_schema not handling result columns properly

### Current Behavior

We have a custom validation which checks data schema. Since new validation was added into DQX, I'm testing it little bit to see what changes will be needed to be performed as it is logical to bump the version to 0.12.0 and use DQX to validate also schema.

I have prepared a fake data set with the following columns:

```python
SCHEMA = """
    prefix                  string,
    first_name              string,
    last_name               string,
    street_address          string,
    city                    string,
    country                 string,
    country_code            string,
    company                 string,
    credit_card             string,
    credit_card_expire      string,
    language_name           string,
    phone_number            string
"""
```

For the DQX engine initialization I'm passing dictionary of renamed errors and warnings columns:

```python
dq_column_names: dict[str, str] = {
    "errors": "audit_dq_errors",
    "warnings": "audit_dq_warnings"
}
extra_parameters = ExtraParams(result_column_names=dq_column_names)
```

**Use Case 1:**

```yaml
- criticality: warn
  check:
    function: has_valid_schema
    arguments:
      expected_schema: "prefix string, first_name string, last_name string, street_address string, city string, country string, country_code string, company string, credit_card string, credit_card_expire string, language_name string"
      strict: True
# , phone_number string
```

Results:

```
[{'name': 'has_invalid_schema', 'message': 'Schema validation failed: Column 'phone_number' in checked data not present in expected schema; Column 'audit_dq_errors' in checked data not present in expected schema', 'columns': None, 'filter': None, 'function': 'has_valid_schema', 'run_time': 2026-01-07 23:49:33.328091, 'run_id': 'a473b57c-d4ae-4152-a51c-0b89dc1e475e', 'user_metadata': {}}]
```

Error column added by DQX into data frame is present in the validation output.

---

**Use Case 2:**

I tried to work around that issue and I've added definition for `audit_dq_errors`:

```yaml
- criticality: warn
  check:
    function: has_valid_schema
    arguments:
      expected_schema: "prefix string, first_name string, last_name string, street_address string, city string, country string, country_code string, company string, credit_card string, credit_card_expire string, language_name string, audit_dq_errors array<struct<name: string,message: string,columns: array<string>,filter: string,function: string,run_time: timestamp,user_metadata: map<string, string>>>"
      strict: False
```

But then I'm getting in the validation output: `Column 'audit_dq_errors' has incompatible type, expected 'array', got 'array''`. So I thought, maybe I need to just give it array without details, but that completely fails: `The definition of \"ARRAY\" type is incomplete.`

---

**Use Case 3:**

Without the strict mode:

```yaml
- criticality: warn
  check:
    function: has_valid_schema
    arguments:
      expected_schema: "prefix string, first_name string, last_name string, street_address string, city string, country string, country_code string, company string, credit_card string, credit_card_expire string, language_name string"
      strict: False
```

It doesn't show that DF has some extra fields. Which is exactly the scenario I want to check because from time to time we get inputs with additional columns. An even though we can merge schema, I would like to know that somebody is sending more data and maybe we need to do something with those data. With our current solution we are not merging schema though but rather removing he extra column. I plan to change that behavior to merge schema though.

---

**Use Case 4:**

With strict mode but criticality as error:

```yaml
- criticality: error
  check:
    function: has_valid_schema
    arguments:
      expected_schema: "prefix string, first_name string, last_name string, street_address string, city string, country string, country_code string, company string, credit_card string, credit_card_expire string, language_name string"
      strict: True
```

Result:

```
[{'name': 'has_invalid_schema', 'message': 'Schema validation failed: Column 'phone_number' in checked data not present in expected schema', 'columns': None, 'filter': None, 'function': 'has_valid_schema', 'run_time': 2026-01-08 00:34:55.474545, 'run_id': '02121b96-3696-40e0-b2db-1959dd4027fe', 'user_metadata': {}}]
```

So it seems like the problem is when criticality is set to warn, then it incorrectly handles the result columns.

### Expected Behavior

I would expect that DQX result columns are not present in the results array (use case 1). These columns are added by DQX framework and therefore shouldn't be considered for schema comparison. With use case 4, it seems to me that the problem is only when criticality is set to warn. With criticality set to error it works as expected.

I want to have an option to set this to warning so afterwards I can handle these extra columns; i.e. by sending some notification to the development team, indicating that we might need to go back to requester and get more details on extra columns.

### Steps To Reproduce

_No response_

### Cloud

AWS

### Operating System

macOS

### Relevant log output

```shell

```

ghanse:
## Changes

This PR adds capabilities in `DQEngine` to ignore metadata columns and internal columns during dataset-level checks (e.g. `has_valid_schema`) which depend on a stable input schema.

I also adds optional exclude_columns which is list of columns in the checked DataFrame schema to ignore for validation.

Check functions can be decorated to register them for schema preselection:
```
@register_for_original_columns_preselection()
def has_valid_schema(
    expected_schema: str | types.StructType | None = None,
    ref_df_name: str | None = None,
    ref_table: str | None = None,
    columns: list[str | Column] | None = None,
    strict: bool = False,
    exclude_columns: list[str] | None = None,
) -> tuple[Column, Callable]:
...
```

Rules decorated in this way will have original dataframe columns preselected. 

### Linked issues

Resolves #989 

### Tests

- [x] manually tested
- [ ] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests


docs/dqx/docs/reference/quality_checks.mdx
```diff
@@ -2456,6 +2466,25 @@ checks = [
...
+          StructField("id", IntegerType(), True),
+          StructField("name", StringType(), True),
+          StructField("age", IntegerType(), True),
+          StructField("contact_info", StructType([
+            StructField("email", StringType(), True),
+            StructField("phone", StringType(), True),
+            StructField("address", StringType(), True),
+          ]), True)
+        ]),
+        "columns": ["last_update_date", "last_updated_by"],
```
mwojtyczka: shouldn't that be `ignore_columns`?

src/databricks/labs/dqx/engine.py
```diff
@@ -345,6 +346,17 @@ def _all_are_dq_rules(checks: list[DQRule]) -> bool:
         """Check if all elements in the checks list are instances of DQRule."""
         return all(isinstance(check, DQRule) for check in checks)
 
+    def _get_checks_with_ignored_result_columns(self, checks: list[DQRule]) -> list[DQRule]:
```
mwojtyczka: We can support ignored columns, but I believe the fix itself requires a different approach (see suggestion in line 354). There are 2 types of issues:

1. Additional `_errors` column coming from checks with criticality error (always applied before warn)

When `has_valid_schema` is configured with warn criticality, applying error-type checks can add an extra `_errors` column to the result. Your fix will eliminate this. We need an integration test for this that uses the `apply_checks` method.

2. Extra columns added by dataset-level checks (edge case)

Many dataset-level checks generate additional columns during execution (suffixed by `unique_str = uuid.uuid4().hex`). If `has_valid_schema` is applied alongside other dataset-level checks, these extra columns may cause schema validation failures. We need an integration test for this that uses the `apply_checks` method.

src/databricks/labs/dqx/engine.py
```diff
@@ -345,6 +346,17 @@ def _all_are_dq_rules(checks: list[DQRule]) -> bool:
...
+        """Get checks with ignored result columns for schema validation checks."""
+        for check in checks:
+            if check.check_func is not check_funcs.has_valid_schema:
+                continue
+            existing_ignored = check.check_func_kwargs.get("ignore_columns") or []
```
```suggestion
            if not check.check_func_kwargs.get("columns"):
                check.check_func_kwargs["columns"] = df.columns
```
This only eliminates the 1 issue described above. To eliminate both we should leverage `columns` kwargs instead.

src/databricks/labs/dqx/engine.py
```diff
@@ -345,6 +346,17 @@ def _all_are_dq_rules(checks: list[DQRule]) -> bool:
...
+            check.check_func_kwargs["ignore_columns"] = list(
```
```suggestion
```
as explained above

src/databricks/labs/dqx/engine.py
```diff
@@ -345,6 +346,17 @@ def _all_are_dq_rules(checks: list[DQRule]) -> bool:
...
+                set(existing_ignored + list(self._result_column_names.values()))
```
```suggestion
```
as explained above

src/databricks/labs/dqx/engine.py
```diff
@@ -345,6 +346,17 @@ def _all_are_dq_rules(checks: list[DQRule]) -> bool:
...
+            )
```
```suggestion
```
as explained above

src/databricks/labs/dqx/engine.py
```diff
@@ -345,6 +346,17 @@ def _all_are_dq_rules(checks: list[DQRule]) -> bool:
         """Check if all elements in the checks list are instances of DQRule."""
         return all(isinstance(check, DQRule) for check in checks)
 
+    def _get_checks_with_ignored_result_columns(self, checks: list[DQRule]) -> list[DQRule]:
+        """Get checks with ignored result columns for schema validation checks."""
```
```suggestion
        """Determine columns for schema validation checks."""
```

src/databricks/labs/dqx/engine.py
```suggestion
    def _preselect_schema_validation_columns(self, checks: list[DQRule]) -> list[DQRule]:
```

mwojtyczka (changes requested):

tests/integration/test_dataset_checks.py
```diff
@@ -2774,3 +2774,27 @@ def test_has_valid_schema_with_ref_df_name(spark: SparkSession):
...
+    actual_condition_df = actual_apply_df.select("a", "b", "c", "d", condition)
+
+    expected_condition_df = spark.createDataFrame(
+        [
+            ["str1", 1, 100.0, "extra", None],
+            ["str2", 2, 200.0, "data", None],
+        ],
+        "a string, b int, c double, d string, has_invalid_schema string",
+    )
+    assert_df_equality(actual_condition_df, expected_condition_df, ignore_nullable=True)
```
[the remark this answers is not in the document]
mwojtyczka: yes, since the comparison is strict, i would expect schema mismatch.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -2003,7 +2012,8 @@ def apply(df: DataFrame, spark: SparkSession, ref_dfs: dict[str, DataFrame]) ->
         else:
             _expected_schema = expected_schema
 
-        actual_schema = df.select(*columns).schema if columns else df.schema
+        base_df = df.select(*columns) if columns else df
+        actual_schema = base_df.drop(*(ignore_column_names or [])).schema
```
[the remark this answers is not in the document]
mwojtyczka: Yes, I think is also better to prefilter and then select, avoiding select + drop. 
```
selected_columns = columns if columns else df.columns

# Remove ignored columns
if ignore_column_names:
    selected_columns = [c for c in selected_columns if c not in ignore_column_names]

# Select final columns and get schema
actual_schema = df.select(*selected_columns).schema
```

src/databricks/labs/dqx/engine.py
```diff
@@ -345,6 +346,25 @@ def _all_are_dq_rules(checks: list[DQRule]) -> bool:
         """Check if all elements in the checks list are instances of DQRule."""
         return all(isinstance(check, DQRule) for check in checks)
 
+    def _preselect_schema_validation_columns(self, df: DataFrame, checks: list[DQRule]) -> list[DQRule]:
+        """Determine columns for schema validation checks."""
+        for check in checks:
```
mwojtyczka: Can we avoid this loop? Instead, execute this within the existing loop before DQRuleManager is called? before the loop we only need to preserve the original columns (df.columns).

src/databricks/labs/dqx/engine.py
```diff
@@ -345,6 +346,25 @@ def _all_are_dq_rules(checks: list[DQRule]) -> bool:
         """Check if all elements in the checks list are instances of DQRule."""
         return all(isinstance(check, DQRule) for check in checks)
 
+    def _preselect_schema_validation_columns(self, df: DataFrame, checks: list[DQRule]) -> list[DQRule]:
+        """Determine columns for schema validation checks."""
```
mwojtyczka: it would be good to describe the 2 edge case scenarios we try to handle here so that it's clear why we need to do this

src/databricks/labs/dqx/engine.py
```diff
@@ -345,6 +346,25 @@ def _all_are_dq_rules(checks: list[DQRule]) -> bool:
...
+        for check in checks:
+            if check.check_func is not check_funcs.has_valid_schema:
+                continue
+
+            # Default columns to all columns of the current DataFrame if not explicitly set
+            if not check.check_func_kwargs.get("columns"):
+                check.check_func_kwargs["columns"] = [
```
mwojtyczka: it may be cleaner to create new instance of DQDatasetRule(DQRule) to keep things immutable. This would also eliminate the need to invalidate cache.

src/databricks/labs/dqx/engine.py
ghanse: Moved this into the existing loop.

src/databricks/labs/dqx/engine.py
ghanse: I added a more detailed method docstring.

src/databricks/labs/dqx/engine.py
ghanse: I refactored the implementation to use `dataclasses.replace(...)` to keep the instances immutable.

src/databricks/labs/dqx/engine.py
```diff
@@ -345,6 +347,29 @@ def _all_are_dq_rules(checks: list[DQRule]) -> bool:
         """Check if all elements in the checks list are instances of DQRule."""
         return all(isinstance(check, DQRule) for check in checks)
 
+    def _preselect_schema_validation_columns(self, df: DataFrame, check: DQRule) -> DQRule:
+        """
+        Determines the selected columns for schema validation checks. This is required when using `has_valid_schema` to
+        ignore columns added during quality checking. This includes:
+        * DQX result columns (e.g. '_warnings' and '_errors')
+        * Internal columns used for aggregate checks
```
```suggestion
        * Internal columns used for dataset-level checks
```

src/databricks/labs/dqx/engine.py
```diff
@@ -345,6 +347,29 @@ def _all_are_dq_rules(checks: list[DQRule]) -> bool:
...
+        # Default columns to all columns of the current DataFrame if not explicitly set
+        if check.check_func_kwargs.get("columns"):
+            return check
+
+        if check.check_func_args and len(check.check_func_args) >= 3:
+            return check
+
+        rule_kwargs = check.check_func_kwargs.copy()
+        rule_kwargs["columns"] = [col for col in df.columns if col not in set(self._result_column_names.values())]
+        return replace(check, check_func_kwargs=rule_kwargs)
```
mwojtyczka: why don't we have to invalidate check cache anymore when using this approach?

src/databricks/labs/dqx/engine.py
```diff
@@ -345,6 +347,29 @@ def _all_are_dq_rules(checks: list[DQRule]) -> bool:
...
+        Args:
+            df: Input DataFrame
+            check: DQRule to be modified
+        """
+
+        # Default columns to all columns of the current DataFrame if not explicitly set
+        if check.check_func_kwargs.get("columns"):
+            return check
+
+        if check.check_func_args and len(check.check_func_args) >= 3:
```
mwojtyczka: I would remove this condition to make it more generic. We will have to use this function for anomaly detection as well, which requires a list of columns.

There are a couple of issues with using check_func_args here:
* makes an assumption of the underlying implementation. Hard to maintain.
* makes an assumption that the function is executed within the context of has_valid_schema. So there is potential for misuse.

tests/integration/test_dataset_checks.py
```diff
@@ -2774,3 +2774,27 @@ def test_has_valid_schema_with_ref_df_name(spark: SparkSession):
...
+    test_df = spark.createDataFrame(
+        [
+            ["str1", 1, 100.0, "extra"],
+            ["str2", 2, 200.0, "data"],
+        ],
+        "a string, b int, c double, d string",
+    )
+
+    expected_schema = "a string, b int, c double"
+    condition, apply_method = has_valid_schema(expected_schema, ignore_columns=["d"], strict=True)
```
mwojtyczka: provide also add a case where column is provided as spark col to increase test coverage: ignore_columns=[F.col("d")]

src/databricks/labs/dqx/engine.py
```diff
@@ -345,6 +347,29 @@ def _all_are_dq_rules(checks: list[DQRule]) -> bool:
         """Check if all elements in the checks list are instances of DQRule."""
         return all(isinstance(check, DQRule) for check in checks)
 
+    def _preselect_schema_validation_columns(self, df: DataFrame, check: DQRule) -> DQRule:
```
```suggestion
    def _preselect_original_columns(self, df: DataFrame, check: DQRule) -> DQRule:
```
I would rename to make it more generic as there will be other checks we want to use it for (e.g. anomaly detection)

src/databricks/labs/dqx/engine.py
```diff
@@ -389,8 +414,13 @@ def _create_results_array(
         current_df = df
 
         for check in checks:
+            normalized_check = (
```
mwojtyczka: I think we should make it more generic, since more checks will require this in the future. This will also allow the creation of custom functions that operate on the full schema.

```
@register_rule("dataset")
@full_schema_rule  # -> register in FULL_SCHEMA_CHECK_FUNC_REGISTRY
def has_valid_schema(
...

@register_rule("dataset")
@full_schema_rule 
def has_no_anomalies(
...

# any custom check function
@register_rule("dataset")
@full_schema_rule 
def custom_func(
```

```
normalized_check = (
    self._preselect_schema_validation_columns(df, check)
    if check.check_func in FULL_SCHEMA_CHECK_FUNC_REGISTRY
    else check
)
```

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -2003,7 +2014,10 @@ def apply(df: DataFrame, spark: SparkSession, ref_dfs: dict[str, DataFrame]) ->
         else:
             _expected_schema = expected_schema
 
-        actual_schema = df.select(*columns).schema if columns else df.schema
+        selected_column_names = column_names if column_names else df.columns
+        if ignore_column_names:
+            selected_column_names = [col for col in selected_column_names if col not in ignore_column_names]
```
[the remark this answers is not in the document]
mwojtyczka: good feedback

src/databricks/labs/dqx/engine.py
ghanse: I used `inspect` here to handle if the schema is passed as an ordinal argument.

src/databricks/labs/dqx/engine.py
ghanse: Updated

PR #1003, opened 2026-01-20
Fixed resolution of pii detection check function

### LINKED ISSUE #998: [BUG]: does_not_contain_pii function is not working

### Current Behavior

We are getting below error while implementing PII detection DQ function.

Image

Have installed the below library

Image

Import is working fine 

Image

Function Definition 

Image

can you please help on this?

### Expected Behavior

We are getting below error while implementing PII detection DQ function.

Image

Have installed the below library

Image

Import is working fine 

Image

Function Definition 

Image

can you please help on this?

### Steps To Reproduce

_No response_

### Cloud

AWS

### Operating System

macOS

### Relevant log output

```shell

```

mwojtyczka:
## Changes

Fixed resolution of pii detection check function in the engine so that i can be used for applying checks defined as metadata.

### Linked issues

Resolves #998 

### Tests

- [x] manually tested
- [ ] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

### DIFF
--- a/tests/unit/test_resolve_check_function.py
+++ b/tests/unit/test_resolve_check_function.py
@@ -1,6 +1,9 @@
 import textwrap
+import sys
+import importlib
 
 import pytest
+from databricks.labs import dqx
 from databricks.labs.dqx.checks_resolver import resolve_check_function, resolve_custom_check_functions_from_path
 from databricks.labs.dqx.errors import InvalidCheckError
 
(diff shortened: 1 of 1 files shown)


src/databricks/labs/dqx/checks_resolver.py
```diff
@@ -7,6 +7,14 @@
 
 from databricks.labs.dqx import check_funcs
 from databricks.labs.dqx.geo import check_funcs as geo_check_funcs
+
+try:
+    from databricks.labs.dqx.pii import pii_detection_funcs as pii_check_funcs
+
+    PII_ENABLED = True
+except ImportError:
+    PII_ENABLED = False
```
alexott, line 16: Should we add a logging that will say what is incorrect?

mwojtyczka: we could log inside the `__init__` when importing the pii module. Currently we only raise the exception. However, since this check is totally optional and requires installation of pii extras, i'm not sure we should log anything by default. It would be confusing for users if we log this on every execution. We could add DEBUG log maybe. If so we should do it in other places as well where we use this pattern: llm and datacontract.

ghanse (approved):

PR #1012, opened 2026-01-24
Feature/partition clustering table support

### LINKED ISSUE #1007: [FEATURE]: save partitioned/clustered table using save_results_in_table

### Problem statement

I am struggling the find a way to create a table with partition or clustering key information when saving the table first time using save_results_in_table.
if i want to save a table with partition, currently I have to create a blank partitioned table, and then save the results into it.

### Proposed Solution

I would like to have the ability to pass the partition information or the cluster information in the OutputConfig(), and it should be able to create the partitioned, or liquid clustered table

### Additional Context

_No response_

sheeluvikas:
## Changes
This change includes:
1. Added a `partition_by`, and `cluster_by` fields in `OutputConfig`
2. Feature added to save the DataFrame into partitioned or clustered table
3. Liquid Clustered table is created when saving the checks first time in liquid clustered table
4. Updated integration tests to check for both `partition_by` and `cluster_by` in the table
5. Updated the relevant documents to reflect the feature and changes

### Linked issues

Resolves #1007

### Tests
integration test to added to test for the partition_by in the table

- [x] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

### DIFF
--- a/tests/e2e/test_save_dataframe_as_table.py
+++ b/tests/e2e/test_save_dataframe_as_table.py
@@ -5,6 +5,7 @@ We need to implement streaming tests for cluster_by using e2e tests in a noteboo
 which is used to run integration tests, does not support setting the Spark property
 'spark.databricks.delta.liquid.eagerClustering.streaming.enabled'.
 """
+
 import logging
 from pathlib import Path
 


tests/integration/test_io.py
ghanse, line 1: We should also test this for streaming.

src/databricks/labs/dqx/config.py
ghanse, line 1: We should also add `cluster_by` if users want to write liquid clustered data. We will need to validate that only 1 of `partition_by` or `cluster_by` is specified by the user.

src/databricks/labs/dqx/io.py
```diff
@@ -124,6 +129,11 @@ def save_dataframe_as_table(df: DataFrame, output_config: OutputConfig) -> Strea
         return _write_stream(stream_writer, output_config)
 
     batch_writer = df.write.format(output_config.format).mode(output_config.mode).options(**output_config.options)
+    if output_config.partition_by:
+        batch_writer = batch_writer.partitionBy(*output_config.partition_by)
+    if output_config.cluster_by:
+        DeltaTable.createIfNotExists().tableName(output_config.location).addColumns(df.schema).clusterBy(*output_config.cluster_by).execute()
```
sheeluvikas: explore cluster_by auto

sheeluvikas: user should be able to pass `clusterByAuto` as one of the options. Hence I don't think we need to handle that scenario

ghanse: Yes, this is one of the limited scenarios where data layout can be controlled via `options`.

docs/dqx/docs/guide/quality_checks_apply.mdx
```diff
@@ -661,8 +661,8 @@ databricks labs dqx logs --workflow quality-checker
 
 The following fields from the [configuration file](/docs/installation/#configuration-file) are used by the quality checker workflow:
 - `input_config`: configuration for the input data. 'location' is autogenerated when the workflow is executed for patterns.
-- `output_config`: configuration for the output data. 'location' is autogenerated when the workflow is executed for patterns.
-- `quarantine_config`: (optional) configuration for the quarantine data. 'location' is autogenerated when the workflow is executed for patterns.
+- `output_config`: configuration for the output data. 'location' is autogenerated when the workflow is executed for patterns. You can set either `partition_by` or `cluster_by` (only one applies);
```
ghanse: I would change the language for this and the other sections from `(only one applies)` to something like `(only one can be specified)`. Hopefully that will make the behavior clearer to the user.

docs/dqx/docs/reference/engine.mdx
```diff
@@ -97,6 +97,8 @@ The 'Supports local execution' in the above table indicates which methods can be
 * `mode`: The write mode for the output data (`overwrite` or `append`, default is `append`).
 * `options`: Additional options for writing the output data, such as schema merge settings or checkpoint locations for streaming.
 * `trigger`: Optional trigger settings for streaming output, such as `trigger={"availableNow": True}` or `trigger={"processingTime": "10 seconds"}`.
+* `partition_by`: Optional list of columns to partition the output data; use either `partition_by` or `cluster_by`.
+* `cluster_by`: Optional list of columns for Liquid Cluster tables; use either `cluster_by` or `partition_by`.
```
ghanse: Same applies here. I think we should tell users explicitly to only pass `cluster_by` or `partition_by`.

src/databricks/labs/dqx/io.py
ghanse: We should be able to call `batch_writer.clusterBy(...)` directly without pre-creating the table using `DeltaTable.createIfNotExists()`.

src/databricks/labs/dqx/io.py
```diff
@@ -124,6 +129,11 @@ def save_dataframe_as_table(df: DataFrame, output_config: OutputConfig) -> Strea
         return _write_stream(stream_writer, output_config)
```
ghanse, line 140: `cluster_by` is ignored with no indication to the user in this case, right? It may be helpful to warn the user when they try to write a streaming DataFrame with an `OutputConfig` that specifies `cluster_by`.

src/databricks/labs/dqx/io.py
sheeluvikas: you are right !
I will update that. Thanks for the feedback.

src/databricks/labs/dqx/io.py
sheeluvikas: Updated the code

src/databricks/labs/dqx/io.py
sheeluvikas: Updated

ghanse (approved):

src/databricks/labs/dqx/io.py
```diff
@@ -167,6 +191,45 @@ def _write_stream(writer: DataStreamWriter, output_config: OutputConfig) -> Stre
...
+    Checks if the execution environment supports liquid clustering on write with structured streaming.
+
+    Returns:
+        True if the environment supports liquid clustering on write with structured streaming, False otherwise
+
+    Note:
+        The `cluster_by` method of Spark's `DataStreamWriter` is supported for creating liquid clustered tables as of
+        Databricks Runtime version 16+ (see: https://docs.databricks.com/aws/en/delta/clustering#create-tables-with-clustering).
+        Users must manually enable `spark.databricks.delta.liquid.eagerClustering.streaming.enabled` to allow clustering
+        on write for streaming workloads. This method is used in `save_dataframe_as_table` to validate that the
```
mwojtyczka:
```suggestion
        on write for streaming workloads. This method is used to validate that the
```
making it easier for maintenance

src/databricks/labs/dqx/io.py
```diff
@@ -167,6 +191,45 @@ def _write_stream(writer: DataStreamWriter, output_config: OutputConfig) -> Stre
...
+        execution environment supports streaming writes with liquid clustering.
+    """
+    environment_version = os.environ.get("DATABRICKS_RUNTIME_VERSION", "")
```
mwojtyczka, line 208: we should also verify that the spark setting is enabled.

`spark.conf.get("spark.databricks.delta.liquid.eagerClustering.streaming.enabled", "false").lower() == "true"`

pyproject.toml
```diff
@@ -113,8 +113,8 @@ dependencies = [
     "ruff~=0.3.4",
     "types-PyYAML~=6.0.3",
     "types-requests~=2.31.0",
-    "databricks-connect~=15.4",
-    "pyspark~=3.5.0",
+    "databricks-connect~=16.4",
+    "pyspark~=4.0.0",
```
mwojtyczka: we need to revert this. It's ok to upgrade the databricks-connect version for testing, but we should continue using pyspark 3.5.x as dependency, because spark 4 is not compatible with databricks-connect 16.4, see:
 https://docs.databricks.com/aws/en/release-notes/dbconnect/#databricks-connect-1640-scala

mwojtyczka: Why do we have to upgrade to pyspark 4? if we do that we also need to further bump the databricks-connect

ghanse: I think we need this because `clusterBy` is only supported for `DataStreamReader` as of PySpark 4.0.

ghanse: This would be validated at runtime by `_supports_streaming_cluster_on_write()`. We need PySpark 4.0 for the static analysis.

mwojtyczka: got it then, let's upgrade the test cluster first

mwojtyczka: let's keep the change, we should have newer testing clusters in few days then we can we continue with this

src/databricks/labs/dqx/io.py
mwojtyczka: @sheeluvikas can you look at this? one problem is that we need access to spark conf. But it would be nice to check this

pyproject.toml
mwojtyczka: This has been resolved now

src/databricks/labs/dqx/io.py
sheeluvikas: sure @mwojtyczka, I will work on this.

src/databricks/labs/dqx/io.py
```diff
@@ -84,7 +85,9 @@ def _read_table_data(spark: SparkSession, input_config: InputConfig) -> DataFram
     return spark.readStream.options(**input_config.options).table(input_config.location)
 
 
-def save_dataframe_as_table(df: DataFrame, output_config: OutputConfig) -> StreamingQuery | None:
+def save_dataframe_as_table(
+    df: DataFrame, output_config: OutputConfig, spark: SparkSession | None = None
```
sheeluvikas: @ghanse do you think it would be better to get the spark from the df itself ?

tests/conftest.py
```diff
@@ -31,7 +31,7 @@
 logger = logging.getLogger(__name__)
 
 
-TEST_CATALOG = "dqx"
+TEST_CATALOG = "gh_catalog_001"
```
mwojtyczka:
```suggestion
TEST_CATALOG = "dqx"
```

tests/conftest.py
```diff
@@ -80,6 +80,15 @@ def skip_if_classic_compute(debug_env):
         pytest.skip("This test requires a serverless compute cluster")
 
 
+@pytest.fixture
+def skip_if_serverless_compute(debug_env):
```
mwojtyczka: Not used anywhere

tests/e2e/notebooks/save_dataframe_as_table_notebook.py
```diff
@@ -0,0 +1,83 @@
+# Databricks notebook source
+
```
mwojtyczka, line 2: this is already covered by integration tests. Is there any reason we need e2e tests?

PR #1017, opened 2026-01-31
Add Decimal support to min_max generator (#1013)

### LINKED ISSUE #1013: [FEATURE]: Improve Generator to support Decimal data in min/max checks

### Problem statement

The `dq_generate_min_max` from `DQGenerator` class only captures checks if the min and max values are from int and float data types.

    def _is_num(value):
         return isinstance(value, (int, float)) 

Looking to the conversation of the pull request to resolve issue https://github.com/databrickslabs/dqx/pull/624, it seems like it would be implemented too, but somehow it's not in the final code.

### Proposed Solution

Make `dq_generate_min_max` from `DQGenerator` class to generate checks to verify min and max values from Decimal data Types.

### Additional Context

_No response_

Jgprog117:
## Summary

Resolves #1013

* Extended `dq_generate_min_max` method to support Python's `Decimal` type in addition to `int` and `float` for min/max validation checks.
* Extended check functions to support Decimal values for limits
* Extended serialization and deserialization of checks to support Decimal during round trip

## Problem

The current implementation of `dq_generate_min_max` only recognizes `int` and `float` as numeric types:

```python
def _is_num(value):
    return isinstance(value, (int, float))
```

## Changes

- Import Decimal from decimal module
- Update _is_num() to include Decimal in isinstance check
- Add comprehensive unit tests for Decimal, int, and float types

This enables proper data quality checks for decimal-precise financial and scientific data where floating-point precision issues would cause false positives.

### Linked issues

Resolves #1013

### Tests

- [X] manually tested
- [X] added unit tests
- [X] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

### DIFF
--- a/docs/dqx/docs/guide/quality_checks_storage.mdx
+++ b/docs/dqx/docs/guide/quality_checks_storage.mdx
@@ -167,7 +167,7 @@ You can load quality checks from various storage locations. You can then apply t
 
 <Admonition type="tip" title="Checks definition">
 For the save methods to work, checks must be defined as a list of dictionaries.
-If you create checks as a list of DQRule objects, you can convert them using the `serialize_checks` method, as described [here](/docs/reference/quality_checks/#converting-dqx-classes-to-metadata).
+If you create checks as a list of DQRule objects, you can convert them using the `serialize_checks` function, as described [here](/docs/reference/quality_checks/#converting-dqx-classes-to-metadata).
 </Admonition>
 
 <Tabs>


tests/unit/test_generator_numeric.py
ghanse, line 1: Can you also add a minimal integration test in `test_generator.py`?

ghanse (changes requested):
Looking good. Requesting some extra tests.

tests/unit/test_generator_numeric.py
```diff
@@ -0,0 +1,61 @@
...
+    assert args["min_limit"] == -273.15
+    assert args["max_limit"] == 1000.0
+
+
+def test_mixed_numeric_types_not_generated():
+    """Test that mixing int/float with Decimal in the same rule returns None."""
+    # int min, Decimal max - should not generate a rule due to type mismatch
+    result = DQGenerator.dq_generate_min_max("mixed_col", **{"min": 10, "max": Decimal("100.00")})
+    # The _same_family check should prevent this combination
+    assert result is None or result["check"]["function"] in ["is_not_greater_than"]
```
ghanse: Shouldn't `_same_family` allow this since both Decimal and integers are considered numeric by `_is_num`?

Jgprog117: Yep, my bad. Changed the test to assert is_in_range for mixed int+Decimal input instead.

tests/unit/test_generator_numeric.py
Jgprog117: Added a Decimal min_max DQProfile test to test_rules with expected output.

tests/integration/test_generator.py
```diff
@@ -156,6 +171,14 @@ def test_generate_dq_rules_warn(ws, spark):
             "name": "d1_isnt_in_range",
             "criticality": "warn",
         },
+        {
+            "check": {
+                "function": "is_in_range",
+                "arguments": {"column": "price", "min_limit": Decimal("0.01"), "max_limit": Decimal("999.99")},
```
ghanse, line 177: I suppose we should also update `is_in_range` and other check functions to explicitly accept `Decimal` arguments.

mwojtyczka: yes, we must add Decimal to the checks as well: `is_in_range`, `is_not_greater_than`, `is_not_less_than`. For consistency, I would also add it to the following: `is_equal_to`, `is_not_equal_to`, `is_not_in_range`, `is_aggr_not_greater_than`, `is_aggr_not_less_than`, `is_aggr_equal`, `is_aggr_not_equal`. This is a simple change

Jgprog117: Added Decimal to all the checks mentioned above :).

mwojtyczka: can you please also add integration tests for each of these check functions where we added Decimal to ensure they actually work with Decimal? easiest probably to extend one of the existing tests by adding decimal column

src/databricks/labs/dqx/profiler/generator.py
```diff
@@ -253,7 +254,7 @@ def dq_generate_min_max(column: str, criticality: str = "error", **params: dict)
             return None
 
         def _is_num(value):
-            return isinstance(value, (int, float))
+            return isinstance(value, (int, float, Decimal))
```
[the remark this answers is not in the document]
ghanse: We call `json.dumps` on dictionary-encoded rules in `checks_serializer.deserialize_checks_to_dataframe`. This will break if the user tries to deserialize any checks with Decimal arguments from a dictionary to a DataFrame (for example to save checks to a table).

I think we have a few options:

1. Use a JSON library with built-in support for Decimal, e.g. [simplejson](https://simplejson.readthedocs.io/en/latest/).
2. Create custom methods to serialize/deserialize Decimal. For example:

```
import json
from decimal import Decimal

class DecimalEncoder(json.JSONEncoder):
    def default(self, obj):
        if isinstance(obj, Decimal):
            return {"__type__": "Decimal", "value": str(obj)}
        return super().default(obj)

def decimal_object_hook(dct):
    if dct.get("__type__") == "Decimal":
        return Decimal(dct["value"])
    return dct

def dumps_decimal(obj) -> str:
    return json.dumps(obj, cls=DecimalEncoder)

def loads_decimal(s):
    return json.loads(s, object_hook=decimal_object_hook)
```

src/databricks/labs/dqx/checks_serializer.py
```diff
@@ -25,263 +27,426 @@
...
+            extension: File extension (e.g., ".json", ".yaml", ".yml").
+        Returns:
+            Serialized checks as bytes.
+        """
+        serializer = SerializerFactory.create_serializer(extension)
+        normalized_checks = ChecksNormalizer.normalize(checks)
+        serialized_str = serializer.serialize(normalized_checks)
+        return serialized_str.encode("utf-8")
+
+
```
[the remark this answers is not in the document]
mwojtyczka: not relevant, users should use load_checks and save_checks

PR #1021, opened 2026-02-02
Fixed serialization/deserialization of row filter parameter for dataset-level rules

bpm1993:
## Changes

This PR fixes a bug in the serialization and deserialization of dataset-level rules that use the `row_filter` argument. 

**Root cause:** During serialization (`serialize_checks_from_dataframe`), the `row_filter` parameter is intentionally excluded from the serialized arguments because it's meant to be resolved from the `filter` field. However, the deserialization logic was not performing the inverse operation - it wasn't restoring the `row_filter` from the `filter` field when deserializing dataset rules.

**The fix:** `row_filter` and not excluded from serialized arguments.. Handling of row filter is only done within the rule. The existing behaviour is preserved: the `filter` condition is pushed down as `row_filter` if the check function supports it. 

### Tests

- [x] manually tested - library demo
- [x] added unit tests
- [x] added integration tests - run from local
- [ ] added end-to-end tests
- [ ] added performance tests


src/databricks/labs/dqx/checks_serializer.py
```diff
@@ -218,7 +219,7 @@ def deserialize_checks(checks: list[dict], custom_checks: dict[str, Callable] |
                         check_func_kwargs=check_func_kwargs,
                         name=name,
                         criticality=criticality,
-                        filter=filter_str,
+                        filter=filter_str or row_filter,
```
ghanse: I would update this to something like:

```
filter_str = filter_str or check_func_kwargs.get("row_filter", None)
dq_rule_checks.append(
                    DQDatasetRule(
                        column=column,
                        check_func_kwargs=check_func_kwargs,
                        name=name,
                        criticality=criticality,
                        filter=filter_str,
                        ...
)
```

mwojtyczka: this has to be solved differently. Preserve all params, and handle the row filter behaviour in the rules.

PR #1025, opened 2026-02-06
Agent skills for Databricks Assistant

grusin-db:
# PR: Add `install_skills()` -- deploy DQX agent skills to Databricks Assistant

**Branch:** `f/agent_skills` → `main` | **3 commits** | **+2109 lines**

---

### 1. New `install_skills()` command (`src/databricks/labs/dqx/skills.py`)

One-liner to deploy DQX agent skills into a Databricks workspace so that Databricks Assistant can use them.

```python
from databricks.labs import dqx

dqx.install_skills()                        # → /Users/{me}/.assistant/skills/dqx/
dqx.install_skills("/Shared/skills/dqx")    # custom folder
```

Reads bundled SKILL.md files and example scripts from package resources via `importlib.resources`, then uploads them with `WorkspaceClient.workspace.upload()`.

### 2. Bundled skill content (`src/databricks/labs/dqx/resources/skills/`)

Six SKILL.md files covering the full DQX surface, structured for the assistant's skill discovery:

| Skill | Scope |
|-------|-------|
| `dqx/SKILL.md` | Root: install, quick-start, routing to sub-skills |
| `checks/SKILL.md` | Check authoring, apply, split, reference DataFrames |
| `checks/row-level/SKILL.md` | 30+ row-level functions with **full parameter reference** (required vs optional) |
| `checks/dataset-level/SKILL.md` | 10+ dataset-level functions with **full parameter reference** |
| `checks/custom/SKILL.md` | SQL expressions, Python row/dataset checks |
| `profiling/SKILL.md` | DQProfiler + manual SQL profiling |

Each SKILL.md includes a parameter table at the top that explicitly marks required vs optional arguments to prevent the assistant from generating invalid check definitions (e.g. calling `is_in_range` without both `min_limit` and `max_limit`).

### 3. Runnable examples (`checks/examples/`)

21 self-contained Python scripts (one per check category) bundled as package resources and deployed alongside skills:

- `01`–`09`: row-level checks (null, list, comparison, range, regex, datetime, network, SQL, complex types)
- `10`–`16`: dataset-level checks (unique, aggregation, FK, compare, freshness, schema, SQL query)
- `17`–`19`: custom checks (SQL window, Python row, Python dataset)
- `20`–`21`: profiling

All SKILL.md files cross-reference relevant examples.

### 4. Package wiring (`__init__.py`, `resources/__init__.py`)

`install_skills` is exported from `databricks.labs.dqx.__init__`, so both import styles work:

```python
from databricks.labs.dqx import install_skills
# or
from databricks.labs import dqx; dqx.install_skills()
```

No `pyproject.toml` build changes needed -- hatchling already includes non-Python files under `src/`.

---

## ⚠️ Breaking Changes

None. This is purely additive.

### Databricks AI Assistant with skills loaded

image

...

image

### Unskilled Databricks AI Assistant
 
image


mwojtyczka (changes requested):
This is really cool! 

Feedback:
* Please sign the commits. 
* We need to reduce the maintenance effort. All the examples should already be available in the docs. We need a better way to manage this to avoid duplication. E.g. store examples in one place and source it for both docs and skills from there; document the examples in each check function. The examples on how to use DQX are also already in the docs. We need to centralize this.

src/databricks/labs/dqx/skills.py
```diff
@@ -0,0 +1,49 @@
...
+
+    Copies bundled SKILL.md files and example notebooks from package
+    resources to the workspace so that Databricks Assistant can discover
+    and use them.
+
+    Args:
+        folder: Target workspace folder. Defaults to
+            ``/Users/{current_user}/.assistant/skills/dqx``.
+    """
+    ws = WorkspaceClient()
```
mwojtyczka, line 23: **`WorkspaceClient()` is not injectable — blocks unit testing**

Constructing the client inside the function means tests must either hit a real Databricks workspace or monkey-patch the constructor. The rest of the codebase (e.g. `DQEngine`, `SettingsManager`) accepts `ws` as a constructor argument precisely to allow mocking.

Suggested signature:

```python
def install_skills(folder: str | None = None, ws: WorkspaceClient | None = None) -> None:
    if ws is None:
        ws = WorkspaceClient()
```

This keeps the zero-argument convenience call while enabling `install_skills(ws=mock_ws)` in tests.

src/databricks/labs/dqx/skills.py
```diff
@@ -0,0 +1,49 @@
...
+
+    if folder is None:
+        user_name = ws.current_user.me().user_name
+        folder = f"/Users/{user_name}/.assistant/skills/dqx"
+
+    skills_root = Path(str(files("databricks.labs.dqx.resources") / "skills" / "dqx"))
```
mwojtyczka, line 29: **`Path(str(Traversable))` + `rglob` breaks for zip-packaged installs**

`files(...)` returns an `importlib.resources.abc.Traversable`, not necessarily a real filesystem `Path`. Converting it with `Path(str(...))` works for editable installs and extracted wheels, but silently fails (or raises `NotADirectoryError`) when the package is imported from inside a zip archive (e.g., a compiled `.egg` or certain CI environments).

For zip-safe resource access the idiomatic approach is to use `Traversable` APIs throughout:

```python
from importlib.resources import files

skills_root = files("databricks.labs.dqx.resources").joinpath("skills").joinpath("dqx")

def _rglob_traversable(node):
    for child in node.iterdir():
        if child.is_dir():
            yield from _rglob_traversable(child)
        else:
            yield child

deploy_files = sorted(_rglob_traversable(skills_root), key=lambda f: str(f))
```

This avoids the `Path` conversion entirely and works correctly regardless of how the package is packaged.

src/databricks/labs/dqx/skills.py
```diff
@@ -0,0 +1,49 @@
...
+
+    if not deploy_files:
+        logger.warning("No skill files found in package resources")
+        return
+
+    logger.info(f"Deploying skills to: {folder}")
+
+    for src_file in deploy_files:
+        relative = src_file.relative_to(skills_root)
+        target_path = f"{folder}/{relative}"
```
mwojtyczka, line 42: **`Path` object in f-string uses OS-native separators — breaks on Windows**

`relative` is a `pathlib.Path`. On Windows, `str(relative)` produces backslash-separated paths (e.g., `checks\examples\01_row_null_empty.py`), so the target workspace path becomes `/.../checks\examples\01_row_null_empty.py`, which the Databricks API will reject.

Use `relative.as_posix()` to always get forward slashes:

```python
target_path = f"{folder}/{relative.as_posix()}"
```

src/databricks/labs/dqx/skills.py
```diff
@@ -0,0 +1,49 @@
...
+        target_dir = "/".join(target_path.split("/")[:-1])
+
+        ws.workspace.mkdirs(target_dir)
```
mwojtyczka, line 45: **`mkdirs` is called once per file instead of once per unique directory**

With 25 example files across ~5 directories, this makes 25 `mkdirs` API calls when only 5 are needed. `mkdirs` is idempotent so it doesn't produce incorrect results, but it generates unnecessary network round-trips.

Collect the unique parent directories before the loop:

```python
unique_dirs = {f"{folder}/{Path(f).as_posix()}".rsplit("/", 1)[0] for f in deploy_files_posix}
for d in sorted(unique_dirs):
    ws.workspace.mkdirs(d)

for src_file in deploy_files:
    ...
    ws.workspace.upload(target_path, ...)
```

or more simply, track which directories have already been created in a `set` inside the loop.

PR #1034, opened 2026-02-12
Fixed App LLM Configuration

mwojtyczka:
## Changes

* Fixed issues in configuring Dspy, to use current token on every request. This fixes issue in the App where current OBO token should be used on every request
* Updated CI/CD to use DBR 17.3 LTS cluster for non-serverless tests execution (use `DATABRICKS_DQX_CLUSTER_ID` if set in the vault)
* Updated databricks-connect to 17.3 and spark 4.0 for testing (required for summary metrics)
* Updated contributing guide
* Updated readme of the app

### Tests

- [x] manually tested
- [ ] added unit tests
- [ ] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests


tests/integration/test_row_checks.py
ghanse, line 1: Are these changes due to the upgrade to Spark 4?

PR #1037, opened 2026-02-17
Fixed profilier rules for empty strings and nulls

### LINKED ISSUE #1036: [BUG]: Fix profiling rules for empty string values

### Current Behavior

The DQX profiler currently generates `is_not_null_or_empty` profiles based on the `max_empty_ratio` specified by the user. This can lead to an issue where an `is_not_null_or_empty` profile is created for a column containing a low ratio of empty values but a high ratio of null values.

### Expected Behavior

Profiles should be generated according to the following rules:

1. If the observed null ratio is less than `max_null_ratio`, create a `is_not_null` profile.
2. If the observed null ratio is less than `max_empty_ratio`, create a `is_not_empty` profile.
3. If both ratios are less than their respective options, create a `is_not_null_or_empty` profile.

### Steps To Reproduce

_No response_

### Cloud

AWS

### Operating System

macOS

### Relevant log output

```shell

```

howardwu-db:
## Changes

Fixed profiler section for empty values and nulls. 

### Linked issues

Resolves #1036 

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests


src/databricks/labs/dqx/profiler/profiler.py
```diff
@@ -475,26 +476,47 @@ def _calculate_metrics(
...
-                        description=f"Column {field_name} has {null_percentage * 100:.1f}% of null values "
-                        f"(allowed {max_nulls * 100:.1f}%)",
-                        filter=opts.get("filter", None),
-                    )
+
+        count_non_empty = dst.filter(F.col(field_name) == "").count()
+        null_percentage = 1 - (1.0 * count_non_null) / total_count
+        empty_percentage = 1 - (1.0 * count_non_empty) / total_count
+
+        if typ == T.StringType() and count_non_empty >= (total_count *(1 - max_empty_ratio)) and count_non_null >= (total_count * (1 - max_null_ratio)):
```
mwojtyczka, line 484: can you please refactor this to a separate methods for readability and document as per https://databrickslabs.github.io/dqx/docs/dev/contributing/#first-contribution

src/databricks/labs/dqx/profiler/generator.py
```diff
@@ -331,6 +331,28 @@ def dq_generate_is_not_null(column: str, criticality: str = "error", **params: d
...
+
+        Returns:
+                A dictionary representing the data quality rule.
+
+        """
+        params = params or {}
+
+        return {
+            "check": {"function": "is_not_empty", "arguments": {"column": column}},
+            "name": f"{column}_is_empty",
```
mwojtyczka, line 352: **Rule name is wrong — will generate broken rule names**

The generated rule name is `f"{column}_is_empty"` but the check function is `is_not_empty`. Every other generator method uses the check-function name as the suffix (e.g. `dq_generate_is_not_null` → `{column}_is_not_null`). Should be:

```python
"name": f"{column}_is_not_empty",
```

src/databricks/labs/dqx/profiler/profiler.py
```diff
@@ -475,26 +476,47 @@ def _calculate_metrics(
...
-                dq_rules.append(
-                    DQProfile(
-                        name="is_not_null",
-                        column=field_name,
-                        description=f"Column {field_name} has {null_percentage * 100:.1f}% of null values "
-                        f"(allowed {max_nulls * 100:.1f}%)",
-                        filter=opts.get("filter", None),
-                    )
+
+        count_non_empty = dst.filter(F.col(field_name) == "").count()
```
mwojtyczka, line 480: **`count_non_empty` is a misnomer — it stores the count of EMPTY strings**

`dst.filter(F.col(field_name) == "")` returns rows where the value **is** the empty string, so `count_non_empty` actually holds the empty-string count. This wrong name ripples into every downstream condition and description. Rename to `count_empty`:

```python
count_empty = dst.filter(F.col(field_name) == "").count()
```

And update all references accordingly. This single naming error is the root cause of the two bugs described in the comments on lines 482 and 484.

src/databricks/labs/dqx/profiler/profiler.py
```diff
@@ -475,26 +476,47 @@ def _calculate_metrics(
...
+        null_percentage = 1 - (1.0 * count_non_null) / total_count
+        empty_percentage = 1 - (1.0 * count_non_empty) / total_count
```
mwojtyczka, line 482: **`empty_percentage` computes the NON-empty fraction, not the empty fraction**

```python
# current (wrong):
empty_percentage = 1 - (1.0 * count_non_empty) / total_count
# count_non_empty is actually count_empty, so this gives: 1 - (empty_count / total)
# = fraction of NON-empty rows
```

The description then displays this as `"has X% of empty values"`, which shows the inverse. Should be:

```python
empty_percentage = (1.0 * count_empty) / total_count  # fraction of empty rows
```

src/databricks/labs/dqx/profiler/profiler.py
mwojtyczka, line 484: **Condition for `is_not_null_or_empty` is inverted — rule is never emitted under normal conditions**

```python
# current (wrong):
if ... and count_non_empty >= (total_count * (1 - max_empty_ratio)) ...
# count_non_empty is actually count_empty, so with default max_empty_ratio=0:
# → count_empty >= total_count → only true when EVERY row is an empty string
```

The original intent (and what the old code did) was: emit the rule when the empty-string count is **within** the allowed threshold, so the rule can enforce that going forward. Should be:

```python
if typ == T.StringType() and count_empty <= (total_count * max_empty_ratio) and count_non_null >= (total_count * (1 - max_null_ratio)):
```

Same inversion issue affects the third `elif` branch (line ~506): `count_non_empty <= (metrics["count"] * max_empty_ratio)` is also backwards once you rename `count_non_empty` → `count_empty`.

mwojtyczka (commented):
The intent here is good — separating `is_not_null`, `is_not_empty`, and `is_not_null_or_empty` into explicit branches is cleaner than the previous approach. However there are several correctness bugs that need fixing before merge, and no tests are included for the new behaviour. See inline comments.

PR #1042, opened 2026-02-21
Feature/fix schema validation

### LINKED ISSUE #1039: [BUG]: has_valid_schema doesn't detect missing columns

### Current Behavior

has_valid_schema should detect missing columns in the data frame. Doesn't work with strict = True nor, strict = False. 

Currently the only way to detect if the field is missing is to apply different function on the column; i.e. is_not_null. However, this is not functionally correct as the column might be optional by design.

### Expected Behavior

has_valid_schema should detect missing columns in strict and non-strict mode.

### Steps To Reproduce

**Strict Mode:**

```python
import yaml

checks = yaml.safe_load("""
- criticality: warn
  check:
    function: has_valid_schema
    arguments:
      expected_schema: "a string, b string, bb string, c string, d string"
      strict: True
""")

test_df = spark.createDataFrame(
    [
        ["str1", "1", "100.0"],
        ["str2", "2", "200.0"],
        ["str3", "3", "300.0"],
    ],
    "a string, b string, c string",
)

status = DQEngine.validate_checks(checks)
assert not status.has_errors

dq_engine = DQEngine(WorkspaceClient())

valid_and_quarantine_df = dq_engine.apply_checks_by_metadata(test_df, checks)
display(valid_and_quarantine_df)
```

**Non-Strict Mode:**

```python
import yaml

checks = yaml.safe_load("""
- criticality: warn
  check:
    function: has_valid_schema
    arguments:
      expected_schema: "a string, b string, bb string, c string, d string"
      strict: False
""")

test_df = spark.createDataFrame(
    [
        ["str1", "1", "100.0"],
        ["str2", "2", "200.0"],
        ["str3", "3", "300.0"],
    ],
    "a string, b string, c string",
)

status = DQEngine.validate_checks(checks)
assert not status.has_errors

dq_engine = DQEngine(WorkspaceClient())

valid_and_quarantine_df = dq_engine.apply_checks_by_metadata(test_df, checks)
display(valid_and_quarantine_df)
```

**Applying Other Check on Missing Column:**

```python
import yaml

checks = yaml.safe_load("""
- criticality: warn
  check:
    function: has_valid_schema
    arguments:
      expected_schema: "a string, b string, c string, d string"
      strict: True

- criticality: error
  check:
    function: is_not_null
    for_each_column:
      - d
""")

test_df = spark.createDataFrame(
    [
        ["str1", "1", "100.0"],
        ["str2", "2", "200.0"],
        ["str3", "3", "300.0"],
    ],
    "a string, b string, c string",
)

status = DQEngine.validate_checks(checks)
assert not status.has_errors

dq_engine = DQEngine(WorkspaceClient())

valid_and_quarantine_df = dq_engine.apply_checks_by_metadata(test_df, checks)
display(valid_and_quarantine_df)
```

### Cloud

AWS

### Operating System

Linux

### Relevant log output

**EDIT:**

Image

fedeflowers:
## Changes
* Fixed a critical bug where `has_valid_schema` silently skipped validation for columns missing from the checked DataFrame.

### Linked issues
Resolves #1039

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

### DIFF
--- a/.github/scripts/fork-sync-pr.sh
+++ b/.github/scripts/fork-sync-pr.sh
@@ -1,0 +1,92 @@
+#!/usr/bin/env bash
+#
+# Sync a fork PR to a branch in databrickslabs/dqx and open a test PR.
+# Acceptance, anomaly, and performance workflows are skipped for PRs from forks.
+# This script pushes the fork's code to a branch in the main repo so those
+# tests can run on the new PR.
+#
+# Usage:
+#   .github/scripts/fork-sync-pr.sh <PR_NUMBER>
+#
+# Prerequisites:
+#   - gh CLI installed and authenticated (gh auth login)
+#   - Run from a clone of databrickslabs/dqx (or your fork with upstream configured)
+#
+# First run: Creates fork-test/pr-<N> branch and opens a test PR.
+# Subsequent runs: Syncs latest changes from the fork PR to the same branch (force push).
+#
+set -e
+
+UPSTREAM_REPO="databrickslabs/dqx"
+BASE_BRANCH="main"
+
+if [ -z "$1" ]; then
+  echo "Usage: $0 <PR_NUMBER>"
+  echo ""
+  echo "Example: $0 123"
+  echo ""
+  echo "Syncs fork PR #123 to branch fork-test/pr-123 in $UPSTREAM_REPO,"
+  echo "creating a test PR so acceptance/anomaly/perf tests can run."
+  exit 1
+fi
+
+PR_NUMBER="$1"
+SYNC_BRANCH="fork-test/pr-${PR_NUMBER}"
+
+# Verify we're in a git repo
+if ! git rev-parse --git-dir >/dev/null 2>&1; then
+  echo "Error: Not in a git repository. Run from a clone of $UPSTREAM_REPO or your fork."
+  exit 1
+fi
+
+# Verify gh is installed
+if ! command -v gh >/dev/null 2>&1; then
+  echo "Error: gh CLI is required. Install from https://cli.github.com/"
+  exit 1
+fi
+
+# Ensure upstream remote points to main repo
+if ! git remote get-url upstream >/dev/null 2>&1; then
+  echo "Adding upstream remote: $UPSTREAM_REPO"
+  git remote add upstream "https://github.com/${UPSTREAM_REPO}.git"
+fi
+
+UPSTREAM_URL=$(git remote get-url upstream)
+if [[ "$UPSTREAM_URL" != *"${UPSTREAM_REPO}"* ]]; then
+  echo "Error: upstream remote does not point to $UPSTREAM_REPO"
+  echo "  current: $UPSTREAM_URL"
+  exit 1
+fi
+
+# Fetch and checkout the fork PR branch (PR lives in upstream repo)
+echo "Fetching PR #${PR_NUMBER} from fork..."
+gh pr checkout "$PR_NUMBER" --repo "$UPSTREAM_REPO"
+
+# Create or update sync branch
+git checkout -B "$SYNC_BRANCH"
+
+# Push to upstream (force to overwrite with latest from fork)
+echo "Pushing to upstream/${SYNC_BRANCH}..."
+git push --force upstream "HEAD:${SYNC_BRANCH}"
+
+# Create test PR if it does not exist
+EXISTING=$(gh pr list --repo "$UPSTREAM_REPO" --head "$SYNC_BRANCH" --state open --json number -q '.[0].number' 2>/dev/null || true)
+if [ -z "$EXISTING" ]; then
+  PR_URL=$(gh pr view "$PR_NUMBER" --repo "$UPSTREAM_REPO" --json url -q '.url')
+  PR_TITLE=$(gh pr view "$PR_NUMBER" --repo "$UPSTREAM_REPO" --json title -q '.title')
+  TEST_PR_TITLE="Fork test: PR #${PR_NUMBER} - ${PR_TITLE}"
+  echo "Creating test PR..."
+  gh pr create --repo "$UPSTREAM_REPO" \
+    --base "$BASE_BRANCH" \
+    --head "$SYNC_BRANCH" \
+    --title "$TEST_PR_TITLE" \
+    --body "Automated sync from fork PR for CI testing.
+
+Original PR: ${PR_URL}
+
+All tests, including acceptance, anomaly, and performance tests, run on this PR (they are skipped for fork PRs)."
+  echo "Test PR created."
+else
+  echo "Test PR already exists: #${EXISTING}"
+  echo "Branch ${SYNC_BRANCH} has been updated with latest changes from fork PR #${PR_NUMBER}."
+fi
--- a/.github/workflows/acceptance.yml
+++ b/.github/workflows/acceptance.yml
@@ -1,7 +1,7 @@
 name: acceptance
 
 on:
-  pull_request_target:
+  pull_request:
     types: [ opened, synchronize, ready_for_review ]
   merge_group:
     types: [ checks_requested ]
@@ -15,20 +15,22 @@ permissions:
   pull-requests: write
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: false  # don't cancel ongoing runs to ensure fixtures are completed and resources terminated
 
 jobs:
 
   integration:
-    if: github.event_name == 'pull_request_target' && !github.event.pull_request.draft
+    # Only run this job for PRs from branches on the main repository and not from forks.
+    # Workflows triggered by PRs from forks don't have access to the tool environment.
+    # PRs from forks to be tested by the reviewer(s) / maintainer(s) before merging.
+    if: github.event_name == 'pull_request' && !github.event.pull_request.draft && !github.event.pull_request.head.repo.fork
     environment: tool
     runs-on: larger
     steps:
       - name: Checkout Code
         uses: actions/checkout@v4
         with:
-          ref: refs/pull/${{ github.event.pull_request.number }}/merge  # ensures checkout works for both repo branches and forks
           fetch-depth: 0
 
       - name: Install Python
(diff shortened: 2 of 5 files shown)


src/databricks/labs/dqx/check_funcs.py
ghanse, line 1: IMO, the right fix is to apply column selection only to the target schema when in _strict_ mode.

We say in our documentation that _strict_ mode enforces an exact match between the source and target schemas:

```
strict: Whether to perform strict schema validation (default: False).
  - False: Validates that all expected columns exist with compatible types (allows extra columns)
  - True: Validates exact schema match (same columns, same order, same types)
```

Column filtering uses the `_get_schema` method. We can modify this method to receive the `strict` parameter and conditionally filter the expected columns.

akshan-main: I think the missing-column bug happens because the engine auto-fills columns for has_valid_schema, and that ends up filtering the expected schema down to only columns that already exist in the DF. So, missing expected cols never get checked. Because of that, I agree with removing the auto-preselection here (or changing it to use exclude_columns instead).

fedeflowers: > IMO, the right fix is to apply column selection only to the target schema when in _strict_ mode.
> 
> We say in our documentation that _strict_ mode enforces an exact match between the source and target schemas:
> 
> ```
> strict: Whether to perform strict schema validation (default: False).
>   - False: Validates that all expected columns exist with compatible types (allows extra columns)
>   - True: Validates exact schema match (same columns, same order, same types)
> ```
> 
> Column filtering uses the `_get_schema` method. We can modify this method to receive the `strict` parameter and conditionally filter the expected columns.

@ghanse While I understand the intent to conditionally filter based on `strict` mode, modifying `_get_schema` doesn't resolve the bug because of how the engine auto-injects arguments behind the scenes.

As @akshan-main correctly pointed out, the root cause is a semantic conflict with the `columns` argument. The `@register_for_original_columns_preselection` decorator forces the `DQEngine` to auto-inject the `columns` kwarg into the check with a list of *all existing columns* in the DataFrame so DQX internal columns are ignored.

However, `has_valid_schema` already uses the `columns` keyword argument to define a specific subset of columns to be validated. Because of the decorator, this user configuration is overridden. When `_get_schema()` is called with this auto-injected list of existing columns, it actively filters the `expected_schema` down to only the columns that currently exist in the DataFrame. 

Because of this, today, even in `strict=False` (permissive mode), missing columns are silently passing. The auto-injected list causes `_get_schema` to silently drop any missing columns from the `expected_schema` before `_get_permissive_schema_comparison` even runs.

If we only selectively applied column selection to the target schema in strict mode as you suggested:

* **In `strict=True`:** `_get_schema` would still filter out the missing columns from the `expected_schema` because of the decorator's injection, resulting in a false PASS for missing columns.
* **In `strict=False`:** `_get_schema` wouldn't filter them out, so the missing columns would finally be correctly flagged.

This would create a paradoxical situation where strict mode is actually *less* strict than permissive mode regarding missing columns.

By removing the decorator, we prevent the engine from spoofing the `columns` kwarg, preserving the user's intent and allowing missing columns to be accurately checked against the full `expected_schema` in *both* modes. To still ignore DQX internal/result columns safely, I've moved the actual column preselection logic directly inside the `apply()` closure where it filters `actual_schema` without mutating the `expected_schema` definition.

Let me know what you think!

ghanse: > If we only selectively applied column selection to the target schema in strict mode as you suggested:
> 
> - In strict=True: _get_schema would still filter out the missing columns from the expected_schema because of the decorator's injection, resulting in a false PASS for missing columns.
> - In strict=False: _get_schema wouldn't filter them out, so the missing columns would finally be correctly flagged.

We should invert this logic and apply column selection to the expected schema only when `strict=True`. The docs mention that non-strict mode will only validate the expected columns and not any extra columns.

fedeflowers: Sorry @ghanse, I need some clarification on the exact requirements to ensure we align with the intended behavior:

1. **`expected_schema` filtering:** Should the `columns` parameter filter the `expected_schema` *only* when `strict=True`? There is an existing test (`has_valid_schema_with_specified_columns`) that explicitly expects `expected_schema` filtering to happen when `strict=False`. If we change this, the test will need to be updated.
2. **Non-strict definition:** In `strict=False`, does "expected columns" refer to the *entire* `expected_schema`, completely ignoring the `columns` parameter?
3. **Superset behavior:** If the `columns` list contains columns not present in the DataFrame (e.g., DF has `a, b, c` but `columns=["a", "b", "c", "d"]`), should validation fail in *both* `strict=True` and `strict=False`? The documentation isn't fully clear on this edge case.

ghanse: Sorry, I meant we should apply column selection on the expected schema only when `strict=False`.

1. **Schema filtering:** Filter the `expected_schema` only when `strict=False`. When `strict=True`, we should filter the DataFrame schema and compare against the unfiltered `expected_schema`.
2. **Non-strict definition:** When `strict=False`, the "expected columns" are the subset of expected columns after filtering using the `columns` parameter. When `strict=True`, the "expected columns" are the full expected schema, either from the `expected_schema`, `ref_df_name`, or `ref_table`. We can add this explanation to the docstring.
3. **Superset behavior:** I don't think we currently cover this edge case. IMO the validation should fail in both cases if the DataFrame schema does not contain all `columns`.

I believe this is in line with the docstring:

```
strict: Whether to perform strict schema validation (default: False).
  - False: Validates that all expected columns exist with compatible types (allows extra columns)
  - True: Validates exact schema match (same columns, same order, same types)
```

fedeflowers: I have squashed the commits and updated the PR with a signed final commit. 

Key updates include:
* **Docstring Updates**: Refined the documentation to reflect the recent changes.
* **Test Case Adjustment**: Updated `test_has_valid_schema_with_specified_columns`. 
    * *Reason*: Previously, the test filtered the expected schema even when `strict=True`, which failed to correctly identify the missing column `c`. I've updated the expected result to ensure it correctly validates the schema under strict mode.

Please let me know if this addresses the requirements!

ghanse: We should use `register_for_original_columns_preselection` instead of the approach based on `dq_result_schema`.

tests/integration/test_apply_checks.py
ghanse, line 1: Can you add a test where `columns` contains extra columns not present in the input/reference DataFrame schemas?

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -2140,11 +2143,17 @@ def apply(df: DataFrame, spark: SparkSession, ref_dfs: dict[str, DataFrame]) ->
 
         if ref_df_name or ref_table:
             ref_df = _get_ref_df(ref_df_name, ref_table, ref_dfs, spark)
-            _expected_schema = _get_schema(ref_df.schema, column_names)
+            _expected_schema = _get_schema(ref_df.schema, column_names if not strict else None)
         else:
             _expected_schema = expected_schema
 
         selected_column_names = column_names if column_names else df.columns
+        if not column_names:
```
mwojtyczka: **Falsy check treats `[]` same as `None`**

`if not column_names` is falsy for both `None` (not provided) and `[]` (explicitly empty list passed by the caller). If someone passes an empty list to opt out of column filtering, this condition would still enter the block and potentially modify `selected_column_names`.

Consider using `if column_names is None:` to distinguish between "not provided" and "provided as empty list".

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -2140,11 +2143,17 @@ def apply(df: DataFrame, spark: SparkSession, ref_dfs: dict[str, DataFrame]) ->
...
+            dq_meta_cols = {field.name for field in df.schema.fields if field.dataType == dq_result_schema}
```
mwojtyczka: **Struct type equality comparison may be fragile**

`field.dataType == dq_result_schema` compares a `DataType` instance against a `StructType`. While Spark does implement `__eq__` for `DataType`, this relies on exact structural equality of the schema. If `dq_result_schema` ever changes (e.g., new nullable flags or metadata), this filter would silently stop working.

A more robust approach would be to filter by known DQX metadata column names (e.g., a set of `DQX_META_COLUMNS`) rather than by schema shape.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -2140,11 +2143,17 @@ def apply(df: DataFrame, spark: SparkSession, ref_dfs: dict[str, DataFrame]) ->
...
+            selected_column_names = [
+                col for col in selected_column_names if col not in dq_meta_cols and not col.startswith("__")
```
mwojtyczka: **`startswith("__")` may silently drop legitimate user columns**

The additional `not col.startswith("__")` guard will silently exclude any user column whose name starts with double underscores (e.g., `__version`, `__created_at`, `__id`). These are not uncommon in Delta tables using conventions like dbt or CDC patterns.

If the intent is to exclude internal DQX columns, prefer filtering by an explicit allowlist/denylist of known DQX column name prefixes (e.g., `_rescued_data`) rather than a generic double-underscore heuristic.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -2140,11 +2143,15 @@ def apply(df: DataFrame, spark: SparkSession, ref_dfs: dict[str, DataFrame]) ->
...
         if ref_df_name or ref_table:
             ref_df = _get_ref_df(ref_df_name, ref_table, ref_dfs, spark)
-            _expected_schema = _get_schema(ref_df.schema, column_names)
+            _expected_schema = _get_schema(ref_df.schema, column_names if not strict else None)
         else:
             _expected_schema = expected_schema
 
         selected_column_names = column_names if column_names else df.columns
+        if column_names is None:
+            dqx_column_names = {col_name.value for col_name in DefaultColumnNames}
```
fedeflowers: ### Metadata Column Handling

Instead of retaining all columns from the dataframe, I will exclude the specific **dqx metadata columns** identified in `DefaultColumnNames`:

* `_errors`
* `_warnings`
* `_dq_info`

This ensures the output remains focused on the core data schema without carrying over validation metadata.

If the user uses other functions that add more internal columns, he/she should put those newly generated columns in the exclude_columns parameter or add them to the expected schema

ghanse: Because the user can [customize metadata column names](https://databrickslabs.github.io/dqx/docs/guide/additional_configuration/#customizing-result-columns), we will need to dynamically create this list during the check execution instead of using `DefaultColumnNames`.

This was the purpose of `@register_for_original_columns_preselection()`. We will need to keep this same pattern or use a similar pattern.

ghanse (changes requested):
Looking good overall. Left a few comments.

PR #1043, opened 2026-02-23
Enhance data contract quality rules generation with schema validation support

### LINKED ISSUE #1016: [FEATURE]: Data Contract Implicit Checks to Include Schema Validation

### Problem statement

the data contracts to dqx check generator should also generate schema validation checks based on what's defined in the ODCS data contracts

### Proposed Solution

1. Add a schema validation rule type: for each ODCS schema, generate one dataset-level rule that calls has_valid_schema with the expected schema derived from the contract (column names and Unity Catalog physicalTypes).
2. Use strict mode only: exact column set, order, and types (no permissive/type-widening).
3. Require every property to have Unity Catalog physicalType (e.g. STRING, INT, DECIMAL(10,2), ARRAY&lt;STRING&gt;). No ODCS→Unity mapping; missing or invalid physicalType raises an error.
4. Make it opt-in by default: new parameter generate_schema_validation=True; set to False to skip these rules.
5. Emit rules with user_metadata.rule_type: "schema_validation" and the expected DDL string so they can be filtered or reported separately.

### Additional Context

_No response_

vb-dbrks:
## Changes

1. Updated DQGenerator to support schema validation rules generation from ODCS contracts, ensuring dataset schemas match contract definitions.
2. Added generate_schema_validation parameter to generate_rules_from_contract method, defaulting to True.
3. Enhanced documentation to reflect the new schema validation feature and its requirements.
4. Introduced InvalidPhysicalTypeError for better error handling when physical types are missing or invalid in schema properties.
5. Updated integration tests to cover schema validation scenarios, ensuring rules are generated correctly.

### Linked issues
Resolves #1016 

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests


demos/dqx_demo_datacontract_odcs.py
```diff
@@ -428,10 +434,11 @@
 # MAGIC | `minLength`/`maxLength` | `logicalTypeOptions` | `sql_expression` | String length validation |
 # MAGIC | `format` | `logicalTypeOptions` | `is_valid_date`, `is_valid_timestamp` | Date/timestamp formats |
 # MAGIC
-# MAGIC ### Three Types of Rules
+# MAGIC ### Four Types of Rules
```
ghanse: Maybe just change it to _"Rule Types"_ to avoid need to modify in the future. :)

docs/dqx/docs/guide/data_contract_quality_rules_generation.mdx
```diff
@@ -12,8 +12,9 @@ DQX provides the capability to generate data quality rules directly from data co
 
 ## Overview
 
-The data contract-based quality rules generation is available via the `DQGenerator` class and supports three types of rule generation:
+The data contract-based quality rules generation is available via the `DQGenerator` class and supports four types of rule generation:
```
ghanse: I would also change this to something like _"supports generation of the following rules types:"_

docs/dqx/docs/guide/data_contract_quality_rules_generation.mdx
```diff
@@ -121,6 +123,23 @@ You can also pass a pre-loaded `DataContract` object from `datacontract-cli`:
   </TabItem>
 </Tabs>
 
+### Schema validation
+
+When `generate_schema_validation=True` (default), DQX generates one **dataset-level** rule per ODCS schema that calls `has_valid_schema` with the expected schema derived from the contract. This ensures the dataset's columns and types match the contract. Schema validation is always **strict** (exact columns, order, and types). To skip these rules, pass `generate_schema_validation=False`:
+
+<Admonition type="note" title="Physical types required for schema validation">
+Schema validation requires every property in the contract to have a `physicalType` field set to a valid Unity Catalog data type. Properties with only `logicalType` will cause an `InvalidPhysicalTypeError`.
+</Admonition>
```
ghanse, line 132: Could we get around this limitation by parsing the logical types to a wide Spark type and using non-strict schema validation?

src/databricks/labs/dqx/datacontract/contract_rules_generator.py
```diff
@@ -85,6 +87,7 @@ def generate_rules_from_contract(
         contract_format: str = "odcs",
         generate_predefined_rules: bool = True,
         process_text_rules: bool = True,
+        generate_schema_validation: bool = True,
```
ghanse, line 87: I think we should use `generate_schema_validation_rules` for consistency.

src/databricks/labs/dqx/datacontract/contract_rules_generator.py
```diff
@@ -269,6 +282,119 @@ def _validate_generated_rules(self, dq_rules: list[dict]) -> list[dict]:
...
+    def _schema_object_to_ddl(self, schema_obj: SchemaObject, schema_name: str = "") -> str:
+        """Build a Spark/Unity Catalog DDL string from an ODCS schema object.
+
+        Every property must have physicalType set to a Unity Catalog type. Raises
+        InvalidPhysicalTypeError (with schema and property name) when missing or invalid.
+        """
+        parts: list[str] = []
+        for prop in schema_obj.properties or []:
+            if not prop.name:
+                continue
```
ghanse, line 489: Should we warn in this case?

src/databricks/labs/dqx/datacontract/contract_rules_generator.py
```diff
@@ -269,6 +282,119 @@ def _validate_generated_rules(self, dq_rules: list[dict]) -> list[dict]:
...
+                "See: https://learn.microsoft.com/en-gb/azure/databricks/sql/language-manual/sql-ref-datatypes"
+            )
+        type_upper = type_str_stripped.upper()
+        if type_upper in cls._UNITY_SIMPLE_TYPES:
+            return type_upper
+        decimal_match = cls._UNITY_DECIMAL_PATTERN.match(type_str_stripped)
+        if decimal_match:
+            return f"DECIMAL({decimal_match.group(1)},{decimal_match.group(2)})"
+        if any(type_upper.startswith(prefix) for prefix in cls._UNITY_COMPLEX_PREFIXES):
+            return type_upper
```
ghanse, line 467: What if the inner types are invalid (e.g. `ARRAY<NOT_A_VALID_TYPE>`)? Should we recursively apply this method to the inner types?

src/databricks/labs/dqx/datacontract/contract_rules_generator.py
```diff
@@ -269,6 +282,119 @@ def _validate_generated_rules(self, dq_rules: list[dict]) -> list[dict]:
...
+            raise InvalidPhysicalTypeError(
+                "physicalType must be set to a Unity Catalog data type (e.g. STRING, INT, ARRAY<STRING>). "
+                "See: https://learn.microsoft.com/en-gb/azure/databricks/sql/language-manual/sql-ref-datatypes"
+            )
+        type_upper = type_str_stripped.upper()
+        if type_upper in cls._UNITY_SIMPLE_TYPES:
+            return type_upper
+        decimal_match = cls._UNITY_DECIMAL_PATTERN.match(type_str_stripped)
+        if decimal_match:
+            return f"DECIMAL({decimal_match.group(1)},{decimal_match.group(2)})"
```
ghanse: Should we validate the scale/precision parsed from the regex? Spark's decimal types are limited to 38 digits ([discussion](https://issues.apache.org/jira/browse/SPARK-28316?focusedCommentId=16884398&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16884398)).

tests/datacontract_helpers.py
ghanse, line 1: I would add this to `conftest.py` and import from there.

demos/dqx_demo_datacontract_odcs.py
vb-dbrks: Good point!

docs/dqx/docs/guide/data_contract_quality_rules_generation.mdx
vb-dbrks: I don't think that's a good idea @ghanse ... better to be explicit and have users define the exact data types via Physical Data Types based on their implementation of the contract in Databricks, especially since in ODCS they have their own way of defining what a logical schematype is to keep the definition platform-agnostic... Mapping will lead to ambiguity and confusion when checks are applied.

src/databricks/labs/dqx/datacontract/contract_rules_generator.py
vb-dbrks: Do you think we should even have this knob? We could always generate the check if there's a schema defined. wdyt?- If we decide to keep your suggestion, it makes sense

src/databricks/labs/dqx/datacontract/contract_rules_generator.py
vb-dbrks: good one! but the arrays might have an evolving schema and might not be feasible to go deeper - same like Variant.

src/databricks/labs/dqx/datacontract/contract_rules_generator.py
ghanse: I don't think we need it. Users can always modify the generated rules. We don't allow this level of customization in other generator methods.

src/databricks/labs/dqx/datacontract/contract_rules_generator.py
ghanse: Variant will not have explicit inner types and can be treated this way, but aren't the subtypes of array, map, and struct columns explicit and checkable?

src/databricks/labs/dqx/datacontract/contract_rules_generator.py
vb-dbrks: Agreed, I'll take it off as I think it's not needed... if we receive any requests in the future we can evaluate.

src/databricks/labs/dqx/datacontract/contract_rules_generator.py
vb-dbrks: Yeah, lets do that - makes sense

src/databricks/labs/dqx/datacontract/contract_rules_generator.py
vb-dbrks: Yes, but I am a bit worried about multi-level recursion and the complexities it brings. let me think about this and test it out...

vb-dbrks: I've added a recursion depth limit, reasonably high so that it should not matter. Also called out in the docs.

tests/datacontract_helpers.py
vb-dbrks: done, was a tiny file anyway. Conftest needs to be looked at in the future though.

src/databricks/labs/dqx/datacontract/contract_rules_generator.py
```diff
@@ -222,6 +225,11 @@ def _generate_all_rules(
         for schema_obj in odcs.schema_ or []:
             schema_name = schema_obj.name or "unknown_schema"
 
+            schema_validation_rules = self._generate_schema_validation_rules_for_schema(
```
mwojtyczka: Schema validation is now always generated with no way to opt out. The PR description says "Added generate_schema_validation parameter to generate_rules_from_contract method, defaulting to True" — but that parameter doesn't exist in the code. There is no change to the `generate_rules_from_contract` signature.

This is a silent breaking change: every existing caller now gets an additional `has_valid_schema` rule they did not ask for. If their contracts have properties without `physicalType` set, calls that previously succeeded now raise `InvalidPhysicalTypeError`.

Please either:
1. Add a `generate_schema_validation: bool = True` parameter to `generate_rules_from_contract` (and thread it through here), or
2. Document clearly that schema validation is now mandatory and bump the minor version.

src/databricks/labs/dqx/datacontract/contract_rules_generator.py
```diff
@@ -269,6 +277,242 @@ def _validate_generated_rules(self, dq_rules: list[dict]) -> list[dict]:
...
+            "BINARY",
+            "VARIANT",
+            "SMALLINT",
+            "TINYINT",
+            "VOID",
+            "OBJECT",
+        }
+    )
+    _UNITY_COMPLEX_PREFIXES: tuple[str, ...] = (
+        "ARRAY<",
```
mwojtyczka: `"ARRAY<"`, `"MAP<"`, and `"STRUCT<"` are unreachable entries here. In `_validate_unity_physical_type`, all three are handled by their dedicated validators (`_validate_array_type`, `_validate_map_type`, `_validate_struct_type`) before the generic `_UNITY_COMPLEX_PREFIXES` fallback is reached. The fallback only ever fires for `GEOGRAPHY(`, `GEOMETRY(`, and `INTERVAL `.

Having them in the set is misleading — it implies the fallback path handles them, but they are intercepted earlier. Remove the three dead entries:
```python
_UNITY_COMPLEX_PREFIXES: tuple[str, ...] = (
    "GEOGRAPHY(",
    "GEOMETRY(",
    "INTERVAL ",
)
```

src/databricks/labs/dqx/datacontract/contract_rules_generator.py
```diff
@@ -269,6 +277,242 @@ def _validate_generated_rules(self, dq_rules: list[dict]) -> list[dict]:
...
+        return ", ".join(parts)
+
+    def _generate_schema_validation_rules_for_schema(
+        self,
+        schema_obj: SchemaObject,
+        schema_name: str,
+        odcs: OpenDataContractStandard,
+        default_criticality: str,
+    ) -> list[dict]:
+        """Generate one dataset-level has_valid_schema rule per ODCS schema. Always strict."""
```
mwojtyczka: `strict=True` means the check will fail if the dataset has *any* columns not declared in the contract. This is reasonable for strongly-typed contracts, but many real-world tables have additive schema evolution (extra columns are fine, missing required columns are not).

There is currently no way for callers to get non-strict schema validation — even passing `strict=False` at a higher level is impossible because this is hardcoded.

Consider accepting a `strict` parameter in `_generate_schema_validation_rules_for_schema` (and threading it from `generate_rules_from_contract`), defaulting to `True` to preserve current behaviour.

tests/integration/test_datacontract_integration.py
```diff
@@ -1,6 +1,100 @@
...
+        try:
+            os.unlink(path)
+        except OSError:
+            pass
+
+
+def _temp_contract_path_with_content(content: str) -> str:
+    """Write raw content to a temp YAML file and return its path. Caller must os.unlink(path)."""
+    with tempfile.NamedTemporaryFile(mode="w", suffix=".yaml", delete=False) as f:
+        f.write(content)
```
mwojtyczka, line 55: The "Caller must os.unlink(path)" comment documents a resource-leak footgun. Looking at the one call site (`test_error_invalid_contract_raises_odcs_error`), it correctly wraps cleanup in `try/finally`, but this is easy to forget in future tests.

Compare with `_generate_rules_from_temp_contract` just above, which handles cleanup internally. A context manager or fixture achieves the same without pushing cleanup responsibility to callers:

```python
@contextlib.contextmanager
def _temp_contract_file(content: str):
    with tempfile.NamedTemporaryFile(mode="w", suffix=".yaml", delete=False) as f:
        f.write(content)
        path = f.name
    try:
        yield path
    finally:
        try:
            os.unlink(path)
        except OSError:
            pass
```

src/databricks/labs/dqx/datacontract/contract_rules_generator.py
vb-dbrks: Hi @mwojtyczka - Good point, I had added the parameter previously, which we then removed it based on previous discussion, considering that Data Contract should always have Schema Validation if the Schema is defined and wanted to remove the extra knobs. I do understand the challenge with an extra rule suddenly being generated and it's very well possible that they have not defined the physicalType. any recommendations?

src/databricks/labs/dqx/datacontract/contract_rules_generator.py
vb-dbrks: @mwojtyczka  - this one is also another trade-off... In the purist sense, Contracts should be enforceable and strict as that's their true purpose... but real world, I understand that customers might be creating 1 Data Product : N Data Contracts with overlapping schema so strict enforcement would become a blocker.... but this becomes one more knob. any thoughts?

ghanse: We should at least generate the checks with a `strict` parameter for clarity. The user can choose whether to customize the generated check for permissive or strict validation. We could also log a message about the chosen schema validation mode.

src/databricks/labs/dqx/datacontract/contract_rules_generator.py
vb-dbrks: as discussed closing this.

ghanse: Would we also consider adding new profile types from the profiler a breaking change? It would be difficult to extend or enhance the profiler and generator long-term.

What if we allowed the user to specify an allow-list (`allowed_rule_types`) or deny list (`disallowed_rule_types`) via options passed to the profiler and generator? We could then document that users should specify the allow-list if they rely on a stable set of rules from the profiler and generator.

This would be simple to integrate with the implementation changes in #1059 but would require some extra work in the generator.

src/databricks/labs/dqx/datacontract/contract_rules_generator.py
```diff
@@ -82,13 +84,18 @@ def generate_rules_from_contract(
```
mwojtyczka, line 81: **Issue: `VOID` and `OBJECT` may not be valid in DDL**

`VOID` and `OBJECT` are included in `_UNITY_SIMPLE_TYPES`, but `StructType.fromDDL()` may not parse them. `VOID` is a Spark internal type (NullType) not typically used in user-facing schemas. `OBJECT` is not a standard Spark SQL type.

If a user specifies `physicalType: VOID` or `physicalType: OBJECT`, the generated DDL will pass validation here but may fail at runtime when `has_valid_schema` tries to parse it.

Consider either removing them or adding a comment explaining that these are intentionally allowed for edge cases.

src/databricks/labs/dqx/datacontract/contract_rules_generator.py
```diff
@@ -266,6 +290,249 @@ def _validate_generated_rules(self, dq_rules: list[dict]) -> list[dict]:
...
+        while i < len(type_str) and depth > 0:
+            if type_str[i] == "<":
+                depth += 1
+            elif type_str[i] == ">":
+                depth -= 1
+            i += 1
+        if depth != 0:
+            raise InvalidPhysicalTypeError(f"physicalType '{type_str}' has unmatched angle brackets.")
+        return type_str[start + 1 : i - 1].strip()
+
```
mwojtyczka, line 341: **Concern: GEOGRAPHY/GEOMETRY/INTERVAL pass through without content validation**

These types hit the `_UNITY_COMPLEX_PREFIXES` branch and return `type_upper` directly. This means `GEOGRAPHY(GARBAGE)` or `INTERVAL NONSENSE` would pass validation and get embedded in the DDL string.

While full validation of these types is complex, consider at minimum logging a warning that these types are passed through without inner validation, or adding a brief comment in the code explaining the design choice.

tests/integration/test_datacontract_integration.py
```diff
@@ -1,6 +1,109 @@
...
 import pytest
+import yaml
+from datacontract.data_contract import DataContract
+from pyspark.sql import types as spark_types
+
+from databricks.sdk.errors import NotFound
+from databricks.labs.dqx.engine import DQEngine
+from databricks.labs.dqx.errors import InvalidPhysicalTypeError, ODCSContractError, ParameterError
 from databricks.labs.dqx.profiler.generator import DQGenerator
+from tests.conftest import get_schema_validation_rules
```
mwojtyczka, line 28: **Convention: Use concrete types instead of `Any`**

Avoid `Any`." These parameters have well-known types:

```python
def _generate_rules_from_temp_contract(
    workspace_client: WorkspaceClient,
    spark: SparkSession,
    contract: dict[str, Any],
    **kwargs: Any,
) -> list[dict[str, Any]]:
```

`workspace_client` should be `WorkspaceClient`, `spark` should be `SparkSession`. The return type `list[dict]` should be `list[dict[str, Any]]` for completeness.

tests/conftest.py
```diff
@@ -38,6 +38,19 @@
 logger = logging.getLogger(__name__)
 
 
+TEST_CATALOG = "dqx"
+
```
mwojtyczka: **Convention: Missing type annotation**

Every parameter and return value should be annotated.

```python
def get_schema_validation_rules(rules: list[dict[str, Any]]) -> list[dict[str, Any]]:
```

tests/conftest.py
```diff
@@ -38,6 +38,19 @@
 logger = logging.getLogger(__name__)
```
mwojtyczka, line 40: **Issue: Duplicate `TEST_CATALOG` constant**

There is already a `TEST_CATALOG = "dqx"` in `tests/constants.py` (used by the anomaly tests and others). This introduces a second definition in `tests/conftest.py`. Should import from `tests.constants` instead to avoid divergence.

src/databricks/labs/dqx/datacontract/contract_rules_generator.py
```diff
@@ -266,6 +290,249 @@ def _validate_generated_rules(self, dq_rules: list[dict]) -> list[dict]:
...
+        if scale < 0 or scale > precision:
+            raise InvalidPhysicalTypeError(f"DECIMAL scale must be between 0 and precision ({precision}), got {scale}.")
+        return f"DECIMAL({precision},{scale})"
+
+    @classmethod
+    def _validate_array_type(cls, type_str_stripped: str, type_str: str, depth: int) -> str:
+        """Validate ARRAY<T> and return normalized string."""
+        inner = cls._extract_content_in_angle_brackets(type_str_stripped)
+        if not inner:
+            raise InvalidPhysicalTypeError(
```
mwojtyczka, line 377: **Nit: `getattr` hides type information**

`getattr(prop, "physicalType", None)` is defensive against missing attributes, but `SchemaProperty` from the ODCS model should have `physicalType` defined (possibly as `None`). Using `prop.physicalType` directly would be clearer and let static analysis catch issues. If the concern is older ODCS model versions that lack the field, a comment explaining that would be helpful.

src/databricks/labs/dqx/datacontract/contract_rules_generator.py
vb-dbrks: removed both. VOID (NullType) can't be stored in Delta Lake and OBJECT can't be stored in table columns; both would fail StructType.fromDDL() at runtime. Users will now get a clear InvalidPhysicalTypeError at contract load time instead.

src/databricks/labs/dqx/datacontract/contract_rules_generator.py
vb-dbrks: Added a code comment above _UNITY_COMPLEX_PREFIXES explaining the design choice: we allow by prefix only; inner content is not validated; malformed values may fail at DDL parse or runtime. GEOGRAPHY(GARBAGE) or INTERVAL NONSENSE is funny XD !

tests/conftest.py
vb-dbrks: get_schema_validation_rules(rules: list[dict[str, Any]]) -> list[dict[str, Any]]

tests/conftest.py
vb-dbrks: Removed the duplicate from tests/conftest.py. All existing usages already import from tests.constants.

src/databricks/labs/dqx/datacontract/contract_rules_generator.py
vb-dbrks: Changed to prop.physicalType directly. Verified SchemaProperty defines physicalType: str | None (default None) in the ODCS Pydantic model, so direct access is safe. Added a short comment noting the field may be None if omitted in the contract.

tests/integration/test_datacontract_integration.py
vb-dbrks: workspace_client is now WorkspaceClient, spark is SparkSession, return type is list[dict[str, Any]]

src/databricks/labs/dqx/datacontract/contract_rules_generator.py
```diff
@@ -266,6 +290,252 @@ def _validate_generated_rules(self, dq_rules: list[dict]) -> list[dict]:
...
+    def _schema_object_to_ddl(self, schema_obj: SchemaObject, schema_name: str = "") -> str:
+        """Build a Spark/Unity Catalog DDL string from an ODCS schema object.
+
+        Every property must have physicalType set to a Unity Catalog type. Raises
+        InvalidPhysicalTypeError (with schema and property name) when missing or invalid.
+        """
+        parts: list[str] = []
+        for prop in schema_obj.properties or []:
+            if not prop.name:
+                logger.warning(f"Skipping property without name in schema '{schema_name}'")
```
mwojtyczka: Please provide unique message, there is another place with the same log

src/databricks/labs/dqx/datacontract/contract_rules_generator.py
```diff
@@ -868,11 +1138,14 @@ def _build_schema_info_from_model(self, schema_obj: SchemaObject) -> str:
...
         columns = []
 
+        schema_name_for_log = schema_obj.name or "unknown"
+
         def _extract_columns(props: list[SchemaProperty] | None, prefix: str = "") -> None:
             """Recursively extract column information from properties."""
             for prop in props or []:
                 # Skip properties without a name
                 if not prop.name:
+                    logger.warning(f"Skipping property without name in schema '{schema_name_for_log}'")
```
mwojtyczka: Please provide unique message, there is another place with the same log

docs/dqx/docs/guide/data_contract_quality_rules_generation.mdx
```diff
@@ -121,6 +123,38 @@ You can also pass a pre-loaded `DataContract` object from `datacontract-cli`:
...
+
+# Permissive schema validation (allow extra columns)
+rules = generator.generate_rules_from_contract(
+    contract_file="/path/to/contract.yaml",
+    strict_schema_validation=False,
+)
+```
+
+<Admonition type="info" title="Type validation rules">
+- **Recursive validation**: Inner types of `ARRAY&lt;T&gt;`, `MAP&lt;K,V&gt;`, and `STRUCT&lt;...&gt;` are validated at every nesting level. Invalid inner types (e.g. `ARRAY&lt;NOT_A_TYPE&gt;`) raise `InvalidPhysicalTypeError` at contract load time.
```
mwojtyczka:
```suggestion
- **Recursive validation**: Inner types of `ARRAY<T>`, `MAP<K,V>`, and `STRUCT<...>` are validated at every nesting level. Invalid inner types (e.g. `ARRAY<NOT_A_TYPE>`) raise `InvalidPhysicalTypeError` at contract load time.
```

docs/dqx/docs/reference/profiler.mdx
```diff
@@ -102,7 +102,7 @@ The `DQGenerator` class converts profiling results into DQX quality rules:
...
+| `generate_rules_from_contract`  | Generates quality rules from ODCS data contract specification. Either `contract` or `contract_file` must be provided. When the contract defines a schema and `generate_schema_validation=True`, one dataset-level `has_valid_schema` rule per schema is generated. | `contract`: Pre-loaded DataContract object (or `None`); `contract_file`: Path to contract YAML file (or `None`); `contract_format`: Contract format (default "odcs"); `generate_predefined_rules`: Generate rules from contract fields (default `True`); `process_text_rules`: Process text expectations with LLM (default `True`); `generate_schema_validation`: Generate has_valid_schema rules from contract schema (default `True`); `strict_schema_validation`: Passed as strict to has_valid_schema (default `True`); `default_criticality`: Default criticality for predefined rules (default "error"). | Yes                      |
```
mwojtyczka:
```suggestion
| `generate_rules_from_contract`  | Generates quality rules from ODCS data contract specification. Either `contract` or `contract_file` must be provided. When the contract defines a schema and `generate_schema_validation=True` (default), one dataset-level `has_valid_schema` rule per schema is generated. | `contract`: Pre-loaded DataContract object (or `None`); `contract_file`: Path to contract YAML file (or `None`); `contract_format`: Contract format (default "odcs"); `generate_predefined_rules`: Generate rules from contract fields (default `True`); `process_text_rules`: Process text expectations with LLM (default `True`); `generate_schema_validation`: Generate has_valid_schema rules from contract schema (default `True`); `strict_schema_validation`: Passed as strict to has_valid_schema (default `True`); `default_criticality`: Default criticality for predefined rules (default "error"). | Yes                      |
```

PR #1035, opened 2026-02-13
Use `WorkspaceClient` to check table existence

ghanse:
## Changes

This PR refactors `spark.catalog` calls to use `WorkspaceClient` to check for table existence when saving checks using the `TableChecksStorageHandler`.

### Linked issues

Resolves #1033 

### Tests

- [x] manually tested
- [ ] added unit tests
- [ ] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

PR #1044, opened 2026-02-24
Versioning of Checks

### LINKED ISSUE #672: [FEATURE]: Versioning of rules

### Problem statement

Tracibility is important for large organization. Ability to track which rules were applied to historical data is especially important in regulated orgs. 

### Proposed Solution

Loose ideas, e.g. add version to the rules and quality results. For quality results we may leverage user_metadata field. Add version to rules stored in a table, and option to provide version number for retrieval.

### Additional Context

Come up with a proposal.

STEFANOVIVAS:
## Changes

Added rule_fingerprint, rule_set_fingerprint, and created_at to checks storage. Added rule_set_fingerprint to summary metrics so that rules can be traced.

BREAKING CHANGE!

Default mode changed from "overwrite" to "append" when saving checks to a delta and lakebase table. Rules are versioned going forward and each new entry has a created_at timestamp, rule_set_fingerprint and rule_fingerprint.

### Linked issues

Resolves #672 

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests


alexott:
### Code review

Found 2 issues:

1. `spark.catalog.tableExists` re-introduced in the `save` path, regressing the fix from PR #1035 which replaced it with `ws.tables.get()` because `spark.catalog.tableExists` fails for tables with special characters and is blocked in Spark Declarative Pipelines. The existing `load` path correctly continues to use `ws.tables.get()`, making the two methods inconsistent.

https://github.com/databrickslabs/dqx/blob/92a5d6b069441a3c055dd89aba853620e333db6d/src/databricks/labs/dqx/checks_storage.py#L148-L153

2. In `_load_checks_from_lakebase`, the "load latest version" path fetches the most recent `rule_set_fingerprint` via a subquery filtered by `run_config_name`, but the outer `SELECT` only filters by `rule_set_fingerprint` without also filtering by `run_config_name`. If two different configs share identical check content (same fingerprint), the query returns rows from both configs mixed together.

https://github.com/databrickslabs/dqx/blob/92a5d6b069441a3c055dd89aba853620e333db6d/src/databricks/labs/dqx/checks_storage.py#L430-L441

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

alexott:
### Additional code review findings (lower confidence)

Found 4 additional issues (scored below primary threshold):

1. **Lakebase deduplication dead code when `mode="overwrite"`** — in `_save_checks_to_lakebase`, the overwrite `DELETE` runs first, then the fingerprint existence check queries for a row that was just deleted. The `SELECT ... WHERE rule_set_fingerprint == ...` always returns empty in overwrite mode, so the early-return skip can never trigger. The existence check should run *before* the delete.

https://github.com/databrickslabs/dqx/blob/92a5d6b069441a3c055dd89aba853620e333db6d/src/databricks/labs/dqx/checks_storage.py#L356-L378

2. **`rule_set_fingerprint = 'None'` on schema migration** — when `latest_row.rule_set_fingerprint` is Python `None` (pre-existing rows whose column was backfilled as NULL via `mergeSchema`), the f-string produces the Spark SQL predicate `rule_set_fingerprint = 'None'` — the literal string `"None"`, not SQL NULL. This silently returns zero rows for any table that has pre-versioning data, breaking `load` after a schema migration.

https://github.com/databrickslabs/dqx/blob/92a5d6b069441a3c055dd89aba853620e333db6d/src/databricks/labs/dqx/checks_serializer.py#L387-L395

3. **Docstring contradicts the new default** — both `TableChecksStorageConfig` and `LakebaseChecksStorageConfig` change `mode` from `"overwrite"` to `"append"`, but `LakebaseChecksStorageConfig`'s docstring still explicitly states `(default is 'overwrite')`. This is a breaking behavioral change for existing callers that relied on the default.

https://github.com/databrickslabs/dqx/blob/92a5d6b069441a3c055dd89aba853620e333db6d/src/databricks/labs/dqx/config.py#L287-L301

4. **SQL injection via f-string for `rule_set_fingerprint`** — the new filter in `TableChecksStorageHandler.save` embeds the user-supplied `rule_set_fingerprint` config field directly into a Spark SQL string. When the fingerprint is computed internally it is always a safe hex-SHA256, but since `TableChecksStorageConfig.rule_set_fingerprint` is an externally-supplied field, a crafted value can bypass the filter. The Column API (`F.col("rule_set_fingerprint") == rule_set_fingerprint`) avoids this, consistent with the rest of the codebase.

https://github.com/databrickslabs/dqx/blob/92a5d6b069441a3c055dd89aba853620e333db6d/src/databricks/labs/dqx/checks_storage.py#L147-L158

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/databricks/labs/dqx/checks_serializer.py
```diff
@@ -361,11 +366,31 @@ def from_dataframe(df: DataFrame, run_config_name: str = "default") -> list[dict
...
             run_config_name: Run configuration name for filtering quality rules, e.g. input table or job name (use "default" if not provided).
+            rule_set_fingerprint: Optional SHA-256 fingerprint of the rule set to load. When provided, filters to rules
+                with this specific fingerprint instead of loading the latest batch.
 
         Returns:
                 List of data quality check specifications as a Python dictionary
         """
-        check_rows = df.where(f"run_config_name = '{run_config_name}'").collect()
+        filtered_df = df.where(f"run_config_name = '{run_config_name}'")
+        if filtered_df.isEmpty():
```
mwojtyczka: `filtered_df.isEmpty()` triggers a full Spark job just to check for zero rows, before you do the actual `.collect()` a few lines below. If the DataFrame is empty, `collect()` already returns `[]` naturally. This early check doubles the number of Spark actions in the common (non-empty) path.

Remove the guard entirely, or if an early log is important, keep it but note the extra cost:
```python
check_rows = filtered_df.collect()
if not check_rows:
    logger.info(f"No checks found for run_config_name '{run_config_name}'.")
    return []
```

src/databricks/labs/dqx/checks_serializer.py
```diff
@@ -361,11 +366,31 @@ def from_dataframe(df: DataFrame, run_config_name: str = "default") -> list[dict
...
+        has_versioning_columns = all(
+            col in df.columns for col in ("created_at", "rule_fingerprint", "rule_set_fingerprint")
+        )
+
+        if has_versioning_columns:
+            if rule_set_fingerprint:
+                filtered_df = filtered_df.where((F.col("rule_set_fingerprint") == rule_set_fingerprint)&(F.col("run_config_name") == run_config_name))
+                    
+            else:
+                rule_set_fingerprint=filtered_df.select(F.col("rule_set_fingerprint")).where(F.col("run_config_name") == run_config_name).orderBy(F.col("created_at").desc()).limit(1).collect()[0][0]
```
mwojtyczka: Three problems on this line:

**1. Crash risk**: `.collect()[0][0]` raises `IndexError` if `created_at` is NULL for all rows or `filtered_df` is empty at this point.

**2. Redundant filter**: `.where(F.col("run_config_name") == run_config_name)` is applied a second time here even though `filtered_df` was already filtered by `run_config_name` earlier (line 374).

**3. Line length**: This is 200+ characters — well over the project's style limit.

Suggested rewrite:
```python
result = (
    filtered_df
    .select(F.col("rule_set_fingerprint"))
    .orderBy(F.col("created_at").desc())
    .limit(1)
    .collect()
)
if not result:
    return []
rule_set_fingerprint = result[0][0]
```

src/databricks/labs/dqx/checks_serializer.py
```diff
@@ -496,3 +528,40 @@ def deserialize_checks(checks: list[dict], custom_checks: dict[str, Callable] |
...
+
+    Returns:
+        A hex-encoded SHA-256 hash string.
+    """
+    fingerprint_data = {
+        "name": check_dict.get("name"),
+        "criticality": check_dict.get("criticality", "error"),
+        "function": check_dict.get("check", {}).get("function"),
+        "arguments": check_dict.get("check", {}).get("arguments"),
+        "filter": check_dict.get("filter"),
```
mwojtyczka: `for_each_column` is not included in `fingerprint_data`. Two rules that differ only in `for_each_column` (e.g. applying the same check to different column lists) will produce the same fingerprint — incorrect deduplication and missed version changes.

```python
fingerprint_data = {
    ...
    "filter": check_dict.get("filter"),
    "for_each_column": check_dict.get("check", {}).get("for_each_column"),  # add this
}
```

src/databricks/labs/dqx/config.py
```diff
@@ -339,11 +259,16 @@ class TableChecksStorageConfig(BaseChecksStorageConfig):
...
+            The default mode is append.
+        rule_set_fingerprint: Optional SHA-256 fingerprint of the rule set to load.
+            When provided, loads rules matching this specific fingerprint instead of the latest batch.
+            When None (default), loads the latest batch.
     """
 
     location: str
     run_config_name: str = "default"  # to filter checks by run config
-    mode: str = "overwrite"
+    mode: str = "append"
```
mwojtyczka, line 355: This is a **breaking behavioural change** for existing users. The previous default (`overwrite`) would replace the checks for the current `run_config_name` on every save; the new default (`append`) will silently accumulate rows instead.

Any user who calls `save_checks()` without explicitly setting `mode` will now get duplicate or stale rows in their table after upgrading — and may not notice until they look at the results.

If `append` is the right default for the versioning model, the change log / migration guide should call this out explicitly, and the PR description should tick off manual testing at minimum.

src/databricks/labs/dqx/checks_storage.py
```diff
@@ -313,20 +352,51 @@ def _save_checks_to_lakebase(self, checks: list[dict], config: LakebaseChecksSto
             logger.info(
                 f"Successfully created or verified table '{config.database_name}.{config.schema_name}.{config.table_name}'."
             )
+            self._ensure_rule_version_columns_exist(conn, config)
+            logger.info("Rule version columns exist or added.")          
 
+            normalized_checks = self._normalize_checks(checks, config)
+            rule_set_fingerprint = normalized_checks[0].get("rule_set_fingerprint")
```
mwojtyczka, line 641: `normalized_checks[0]` raises `IndexError` if `checks` is empty. The same issue exists in the same pattern in the `TableChecksStorageHandler.save` (where `rules_df.select("rule_set_fingerprint").first()` is slightly safer, but still asymmetric).

```python
rule_set_fingerprint = normalized_checks[0].get("rule_set_fingerprint") if normalized_checks else None
```
Or guard earlier:
```python
if not checks:
    logger.info("No checks to save — skipping.")
    return
```

tests/integration/test_save_and_load_checks_from_table.py
```diff
@@ -497,3 +503,54 @@ def test_save_checks_invalid_storage_config(ws, spark):
...
+
+    engine = DQEngine(ws, spark)
+    config_save = TableChecksStorageConfig(location=table_name)
+    engine.save_checks(INPUT_CHECKS[:1], config=config_save)
+
+    engine.save_checks(INPUT_CHECKS[1:], config=config_save)
+
+    config_load = TableChecksStorageConfig(
+        location=table_name,
+        rule_set_fingerprint="e27b1748e670c8bceeb8449ac494f22bd80a934a30a3c86919547de56790bc00",
```
mwojtyczka: Hardcoded hash values in tests are brittle. If the fingerprinting algorithm, input data, or canonical serialisation ever changes, this test will silently fail (wrong fingerprint → empty result → assertion passes on the wrong data) or raise a confusing error.

Compute the expected fingerprint from the same input data instead:
```python
from databricks.labs.dqx.checks_serializer import compute_rule_set_fingerprint

config_load = TableChecksStorageConfig(
    location=table_name,
    rule_set_fingerprint=compute_rule_set_fingerprint(INPUT_CHECKS[:1]),
)
```
Same applies to the equivalent hardcoded hash in `test_save_and_load_checks_from_lakebase_table.py`.

src/databricks/labs/dqx/config.py
STEFANOVIVAS: Hi, @mwojtyczka! This change in default behavior was suggested by you in the issue discussion: "Change default mode to "append". Whenever checks are saved, they will be appended with new created_at timetstamp. This will preserve the history by default." 
I think that with the new functionality for versioning quality rules, the default append behavior would make more sense. It preserves the history and gives you the possibility to return only a portion of the records from your checks table. 
It seems to me that with the current implementation we could have duplicate checks, but they would only be loaded once (only the most recent one would have a rule_set_finerprint). In short, perhaps the impact of the change is too great in relation to the possible benefits.

mwojtyczka: right, you can ignore the comment. Just going back now through out initial design

mwojtyczka (changes requested):
Docs need to be updated, the rest LGTM, e.g. 
* https://databrickslabs.github.io/dqx/docs/guide/quality_checks_apply/#detailed-quality-checking-results
* https://databrickslabs.github.io/dqx/docs/guide/quality_checks_definition/#delta-table-format-declarative-approach
* https://databrickslabs.github.io/dqx/docs/guide/best_practices/#profile-and-tune-rules-over-time

mwojtyczka (approved):
LGTM

xicez:
This breaks pipelines that have user defined schemas

mwojtyczka:
Yes, this adds two new fields to the DQ result struct schema (the inner type of _errors / _warnings array columns). The new fields are nullable=True, and DQX docs surface {"mergeSchema": "true"} as the documented write option which could be used here. But the upgrade requires user action. We put this in the release notes: https://github.com/databrickslabs/dqx/releases/tag/v0.14.0

xicez:
> Yes, this adds two new fields to the DQ result struct schema (the inner type of _errors / _warnings array columns). The new fields are nullable=True, and DQX docs surface {"mergeSchema": "true"} as the documented write option which could be used here. But the upgrade requires user action. We put this in the release notes: https://github.com/databrickslabs/dqx/releases/tag/v0.14.0

Gotcha, thanks! Looks like I need to set up some notifications for the release notes 😅

PR #1050, opened 2026-02-27
Fix save checks in the app to return correct error code for invalid checks

Swayam-arora-2004:
Problem
-------
save_run_checks does not catch InvalidCheckError raised by DQEngine.save_checks,
causing malformed client input to surface as HTTP 500.

Expected behavior
-----------------
Invalid checks should result in a 400 Bad Request, consistent with get_run_checks.

Solution
--------
- Catch InvalidCheckError in save_run_checks
- Map it to HTTP 400
- Add regression test to enforce API contract consistency

### Tests

- [x] manually tested
- [x] added unit tests
- [ ] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

### DIFF
--- a/tests/unit/test_app_backend.py
+++ b/tests/unit/test_app_backend.py
@@ -385,7 +385,7 @@ class _FakeRunConfig:
 class _FakeSerializer:
     """Fake ConfigSerializer that returns a simple run config."""
 
-    def load_run_config(self, run_config_name: str, *args, **kwargs):
+    def load_run_config(self, run_config_name: str, *_args, **_kwargs):
         return _FakeRunConfig(run_config_name)
 
 


app/src/databricks_labs_dqx_app/backend/router.py
```diff
@@ -195,7 +195,11 @@ def save_run_checks(
         raise HTTPException(status_code=404, detail=f"Run config '{name}' not found")
 
     checks_config = InstallationChecksStorageConfig(run_config_name=run_config.name, install_folder=install_folder)
-    engine.save_checks(body.checks, checks_config)
+    try:
```
mwojtyczka, line 198: Consider also catching `InvalidConfigError` here for consistency with `get_run_checks`, which handles both:

```python
except InvalidCheckError as e:
    raise HTTPException(status_code=400, detail=f"Invalid checks format: {e}")
except InvalidConfigError as e:
    raise HTTPException(status_code=400, detail=f"Invalid configuration: {e}")
```

If `save_checks` can raise `InvalidConfigError`, the endpoint would still return 500 without this.

tests/unit/test_save_run_checks.py
```diff
@@ -0,0 +1,65 @@
+"""Unit tests for save_run_checks error handling in the backend router."""
```
mwojtyczka: This test belongs in `tests/unit/test_app_backend.py` as a new `class TestSaveRunChecks` — that's where all backend router tests live. Adding a separate file fragments coverage for the same module.

tests/unit/test_save_run_checks.py
```diff
@@ -0,0 +1,65 @@
...
+    def load_run_config(self, run_config_name: str, install_folder: str):
+        return _FakeRunConfig(run_config_name)
+
+
+class _FakeChecksStorageConfig:
+    """Fake InstallationChecksStorageConfig used only for type compatibility."""
+
+    def __init__(self, run_config_name: str, install_folder: str):
+        self.run_config_name = run_config_name
+        self.install_folder = install_folder
```
mwojtyczka: The stored attributes (`run_config_name`, `install_folder`) are never read by the test. The `__init__` body can be replaced with `pass`, or the whole class can be collapsed to a lambda in the monkeypatch:

```python
monkeypatch.setattr(router, "InstallationChecksStorageConfig", lambda **kw: None)
```

I would generally avoid using monkeypatch. Every time we use it we introduce technical debt.

PR #1054, opened 2026-03-02
Fix save_run_checks returning 500 on invalid checks

Swayam-arora-2004:
###Changes
1. Fix error handling for save_run_checks: wrap DQEngine.save_checks in a try/except and map InvalidCheckError to an HTTP 400 response, aligning behavior with get_run_checks.
2. Add regression unit test: new test test_save_run_checks_maps_invalid_check_error_to_http_400 verifies that when save_run_checks receives invalid checks and DQEngine.save_checks raises InvalidCheckError, the endpoint returns 400 with an appropriate error message.
3. Keep scope minimal: no changes to other endpoints or shared abstractions; only the specific error mapping and its focused test.

### Tests

- [x] manually tested
- [x] added unit tests
- [ ] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests


app/src/databricks_labs_dqx_app/backend/router.py
```diff
@@ -195,7 +195,11 @@ def save_run_checks(
         raise HTTPException(status_code=404, detail=f"Run config '{name}' not found")
 
     checks_config = InstallationChecksStorageConfig(run_config_name=run_config.name, install_folder=install_folder)
-    engine.save_checks(body.checks, checks_config)
+    try:
```
mwojtyczka, line 198: Consider also catching `InvalidConfigError` here for consistency with `get_run_checks`, which handles both:

```python
except InvalidCheckError as e:
    raise HTTPException(status_code=400, detail=f"Invalid checks format: {e}")
except InvalidConfigError as e:
    raise HTTPException(status_code=400, detail=f"Invalid configuration: {e}")
```

If `save_checks` can raise `InvalidConfigError`, the endpoint would still return a 500 without this handler.

tests/unit/test_save_run_checks.py
```diff
@@ -0,0 +1,67 @@
+"""Unit tests for save_run_checks error handling in the backend router."""
```
mwojtyczka, line 1: This file should be added as a class inside `tests/unit/test_app_backend.py` (e.g. `class TestSaveRunChecks`) rather than as a standalone file — that's the established convention for backend router tests in this project.

tests/unit/test_save_run_checks.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class _FakeSerializer:
+    """Fake ConfigSerializer that returns a simple run config."""
+
+    def __init__(self, _ws):
+        pass
+
+    def load_run_config(self, run_config_name: str, install_folder: str):
+        # Use install_folder to satisfy linting while keeping the signature compatible
```
mwojtyczka, line 25: This `assert` inside a test double is fragile — if `install_folder` is ever `None` (a valid call path in the router), this raises a cryptic `AssertionError` instead of a clean test failure. It can safely be removed.

tests/unit/test_save_run_checks.py
```diff
@@ -0,0 +1,67 @@
...
+        assert isinstance(install_folder, str)
+        return _FakeRunConfig(run_config_name)
+
+
+class _FakeChecksStorageConfig:
+    """Fake InstallationChecksStorageConfig used only for type compatibility."""
+
+    def __init__(self, run_config_name: str, install_folder: str):
+        self.run_config_name = run_config_name
+        self.install_folder = install_folder
```
mwojtyczka, line 35: I would avoid using monkeypatch. Every time we patch we introduce a technical debt.

The stored attributes (`run_config_name`, `install_folder`) are never read by the test. The `__init__` body can be replaced with `pass`.

mwojtyczka (changes requested):
Thank you for the PR! Kindly please address the comments

mwojtyczka:
Duplicate of https://github.com/databrickslabs/dqx/pull/1050/

PR #1055, opened 2026-03-02
Add interactive slide deck for row anomaly detection documentation

vb-dbrks:
## Changes
- Introduced a new interactive slide deck in the row anomaly detection guide to visually explain the concept and functionality of anomaly detection in data quality.
- Enhanced the documentation with detailed explanations and visual aids to improve user understanding of the feature.

### Linked issues

Resolves #..

### Tests

- [x] manually tested
- [ ] added unit tests
- [ ] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

### DIFF
--- a/docs/dqx/docusaurus.config.ts
+++ b/docs/dqx/docusaurus.config.ts
@@ -47,7 +47,105 @@ const config: Config = {
       }
     },
     'docusaurus-plugin-image-zoom',
-    'docusaurus-lunr-search'
+    'docusaurus-lunr-search',
+    [
+      // TODO: pinned to pre-release alpha; revisit when a stable 2.0.0 is published
+      // https://www.npmjs.com/package/@signalwire/docusaurus-plugin-llms-txt
+      '@signalwire/docusaurus-plugin-llms-txt',
+      {
+        // Markdown file generation with hierarchical structure
+        markdown: {
+          enableFiles: true,
+          relativePaths: true,
+          includeBlog: false,
+          includePages: true,
+          includeDocs: true,
+          includeVersionedDocs: false,
+          excludeRoutes: [],
+        },
+
+        // llms.txt index file configuration
+        llmsTxt: {
+          enableLlmsFullTxt: true,
+          includeBlog: false,
+          includePages: true,
+          includeDocs: true,
+          excludeRoutes: [],
+
+          // Site metadata
+          siteTitle: 'DQX',
+          siteDescription: 'Simplified Data Quality checking at Scale for PySpark Workloads on streaming and standard DataFrames.',
+
+          // Auto-section organization (set to 1 to minimize auto-sections)
+          autoSectionDepth: 1, // Group by first path segment only
+          autoSectionPosition: 100, // Auto-sections appear after manual sections
+
+          // Manual section organization
+          sections: [
+            {
+              id: 'getting-started',
+              name: 'Getting Started',
+              description: 'Installation and motivation for using DQX',
+              position: 1,
+              // Intentionally hardcoded: installation and motivation live at the
+              // top-level docs/ path, not under a getting-started/ subdirectory.
+              // If a new getting-started page is added, include it explicitly here.
+              routes: [
+                { route: '/dqx/docs/installation' },
+                { route: '/dqx/docs/motivation' }
+              ],
+            },
+            {
+              id: 'user-guide',
+              name: 'User Guide',
+              description: 'Complete guide for using DQX features',
+              position: 2,
+              routes: [
+                { route: '/dqx/docs/guide/**' }
+              ],
+            },
+            {
+              id: 'reference',
+              name: 'Reference',
+              description: 'API reference, CLI commands, and technical documentation',
+              position: 3,
+              routes: [
+                { route: '/dqx/docs/reference/**' }
+              ],
+            },
+            {
+              id: 'development',
+              name: 'Development',
+              description: 'Contributing and development documentation',
+              position: 4,
+              routes: [
+                { route: '/dqx/docs/dev/**' }
+              ],
+            },
+            {
+              id: 'demos',
+              name: 'Demos',
+              description: 'Example implementations and demos',
+              position: 5,
+              routes: [
+                { route: '/dqx/docs/demos' },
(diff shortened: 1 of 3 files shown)


docs/dqx/src/components/SpotTheOddBanana.tsx
```diff
@@ -0,0 +1,141 @@
+import React, { useState, useEffect } from 'react';
```
mwojtyczka: This file appears to be dead code — it is not imported anywhere in the docs site (`index.mdx`, `Deck.tsx`, or any other file in this PR). If it was an earlier prototype that got superseded by `Deck.tsx` + `SlideVisuals.tsx`, it should be deleted. If it's intended as a fallback/alternative, it needs a consumer.

docs/dqx/src/components/SpotTheOddBanana.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+type VisualKey =
```
mwojtyczka: `VisualKey` is already exported from `SlideVisuals.tsx`. This local re-definition will silently diverge if a new visual is added to `SlideVisuals.tsx` but not here. Import it instead:

```ts
import type { VisualKey } from './SlideVisuals';
```

docs/dqx/src/components/Deck.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+  const [slideIndex, setSlideIndex] = useState(0);
+  const totalSlides = slideEls.length;
+  const current = slideEls[slideIndex];
+  const isLast = slideIndex === totalSlides - 1;
+  const isFirst = slideIndex === 0;
+
+  useEffect(() => {
+    const handleKeyDown = (e: KeyboardEvent) => {
+      if (e.key === 'ArrowRight' || e.key === ' ') {
```
mwojtyczka: Attaching this to `window` and calling `e.preventDefault()` on `' '` (Space) is page-global. This suppresses the browser's default page-scroll behaviour any time the user presses Space anywhere on the docs page — even when focused on an input or reading text above/below the deck.

Consider scoping the handler to the deck element instead, so it only fires when the deck region has focus:

```tsx
// on the outer div, add tabIndex={0} + onKeyDown
<div
  className="banana-deck margin-vert--lg"
  tabIndex={0}
  onKeyDown={(e) => {
    if (e.key === 'ArrowRight' || e.key === ' ') { e.preventDefault(); goNext(); }
    else if (e.key === 'ArrowLeft') { e.preventDefault(); goPrev(); }
  }}
>
```

This also removes the need for the `useEffect`/`window` listener entirely.

docs/dqx/src/components/SlideVisuals.tsx
```diff
@@ -0,0 +1,577 @@
...
+        </div>
+      </div>
+      <p className="bv-known__footnote">It&apos;s what you don&apos;t know you don&apos;t know that gets you.</p>
+    </div>
+  );
+}
+
+const DQX_LINE_STAGGER_S = 0.78;
+
+function HighlightedLine({ line, getTokenProps }: { line: any[]; getTokenProps: any }) {
```
mwojtyczka: `any[]` and `any` can be replaced with the proper types from `prism-react-renderer`:

```ts
import type { Token, RenderProps } from 'prism-react-renderer';

function HighlightedLine({
  line,
  getTokenProps,
}: {
  line: Token[];
  getTokenProps: RenderProps['getTokenProps'];
})
```

docs/dqx/src/components/SlideVisuals.tsx
```diff
@@ -0,0 +1,577 @@
...
+];
+
+const BANANA_FEATURES = ['size', 'colour', 'shape', 'flavour', 'smell'] as const;
+
+const DQX_RULES_CODE = `from databricks.labs.dqx.rule import DQRowRule
+from databricks.labs.dqx.check_funcs import is_not_null, is_in_range, is_in_list
+
+# One banana: check size, colour, flavour
+rules = [
+    DQRowRule(check_func=is_not_null, check_func_kwargs={"column": "banana"}),
```
mwojtyczka: Please use the syntax where column is a parameter of DQRowRule, and not check_func_kwargs, e.g.
`DQRowRule(check_func=is_not_null, column="banana"),`

docs/dqx/docs/guide/row_anomaly_detection/index.mdx
```diff
@@ -5,11 +5,48 @@ sidebar_position: 308
...
+    Row anomaly detection in data quality — in a few minutes.
+  </Slide>
+  <Slide title="Known unknowns vs unknown unknowns." visual="known-unknowns">
+    **Rules** catch things we already expect — like "bananas should be yellow" or "a batch must have at least 100 rows". But what about surprises we never thought to check for? That's where **anomaly detection** comes in.
+  </Slide>
+  <Slide title="DQX handles rules really well." visual="dqx-rules">
+    For a single banana, we can write rules: check size, colour, flavour. These rules live in code — clear, testable, versioned.
+  </Slide>
+  <Slide title="DQM built-in anomaly detection" visual="trucks-dqm">
+    Databricks DQM watches the big picture: did the data arrive? Is it fresh? Are the row counts about right? Think of it as checking the delivery trucks.
```
mwojtyczka:
```suggestion
    Databricks Data Quality Monitoring (DQM) watches the big picture: did the data arrive? Is it fresh? Are the row counts about right? Think of it as checking the delivery trucks.
```

docs/dqx/src/components/SpotTheOddBanana.tsx
vb-dbrks: Will do, left over from the first prototype.

docs/dqx/src/components/Deck.tsx
vb-dbrks: Good catch

docs/dqx/src/components/SlideVisuals.tsx
vb-dbrks: Good catch. Should we consider adding ESLint configuration to the Docusaurus component to catch these issues in the future?

PR #1059, opened 2026-03-05
Update profiler implementation

### LINKED ISSUE #1036: [BUG]: Fix profiling rules for empty string values

ghanse:
## Changes

This PR refactors the DQX profiler implementation. Profiles are generated via a rules-based approach.

### Profiler Builder

To create a new profile type, users should implement and register a profile builder:

```
@register_profile_builder("null_or_empty")
def make_null_or_empty_profile(
    df: DataFrame,
    column_name: str,
    column_type: T.DataType,
    profiler_metrics: dict[str, Any],
    profiler_options: dict[str, Any],
) -> DQProfile | None:
  # Rules-based generation of a "null_or_empty" profile
```

#### String Types
`is_in` (value-set) profiles are now supported for CharType and VarcharType as well as StringType.

#### Null Value Handling
This PR also fixes an issue where `is_not_null_or_empty` profiles were incorrectly generated for columns where the null value thresholds were exceeded but the empty value thresholds were not exceeded. Nullability and emptiness are now implemented separately. Use can override default null/empty thresholds via profiler config (max_null_ratio and/or max_empty_ratio) so the profiler suggests not-null or not-empty checks on columns with higher null/empty rates.

### Linked issues

Resolves #1036

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

### DIFF
--- a/src/databricks/labs/dqx/profiler/profiler_runner.py
+++ b/src/databricks/labs/dqx/profiler/profiler_runner.py
@@ -116,16 +116,16 @@ class ProfilerRunner:
             product: Product name for the installation.
             max_parallelism: Maximum number of parallel threads to use for profiling.
         """
-        options = [
-            {
-                "table": "*",  # Matches all tables
-                "options": {
-                    "sample_fraction": run_config.profiler_config.sample_fraction,
-                    "sample_seed": run_config.profiler_config.sample_seed,
-                    "limit": run_config.profiler_config.limit,
-                },
-            }
-        ]
+        pattern_options = {
+            "sample_fraction": run_config.profiler_config.sample_fraction,
+            "sample_seed": run_config.profiler_config.sample_seed,
+            "limit": run_config.profiler_config.limit,
+        }
+        if run_config.profiler_config.max_null_ratio is not None:
+            pattern_options["max_null_ratio"] = run_config.profiler_config.max_null_ratio
+        if run_config.profiler_config.max_empty_ratio is not None:
+            pattern_options["max_empty_ratio"] = run_config.profiler_config.max_empty_ratio
+        options = [{"table": "*", "options": pattern_options}]
         logger.info(f"Using options: \n{options}")
 
         # Include tables matching the patterns, but skip existing output and quarantine tables.


src/databricks/labs/dqx/profiler/profile_builder.py
```diff
@@ -0,0 +1,603 @@
...
+    """
+    Validates that the input column type is a Spark text type.
+
+    Args:
+        column_type: Input column type
+
+    Returns:
+        True if the column is a Spark text type, otherwise False
+    """
+    return column_type in [T.CharType, T.StringType, T.VarcharType]
```
mwojtyczka: **Bug: instance compared to class — always `False`**

`column_type` is a DataType *instance* (e.g. `T.StringType()`), but the list contains the *classes* themselves (no parentheses). In Python, `T.StringType() in [T.StringType]` evaluates to `False` because PySpark's `DataType.__eq__` compares `type(self) == type(other)`, and `type(T.StringType)` is the metaclass `type`, not `StringType`.

**Impact:** `_is_text` always returns `False`, so every column is routed through `_make_null_profile` (the non-text path). This silently prevents `is_not_null_or_empty` and `is_not_empty` profiles from ever being generated for string columns.

Fix using `isinstance`:
```python
def _is_text(column_type: T.DataType) -> bool:
    return isinstance(column_type, (T.CharType, T.StringType, T.VarcharType))
```

src/databricks/labs/dqx/profiler/profile_builder.py
```diff
@@ -0,0 +1,603 @@
...
+    """
+    Validates that the input column type supports distinct operations.
+
+    Args:
+        column_type: Input column type
+
+    Returns:
+        True if the column supports distinct operations, otherwise False
+    """
+    return column_type in [T.IntegerType, T.LongType, T.StringType]
```
mwojtyczka: **Bug: same instance-vs-class comparison issue**

Same problem as `_is_text` — `column_type in [T.IntegerType, T.LongType, T.StringType]` always evaluates to `False`. The old code correctly used `== T.IntegerType()` (with parentheses).

**Impact:** `_supports_distinct` always returns `False`, meaning `is_in` profiles are never generated for any column.

Fix:
```python
def _supports_distinct(column_type: T.DataType) -> bool:
    return isinstance(column_type, (T.IntegerType, T.LongType, T.StringType))
```

src/databricks/labs/dqx/profiler/profile_builder.py
```diff
@@ -0,0 +1,603 @@
...
+    """
+    Validates that the input column type supports min and max operations.
+
+    Args:
+        column_type: Input column type
+
+    Returns:
+        True if the column supports min and max operations, otherwise False
+    """
+    return isinstance(column_type, T.NumericType) or column_type in [T.DateType, T.TimestampType, T.TimestampNTZType]
```
mwojtyczka: **Bug: `column_type in [T.DateType, ...]` always `False` for Date/Timestamp columns**

The `isinstance(column_type, T.NumericType)` part is correct, but `column_type in [T.DateType, T.TimestampType, T.TimestampNTZType]` compares instances to class references (same issue as `_is_text`). The old code used `typ == T.DateType()` (with parentheses).

**Impact:** min/max profiles are not generated for Date or Timestamp columns.

Fix:
```python
return isinstance(column_type, T.NumericType) or isinstance(column_type, (T.DateType, T.TimestampType, T.TimestampNTZType))
```

src/databricks/labs/dqx/profiler/profile_builder.py
```diff
@@ -0,0 +1,603 @@
...
+        return None
+
+    total_count = profiler_metrics.get("total_count", 0)
+    max_in_count = profiler_options.get("max_in_count", 0)
+    max_distinct_ratio = profiler_options.get("distinct_ratio", 0.0)
+
+    col = df.columns[0]
+    distinct_df = df.select(col).distinct()
+    distinct_count = distinct_df.count()
+    distinct_ratio = (1.0 * distinct_count) / total_count
```
mwojtyczka, line 95: **Bug: `ZeroDivisionError` when `total_count == 0`**

`total_count` defaults to `0` from `profiler_metrics.get("total_count", 0)`. If the DataFrame is empty, `distinct_ratio = (1.0 * distinct_count) / total_count` raises `ZeroDivisionError`.

Add a guard before this line:
```python
if total_count == 0:
    return None
```

src/databricks/labs/dqx/profiler/profile_builder.py
```diff
@@ -0,0 +1,603 @@
...
+        column_name: Input column name
+        profiler_metrics: Column-level statistics computed by the DQProfiler
+        profiler_options: Configuration options for the DQProfiler
+
+    Returns:
+        A DQProfile if the correct conditions are met, otherwise None
+    """
+    null_count = profiler_metrics.get("null_count", 0)
+    total_count = profiler_metrics.get("total_count", 0)
+    null_ratio = null_count / total_count
```
mwojtyczka, line 249: **Bug: `ZeroDivisionError` when `total_count == 0`**

Unlike `_make_null_or_empty_profile` (which guards with `if total_count == 0: return None`), this function has no such guard. If called with an empty DataFrame, `null_count / total_count` raises `ZeroDivisionError`.

Add a guard:
```python
if total_count == 0:
    return None
null_ratio = null_count / total_count
```

src/databricks/labs/dqx/profiler/profile_builder.py
```diff
@@ -0,0 +1,603 @@
...
+
+    Returns:
+        A tuple containing the minimum limit, maximum limit, and description.
+    """
+
+    min_value = aggregates.get("min_value")
+    max_value = aggregates.get("max_value")
+    mean_value = aggregates.get("mean_value")
+    stddev_value = aggregates.get("stddev_value")
+    num_sigmas = profiler_options.get("num_sigmas", 3)
```
mwojtyczka, line 470: **Breaking change: option key renamed from `sigmas` to `num_sigmas`**

The old code read `opts.get("sigmas", 3)`. Any user passing `{"sigmas": 5}` in their profiler options will silently fall back to the default of `3` with no warning or error.

If backward compatibility matters, read both keys:
```python
num_sigmas = profiler_options.get("num_sigmas", profiler_options.get("sigmas", 3))
```
Otherwise, document this as a breaking change in the changelog.

src/databricks/labs/dqx/profiler/profile_builder.py
```diff
@@ -0,0 +1,603 @@
...
+import math
+from typing import Any
+
+from pyspark.sql import DataFrame
+from pyspark.sql import types as T, functions as F
+from databricks.labs.dqx.errors import InvalidParameterError
+from databricks.labs.dqx.profiler.profile import DQProfile, DQProfileType
+
+
+PROFILE_BUILDER_REGISTRY = {}
```
mwojtyczka: **Minor: missing type annotation on the registry dict**

```python
PROFILE_BUILDER_REGISTRY: dict[str, DQProfileType] = {}
```
This makes the content type explicit and enables better mypy checking for code that reads from the registry.

src/databricks/labs/dqx/profiler/profile_builder.py
ghanse: Fixed: now uses `isinstance(column_type, (T.CharType, T.StringType, T.VarcharType))`.

src/databricks/labs/dqx/profiler/profile_builder.py
ghanse: Fixed: now uses `isinstance(column_type, (T.IntegerType, T.LongType, T.StringType))`.

src/databricks/labs/dqx/profiler/profile_builder.py
ghanse: Fixed: merged into a single isinstance call — `isinstance(column_type, (T.NumericType, T.DateType, T.TimestampType, T.TimestampNTZType))`.

src/databricks/labs/dqx/profiler/profile_builder.py
ghanse: Fixed: added early return guard when `total_count == 0`.

src/databricks/labs/dqx/profiler/profile_builder.py
ghanse: Fixed: added the same `total_count == 0` guard here as well.

src/databricks/labs/dqx/profiler/profile_builder.py
ghanse: Fixed: now falls back to `sigmas` key for backward compat.

src/databricks/labs/dqx/profiler/profile_builder.py
ghanse: Fixed: added type annotation — `PROFILE_BUILDER_REGISTRY: dict[str, DQProfileType] = {}`.

src/databricks/labs/dqx/profiler/profile_builder.py
ghanse: Fixed — switched to `isinstance(column_type, (T.CharType, T.StringType, T.VarcharType))`.

src/databricks/labs/dqx/profiler/profile_builder.py
ghanse: Fixed — switched to `isinstance(column_type, (T.IntegerType, T.LongType, T.StringType))`.

src/databricks/labs/dqx/profiler/profile_builder.py
ghanse: Fixed — merged into single `isinstance(column_type, (T.DateType, T.NumericType, T.TimestampNTZType, T.TimestampType))`.

src/databricks/labs/dqx/profiler/profile_builder.py
ghanse: Fixed — added `if total_count == 0: return None` guard.

src/databricks/labs/dqx/profiler/profile_builder.py
ghanse: Fixed — added `if total_count == 0: return None` guard.

src/databricks/labs/dqx/profiler/profile_builder.py
ghanse: Fixed — now reads both keys for backward compat: `profiler_options.get("num_sigmas", profiler_options.get("sigmas", 3))`.

src/databricks/labs/dqx/profiler/profile_builder.py
ghanse: Fixed — added `dict[str, DQProfileType]` type annotation.

src/databricks/labs/dqx/profiler/profile.py
```diff
@@ -0,0 +1,25 @@
...
+    parameters: dict[str, Any] | None = None
+    filter: str | None = None
+
+
+@dataclass(frozen=True)
+class DQProfileType:
+    """Data quality profile type class representing a data quality rule candidate type."""
+
+    name: str
+    builder: Callable[[DataFrame, str, DataType, dict[str, Any], dict[str, Any]], DQProfile | None]
```
mwojtyczka, line 43: it would be good to provide a comment to explain each arg

src/databricks/labs/dqx/profiler/profile_builder.py
```diff
@@ -0,0 +1,640 @@
...
+        column_type: Input column type
+        profiler_metrics: Column-level statistics (includes summary stats min/max/mean/stddev for numeric types)
+        profiler_options: Configuration options for the DQProfiler
+
+    Returns:
+        A 'min_max' DQProfile
+    """
+    column_alias = df.columns[0]
+    if isinstance(column_type, T.DateType):
+        cast_df = df.select(F.col(column_alias).cast("timestamp").cast("bigint").alias(column_alias))
```
mwojtyczka, line 342: it's not super obvious that we are converting to Unix epoch time. A comment here to explain cast to bigint would be nice

src/databricks/labs/dqx/profiler/profile.py
```diff
@@ -19,7 +19,25 @@ class DQProfile:
 
 @dataclass(frozen=True)
 class DQProfileType:
```
ghanse: Should it be renamed to `DQProfileBuilder`?

src/databricks/labs/dqx/profiler/profile_builder.py
```diff
@@ -266,7 +271,7 @@ def _supports_distinct(column_type: T.DataType) -> bool:
     Returns:
         True if the column supports distinct operations, otherwise False
     """
-    return isinstance(column_type, (T.IntegerType, T.LongType, T.StringType))
+    return isinstance(column_type, (T.CharType, T.IntegerType, T.LongType, T.StringType, T.VarcharType))
```
ghanse: Should we note this change in the release notes? These are additional types that have been added which are not present in the previous implementation: https://github.com/databrickslabs/dqx/blob/280ccd2343986e7980e825c869fc8ca7524372ee/src/databricks/labs/dqx/profiler/profiler.py#L855

src/databricks/labs/dqx/profiler/profiler.py
```diff
@@ -383,21 +387,28 @@ def _profile(
 
             column_df = df.select(field_name).dropna()
             column_label = column_df.columns[0]
-            if isinstance(field_type, T.StringType) and trim_strings:
+            is_text = isinstance(field_type, (T.CharType, T.StringType, T.VarcharType))
```
ghanse: Maybe abstracting this to a type variable is cleaner?

```
TextType = T.CharType | T.StringType | T.VarcharType
...

if isinstance(typ, TextType):
  ...
```

src/databricks/labs/dqx/profiler/profile_builder.py
mwojtyczka: added to the PR description, will include in release notes when making the release

src/databricks/labs/dqx/profiler/profiler.py
mwojtyczka: refactore

src/databricks/labs/dqx/profiler/profile.py
mwojtyczka: renamed

PR #1061, opened 2026-03-06
feat(geo): add are_polygons_mutually_disjoint check function

### LINKED ISSUE #606: [FEATURE]: Check if a column of polygons is mutually disjoint

### Problem statement

A column with polygons can, e.g., represent a partitioning of an area (think of the U.S and all of its different states). We want to check if all polygons are mutually disjoint, i.e., if we take any two polygons from the column, the intersection of those polygons should be empty.

### Proposed Solution

Create a dataset rule that captures this check

### Additional Context

_No response_

fedeflowers:
Implement a new geo DQ check that validates whether polygons in a column are mutually disjoint using `ST_Intersects`. Supports `row_filter`, and handles nulls and invalid geometries gracefully. Includes comprehensive test coverage.

## Changes

* **Implementation:** Added `are_polygons_mutually_disjoint` using native Spark spatial intersections.
* **H3 vs Native Optimization:** I explored an alternative approach using H3 indexing with a configurable `h3_resolution` parameter. However, this was discarded for the following reasons:
    * **Incompatibility with Photon:** H3 indexing does not leverage Photon's native spatial optimizations for `st_intersects`, negating a key performance benefit.
    * **Usability concerns:** It requires users to understand and correctly tune the resolution parameter — setting it too high can generate an excessive number of H3 cells, leading to dropped rows and compromised accuracy.
* **Bounding Filters:** Omitted manual pre-bound spatial filtering. Photon natively optimizes spatial data and joins. If a user runs this without Photon, the algorithm will be slower but will remain completely accurate.
* **Intersection Strictness:** Polygons are evaluated strictly. If any point is shared—including touching edges or boundaries—the polygons are considered intersecting, not disjoint. 
* **Tests:** Added only integration tests since databricks runtime is required

### Linked issues

Resolves #606

### Tests

- [ ] manually tested
- [ ] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -580,7 +580,7 @@ indent-string = "    "
 max-line-length = 100
 
 # Maximum number of lines in a module.
-max-module-lines = 10000
+max-module-lines = 11000
 
 # Allow the body of a class to be on the same line as the declaration if body
 # contains single statement.


src/databricks/labs/dqx/geo/check_funcs.py
ghanse, line 1: Do you think this is clearer? I am not sure which will be faster.

```
df.createOrReplaceTempView("_is_mutually_disjoint")
disjoint_df = spark.sql("""
  select 
    a.id,
    a.geom_col,
    not exists (
      select 1 from _is_mutually_disjoint b
      where b.id > a.id and st_intersects(b.geom_col, a.geom_col)
    ) as is_disjoint
  from _is_mutually_disjoint a
""")
...
```

fedeflowers: I've updated the intersection logic to use the native DataFrame API rather than managing Spark SQL temp views. 

**Reasons:**
* **Efficient ID Unpacking:** Using `explode(array(a.id, b.id))` extracts both sides of the colliding pair into a single column in one step, avoiding the need for `UNION` operations.
* **Safer Execution:** Relying strictly on the DataFrame API removes the need for `createOrReplaceTempView`, keeping the code cleaner and eliminating any risk of namespace collisions or concurrency issues in shared Spark sessions.

If you think sql is better for readability and maintainability I can change it

ghanse: I think the DataFrame APIs are better honestly. With SQL we introduce the side effect of creating/updating a view and polluting the session.

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -821,3 +822,101 @@ def _compare_spatial_sql_function_result(
...
+            or other row-count limits before using this rule.
+
+        Args:
+            df: Input DataFrame containing the geometry column to check.
+
+        Returns:
+            DataFrame with an additional boolean column marking rows whose polygon
+            intersects with at least one other polygon in the dataset.
+        """
+        df_with_id = df.withColumn(row_id_col, F.monotonically_increasing_id())
```
mwojtyczka: **`monotonically_increasing_id()` is non-deterministic — correctness risk in self-join**

`monotonically_increasing_id()` is a non-deterministic transformation. When `prep_df.alias("a")` and `prep_df.alias("b")` are both scanned in the self-join, Spark may re-evaluate the underlying plan for each side independently, assigning **different IDs to the same row** on each scan. This means the back-join to `df_with_id` at line ~903 can silently produce wrong results — rows matched to the wrong intersecting ID, or missed entirely.

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -821,3 +822,101 @@ def _compare_spatial_sql_function_result(
...
+    native Spark spatial intersections are used.
+
+    Args:
+        column: Column to check; can be a string column name or a column expression
+        row_filter: Optional SQL expression to filter rows before checking for intersections.
+
+    Returns:
+        Column object and a callable to apply the check to a DataFrame.
+
+    Note:
```
mwojtyczka, line 846: **O(n²) performance warning belongs on the public function, not only on the inner `apply()`**

The warning about the self-join being O(n²) is currently only in the docstring of the nested `apply()` function, which is invisible to users calling `are_polygons_mutually_disjoint()` and reading its signature in IDE hints or generated docs.

Move (or copy) the warning to the outer function's `Note` section:

```
Note:
    This function requires Databricks runtime 17.1 or above.
    This check performs a self-join over all valid polygons in the input DataFrame,
    which is O(n²). Use ``row_filter`` to limit the rows evaluated on large tables.
    Photon activation is suggested to use optimised spatial computation.
```

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -821,3 +822,99 @@ def _compare_spatial_sql_function_result(
...
+
+        Returns:
+            DataFrame with an additional boolean column marking rows whose polygon
+            intersects with at least one other polygon in the dataset.
+        """
+        # Hash all rows first (needed for final join to preserve all rows)
+        df_with_hash = df.withColumn(row_id_col, F.xxhash64(F.struct("*")))
+
+        # Apply filter after hashing (so hash is available for final join to preserve all rows)
+        filtered_df = df_with_hash.filter(F.expr(row_filter)) if row_filter is not None else df_with_hash
```
fedeflowers, line 878: should add sql injection warning in the docstring

Warning:
    The `row_filter` argument is evaluated directly as a Spark SQL expression via `F.expr()`. To prevent SQL injection, ensure this value is fully sanitized or strictly comes from a trusted source.

mwojtyczka: we have a separate ticket for this. You can skip. We need unified way of checking this across all check functions

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -821,3 +822,99 @@ def _compare_spatial_sql_function_result(
         F.when(is_valid_cond, is_valid_message).otherwise(compare_message),
         f"{col_str_norm}_{spatial_quantity_name}_{compare_op_name}_limit",
     )
+
+
+@register_rule("dataset")
+def are_polygons_mutually_disjoint(
```
mwojtyczka, line 828: @fedeflowers can you please also update documentation?

PR #1062, opened 2026-03-09
Anomaly tests, minor improvements

vb-dbrks:
## Changes
- Refactored MLflow experiment management to cache experiments per worker process, ensuring unique experiments for parallel test execution.
- Updated integration tests to use a consistent anomaly threshold for deterministic results, enhancing test reliability.

These changes streamline the testing process and improve the management of MLflow experiments across different test sessions.

### Linked issues
No linked issues, flaky and long running tests for every PR

Resolves #..

### Tests
Locally run

### DIFF
--- a/tests/integration_anomaly/conftest.py
+++ b/tests/integration_anomaly/conftest.py
@@ -30,6 +30,9 @@ logging.getLogger("tests").setLevel("DEBUG")
 logging.getLogger("databricks.labs.dqx").setLevel("DEBUG")
 
 logger = logging.getLogger(__name__)
+# Process-scoped cache for MLflow experiment (one per xdist worker); lifecycle cleared by
+# _cleanup_mlflow_worker_experiment at session end. Not thread-safe.
+_MLFLOW_WORKER_EXPERIMENT_CACHE: dict[str, str | None] = {"id": None, "path": None}
 
 
 # -----------------------------------------------------------------------------
@@ -398,9 +401,11 @@ def _delete_mlflow_experiment_by_name(experiment_path: str) -> None:
         logger.warning(msg)
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture
 def mlflow_worker_experiment(ws):
-    """Create one MLflow experiment per worker session; reuse across tests and clean up on teardown."""
+    """Create one MLflow experiment per xdist worker (via module cache); reuse across tests.
+    Cleanup is done by _cleanup_mlflow_worker_experiment at session end. Function-scoped so
+    we can depend on ws (pytester provides ws as function-scoped)."""
     tracking_uri = os.environ.get("MLFLOW_TRACKING_URI")
     if not tracking_uri:
         local_mlflow_db = os.environ.get("MLFLOW_LOCAL_DB")
@@ -414,6 +419,10 @@ def mlflow_worker_experiment(ws):
     mlflow.set_tracking_uri(tracking_uri)
     mlflow.set_registry_uri(registry_uri)
 
+    cached_path = _MLFLOW_WORKER_EXPERIMENT_CACHE["path"]
+    if cached_path is not None:
+        return (_MLFLOW_WORKER_EXPERIMENT_CACHE["id"], cached_path)
+
     os.environ.pop("MLFLOW_EXPERIMENT_ID", None)
     user_name = ws.current_user.me().user_name
     worker_id = os.environ.get("PYTEST_XDIST_WORKER", "main")
@@ -422,15 +431,27 @@ def mlflow_worker_experiment(ws):
     experiment = mlflow.set_experiment(experiment_path)
     experiment_id = getattr(experiment, "experiment_id", None) if experiment else None
     logger.debug(f"Created MLflow experiment {experiment_path} (id={experiment_id})")
+    _MLFLOW_WORKER_EXPERIMENT_CACHE["id"] = experiment_id
+    _MLFLOW_WORKER_EXPERIMENT_CACHE["path"] = experiment_path
+    return (experiment_id, experiment_path)
 
-    yield (experiment_id, experiment_path)
 
+@pytest.fixture(autouse=True, scope="session")
+def _cleanup_mlflow_worker_experiment():
+    """Clear MLflow experiment and cache at end of worker session."""
+    yield
+    experiment_path = _MLFLOW_WORKER_EXPERIMENT_CACHE["path"]
+    if experiment_path is None:
+        return
+    experiment_id = _MLFLOW_WORKER_EXPERIMENT_CACHE["id"]
     os.environ.pop("MLFLOW_EXPERIMENT_ID", None)
     os.environ.pop("MLFLOW_EXPERIMENT_NAME", None)
     if experiment_id:
         _delete_mlflow_experiment(experiment_id)
     else:
         _delete_mlflow_experiment_by_name(experiment_path)
+    _MLFLOW_WORKER_EXPERIMENT_CACHE["id"] = None
+    _MLFLOW_WORKER_EXPERIMENT_CACHE["path"] = None
 
 
 @pytest.fixture(autouse=True)


tests/integration_anomaly/conftest.py
```diff
@@ -2,24 +2,25 @@
 
 import logging
 import os
-from typing import Any
 from uuid import uuid4
 
 import mlflow
```
mwojtyczka, line 8: **Concern: Module-level mutable state for caching**

`_MLFLOW_WORKER_EXPERIMENT_CACHE` is a module-level mutable dict. While this works with xdist (each worker is a separate process), it creates an implicit coupling between the `mlflow_worker_experiment` fixture and this global. Consider documenting that this is intentionally process-scoped (not thread-safe) — or use a `session`-scoped fixture to hold the state instead, which makes the lifecycle explicit via pytest's fixture system rather than relying on module globals.

tests/integration_anomaly/conftest.py
```diff
@@ -30,6 +31,7 @@
 logging.getLogger("databricks.labs.dqx").setLevel("DEBUG")
 
 logger = logging.getLogger(__name__)
+_MLFLOW_WORKER_EXPERIMENT_CACHE: dict[str, str | None] = {"id": None, "path": None}
```
mwojtyczka, line 368: **Suggestion: Make fixture scope explicit**

This fixture is function-scoped (default) but its purpose is to create-once-per-worker via the module cache. The first call creates the experiment, subsequent calls return cached values. This is effectively session-scoped behavior implemented via a module global.

Consider either:
1. Making this `@pytest.fixture(scope="session")` and storing state in the fixture itself (eliminating the module-level cache), or
2. Adding a comment explaining why function scope + module cache is preferred over session scope

Option 1 would be cleaner:
```python
@pytest.fixture(scope="session")
def mlflow_worker_experiment(ws):
    # ... create experiment once per session ...
    yield (experiment_id, experiment_path)
    # cleanup here instead of separate fixture
```

tests/integration_anomaly/test_helpers_data.py
```diff
@@ -0,0 +1,58 @@
+"""Lightweight data-only helpers for anomaly tests (no PySpark/dqx). Use for unit tests and fast imports."""
+
+# Match tests.integration_anomaly.constants for test point "clear_anomaly"
+_OUTLIER_AMOUNT = 9999.0
+_OUTLIER_QUANTITY = 1.0
```
mwojtyczka: **Issue: Duplicated constants**

`_OUTLIER_AMOUNT = 9999.0` and `_OUTLIER_QUANTITY = 1.0` duplicate values from `constants.py` (`OUTLIER_AMOUNT`, `OUTLIER_QUANTITY`). If these values change in `constants.py`, this file will silently diverge.

Since this module is meant to be "lightweight, no PySpark/dqx", the intent to avoid importing from constants is understandable — but `constants.py` also has no heavy dependencies. Consider importing from `constants.py` directly to avoid the duplication risk:
```python
from tests.integration_anomaly.constants import OUTLIER_AMOUNT, OUTLIER_QUANTITY
```

.github/workflows/anomaly.yml
```diff
@@ -91,19 +91,8 @@ jobs:
         timeout-minutes: 200
         env:
           COVERAGE_FILE: ${{ github.workspace }}/.coverage
-          DATABRICKS_HOST: ${{ env.DATABRICKS_HOST }}
```
mwojtyczka, line 94: **Verify: Env vars are inherited from job-level scope**

This removes explicit env var forwarding (`DATABRICKS_HOST`, `DATABRICKS_CLUSTER_ID`, `MLFLOW_TRACKING_URI`, `MLFLOW_REGISTRY_URI`, `MLFLOW_HTTP_REQUEST_TIMEOUT`, `MLFLOW_HTTP_REQUEST_MAX_RETRIES`, etc.) from the step.

Please confirm these are set at the job or workflow level and inherited by steps. In particular, `MLFLOW_REGISTRY_URI: databricks-uc`, `MLFLOW_HTTP_REQUEST_TIMEOUT: 600`, and `MLFLOW_HTTP_REQUEST_MAX_RETRIES: 10` were hardcoded values (not references to `env.*`) — removing them means the tests now rely on defaults. Is that intentional? If MLflow defaults to shorter timeouts, this could cause flaky failures on slow registrations.

tests/integration_anomaly/test_anomaly_apply_checks_by_metadata.py
```diff
@@ -165,9 +169,21 @@ def _has_is_not_null_error(err) -> bool:
     row1_errors = rows[1]["_errors"] if rows[1]["_errors"] is not None else []
```
mwojtyczka, line 169: **Good pattern, but consider clarifying the contract**

The shift from `assert len(row2_errors) > 0` to checking `_dq_info` presence is a good resilience improvement. However, the conditional `if anomaly_info.is_anomaly: assert len(row2_errors) > 0` means this test can pass silently even when the anomaly detector fails to flag a clear outlier (9999, 1).

Consider adding a soft signal, e.g.
```
# Anomaly row (transaction_id=3): identify by key; require at least one row has an error
anomaly_row = next((r for r in rows if r["transaction_id"] == 3), None)
assert anomaly_row is not None, "Expected row with transaction_id=3 (anomaly row)"
assert any(
   len(r["_errors"] or []) > 0 for r in rows
), "At least one row should have at least one error (anomaly check or other)"**
```

tests/integration_anomaly/test_helpers.py
```diff
@@ -0,0 +1,298 @@
...
+
+from databricks.labs.dqx.anomaly.anomaly_engine import AnomalyEngine
+from databricks.labs.dqx.anomaly.check_funcs import has_no_row_anomalies
+from databricks.labs.dqx.config import AnomalyParams
+from databricks.labs.dqx.rule import DQDatasetRule
+from tests.integration_anomaly.constants import DEFAULT_SCORE_THRESHOLD, OUTLIER_AMOUNT, OUTLIER_QUANTITY
+from tests.integration_anomaly.test_helpers_data import (
+    qualify_model_name,
+)
+
```
mwojtyczka: **Nit: Re-export creates indirection**

`qualify_model_name` lives in `test_helpers_data.py` but is imported here and used by other functions in this module. Some test files import it from `test_helpers`, others from `test_helpers_data`. This split could be confusing.

Since `qualify_model_name` is a pure string utility with no heavy deps, consider keeping it in `test_helpers_data.py` only and having all callers import from there. The re-export here is not in `__all__` and adds an unnecessary indirection.

.github/workflows/anomaly.yml
```diff
@@ -91,19 +91,8 @@ jobs:
...
-          DATABRICKS_AUTH_TYPE: ${{ env.DATABRICKS_AUTH_TYPE }}
-          DATABRICKS_CONFIG_FILE: ${{ env.DATABRICKS_CONFIG_FILE }}
-          MLFLOW_ENABLE_DB_SDK: "true"
-          MLFLOW_TRACKING_URI: ${{ env.MLFLOW_TRACKING_URI }}
-          MLFLOW_REGISTRY_URI: "databricks-uc"
-          MLFLOW_HTTP_REQUEST_TIMEOUT: "600"
-          MLFLOW_HTTP_REQUEST_MAX_RETRIES: "10"
         run: |
-          hatch run pytest tests/integration_anomaly/ -v -rs -n 10 --cov --cov-report=xml --timeout=2000 --reruns 1 --reruns-delay 5
+          hatch run pytest tests/integration_anomaly/ -v -rs -n 10 --cov --cov-report=xml --timeout=2000 --reruns 1 --reruns-delay 5 --durations=10
```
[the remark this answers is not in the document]
mwojtyczka: yes, nightly.yml should use the same process

tests/integration_anomaly/conftest.py
```diff
@@ -138,15 +81,53 @@ def configure_mlflow_tracking(ws, _mlflow_session_state):
...
+
+
+@pytest.fixture(autouse=True)
+def configure_mlflow_tracking(mlflow_worker_experiment):
+    """Set MLflow experiment env for each test from the worker-scoped experiment fixture."""
+    experiment_id, experiment_path = mlflow_worker_experiment
+    if experiment_id:
+        os.environ["MLFLOW_EXPERIMENT_ID"] = experiment_id
+    os.environ["MLFLOW_EXPERIMENT_NAME"] = experiment_path
+    logger.debug(f"MLflow configured: experiment={experiment_path}")
```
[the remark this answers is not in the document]
mwojtyczka: yes

tests/integration_anomaly/conftest.py
```diff
@@ -419,340 +400,3 @@ def _train(
         return full_model_name, registry_table, columns
 
     return _train
-
```
mwojtyczka, line 422: I personally prefer to keep reusable functions for tests n conftest.py instead of creating many small helper modules

tests/integration_anomaly/conftest.py
vb-dbrks: session scoped mlflow_worker_experiment(ws) causes ScopeMismatch. I've Kept function-scoped fixture + module cache for now and a comment on the fixture explaining why... i.e. that it depends on ws; cleanup via _cleanup_mlflow_worker_experiment...

.github/workflows/anomaly.yml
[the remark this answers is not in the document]
vb-dbrks: done.

tests/integration_anomaly/conftest.py
vb-dbrks: Ok, makes sense. I've consolidated the helpers again.

tests/integration_anomaly/conftest.py
vb-dbrks: I've documented it as code comment.

PR #1063, opened 2026-03-10
feat: add suppress_skipped option and skipped flag for skipped checks

### LINKED ISSUE #953: [FEATURE]: Option to skip checks quietly and improve logging of skipped checks

### Problem statement

Ability to skip checks as introduced with #608 and not stop check execution if check is not valid because of missing columns is great. I love this idea. It allows a lot of simplification in our check process, because removes the need to pre-select which checks are applicable for a data frame. 

But if a check is skipped, then all rows of the data frame are added to the invalid dataframe and a log entry is added to eather `_errors` resp `_warnings` column. This is not a useful behaviour.

That a check was skipped should not be seen as invalidation of a row. Rows should not be added to invalid df just because of skipped checks.

And btw. these log entries are also not nicely identifyable in the array in `_errors` resp `_warnings` column. You need to text parse the message attribute, if the text starts with "Check evaluation skipped due to". There should be a structured way to identify these log entries

### Proposed Solution

Prio 1: Provide a configuration to skip quietly, no entry added to `_errors` resp `_warnings` column for skipped checks then and only rows with true issues would be included in invalid df. e.g.

`valid_df, invalid_df = dq_engine.apply_checks_by_metadata_and_split(test_df, checks, ref_dfs=ref_dfs, skip_quietly=true)` 

Prio 2 (in addition or instead of Prio 1): Introduce a new output data frame, which just contains the list of skipped checks similar to the current log messages at command line
`08:49:37  WARN [d.l.dqx.manager] Skipping check 'dmo_ae_faae_fraction' due to invalid check filter: 'AECATTT = 'FRACTION''`

`valid_df, invalid_df, skipped_checks_df = dq_engine.apply_checks_by_metadata_and_split(test_df, checks, ref_dfs=ref_dfs)` 

Prio 3: Add a new attribute `skipped = true` to log entry in `_errors` resp `_warnings` to enable clear identification of these log entries avoid need to look for message attribute starting with "Check evaluation skipped due to" to identify skipped (should be done anyhow)  
```json
{
  "name": "dmo_ae_faae_fraction",
  "skipped": true,
  "message": "Check evaluation skipped due to invalid check filter: 'AECATTT = 'FRACTION''",
  "columns": ["SUBJECTNAME", "AEGRPID", "AECAT"],
  "filter": "AECATTT = 'FRACTION'",
  "function": "foreign_key",
  "run_time": "2025-11-28T08:49:37.886Z",
  "run_id": "c033c831-9b1e-44c3-9562-063bc0dac94c",
  "user_metadata": {}
}  
``` 

### Additional Context

_No response_

Roshan1299:
## Changes

Implements Prio 1 and Prio 3 from #953:

- Added `suppress_skipped: bool = False` to `ExtraParams`. When `True`, checks skipped due to missing columns or invalid filters produce no entry in `_errors`/`_warnings` and do not cause rows to appear in the invalid DataFrame.
- Added `skipped` boolean field to `dq_result_item_schema`. Skipped checks now have `skipped=True` in their result entry, enabling structured identification without string-parsing the message.

### Linked issues

Resolves #953 

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

### DIFF
--- a/tests/integration/test_apply_checks.py
+++ b/tests/integration/test_apply_checks.py
@@ -9647,6 +9647,7 @@ def test_apply_checks_with_has_valid_schema_extra_columns_in_params(ws, spark):
         "run_time": RUN_TIME,
         "run_id": RUN_ID,
         "user_metadata": {},
+        "skipped": True,
     }
 
     expected_skip_permissive = {
@@ -9658,6 +9659,7 @@ def test_apply_checks_with_has_valid_schema_extra_columns_in_params(ws, spark):
         "run_time": RUN_TIME,
         "run_id": RUN_ID,
         "user_metadata": {},
+        "skipped": True,
     }
 
     expected = spark.createDataFrame(


src/databricks/labs/dqx/config.py
```diff
@@ -215,6 +215,7 @@ class ExtraParams:
     user_metadata: dict[str, str] = field(default_factory=dict)
     run_time_overwrite: str | None = None
     run_id_overwrite: str | None = None
+    skip_quietly: bool = False
```
mwojtyczka:
```suggestion
    suppress_skipped: bool = False
```
Reads more clearly

src/databricks/labs/dqx/manager.py
```diff
@@ -127,8 +128,11 @@ def process(self) -> DQCheckResult:
         """
         invalid_cols_message = self._get_invalid_cols_message()
         if invalid_cols_message:
-            # overwrite message but preserve all other fields in the result
-            result_struct = self._build_result_struct(condition=F.lit(invalid_cols_message))
+            if self.skip_quietly:
+                # return null condition so this check produces no entry in _errors/_warnings
+                return DQCheckResult(condition=F.lit(None), check_df=self.df)
```
mwojtyczka: **`F.lit(None)` is untyped — cast it to the result item schema**

All other `check_conditions` elements are of the struct type defined by `dq_result_item_schema`. `F.lit(None)` without a cast has `NullType`, which is a different Spark type. While PySpark usually coerces `NullType` in `F.array()`, it is fragile and inconsistent with how the rest of the codebase handles null conditions (see `engine.py` line 461 where `F.lit(None).cast(dq_result_schema)` is used explicitly).

Fix:
```python
from databricks.labs.dqx.schema.dq_result_schema import dq_result_item_schema
return DQCheckResult(condition=F.lit(None).cast(dq_result_item_schema), check_df=self.df)
```

src/databricks/labs/dqx/manager.py
```diff
@@ -158,6 +162,7 @@ def _build_result_struct(self, condition: Column) -> Column:
             F.create_map(*[item for kv in self.user_metadata.items() for item in (F.lit(kv[0]), F.lit(kv[1]))]).alias(
                 "user_metadata"
             ),
+            F.lit(True if skipped else None).alias("skipped"),
```
mwojtyczka: **Minor: `True if skipped else None` can be simplified**

`True if skipped else None` is equivalent to `skipped or None` in Python (since `False or None` evaluates to `None` and `True or None` evaluates to `True`):

```python
F.lit(skipped or None).alias("skipped"),
```

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Version changelog
 
+## 0.14.0
```
mwojtyczka: pls remove, we create the change log during release

docs/dqx/docs/guide/additional_configuration.mdx
```diff
@@ -129,3 +129,45 @@ You can customize the names of these result columns by specifying extra paramete
...
+
+When a check is skipped and `suppress_skipped` is not enabled, the result struct includes a `skipped` boolean field set to `True`. This allows downstream consumers to distinguish skipped checks from actual data quality failures without parsing the message string.
+
+```python
+checked_df = dq_engine.apply_checks(df, checks)
+
+# filter to only skipped entries in _errors
+from pyspark.sql import functions as F
+
+skipped = checked_df.select(F.explode("_errors").alias("e")).filter("e.skipped = true")
```
mwojtyczka:
```suggestion
skipped = checked_df.select(F.explode("_errors").alias("e")).filter(F.col("e.skipped") == True)
```

PR #1066, opened 2026-03-12
Support custom folder in CLI uninstall

### LINKED ISSUE #599: [FEATURE]: Enhance CLI to be able to uninstall dqx from a custom folder

### Problem statement

DQX provides a method to install it in the workspace in a custom folder. But the uninstallation method only works for standard user's home or global folder. 

### Proposed Solution

The CLI uninstallation method should prompt the user for custom installation folder similar to the installation.

### Additional Context

Relevant code: https://github.com/databrickslabs/dqx/blob/main/src/databricks/labs/dqx/installer/uninstall.py

moomindani:
## Summary
- Added `install_folder` parameter to `InstallationService.current()` so it can locate installations in custom workspace folders
- Updated `uninstall.py` to prompt for custom folder path, mirroring the install flow
- Added unit test for `InstallationService` with custom install folder

Resolves #599

## Test plan
- [x] Unit tests pass (`make test` — 785 passed)
- [x] Linting passes (`make lint`)
- [x] E2E verified on live workspace: install to custom folder → uninstall from custom folder → folder removed

### DIFF
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -156,7 +156,7 @@ jobs:
           echo "Benchmark report updated at $OUT"
           echo "BENCHMARK_REPORT=$OUT" >> $GITHUB_ENV
 
-      # for troubleshooting issues only
+      # For troubleshooting and as fallback for fork PRs (where commit-back is skipped so we can manually commit the baseline)
       - name: Upload baseline artifacts
         uses: actions/upload-artifact@v4
         with:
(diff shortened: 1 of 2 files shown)


src/databricks/labs/dqx/installer/uninstall.py
```diff
@@ -9,7 +11,20 @@
 
 
 if __name__ == "__main__":
-    logger.setLevel("INFO")
+    logger = get_logger(__file__)
```
mwojtyczka: do we need this, it is not used anywhere?

src/databricks/labs/dqx/installer/uninstall.py
mwojtyczka: I think we don't need this

PR #1078, opened 2026-03-15
feat: add variable substitution support for check definitions

### LINKED ISSUE #967: [FEATURE]: Parametrization of rules

### Problem statement

Currently the rules are defined statically and there is no way to parametrize them.

### Proposed Solution

Introduce a way to parametrize rules via variables substitution to make them more generic and scalable so that a rule can be used for across many tables and columns. 

Parametrization (variable substitiution) could be supported in:
* all string arguments of a function
* name
* filter
* message

Example: 
```
checks = [
      {
            "criticality": "error",
            "name": "{{ param1 }}_greater_than_{{ threshold }}",
            "check": {
                "function": "sql_expression",
                "arguments": {"expression": "{{ param1 }} > {{ threshold }}"},
            },
            "filter": "{{ param2 }} = 'column2'",
        },
        {
            "criticality": "error",
            "check": {
                "function": "is_not_null",
                "arguments": {"column": "{{ param2 }}"},
            },
        }
]

# Pass params
variables = {"param1": "column1", "threshold": 10m, "param2": "column2"}

valid_and_quarantine_df = engine.apply_checks(df, checks, ref_dfs=ref_dfs, variables=variables)
```

### Additional Context

Requested by multiple customers.

fedeflowers:
### Changes
Add variable substitution support for check definitions. This allows users to define reusable check templates with `{{ placeholder }}` syntax and resolve them prior to execution by passing a [variables] dictionary at load time.
**New functionality:**
* `resolve_variables()` in `utils.py` — recursively replaces `{{ key }}` placeholders in all string values of check definitions using a highly efficient single-pass substitution.
* New optional `variables` parameter on `load_checks()` and `load_checks_from_local_file()`. This delegates the templating to the data-loading layer so the engine always executes clean and self-contained rules.
* **Default Variables in `ExtraParams`**: The `DQEngine` now accepts default variables via `ExtraParams`. These are applied to all checks loaded by that engine instance unless overridden in a specific `load_checks`
* Supports scalar variable types (`str`, `int`, `float`, `bool`, `Decimal`, `datetime.date`, `datetime.datetime`, `datetime.time`).
* Logs warnings for any unresolved placeholders after substitution is complete.
* Rejects collection types (`list`, `dict`, `set` `tuple` and `None` with an explicit `InvalidParameterError`.
**Example usage:**
Using `load_checks` from a file or table:
```python
# checks.yml
# - check:
#     function: is_not_null
#     arguments:
#       column: "{{ col }}"
# Resolving templates at load time:
checks = engine.load_checks_from_local_file('checks.yml', variables={"col": "email"})
# Engine executes the resolved rules
engine.apply_checks_by_metadata(df, checks)
```
If building checks programmatically using raw dictionaries:

```
from databricks.labs.dqx.utils import resolve_variables
raw_checks = [{"check": {"function": "is_not_null", "arguments": {"column": "{{ col }}"}}}]
resolved_checks = resolve_variables(raw_checks, variables={"col": "email"})
engine.apply_checks_by_metadata(df, resolved_checks)
```
### Linked issues

Resolves #967

### Tests

* [x] manually tested
* [x] added unit tests
* [x] added integration tests
* [ ] added end-to-end tests
* [ ] added performance tests


ghanse (changes requested):
I think substitution should happen when the user loads checks. We should allow the user to pass variables to `load_checks` and delegate substitution to that method instead of adding to the `apply_checks` methods.

mwojtyczka (commented):
## Code Review

### Overview
Adds `{{ placeholder }}` variable substitution for check definitions. The core logic in `utils.py` is clean and well-tested. A few issues need addressing before merge.

### Summary of findings
- **`Any` in non-anomaly modules** — AGENTS.md explicitly forbids this in new code outside `anomaly/`; use a concrete union type instead
- **Lazy logging** — the warning log uses an f-string; project convention is `%s` lazy formatting
- **Double traversal** — `apply_variables` scans the check structure twice (once to substitute, once to find unresolved placeholders); the warning check can be folded into `_replace_template`
- **Integration tests bypass the engine API** — tests call `apply_variables()` directly instead of through engine methods
- **PR description vs. implementation mismatch** — the description says `variables` is added to `apply_checks_by_metadata()` and related methods, but only `load_checks()` and `load_checks_from_local_file()` received the parameter. Either extend the API or update the description.

src/databricks/labs/dqx/utils.py
```diff
@@ -541,6 +544,132 @@ def missing_required_packages(packages: list[str]) -> bool:
...
+    """
+    if not variables:
+        return checks
+
+    _validate_variable_types(variables)
+    str_variables = {k: str(v) for k, v in variables.items()}
+    resolved: list[dict] = _substitute_variables(checks, str_variables)
+
+    # Warn about any remaining unresolved placeholders
+    for check_def in resolved:
```
mwojtyczka: **Double traversal of the check structure**

The structure is already visited in full by `_substitute_variables` above. This second pass adds O(N) overhead for every `apply_variables` call.

Consider folding the unresolved-placeholder warning into `_replace_template`: after all keys have been substituted, check `_UNRESOLVED_PLACEHOLDER_PATTERN.search(text)` once per string and warn there. That keeps the traversal to a single pass.

tests/integration/test_apply_checks.py
```diff
@@ -9990,3 +9991,162 @@ def test_apply_checks_by_metadata_skip_checks_with_missing_columns(ws, spark):
...
+    checks = [
+        {
+            "criticality": "error",
+            "check": {
+                "function": "is_not_null_and_not_empty",
+                "arguments": {"column": "{{ col }}"},
+            },
+        },
+    ]
+    checks = apply_variables(checks, {"col": "b"})
```
mwojtyczka: This exercises `apply_variables` in isolation (already fully covered by the unit tests in `test_utils.py`) rather than testing real engine integration. Integration tests for this function are not needed.

tests/integration/test_apply_checks_and_save_in_table.py
```diff
@@ -2255,3 +2256,59 @@ def test_apply_checks_by_metadata_and_save_in_table_loads_checks_from_table(ws,
...
+    test_df.write.format("delta").mode("overwrite").saveAsTable(input_table)
+
+    checks = [
+        {
+            "name": "{{ col }}_is_null",
+            "criticality": "{{ crit }}",
+            "check": {"function": "is_not_null", "arguments": {"column": "{{ col }}"}},
+        },
+    ]
+    checks = apply_variables(checks, {"col": "a", "crit": "error"})
```
mwojtyczka: Integration testing should cover invoking `load_checks` with variable substitution. Using one of supported storage should be enough.

tests/unit/test_load_checks.py
```diff
@@ -84,3 +89,96 @@ def test_file_download_contents_read_none():
...
+    file_path.write_text(content, encoding="utf-8")
+
+    checks = DQEngineCore.load_checks_from_local_file(str(file_path), variables=None)
+
+    assert checks == [
+        {"criticality": "error", "check": {"function": "is_not_null", "arguments": {"column": "id"}}},
+    ]
+
+
+def test_load_checks_from_local_file_variables_empty(tmp_path):
```
mwojtyczka, line 129: Add also a negative test, where variable value is not provided, but the parameter is defined in the check

tests/integration/test_apply_checks_and_save_in_table.py
```diff
@@ -2255,3 +2256,59 @@ def test_apply_checks_by_metadata_and_save_in_table_loads_checks_from_table(ws,
         schema=expected_schema,
     )
```
mwojtyczka, line 2264: Cover edge cases, e.g. parameter present in the check but variable value not provided

tests/unit/test_utils.py
```diff
@@ -1,3 +1,4 @@
+import logging
 from datetime import date, datetime
 from decimal import Decimal
```
mwojtyczka, line 3: Please cover the following edge cases:

1. None value in variables is not tested:
`variables={"col": None}  # None as a value, not as the variables dict`

`None` is not in `_SCALAR_VARIABLE_TYPES` so it should raise InvalidParameterError, but there's no test confirming this. A user writing variables={"col": None} to "clear" a placeholder would get an error with no obvious message.

2. set and tuple rejection not tested. Only list and dict rejection are tested. Set and tuple also fail the scalar check but are not explicitly covered.
3. Dict keys are never substituted — not explicitly tested
The docstring says "Dict keys are not substituted" but no test verifies this. A check like {"{{ col }}": "value"} should leave the key unchanged; this should be an explicit assertion.
4. False bool → "False" not tested
Only True is covered. False → "False" is the expected output, but also worth noting that in SQL contexts neither "True" nor "False" are valid boolean literals — a user expecting "true" / "false" or "1" / "0" will be surprised.
5. float('nan') and float('inf') pass type validation
Both are float instances, so they pass _validate_variable_types and produce "nan" / "inf" — syntactically invalid in SQL expressions. No test documents this footgun.
6. Multiple unresolved placeholders in one string logs only one warning
"{{ a }} and {{ b }}" with only a resolved produces one warning for the entire string, not one per unresolved token. The warning message shows the full string, not the specific unresolved placeholder. No test confirms this behaviour or the message content.
7. No warning when variables is None/empty but placeholders exist
apply_variables([{"col": "{{ x }}"}], None) returns the original list unchanged with no warning. This is silent — a user who forgets to pass variables gets no feedback. Whether this is intentional is not documented by a test.
8. Variable key with leading/trailing whitespace
`variables={" col ": "value"}` — the regex pattern `\{\{\s* + re.escape(" col ") + \s*\}\}` would not match `{{ col }}` since the key itself has spaces. Behaviour is undefined.
9. Unicode variable values
`variables={"col": "prénom"} `— likely works fine with Python's re.sub, but untested.
10. Placeholder in top-level check key (e.g. "criticality") vs nested key — the existing tests cover criticality values but do not test that a placeholder in the literal string "check" (the key name) is left alone.

src/databricks/labs/dqx/utils.py
```diff
@@ -541,6 +544,132 @@ def missing_required_packages(packages: list[str]) -> bool:
...
+    """Raise :class:`InvalidParameterError` if any variable value is not a supported scalar type."""
+    for key, val in variables.items():
+        if not isinstance(val, _SCALAR_VARIABLE_TYPES):
+            raise InvalidParameterError(
+                f"Variable '{key}' has unsupported type '{type(val).__name__}'. "
+                f"Only scalar types are supported: str, int, float, bool, Decimal."
+            )
+
+
+def apply_variables(checks: list[dict], variables: dict[str, Any] | None) -> list[dict]:
```
mwojtyczka:
```suggestion
def resolve_variables(checks: list[dict], variables: dict[str, Any] | None) -> list[dict]:
```
It reads more naturally. apply_variables collides with the dominant apply_* pattern in DQX. Every other apply_* in the engine applies something to data; apply_variables applies something to check definitions — a fundamentally different subject. The project uses "resolve" as the verb for turning raw definitions into executable form, making resolve_variables the natural fit with existing naming conventions

src/databricks/labs/dqx/utils.py
```diff
@@ -541,6 +544,132 @@ def missing_required_packages(packages: list[str]) -> bool:
     return not all(find_spec(spec) for spec in packages)
 
 
+def _literal_replacer(val: str) -> Callable[[re.Match], str]:
```
mwojtyczka: **`_literal_replacer` is unnecessary**

The named higher-order function exists solely to avoid backslash interpretation in re.sub replacements — a lambda does the same thing in one line:
text = re.sub(pattern, lambda _: val, text)
Seven lines of code and a docstring can be deleted. The docstring on _replace_template already explains why a callable is used.

src/databricks/labs/dqx/utils.py
```diff
@@ -541,6 +544,132 @@ def missing_required_packages(packages: list[str]) -> bool:
...
+    """Return a ``re.sub`` replacer that always returns *val* literally."""
+
+    def replacer(_: re.Match) -> str:
+        return val
+
+    return replacer
+
+
+def _replace_template(text: str, variables: dict[str, str]) -> str:
```
mwojtyczka, line 556: **O(N × M) substitution complexity**

`_replace_template` loops over every variable for every string. With N variables and M strings across the check definitions, that's N × M regex compilations per call. For large check files or many variables this adds up. All variables could be matched in a single pass using a combined pattern, e.g. 

```
pattern = r"\{\{\s*" + re.escape(key) + r"\s*\}\}"
pattern = r"\{\{[ \t]*" + re.escape(key) + r"[ \t]*\}\}" # safer
        text = re.sub(pattern, lambda _: val, text)
```

src/databricks/labs/dqx/utils.py
```diff
@@ -541,6 +544,132 @@ def missing_required_packages(packages: list[str]) -> bool:
...
+            If ``None`` or empty the checks are returned unchanged.
+
+    Returns:
+        A new list of check dicts with placeholders resolved, or the original list
+        when no substitution is needed.
+
+    Raises:
+        InvalidParameterError: If any variable value is not a supported scalar type.
+    """
+    if not variables:
```
mwojtyczka, line 674: **SQL injection risk — document in the docstring**

Variable values substituted into `sql_expression` checks are passed directly to `F.expr()` with no safety check (unlike `sql_query`, which calls `is_sql_query_safe`). If `variables` is populated from an untrusted source (user input, external API, config file), a malicious value can inject arbitrary SQL:

```python
# template
{"expression": "{{ col }} > 0"}

# malicious variable
variables = {"col": "(SELECT secret FROM sensitive_table LIMIT 1)"}

# resolved -- passed verbatim to F.expr()
"(SELECT secret FROM sensitive_table LIMIT 1) > 0"
```

In Databricks the injected SQL runs under the caller's own identity so it cannot escalate privileges, but it could leak data from tables the Spark session has access to.

As a minimum there should be a security note to the `resolve_variables` docstring (and `load_checks` / `load_checks_from_local_file`) stating that variable values substituted into `sql_expression` checks are not sanitised, and callers are responsible for ensuring values come from trusted sources.

mwojtyczka (commented):
Two follow-up inline notes: a VariableValue type alias to eliminate Any from the public API, and a SQL injection caveat for sql_expression checks.

mwojtyczka (changes requested):
Going in the right direction. Left comments. In addition:
* Updated PR description
* Update docs: both cases load_checks and when rules are defined programmatically as dicts where `resolve_variables` func must be used

src/databricks/labs/dqx/base.py
```diff
@@ -1,7 +1,7 @@
 import abc
 from collections.abc import Callable
 from functools import cached_property
-from typing import final
+from typing import Any, final
```
mwojtyczka: **Use a `VariableValue` type alias instead of `Any`**

The accepted types are already captured in `_SCALAR_VARIABLE_TYPES`, so a type alias keeps the two in sync:

```python
VariableValue = str | int | float | bool | Decimal

def apply_variables(checks: list[dict], variables: dict[str, VariableValue] | None) -> list[dict]:
```

Apply the same alias to `_validate_variable_types`, `load_checks_from_local_file` (base + engine), and `load_checks`.

src/databricks/labs/dqx/utils.py
```diff
@@ -541,6 +544,132 @@ def missing_required_packages(packages: list[str]) -> bool:
...
+            raise InvalidParameterError(
+                f"Variable '{key}' has unsupported type '{type(val).__name__}'. "
+                f"Only scalar types are supported: str, int, float, bool, Decimal."
+            )
+
+
+def apply_variables(checks: list[dict], variables: dict[str, Any] | None) -> list[dict]:
+    """Apply variable substitution to check definitions.
+
+    Replaces ``{{ key }}`` placeholders in all string values of *checks* with the
```
mwojtyczka:
```suggestion
    Replaces **{{ key }}** placeholders in all string values of *checks* with the
```

Avoid using backticks in docstrings, as this can cause issues when rendering API documentation. See: https://databrickslabs.github.io/dqx/docs/dev/contributing/#writing-docstrings

src/databricks/labs/dqx/utils.py
```diff
@@ -32,6 +33,8 @@
 COLUMN_NORMALIZE_EXPRESSION = re.compile("[^a-zA-Z0-9]+")
 COLUMN_PATTERN = re.compile(r"Column<'(.*?)(?: AS (\w+))?'>$", re.DOTALL)
 INVALID_COLUMN_NAME_PATTERN = re.compile(r"[\s,;{}\(\)\n\t=]+")
+_UNRESOLVED_PLACEHOLDER_PATTERN = re.compile(r"\{\{.*?\}\}")
```
mwojtyczka: **Potential ReDoS**

```suggestion
_UNRESOLVED_PLACEHOLDER_PATTERN = re.compile(r"\{\{[^}]*\}\}")
```

Is uses .*? (non-greedy dot). On a pathological string with many `{{ `that never close, the regex engine can backtrack excessively: `"{{ " * 10000  # no closing }}`

Python's re module uses a backtracking engine so this can cause the warning scan loop to hang. The substitution loop in `_replace_template` has the same exposure for each variable's pattern.

The [^}]* character class cannot backtrack across } characters, making it linear time.

src/databricks/labs/dqx/utils.py
fedeflowers: I did like this first but got problems with variable declaration through make lint, so I created _literal_replacer to avoid these checks, but I understand it's useless code, is using # type: ignore the best way to approach this?

src/databricks/labs/dqx/utils.py
```diff
@@ -32,6 +33,8 @@
...
+_SCALAR_VARIABLE_TYPES = (str, int, float, bool, Decimal)
```
ghanse: Do we need to include `datetime.date`, `datetime.datetime`, or any other potential check function argument types?

ghanse (commented):
Should we allow users to pass a default set of variables for all checks loaded by a single `DQEngine` instance? For example if the user wants to create a single engine instance, set a run date variable, then load checks for Table A, Table B, ... with the run date substituted across datasets?

We do have `ExtraParams` which could be used to set this by default. Users should be allowed to override the default configuration IMO.

src/databricks/labs/dqx/utils.py
fedeflowers: Lazy logging: I see in core module engine.py that f string formatting is used, should I refactor to use %s formatting?

src/databricks/labs/dqx/utils.py
```diff
@@ -541,6 +546,112 @@ def missing_required_packages(packages: list[str]) -> bool:
...
+    Args:
+        text: Input string potentially containing **{{ key }}** placeholders.
+        variables: Pre-stringified mapping of placeholder names to values.
+
+    Returns:
+        String with all matching placeholders replaced.
+    """
+    if not variables:
+        if _UNRESOLVED_PLACEHOLDER_PATTERN.search(text):
+            logger.warning("Unresolved placeholder found: '%s'", text)  # pylint: disable=logging-too-many-args
```
ghanse: Can we use an f-string here and remove the pylint disable comment?

src/databricks/labs/dqx/utils.py
```diff
@@ -541,6 +546,112 @@ def missing_required_packages(packages: list[str]) -> bool:
...
+        key = match_obj.group(0).strip("{} \t")
+        if key in variables:
+            return variables[key]
+        unresolved.append(key)
+        return match_obj.group(0)
+
+    unresolved: list[str] = []
+    output = _UNRESOLVED_PLACEHOLDER_PATTERN.sub(_resolve, text)
+    if unresolved:
+        logger.warning("Unresolved placeholders found: %s", unresolved)  # pylint: disable=logging-too-many-args
```
ghanse: Same here.

src/databricks/labs/dqx/utils.py
fedeflowers: > ## Code Review
> ### Overview
> Adds `{{ placeholder }}` variable substitution for check definitions. The core logic in `utils.py` is clean and well-tested. A few issues need addressing before merge.
> 
> ### Summary of findings
> * **`Any` in non-anomaly modules** — AGENTS.md explicitly forbids this in new code outside `anomaly/`; use a concrete union type instead
> * **Lazy logging** — the warning log uses an f-string; project convention is `%s` lazy formatting
> * **Double traversal** — `apply_variables` scans the check structure twice (once to substitute, once to find unresolved placeholders); the warning check can be folded into `_replace_template`
> * **Integration tests bypass the engine API** — tests call `apply_variables()` directly instead of through engine methods
> * **PR description vs. implementation mismatch** — the description says `variables` is added to `apply_checks_by_metadata()` and related methods, but only `load_checks()` and `load_checks_from_local_file()` received the parameter. Either extend the API or update the description.

I've also seen that f-string format is used elsewhere in the core components, so which one is the one to use?

mwojtyczka: Yes, we use f-string, but %s (lazy formatting) is generally better for logging since formatting happens only if log level is enabled. However, f-strings are very fast so this is negligible. 

From design perspective, the logging API is explicitly designed like this: `logger.info(msg, *args)`
where:
msg = format string
args = values to interpolate later

That’s why %s is the intended pattern. 

If we need to disable pylint because of this, let's use f-string. Please don't disable pylint.

mwojtyczka:
```suggestion
            logger.warning("Unresolved placeholder found: '%s'", text)
```
pls don't disable pylint, if it doesn't work then use f-string

tests/integration/test_apply_checks_and_save_in_table.py
```diff
@@ -2255,3 +2255,62 @@ def test_apply_checks_by_metadata_and_save_in_table_loads_checks_from_table(ws,
...
+          check:
+            function: is_not_null
+            arguments:
+              column: "{{ col }}"
+        """
+    checks_file = tmp_path / "checks.yml"
+    checks_file.write_text(checks_yaml, encoding="utf-8")
+    checks = DQEngineCore.load_checks_from_local_file(str(checks_file), variables={"col": "a", "crit": "error"})
+
+    engine = DQEngine(ws, spark=spark, extra_params=EXTRA_PARAMS)
```
mwojtyczka: we should just validate the checks directly after loading without applying the checks.

tests/integration/test_apply_checks_variables.py
```diff
@@ -0,0 +1,334 @@
...
+          check:
+            function: is_not_null_and_not_empty
+            arguments:
+              column: "{{ col }}"
+        """
+    checks_file = tmp_path / "checks.yml"
+    checks_file.write_text(checks_yaml, encoding="utf-8")
+    checks = DQEngineCore.load_checks_from_local_file(str(checks_file), variables={"col": "b"})
+
+    checked = dq_engine.apply_checks_by_metadata(test_df, checks)
```
mwojtyczka: same as above, no need to apply the checks, just verify them against static expected checks

tests/integration/test_apply_checks_variables.py
```diff
@@ -0,0 +1,334 @@
...
+            arguments:
+              expression: "{{ expr_col }} > {{ threshold }}"
+        """
+    checks_file = tmp_path / "checks.yml"
+    checks_file.write_text(checks_yaml, encoding="utf-8")
+    checks = DQEngineCore.load_checks_from_local_file(
+        str(checks_file), variables={"col": "b", "expr_col": "a", "threshold": 1}
+    )
+
+    good, bad = dq_engine.apply_checks_by_metadata_and_split(test_df, checks)
```
mwojtyczka: same as above

tests/integration/test_apply_checks_variables.py
```diff
@@ -0,0 +1,334 @@
...
+              expression: "{{ col }} > {{ threshold }}"
+          filter: "{{ filter_col }} IS NOT NULL"
+        """
+    checks_file = tmp_path / "checks.yml"
+    checks_file.write_text(checks_yaml, encoding="utf-8")
+    checks = DQEngineCore.load_checks_from_local_file(
+        str(checks_file), variables={"col": "a", "threshold": 1, "filter_col": "a"}
+    )
+
+    checked = dq_engine.apply_checks_by_metadata(test_df, checks)
```
mwojtyczka: same as above

tests/integration/test_apply_checks_variables.py
```diff
@@ -0,0 +1,334 @@
...
+            function: is_not_null
+            arguments:
+              column: b
+        """
+    checks_file = tmp_path / "checks.yml"
+    checks_file.write_text(checks_yaml, encoding="utf-8")
+    checks = DQEngineCore.load_checks_from_local_file(str(checks_file), variables={"crit": "not_a_valid_criticality"})
+
+    status = DQEngine.validate_checks(checks)
+    assert status.has_errors
```
mwojtyczka: it would be good to validate the exact error

tests/integration/test_apply_checks_variables.py
```diff
@@ -0,0 +1,334 @@
...
+            "criticality": "{{ crit }}",
+            "check": {
+                "function": "is_not_null",
+                "arguments": {"column": "b"},
+            },
+        },
+    ]
+
+    status = DQEngine.validate_checks(checks)
+    assert status.has_errors
```
mwojtyczka: it would be good to validate the exact error

tests/integration/test_apply_checks_variables.py
```diff
@@ -0,0 +1,334 @@
...
+    config = FileChecksStorageConfig(location=str(checks_file))
+    checks = dq_engine.load_checks(config, variables={"environment": "prod", "nested_var": "new"})
+
+    # Verify substitution (Structural check)
+    assert checks[0]["check"]["arguments"]["column"] == "id"
+    assert checks[0]["user_metadata"]["env"] == "prod"
+    assert checks[0]["user_metadata"]["rule_id"] == "new"
+
+    # Apply checks to DataFrame (Functional check)
+    checked_df = dq_engine.apply_checks_by_metadata(df, checks)
```
mwojtyczka: same as above, no need to apply

tests/integration/test_apply_checks_variables.py
```diff
@@ -0,0 +1,334 @@
...
+    checks_file.write_text(checks_yaml, encoding="utf-8")
+
+    # Load file, which will warn and leave the placeholder
+    checks = DQEngineCore.load_checks_from_local_file(str(checks_file), variables={"different_var": "val"})
+
+    # Assert that the placeholder was left in the metadata (unresolved variable)
+    assert checks[0]["check"]["arguments"]["column"] == "{{ missing_col }}"
+
+    # Check function apply should not raise an exception, but instead skip the check and report it in the results
+    checked = dq_engine.apply_checks_by_metadata(test_df, checks)
```
mwojtyczka: same as above

src/databricks/labs/dqx/config.py
```diff
@@ -215,7 +216,7 @@ class ExtraParams:
     user_metadata: dict[str, str] = field(default_factory=dict)
     run_time_overwrite: str | None = None
     run_id_overwrite: str | None = None
-    suppress_skipped: bool = False
```
mwojtyczka, line 219: why is this removed? please revert this change

mwojtyczka (changes requested):

fedeflowers (commented):
- Fixed tests to only use load_cheks
- Removed pylint ignore
- Reverted change in config.py

mwojtyczka (commented):
## Review: Variable Substitution — Test Placement

The integration test file `tests/integration/test_apply_checks_variables.py` contains **55 tests**, but the vast majority don't need Spark or a workspace. This significantly slows CI — integration tests run against a live workspace, one at a time via the acceptance action.

### Tests that are duplicated across integration AND unit files

These tests exist in both `tests/integration/test_apply_checks_variables.py` and `tests/unit/test_utils.py` (or `test_load_checks.py`). **Remove them from the integration file:**

- All 37 `test_resolve_variables_*` tests — pure dict manipulation, no Spark
- `test_load_checks_from_local_file_with_variables` — reads YAML, calls static method
- `test_load_checks_from_local_file_variables_none`
- `test_load_checks_from_local_file_variables_empty`
- `test_load_checks_from_local_file_unresolved_placeholder`
- `test_load_checks_with_variables` — uses mocked SparkSession
- `test_load_checks_variables_none`
- `test_load_checks_with_engine_default_variables`
- `test_load_checks_per_call_overrides_engine_defaults`

### Tests that should move from integration to unit

These tests are only in the integration file but don't use `ws` or `spark` fixtures:

- `test_load_checks_by_metadata_with_variables` — only calls `DQEngineCore.load_checks_from_local_file()` (static method, no Spark)
- `test_load_checks_by_metadata_and_split_with_variables` — same, just variable substitution + assertion
- `test_load_checks_by_metadata_with_variables_name_and_filter` — same
- `test_validate_checks_with_variables` — calls `DQEngine.validate_checks()` (static, no Spark)
- `test_validate_checks_with_variables_invalid_after_substitution` — same
- `test_validate_checks_without_variables_fails_on_placeholders` — same
- `test_load_checks_with_missing_variable` — calls static `load_checks_from_local_file`, no Spark

### Tests that correctly belong in integration

Only **3 tests** actually need `ws` + `spark`:

- `test_extra_params_variables_substitution_and_overrides` — creates DQEngine with spark, applies checks to DataFrame
- `test_extra_params_variables_conflict_resolution` — same
- `test_extra_params_variables_fallback_to_defaults` — same

Even these 3 could potentially be unit tests using `create_autospec(SparkSession)` and a mocked `checks_handler_factory` (similar to the existing unit tests in `test_load_checks.py`), since they only test variable merging logic, not Spark execution.

### Summary

| Category | Count | Action |
|----------|-------|--------|
| Duplicated (exist in both unit + integration) | ~45 | Remove from integration |
| Should be unit tests | 7 | Move to unit |
| Correctly integration | 3 | Keep (or convert to unit with mocks) |

Moving these tests to unit would reduce integration test time significantly — each integration test runs against a live workspace.

tests/integration/test_apply_checks_variables.py
```diff
@@ -0,0 +1,275 @@
+import dataclasses
```
mwojtyczka: **MEDIUM — Most tests in this file don't need integration.** Of the ~55 tests here, only 3 actually use `ws`/`spark` fixtures (`test_extra_params_variables_substitution_and_overrides`, `test_extra_params_variables_conflict_resolution`, `test_extra_params_variables_fallback_to_defaults`). The rest are pure Python logic that should be unit tests.

tests/integration/test_apply_checks_variables.py
```diff
@@ -0,0 +1,275 @@
...
+from databricks.labs.dqx.engine import DQEngine, DQEngineCore
+from databricks.labs.dqx.config import FileChecksStorageConfig
+from tests.integration.conftest import (
+    REPORTING_COLUMNS,
+    EXTRA_PARAMS,
+)
+
+SCHEMA = "a: int, b: int, c: int"
+EXPECTED_SCHEMA = SCHEMA + REPORTING_COLUMNS
+
```
mwojtyczka: **Move to unit:** `test_load_checks_by_metadata_with_variables` only calls `DQEngineCore.load_checks_from_local_file()` which is a static method that reads YAML and applies variable substitution — no Spark, no workspace. Same for `test_load_checks_by_metadata_and_split_with_variables` and `test_load_checks_by_metadata_with_variables_name_and_filter`.

tests/integration/test_apply_checks_variables.py
```diff
@@ -0,0 +1,275 @@
...
+    )
+
+    assert checks == [
+        {
+            "criticality": "error",
+            "name": "b_null_check",
+            "check": {
+                "function": "is_not_null_and_not_empty",
+                "arguments": {"column": "b"},
+            },
```
mwojtyczka: **Move to unit:** `test_validate_checks_with_variables`, `test_validate_checks_with_variables_invalid_after_substitution`, and `test_validate_checks_without_variables_fails_on_placeholders` only call `DQEngine.validate_checks()` which is a static method — no Spark needed.

mwojtyczka (commented):
Inline comments on test placement.

tests/integration/test_apply_checks_variables.py
```diff
@@ -0,0 +1,275 @@
...
+    checks_file.write_text(checks_yaml, encoding="utf-8")
+
+    # Setup DQEngine with ExtraParams variables (Default values)
+    # Default variables: target_col=id, environment=dev, nested_var=old
+    extra_params = dataclasses.replace(
+        EXTRA_PARAMS,
+        variables={
+            "target_col": "id",
+            "environment": "dev",
+            "nested_var": "old",
```
mwojtyczka: **Remove — duplicated:** `test_load_checks_from_local_file_with_variables`, `test_load_checks_from_local_file_variables_none`, `test_load_checks_from_local_file_variables_empty`, `test_load_checks_with_variables`, `test_load_checks_variables_none`, `test_load_checks_from_local_file_unresolved_placeholder`, `test_load_checks_with_engine_default_variables`, `test_load_checks_per_call_overrides_engine_defaults` — all already exist in `tests/unit/test_load_checks.py`.

src/databricks/labs/dqx/utils.py
```diff
@@ -541,6 +546,112 @@ def missing_required_packages(packages: list[str]) -> bool:
...
+
+    Logs a warning for any **{{ ... }}** placeholders that remain unresolved after
+    substitution (e.g. misspelled variable names).
+
+    **Security note:** variable values substituted into **sql_expression** checks are
+    not sanitized and are passed directly to **F.expr()**. Callers must ensure that
+    variable values come from trusted sources to prevent SQL injection.
+
+    Args:
+        checks: List of check definition dictionaries (metadata format).
```
mwojtyczka, line 663: **Nit:** The `# type: ignore[return-value]` on `_substitute_variables` is needed because the function signature returns `object` but we know the input is `list[dict]` so the output will be too. Consider narrowing the return type with `@overload` to avoid the ignore, or accept it as-is since it's internal.

mwojtyczka (commented):
Additional inline comments on the source code changes.

src/databricks/labs/dqx/base.py
```diff
@@ -2,11 +2,14 @@
...
 from typing import final
+
 from pyspark.sql import DataFrame, Observation
+
+from databricks.labs.dqx.__about__ import __version__
 from databricks.labs.dqx.checks_validator import ChecksValidationStatus
 from databricks.labs.dqx.rule import DQRule
+from databricks.labs.dqx.utils import VariableValue
 from databricks.sdk import WorkspaceClient
-from databricks.labs.dqx.__about__ import __version__
```
mwojtyczka, line 9: any reason this has been reordered?

mwojtyczka (changes requested):
Left some small comments. Can you please also update documentation?

src/databricks/labs/dqx/base.py
fedeflowers: this follows alphabetical order with _ first

demos/dqx_quick_start_demo_library.py
ghanse, line 1: As this is more of an advanced/optional topic, I think it's best to leave the examples to `dqx_demo_library.py` and remove this from the intro demo.

docs/dqx/docs/guide/quality_checks_definition.mdx
```diff
@@ -720,6 +720,60 @@ Example checks saved in a Delta or Lakebase table (compact format — `for_each_
 
 If `run_config_name` is not provided, "default" is used. Typically, the input table or job name is used for run config name to establish a one-to-one mapping between tables or jobs and checks.
 
+## Variable Substitution
+
+DQX supports variable substitution in declarative check definitions (YAML, JSON, or Delta tables). This allows you to parameterize your quality rules and inject values at **load time** via the `variables` parameter in `load_checks`.
```
ghanse: Should we also document that default variable values can be added to the `ExtraParams` and that the default `DQEngine` values can be used at runtime?

ghanse (changes requested):
Overall looking good. Left a few small comments.

docs/dqx/docs/guide/quality_checks_definition.mdx
fedeflowers: yes, makes sense

docs/dqx/docs/guide/additional_configuration.mdx
```diff
@@ -171,3 +171,75 @@ from pyspark.sql import functions as F
...
+    )
+
+    # Load checks - uses 'min_temp' and 'max_temp' from defaults,
+    # but overrides 'region' specifically for this call.
+    resolved_checks = dq_engine.load_checks(
+        config=FileChecksStorageConfig(location="checks.yml"),
+    )
+    ```
+  </TabItem>
+  <TabItem value="Workflows" label="Workflows">
```
fedeflowers: I am afraid that variable resolution will not work on DQX workflows, as it's not called by the method `_apply_checks_for_run_config`

docs/dqx/docs/guide/additional_configuration.mdx
```diff
@@ -171,3 +171,75 @@ from pyspark.sql import functions as F
...
+        min_temp: 0
+        max_temp: 50
+        region: GLOBAL
+    ```
+  </TabItem>
+</Tabs>
+
+## Overwriting run metadata
+
+By default, DQX automatically generates a unique `run_id` for each engine instance and uses the current timestamp as the `run_time`. You can manually overwrite these values using `ExtraParams` if you need to align DQX results with external systems or re-run checks for a specific historical point in time.
```
fedeflowers, line 224: I saw that `run_id_overwrite` and `run_time_overwrite` explanation was missing in the additional param doc

fedeflowers (commented):
Added doc, left a few comments

docs/dqx/docs/guide/additional_configuration.mdx
ghanse: We can open a follow-up issue to add optional variable substitution to the installable workflows. It may be best to call this out as a warning in the docs until we can support this.

tests/unit/test_utils.py
```diff
@@ -520,3 +522,314 @@ def test_get_file_extension_with_path_object():
...
+    with caplog.at_level(logging.WARNING):
+        result = resolve_variables(checks, {"a": "x"})
+    assert result[0]["expr"] == "x and {{ b }}"
+    assert any("Unresolved placeholder" in msg for msg in caplog.messages)
+
+
+def test_resolve_variables_none_vars_no_warning(caplog):
+    checks = [{"col": "{{ x }}"}]
+    with caplog.at_level(logging.WARNING):
+        result = resolve_variables(checks, None)
```
ghanse, line 818: Maybe add another call here to test an empty dictionary.

```
with caplog.at_level(logging.WARNING):
  result = resolve_variables(checks, {})

assert result[0]["col"] == "{{ x }}"
assert not any("Unresolved placeholder" in msg for msg in caplog.messages)
```

Can go in the same test function IMO.

ghanse (approved):
LGTM

docs/dqx/docs/guide/additional_configuration.mdx
```diff
@@ -171,3 +171,69 @@ from pyspark.sql import functions as F
...
+    # but overrides 'region' specifically for this call.
+    resolved_checks = dq_engine.load_checks(
+        config=FileChecksStorageConfig(location="checks.yml"),
+    )
+    ```
+  </TabItem>
+</Tabs>
+
+<Admonition type="warning" title="Variable substitution in workflows">
+Variable substitution is not currently supported in DQX installable workflows. Variables cam be defined and stored as YAML in the configuration file but will not be applied during workflow execution.
```
mwojtyczka:
```suggestion
Variable substitution is not currently supported in DQX installable workflows. Variables can be defined and stored as YAML in the configuration file but will not be applied during workflow execution.

Variable substitution is only available when defining checks declaratively (as dictionaries or in files/tables). It is not supported when using DQX classes (e.g., `DQRowRule`) directly.
```

docs/dqx/docs/guide/additional_configuration.mdx
```diff
@@ -171,3 +171,69 @@ from pyspark.sql import functions as F
...
+                "max_temp": 50,
+                "region": "GLOBAL"
+            }
+        )
+    )
+
+    # Load checks - uses 'min_temp' and 'max_temp' from defaults,
+    # but overrides 'region' specifically for this call.
+    resolved_checks = dq_engine.load_checks(
+        config=FileChecksStorageConfig(location="checks.yml"),
```
mwojtyczka, line 201:
```suggestion
        config=FileChecksStorageConfig(location="checks.yml"),  variables={ "region": "EMEA" }
```

docs/dqx/docs/guide/quality_checks_definition.mdx
```diff
@@ -720,6 +720,66 @@ Example checks saved in a Delta or Lakebase table (compact format — `for_each_
...
```
mwojtyczka:
```suggestion
DQX supports variable substitution in declarative check definitions (YAML, JSON, or Delta tables). This allows you to parameterize your quality rules and inject values at **load time** from engine-level defaults and/or via the `variables` parameter in `load_checks`.
```

docs/dqx/docs/guide/quality_checks_storage.mdx
```diff
@@ -180,6 +180,13 @@ If you create checks as a list of DQRule objects, you can convert them using the
     # also works for absolute and relative workspace paths if invoked from Databricks notebook or job
     checks: list[dict] = dq_engine.load_checks(config=FileChecksStorageConfig(location="checks.yml"))
 
+    # load checks from a local file with variable substitution
+    # see more on variable substitution [here](/docs/guide/quality_checks_definition/#variable-substitution)
```
mwojtyczka:
```suggestion
```

docs/dqx/docs/guide/quality_checks_storage.mdx
mwojtyczka, line 217:
```suggestion
    ```
    <Admonition type="tip" title="Checks definition">
    For technical details and configuration examples on how to use variables substitution, see the guide [here](/docs/guide/quality_checks_definition/#variable-substitution).
    </Admonition>
```

src/databricks/labs/dqx/engine.py
```diff
@@ -1186,7 +1215,23 @@ def load_checks(self, config: BaseChecksStorageConfig) -> list[dict]:
             InvalidConfigError: If the configuration type is unsupported.
         """
         handler = self._checks_handler_factory.create(config)
-        return handler.load(config)
+        checks = handler.load(config)
+        merged = self._merge_variables(variables)
```
```suggestion
        merged_variables = self._merge_variables(variables)
```

src/databricks/labs/dqx/engine.py
```diff
@@ -1186,7 +1215,23 @@ def load_checks(self, config: BaseChecksStorageConfig) -> list[dict]:
...
+        return resolve_variables(checks=checks, variables=merged)
```
```suggestion
        return resolve_variables(checks=checks, variables=merged_variables)
```

mwojtyczka (changes requested):

tests/unit/test_checks_validation.py
```diff
@@ -456,3 +456,60 @@ def test_is_in_range_float_arguments():
     ]
     status = DQEngine.validate_checks(checks)
     assert not status.has_errors
+
+
+def test_validate_checks_with_variables(tmp_path):
```
mwojtyczka, line 491: we need to add test with sql_query to verify the behaviour as it can have  {{ input_placeholder }} inside the query string

mwojtyczka: added, so the behaviour is that a warning message will be shown. I extended the warning message to explain that for certain rules, the vars are replaced during runtime

mwojtyczka (changes requested):
This must work for save_checks method to. Currently saving errors out. We need to resolve vars during saving. I'm updating accordingly

mwojtyczka (approved):
LGTM

ghanse (approved):
LGTM

PR #1090, opened 2026-03-20
App MVP

### LINKED ISSUE #1058: [FEATURE]: No-Spark dependency for the DQX App (Reduce API scopes)

### Problem statement

The DQX App (`app/`) currently depends on PySpark / Spark Connect for all data operations, including table metadata retrieval, schema inspection, and SQL execution. This introduces a heavy dependency (`pyspark`, `databricks-connect`) into the app's runtime environment, which:

1. **Increases cold start time** — Establishing a Spark Connect session (`DatabricksSession.builder...serverless().getOrCreate()`) adds significant latency to every authenticated request.
2. **Increases resource consumption** — The app must allocate serverless compute for operations that are fundamentally lightweight metadata queries.
3. **Complicates the dependency graph** — PySpark pulls in a large transitive dependency tree that is unnecessary for the app's actual workload (config management, checks CRUD, AI-assisted rule generation).
4. Complicates the scopes needed for OBO Access

The architecture already supports this change. `TableDataProvider` is defined as a `Protocol` in `src/databricks/labs/dqx/llm/table_manager.py`, and `TableManager` accepts an injectable `repository` parameter. The `SparkTableDataProvider` is the only concrete implementation today.

### Proposed Solution

Create a new `SDKTableDataProvider` class that implements the `TableDataProvider` protocol using the **Databricks SDK Unity Catalog REST APIs** instead of `SparkSession.sql()`. The primary API to use is `WorkspaceClient.tables.get()` (`databricks.sdk.service.catalog.TablesAPI`), supplemented by `WorkspaceClient.statement_execution.execute_statement()` only where SQL execution is actually needed.

### Why `TablesAPI.get()` over Statement Execution for metadata

`TablesAPI.get(full_name, include_delta_metadata=True)` returns a `TableInfo` object that already contains:

- **`columns`**: `List[ColumnInfo]` — column names, data types, comments, nullability, partition index
- **`properties`**: `Dict[str, str]` — table properties (numRows, totalSize, constraints, etc.)
- **`table_constraints`**: `List[TableConstraint]` — primary key and foreign key constraints
- **`delta_runtime_properties_kvpairs`**: Delta-specific metadata (when `include_delta_metadata=True`)
- **`table_type`**, **`data_source_format`**, **`storage_location`**, **`owner`**, etc.

This means most `TableDataProvider` methods can be satisfied by a **single REST API call** with no SQL execution or warehouse required — faster, cheaper, and simpler.

### Key changes

1. **New class `SDKTableDataProvider`** in `src/databricks/labs/dqx/llm/table_manager.py` that:
   - Takes a `WorkspaceClient` (and optionally a `warehouse_id` for SQL execution) in its constructor
   - Implements all `TableDataProvider` protocol methods
   - Returns `pandas.DataFrame` results (matching the existing protocol contract)
   - Uses `TablesAPI.get()` for metadata operations and `StatementExecutionAPI` only for `execute_query`

2. **Update `DQLLMEngine`** (`src/databricks/labs/dqx/llm/llm_engine.py`):
   - Accept an optional `TableDataProvider` / `TableManager` instead of requiring `SparkSession`
   - Fall back to `SparkTableDataProvider` when Spark is available (backward compatible)

3. **Update `DQGenerator`** (`src/databricks/labs/dqx/profiler/generator.py`):
   - Accept an optional `TableDataProvider` / repository parameter
   - `get_column_metadata` in `llm_utils.py` currently uses `read_input_data(spark, input_config)` which requires Spark — provide an SDK-based alternative using `TablesAPI.get().columns`

4. **Update app dependencies** (`app/src/databricks_labs_dqx_app/backend/dependencies.py`):
   - Replace `get_spark` dependency with `SDKTableDataProvider` for the AI-generation endpoint
   - For `DQEngine` (used by checks load/save with `InstallationChecksStorageConfig` on workspace files), evaluate whether Spark is still needed
   - Remove Spark Connect session creation from request path where possible

### Protocol method mapping

| Method | Current (Spark SQL) | Proposed (SDK) |
|--------|-------------------|---------------|
| `get_table_columns` | `DESCRIBE TABLE EXTENDED {table}` | `TablesAPI.get(table).columns` → build DataFrame from `ColumnInfo` list |
| `get_existing_primary_key` | `SHOW TBLPROPERTIES {table}` + scan for 'primary' | `TablesAPI.get(table).table_constraints` → filter for `PrimaryKeyConstraint` |
| `get_table_properties` | `SHOW TBLPROPERTIES {table}` | `TablesAPI.get(table).properties` → convert `Dict[str, str]` to DataFrame |
| `get_column_statistics` | `DESCRIBE TABLE EXTENDED {table}` | `TablesAPI.get(table, include_delta_metadata=True).columns` + `delta_runtime_properties_kvpairs` |
| `get_table_column_names` | `spark.table(table).columns` | `[col.name for col in TablesAPI.get(table).columns]` |
| `execute_query` | `spark.sql(query)` | `StatementExecutionAPI.execute_statement(warehouse_id, statement=query)` |

### Optimization: cache the `TablesAPI.get()` call

Since `get_table_columns`, `get_existing_primary_key`, `get_table_properties`, `get_column_statistics`, and `get_table_column_names` all operate on the same table, the `SDKTableDataProvider` could cache the `TableInfo` result per table name to avoid redundant REST calls when `TableManager` calls multiple methods for the same table.

## Benefits

- **No warehouse needed for metadata** — 5 of 6 protocol methods can work without a SQL warehouse, using only the Unity Catalog REST API
- **Faster app response times** — REST API call vs Spark Connect session establishment + SQL execution
- **Lighter deployment footprint** — Potential to remove `pyspark`/`databricks-connect` from the app's dependencies entirely
- **Better separation of concerns** — The app uses the SDK for REST API calls, Spark for large-scale data processing in notebooks/jobs
- **Backward compatible** — The existing `SparkTableDataProvider` remains available for notebook/job usage; the Protocol pattern ensures both implementations are interchangeable

### Additional Context

- The `TableDataProvider` Protocol is already defined and the `TableManager` facade already supports dependency injection via its `repository` parameter — the architecture is ready for this change.
- The app's `DQEngine` usage (`load_checks` / `save_checks` with `InstallationChecksStorageConfig`) operates on workspace files, not Spark tables — these paths may not need Spark at all.
- The `TableConstraintsAPI` (`w.table_constraints`) can also be used to read/write primary key and foreign key constraints directly via REST, which would benefit the PK detection workflow.
- This would also benefit any headless/API-only deployment scenario where PySpark is not available.

laurencewells:
### Linked issues

This pull request introduces major improvements to the DQX App's documentation and deployment configuration, focusing on clarifying the architecture, authentication model, and deployment steps, as well as enhancing the Databricks Asset Bundle (`databricks.yml`) to provision all required resources automatically. The changes also update development dependencies and provide a comprehensive product overview in a new `CLAUDE.md` file.

**Key changes include:**

**1. Documentation and Product Context**
- Added `CLAUDE.md` with a detailed product overview, architecture diagram, personas, user journeys, and internal storage design for the DQX App. This helps new contributors and users understand the app's purpose, scope, and design decisions.
- Expanded and reorganized `README.md` to clarify the authentication model (OBO vs. SP), async job pattern for Spark operations, deployment steps, troubleshooting, and removed outdated configuration steps. [[1]](diffhunk://#diff-784b9d23d8736e81678a62849e887188883174aed313c3696867f30fbe8ca040L23-R95) [[2]](diffhunk://#diff-784b9d23d8736e81678a62849e887188883174aed313c3696867f30fbe8ca040L79-R149) [[3]](diffhunk://#diff-784b9d23d8736e81678a62849e887188883174aed313c3696867f30fbe8ca040R202-R203) [[4]](diffhunk://#diff-784b9d23d8736e81678a62849e887188883174aed313c3696867f30fbe8ca040L299-R345) [[5]](diffhunk://#diff-784b9d23d8736e81678a62849e887188883174aed313c3696867f30fbe8ca040L329-R367) [[6]](diffhunk://#diff-784b9d23d8736e81678a62849e887188883174aed313c3696867f30fbe8ca040L373-R411) [[7]](diffhunk://#diff-784b9d23d8736e81678a62849e887188883174aed313c3696867f30fbe8ca040L384-R429) [[8]](diffhunk://#diff-784b9d23d8736e81678a62849e887188883174aed313c3696867f30fbe8ca040L419-R441) [[9]](diffhunk://#diff-784b9d23d8736e81678a62849e887188883174aed313c3696867f30fbe8ca040L453-R475) [[10]](diffhunk://#diff-784b9d23d8736e81678a62849e887188883174aed313c3696867f30fbe8ca040R518-R525)

**2. Deployment Automation and Resource Provisioning**
- Enhanced `databricks.yml` to:
  - Define variables for catalog and schema names, and centralize app environment configuration.
  - Automatically provision required resources: SQL warehouse, Delta schema, and Databricks Job for async Spark tasks (profiler and dry-run), and wire their IDs into the app config. [[1]](diffhunk://#diff-1496165d7591083dfda9aa237adebed69d814ad60089756c941bacf9e74a81fbR45) [[2]](diffhunk://#diff-1496165d7591083dfda9aa237adebed69d814ad60089756c941bacf9e74a81fbL27-R137)
  - Grant necessary permissions and set up API scopes for workspace and catalog access.

**3. Dependency Updates**
- Replaced `databricks-connect` with `databricks-sql-connector[pyarrow]` in `pyproject.toml`, reflecting the new architecture where Spark operations are offloaded to a Databricks Job rather than running in-process.

**4. Development and Troubleshooting**
- Updated documentation to cover the new async job pattern, environment variables, and troubleshooting steps for profiler/dry-run operations. [[1]](diffhunk://#diff-784b9d23d8736e81678a62849e887188883174aed313c3696867f30fbe8ca040L23-R95) [[2]](diffhunk://#diff-784b9d23d8736e81678a62849e887188883174aed313c3696867f30fbe8ca040R518-R525)

These changes make the app easier to deploy, maintain, and understand, and set a strong foundation for future development.

---

**References:**  
[[1]](diffhunk://#diff-cb82e5c2bd201ba350a51a9ba43c9ccd760d7de7d64f37e39728880e0207fabeR1-R87) [[2]](diffhunk://#diff-784b9d23d8736e81678a62849e887188883174aed313c3696867f30fbe8ca040L23-R95) [[3]](diffhunk://#diff-784b9d23d8736e81678a62849e887188883174aed313c3696867f30fbe8ca040L79-R149) [[4]](diffhunk://#diff-784b9d23d8736e81678a62849e887188883174aed313c3696867f30fbe8ca040R202-R203) [[5]](diffhunk://#diff-784b9d23d8736e81678a62849e887188883174aed313c3696867f30fbe8ca040L299-R345) [[6]](diffhunk://#diff-784b9d23d8736e81678a62849e887188883174aed313c3696867f30fbe8ca040L329-R367) [[7]](diffhunk://#diff-784b9d23d8736e81678a62849e887188883174aed313c3696867f30fbe8ca040L373-R411) [[8]](diffhunk://#diff-784b9d23d8736e81678a62849e887188883174aed313c3696867f30fbe8ca040L384-R429) [[9]](diffhunk://#diff-784b9d23d8736e81678a62849e887188883174aed313c3696867f30fbe8ca040L419-R441) [[10]](diffhunk://#diff-784b9d23d8736e81678a62849e887188883174aed313c3696867f30fbe8ca040L453-R475) [[11]](diffhunk://#diff-784b9d23d8736e81678a62849e887188883174aed313c3696867f30fbe8ca040R518-R525) [[12]](diffhunk://#diff-1496165d7591083dfda9aa237adebed69d814ad60089756c941bacf9e74a81fbR4-R33) [[13]](diffhunk://#diff-1496165d7591083dfda9aa237adebed69d814ad60089756c941bacf9e74a81fbR45) [[14]](diffhunk://#diff-1496165d7591083dfda9aa237adebed69d814ad60089756c941bacf9e74a81fbL27-R137) [[15]](diffhunk://#diff-f79644217c104a26d0814d8faf2afc24c0239350806e80ff239b93271888e465L13-R13)

Resolves #..

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests


app/src/databricks_labs_dqx_app/backend/common/authorization.py
```diff
@@ -0,0 +1,53 @@
...
+from typing import Annotated
+
+from fastapi import Depends, Header, HTTPException, status
+
+logger = logging.getLogger(__name__)
+
+
+class UserRole(StrEnum):
+    ADMIN = "admin"
+    DATA_STEWARD = "data_steward"
```
mwojtyczka: Data Engineer persona is missing

mwojtyczka (commented):
## Code Review: PR #1090 — App Refactor

Good architectural direction — service layer, versioned routes, async job pattern, caching. Main concerns below.

---

### HIGH — SQL Injection via String Interpolation

**Files:** `services/rules_catalog_service.py`, `app_settings_service.py`, `view_service.py`, `job_service.py`

All SQL uses f-strings with user input. The escaping `value.replace("'", "\\'")` is **wrong for Databricks SQL** — it uses ANSI `''` not backslash escaping. The migration code correctly uses `chr(39)*2` but the services use `\\'` — inconsistent.

Key vectors:
- `view_service.py`: `f"CREATE OR REPLACE VIEW {view_name} AS SELECT * FROM {source_table_fqn}"` — no validation
- `rules_catalog_service.py`: `table_fqn` from route path params used in SQL
- `app_settings_service.py`: config values used in SQL

**Fix:** Parameterized queries, or at minimum validate identifiers match `^[a-zA-Z0-9_]+(\.[a-zA-Z0-9_]+)*$` and use `''` for string escaping.

---

### HIGH — No `table_fqn` Validation

`table_fqn` in path params/request bodies is never validated as a three-part identifier. Add `@field_validator` on Pydantic models.

---

### MEDIUM — Cache Doesn't Store Falsy Values (Bug)

**File:** `cache.py` — checks `if result is not None`, so `[]`, `{}`, `0`, `False` are never cached. Use a sentinel `_MISSING = object()`.

---

### MEDIUM — Authorization Stubbed

`get_user_role` always returns `ADMIN`. `require_role` missing on `approve_rules`/`reject_rules` despite docstrings saying "admin only".

---

### MEDIUM — Migration Failure Silently Swallowed

**File:** `app.py` — comment says "will retry on first request" but no retry exists. App proceeds with missing tables.

---

### MEDIUM — `account users` Gets Write Access to tmp Schema

**File:** `databricks.yml` — `dqx_tmp_schema` grants `CREATE_TABLE` to `account users`. Restrict to SP or specific group.

---

### MEDIUM — SP Application ID Defaults to All-Zeros

**File:** `databricks.yml` — deployer who forgets to override gets confusing runtime failures. Remove default or make required.

---

### MEDIUM — `view_fqn` in `spark.sql()` Unsanitized

**File:** `tasks/dqx_task_runner.py` — validate matches expected pattern before using in SQL.

---

### LOW — Missing Tests

No unit tests for: `AiRulesService`, `AppSettingsService`, `authorization.py`, `dqx_task_runner.py`, route handlers for config/settings/generate/discovery.

---

### LOW — Duplicate SQL Boilerplate

Four classes have identical `_execute`/`_query` helpers. Extract shared utility.

---

### LOW — Cache `_locks` Dict Grows Unboundedly

Locks created per key, never cleaned up after expiry.

mwojtyczka (commented):
Inline review comments below.

app/src/databricks_labs_dqx_app/backend/services/view_service.py
```diff
@@ -0,0 +1,70 @@
...
+        Returns the fully qualified view name.  The view is created using
+        the caller's OBO token so that the user's table permissions are
+        enforced.
+        """
+        view_id = uuid4().hex[:12]
+        view_name = f"{self._catalog}.{self._schema}.tmp_view_{view_id}"
+        limit_clause = f" LIMIT {sample_limit}" if sample_limit else ""
+        sql = f"CREATE OR REPLACE VIEW {view_name} AS SELECT * FROM {source_table_fqn}{limit_clause}"
+
+        logger.info("Creating view %s from %s", view_name, source_table_fqn)
```
mwojtyczka, line 78: **HIGH - SQL Injection:** `source_table_fqn` is user input passed directly into SQL via f-string with no validation.

An attacker could pass `x; DROP TABLE important_data; --` as the table name.

**Fix:** Validate `source_table_fqn` matches a strict three-part identifier pattern before using in SQL, or use backtick-quoting for identifiers.

app/src/databricks_labs_dqx_app/backend/services/rules_catalog_service.py
```diff
@@ -0,0 +1,261 @@
+from __future__ import annotations
+
+import json
+import logging
+from datetime import datetime
```
mwojtyczka: **HIGH - SQL Escaping:** This file uses backslash escaping for SQL string literals. This is incorrect for Databricks SQL which follows the ANSI standard of doubled single quotes. The migration code correctly uses `chr(39)*2` but the services use backslash escaping - inconsistent.

Also, `table_fqn` from route path parameters is used directly in SQL without identifier validation.

app/src/databricks_labs_dqx_app/backend/services/app_settings_service.py
```diff
@@ -0,0 +1,117 @@
+import json
+import logging
+
+from databricks.labs.blueprint.installation import Installation
+from databricks.labs.dqx.config import WorkspaceConfig
```
mwojtyczka, line 5: **HIGH - SQL Escaping:** Same issue as `rules_catalog_service.py` - uses backslash escaping instead of ANSI doubled single quotes for Databricks SQL string literals.

app/src/databricks_labs_dqx_app/backend/cache.py
```diff
@@ -0,0 +1,191 @@
+"""Async in-memory cache with TTL support.
+
+Provides a singleton ``app_cache`` instance shared across the application,
+with a ``cached`` decorator for transparent cache-aside on async functions
+and methods.  No external dependencies — uses only stdlib asyncio and time.
```
mwojtyczka, line 5: **MEDIUM - Bug:** The cache checks `if result is not None` before storing. Falsy values like empty list, empty dict, 0, False are never cached and always re-fetched. `DiscoveryService.list_catalogs_async` can legitimately return an empty list.

**Fix:** Use a sentinel `_MISSING = object()` to distinguish not-in-cache from cached-falsy-value.

Also: `_locks` dict grows unboundedly - locks created per key are never cleaned up after cache entry expiry.

app/src/databricks_labs_dqx_app/backend/common/authorization.py
```diff
@@ -0,0 +1,53 @@
+import logging
+from enum import StrEnum
+from typing import Annotated
+
+from fastapi import Depends, Header, HTTPException, status
```
mwojtyczka, line 5: **MEDIUM - Stubbed auth:** `get_user_role` always returns `UserRole.ADMIN`. The `require_role` decorator is used on config routes but not on `approve_rules`/`reject_rules` endpoints in `routes/v1/rules.py` despite their docstrings saying admin only.

**Fix:** Add `require_role(UserRole.ADMIN)` to approve/reject routes. Add unit tests for authorization module.

app/src/databricks_labs_dqx_app/backend/app.py
```diff
@@ -4,21 +4,34 @@
 from fastapi.staticfiles import StaticFiles
 
 from .config import conf
+from .dependencies import get_migration_runner, get_sp_ws
 from .logger import logger
-from .router import api
+from .routes import api_router
 from .utils import add_not_found_handler
```
mwojtyczka, line 20: **MEDIUM - Silent failure:** Migration failure is caught with `logger.warning` and the comment says will retry on first request, but no retry mechanism exists. If migrations fail, the app proceeds with missing tables and will fail on first API call with confusing SQL errors.

**Fix:** Either implement retry-on-first-request or make migration failure fatal.

app/databricks.yml
```diff
@@ -1,20 +1,62 @@
 bundle:
   name: "databricks-labs-dqx-app"
+  
+
+variables:
+  catalog_name:
+    description: "Name of the catalog for DQX app storage"
+    default: "dqx"
+  schema_name:
+    description: "Name of the schema for DQX app storage"
```
mwojtyczka, line 10: **MEDIUM - Two issues:**

1. `dqx_tmp_schema` grants `USE_SCHEMA` and `CREATE_TABLE` to `account users` - any user in the account can create tables in the tmp schema. Can we restrict this?

2. `dqx_service_principal_application_id` defaults to all-zeros. If not overridden, deployment succeeds but runtime fails with confusing errors. Remove the default or make it required.

app/tasks/src/dqx_task_runner/runner.py
```diff
@@ -0,0 +1,315 @@
+"""DQX Task Runner — executed inside a serverless Databricks Job.
+
+This script is deployed as a workspace file via DABs and runs profiler
+or dry-run operations on behalf of the DQX App.  The app creates a
+temporary VIEW (using the user's OBO token) and then submits this job
```
mwojtyczka, line 5: **MEDIUM - Unsanitized SQL:** `view_fqn` from job parameters is used directly in `spark.sql()`. While app-generated (uuid-based), job params are visible in the UI and could be tampered with.

**Fix:** Validate `view_fqn` matches expected pattern before using in SQL.

app/src/databricks_labs_dqx_app/backend/services/ai_rules_service.py
```diff
@@ -0,0 +1,132 @@
...
+from databricks_langchain import ChatDatabricks  # type: ignore[import-untyped]
+from langchain_core.messages import AIMessage, BaseMessage, HumanMessage, SystemMessage
+
+from databricks.labs.dqx.llm.llm_utils import get_required_check_functions_definitions
+
+from databricks_labs_dqx_app.backend.config import conf
+
+logger = logging.getLogger(__name__)
+
+_SYSTEM_TEMPLATE = """\
```
mwojtyczka, line 21: we should reuse the existing instructions inside llm module. This will not be maintainable otherwise.

app/src/databricks_labs_dqx_app/backend/services/ai_rules_service.py
```diff
@@ -0,0 +1,132 @@
...
+
+    @classmethod
+    def _get_few_shot_messages(cls) -> list[BaseMessage]:
+        if cls._few_shot_messages is None:
+            cls._few_shot_messages = cls._build_few_shot_messages()
+        return cls._few_shot_messages
+
+    @classmethod
+    def _build_few_shot_messages(cls) -> list[BaseMessage]:
+        resource = Path(str(files("databricks.labs.dqx.llm.resources") / "training_examples.yml"))
```
mwojtyczka, line 74: This is a separate implementation from our current llm modules. We really need to use existing llm modules. It's ok as a first version, but in the next iteration, let's refactor the existing llm module to use workspace client instead of spark. We will be able to use it then.

https://github.com/databrickslabs/dqx/issues/1095

app/src/databricks_labs_dqx_app/backend/config.py
```diff
@@ -21,13 +22,30 @@ class AppConfig(BaseSettings):
...
     )
     app_name: str = Field(default=app_name)
     api_prefix: str = Field(default="/api")
+    catalog: str = Field(default="dqx")
+    schema_name: str = Field(default="dqx_app", validation_alias="DQX_SCHEMA")
+    tmp_schema_name: str = Field(default="dqx_app_tmp", validation_alias="DQX_TMP_SCHEMA")
+    job_id: str = Field(default="", validation_alias="DQX_JOB_ID")
+    llm_endpoint: str = Field(default="databricks-claude-sonnet-4-5", validation_alias="DQX_LLM_ENDPOINT")
+    profiler_max_sample_limit: int = Field(default=100_000)
+    profiler_default_sample_limit: int = Field(default=50_000)
```
mwojtyczka, line 41: Should we align this with the default profiler setting? 
 The PR sets profiler_default_sample_limit to 50,000, while the actual ProfilerConfig.limit default is 1,000. That's a 50x discrepancy.

app/README.md
mwojtyczka: After refactor in this PR this step is not needed anymore?

mwojtyczka: do we still need to provide source code path?

app/README.md
```diff
@@ -381,30 +419,14 @@ https://<your-workspace-url>/apps/databricks-labs-dqx-app
...
-   - Or choose a custom location (e.g., `/Workspace/Shared/dqx-config`)
-3. **Save Settings**: Click **Save** to persist your settings
-
-**What Gets Created:**
-- **App Settings**: Stored in `/Users/<your-username>/.dqx/app.yml` (contains the install folder path)
-- **Install Folder**: Created automatically if it doesn't exist
-- **Default config.yml**: Automatically created in your install folder with an empty configuration if it doesn't already exist
-  - You can then add run configurations through the Configuration page in the app
-
 ### Complete Deployment Script
```
mwojtyczka: do we need this section, since the steps are already described in the previous section?

mwojtyczka (changes requested):

app/src/databricks_labs_dqx_app/backend/services/view_service.py
```diff
@@ -0,0 +1,230 @@
...
+        """Create a temporary view whose body is an arbitrary SQL query.
+
+        Used for cross-table SQL checks where the query itself returns
+        the violation rows.  Returns the fully qualified view name.
+        """
+        self._ensure_schema()
+
+        view_id = uuid4().hex[:12]
+        view_name = f"{self._catalog}.{self._schema}.tmp_view_{view_id}"
+        sql = f"CREATE OR REPLACE VIEW {view_name} AS {sql_query}"
```
mwojtyczka: **HIGH — arbitrary SQL execution.** `sql_query` is interpolated directly into `CREATE OR REPLACE VIEW ... AS {sql_query}` with no validation. A RULE_AUTHOR who saves an SQL-check rule can smuggle `SELECT 1; DROP TABLE x` or exfiltrate rows from tables the OBO user can read but shouldn't surface. Per AGENTS.md, user/template SQL must pass `is_sql_query_safe()` and raise `UnsafeSqlQueryError` if unsafe. Validate at save time (in `RulesCatalogService.save`) **and** again here as defence-in-depth.

app/src/databricks_labs_dqx_app/backend/services/scheduler_service.py
```diff
@@ -0,0 +1,601 @@
...
+        self._grant_view(view_name)
+        if not self._view_exists(view_name):
+            raise RuntimeError(f"Scheduler: view creation succeeded but view not found: {view_name}")
+        return view_name
+
+    def _create_view_from_sql(self, sql_query: str) -> str:
+        view_id = uuid4().hex[:12]
+        view_name = f"{self._catalog}.{self._tmp_schema}.tmp_view_{view_id}"
+        self._ensure_tmp_schema()
+        sql = f"CREATE OR REPLACE VIEW {view_name} AS {sql_query}"
```
mwojtyczka: **HIGH — same arbitrary-SQL issue as `ViewService.create_view_from_sql`**, but this path runs as the **Service Principal**, not the user — so unvalidated `sql_query` gets SP-level privilege. Must call `is_sql_query_safe()` before embedding, and reject on failure. Ideally also capture the approver's reviewed SQL fingerprint and verify the body hasn't changed between approval and execution.

app/src/databricks_labs_dqx_app/backend/sql_utils.py
```diff
@@ -0,0 +1,46 @@
...
+
+All SQL string escaping MUST use these functions instead of inline
+.replace() calls to ensure consistent, correct Databricks SQL escaping.
+"""
+
+from __future__ import annotations
+
+import re
+
+_FQN_RE = re.compile(r"^[a-zA-Z0-9_`\-]+\.[a-zA-Z0-9_`\-]+\.[a-zA-Z0-9_`\-]+$")
```
mwojtyczka: **HIGH — regex too permissive.** The character class allows backticks (`` ` ``) and hyphens, so a crafted identifier like `` a`.b.`c``SELECT… `` satisfies `validate_fqn` and then gets embedded verbatim into DDL by callers (`view_service.create_view`, `scheduler_service._create_view`, several services). Tighten to something like `^[A-Za-z_]\w*\.[A-Za-z_]\w*\.[A-Za-z_]\w*$` and, where identifiers truly need non-standard characters, quote each part with backticks **after** validation.

app/src/databricks_labs_dqx_app/backend/common/authentication/sql.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def __init__(self, bearer: str | None = None) -> None:
+        self._bearer = bearer
+
+    @property
+    def access_token(self) -> str:
+        if self._bearer:
+            return self._bearer
+
+        env_token = os.environ.get("DATABRICKS_TOKEN")
```
mwojtyczka: **HIGH — privilege-escalation hazard.** Falling back to `os.environ["DATABRICKS_TOKEN"]` means any request missing the OBO header silently uses the SP/app token for SQL. Combined with `get_sql_connector` being wired into routes that don't all enforce `require_role`, an unauthenticated caller can end up issuing SQL with elevated privilege. Either remove the env fallback entirely, or restrict it to a tightly-scoped startup/bootstrapping code path (migrations) and never expose it to HTTP handlers.

app/src/databricks_labs_dqx_app/backend/common/authorization.py
```diff
@@ -0,0 +1,82 @@
...
+def get_permissions_for_role(role: UserRole) -> list[str]:
+    """Return the list of permissions granted to a role."""
+    return PERMISSIONS.get(role, [])
+
+
+def get_user_email(
+    x_forwarded_email: Annotated[str | None, Header()] = None,
+    x_forwarded_user: Annotated[str | None, Header()] = None,
+) -> str:
+    email = x_forwarded_email or x_forwarded_user
```
mwojtyczka: **HIGH — spoofable identity.** `X-Forwarded-User` is not a Databricks Apps-verified header and can be set by any proxy in front of the app. Identity must come from `X-Forwarded-Email` (verified by Apps) or, for defence-in-depth, `obo_ws.current_user.me().user_name`. Drop the `x_forwarded_user` fallback.

app/src/databricks_labs_dqx_app/backend/routes/v1/dryrun.py
```diff
@@ -0,0 +1,342 @@
...
+
+
+@router.get("/runs/{run_id}/status", response_model=RunStatusOut, operation_id="getDryRunStatus")
+def get_dry_run_status(
+    run_id: str,
+    job_run_id: int,
+    job_svc: Annotated[JobService, Depends(get_job_service)],
+    view_svc: Annotated[ViewService, Depends(get_view_service)],
+    app_conf: Annotated[AppConfig, Depends(get_conf)],
+    view_fqn: str | None = None,
```
mwojtyczka: **HIGH — unauthorized cross-user action.** `view_fqn` is taken as a query param and fed directly into `view_svc.drop_view(view_fqn)` (line 238). Any authenticated user who knows or guesses another user's temp-view name can drop it while that user's job is still running. Same issue in `profiler.py`. Fix: look up `view_fqn` from `dq_validation_runs` using `(run_id, requesting_user)` rather than trusting the client.

app/src/databricks_labs_dqx_app/backend/routes/v1/dryrun.py
```diff
@@ -0,0 +1,342 @@
...
+        if view_fqn and is_terminal:
+            try:
+                view_svc.drop_view(view_fqn)
+                view_cleaned_up = True
+                logger.info("Cleaned up temporary view: %s", view_fqn)
+            except Exception as cleanup_err:
+                logger.warning("Failed to clean up view %s: %s", view_fqn, cleanup_err)
+
+        if is_terminal and status.state != "TERMINATED":
+            update_run_status(
```
mwojtyczka, line 345: **HIGH — unauthorized run-state corruption.** `job_run_id` is a client-supplied int that is never verified against `run_id`; the subsequent `update_run_status(..., status="FAILED", ...)` call can therefore flip **someone else's** run to FAILED if a user passes that user's `job_run_id`. Persist `(run_id → job_run_id, requesting_user)` in `dq_validation_runs` and verify the tuple before any status update.

app/src/databricks_labs_dqx_app/backend/app.py
```diff
@@ -1,34 +1,178 @@
...
+
+            if uploaded:
+                try:
+                    await _update_job_wheels(sp_ws, conf.job_id, uploaded)
+                except Exception as e:
+                    logger.warning("Could not update job environment with uploaded wheels: %s", e, exc_info=True)
+
+    if sp_ws is not None and conf.job_id:
+        try:
+            _scheduler = SchedulerService(
```
mwojtyczka, line 228: **HIGH — scheduler double-fires under `--workers 2`.** `databricks.yml` runs uvicorn with 2 workers, and each worker runs this `lifespan` and starts its own `SchedulerService`. The `_upsert_tracker` MERGE has no `WHERE next_run_at = @expected` guard, so both workers can `jobs.run_now` for the same schedule in the same tick. Options: (a) gate scheduler startup to worker index 0 via an env var / process lock, (b) add a DB-level lease (`WHERE next_run_at < now() AND (last_run_id IS NULL OR last_run_id = :prev)`) to the MERGE, or (c) run the scheduler as a separate long-running job and let the app be stateless.

app/src/databricks_labs_dqx_app/backend/app.py
```diff
@@ -1,34 +1,178 @@
...
+            try:
+                uploaded = await _upload_wheels_to_volume(sp_ws, conf.wheels_volume)
+            except Exception as e:
+                logger.warning(
+                    "Could not upload wheels to volume — job environment will not be updated: %s", e, exc_info=True
+                )
+
+            if uploaded:
+                try:
+                    await _update_job_wheels(sp_ws, conf.job_id, uploaded)
```
mwojtyczka: **Perf/noise — gate wheel re-upload on content hash.** Every cold start re-scans the cwd, uploads wheels to the volume, and calls `jobs.update` on the environment. That's slow, writes noisy audit entries, and can thrash the job config. Compute a hash of the wheel set and only upload/update when it changes.

app/databricks.yml
```diff
@@ -24,10 +76,144 @@ resources:
...
 
 targets:
   dev:
-    mode: development
-    default: true
+    workspace:
+      profile: dqx-app-demo
+    variables:
+      catalog_name: tsh_dqx_app_demo_catalog
+      dqx_service_principal_application_id: "4174f306-d722-4cea-8745-85416017ee92"
```
mwojtyczka: **LOW — committed identifier.** `dqx_service_principal_application_id` looks like a real UUID for the `dqx-app-demo` workspace. Even if it's only a dev workspace, committing SP app IDs (and the profile name) into the repo is against AGENTS.md's "sensitive content" guidance. Move to a `.env` file or secret-scope reference and ship a placeholder here.

app/databricks.yml
```diff
@@ -1,20 +1,72 @@
...
+    default: "00000000-0000-0000-0000-000000000000"
+  app_config:
+    type: complex
+    description: "Configuration for dqx app"
+    default:
+      command:
+        - "uvicorn"
+        - "databricks_labs_dqx_app.backend.app:app"
+        - "--workers"
+        - "2"
```
mwojtyczka: **Correctness** — `--workers 2` is fine for the HTTP path but interacts badly with the in-process `SchedulerService` and the in-memory `CacheFactory` (each worker has its own cache/scheduler). Consider noting this constraint explicitly and either (a) running a single worker with async concurrency, or (b) externalizing cache (Redis/Delta) and scheduler (separate process).

app/src/databricks_labs_dqx_app/backend/services/scheduler_service.py
mwojtyczka: **Convention — don't hand-roll `.replace("'", "''")`.** Several places in this file bypass `escape_sql_string` from `sql_utils.py`. Use the single source of truth; it's what AGENTS.md calls out. Same pattern appears in `app_settings_service.py`.

app/src/databricks_labs_dqx_app/backend/services/view_service.py
```diff
@@ -0,0 +1,230 @@
...
+                state = self._wait_for_completion(statement_id, timeout_seconds=60)
+
+            exists = state == StatementState.SUCCEEDED
+            logger.info("View existence check for %s: %s", view_fqn, "exists" if exists else "not found")
+            return exists
+        except Exception as e:
+            logger.warning("View existence check failed for %s: %s", view_fqn, e)
+            return False
+
+    def create_view_from_sql(self, sql_query: str) -> str:
```
mwojtyczka, line 119: **Test coverage gap.** This method and `create_view` have no unit tests, and the integration test file was net-shrunk by 207 lines. Given these are the primary SQL-interpolation surfaces in the PR, please add at minimum: (1) a unit test asserting `is_sql_query_safe()` is called, (2) negative tests covering rejected payloads (multiple statements, DDL, comment-based escapes), (3) an integration test covering the OBO happy path.

mwojtyczka (changes requested):
Thanks for the large refactor — the move to an out-of-process task runner + SQL Warehouse is the right direction. Leaving inline comments on the blocking issues. The biggest themes:

1. **Unvalidated SQL interpolated into `CREATE VIEW AS …`** — any RULE_AUTHOR can execute arbitrary SQL under OBO and (via the scheduler) under SP.
2. **Auth-bypass fallbacks** — `DATABRICKS_TOKEN` env fallback in the SQL auth resolver, and trusting `X-Forwarded-User` as identity.
3. **Unauthorized query params** (`view_fqn`, `job_run_id`) on dry-run/profiler polling endpoints let one user drop another user's view and flip another user's run to FAILED.
4. **Scheduler doubles-fires** under `--workers 2` — both uvicorn workers run `lifespan` and start a `SchedulerService`.
5. **Tests regressed** — `tests/integration/test_app_backend.py` net-drops 207 lines; the new routes/services (rules catalog, scheduler, roles, view service, job service, migrations, task runner) have little to no coverage.

Additional non-blocking items (convention/perf) noted inline where useful; happy to discuss any of these.

app/tasks/src/dqx_task_runner/runner.py
```diff
@@ -0,0 +1,707 @@
...
+    a Spark-local temp view is created here.
+    """
+    checks = config.get("checks", [])
+    source_table_fqn = config.get("source_table_fqn", "")
+
+    if config.get("is_sql_check"):
+        sql_query = config.get("sql_query")
+        if sql_query:
+            tmp_name = f"_dqx_scheduled_sql_{run_id}"
+            spark.sql(f"CREATE OR REPLACE TEMP VIEW {tmp_name} AS {sql_query}")
```
mwojtyczka, line 525: **Defense-in-depth — validate `sql_query` here too.** `sql_query` arrives through `config_json` job parameters. The app-side paths (`ViewService.create_view_from_sql`, `SchedulerService._create_view_from_sql`) now gate it with `is_sql_query_safe()`, but anyone with direct `CAN_MANAGE_RUN` on the task-runner job could bypass those gates and pass arbitrary DDL/DML here. Mirror the check:

```python
from databricks.labs.dqx.utils import is_sql_query_safe
from databricks.labs.dqx.errors import UnsafeSqlQueryError

if sql_query and not is_sql_query_safe(sql_query):
    raise UnsafeSqlQueryError("Refusing to execute unsafe sql_query in task runner")
```

Cheap belt-and-braces; matches the pattern already set elsewhere.

app/tasks/src/dqx_task_runner/runner.py
```diff
@@ -0,0 +1,707 @@
...
+                str(exc),
+            )
+        except Exception as write_exc:
+            logger.error("Failed to write error result: %s", write_exc, exc_info=True)
+        sys.exit(1)
+    finally:
+        if args.task_type != "scheduled":
+            try:
+                logger.info("Dropping temporary view: %s", args.view_fqn)
+                spark.sql(f"DROP VIEW IF EXISTS {args.view_fqn}")  # noqa: S608
```
mwojtyczka: **Quote `args.view_fqn` before DDL.** `args.view_fqn` comes from job parameters and is interpolated raw into `DROP VIEW IF EXISTS {args.view_fqn}`. Same access model as above (`CAN_MANAGE_RUN` required), but we've already built `validate_fqn` + `quote_fqn` for exactly this — use them here and drop the `# noqa: S608`:

```python
from databricks_labs_dqx_app.backend.sql_utils import validate_fqn, quote_fqn
validate_fqn(args.view_fqn)
spark.sql(f"DROP VIEW IF EXISTS {quote_fqn(args.view_fqn)}")
```

mwojtyczka (commented):
Another pass — thanks for the hardening commit (`a92d92f`), I've resolved the threads it fixed. Two small defense-in-depth items left in the task runner; otherwise the security posture looks solid. Test coverage added for the SQL safety gates is a big improvement.

app/src/databricks_labs_dqx_app/backend/routes/v1/profiler.py
```diff
@@ -0,0 +1,340 @@
...
+            )
+            for row in rows
+        ]
+    except Exception as e:
+        logger.error("Failed to list profile runs: %s", e, exc_info=True)
+        raise HTTPException(status_code=500, detail=f"Failed to list profile runs: {e}")
+
+
+@router.post("/run", response_model=ProfileRunOut, operation_id="submitProfileRun")
+def submit_profile_run(
```
mwojtyczka, line 84: **B1 (Blocker) — RBAC missing.** No `require_role(...)` dependency on this router. Any authenticated user — including VIEWER — can submit a profiler job that runs as the SP. Apply a role guard (RULE_AUTHOR+) at the router or per-handler, matching `common/authorization.py`.

app/src/databricks_labs_dqx_app/backend/routes/v1/generate.py
```diff
@@ -0,0 +1,37 @@
...
+from databricks_labs_dqx_app.backend.dependencies import get_ai_rules_service
+from databricks_labs_dqx_app.backend.logger import logger
+from databricks_labs_dqx_app.backend.models import GenerateChecksIn, GenerateChecksOut
+from databricks_labs_dqx_app.backend.services.ai_rules_service import AiRulesService
+
+router = APIRouter()
+
+
+@router.post("/generate-checks", response_model=GenerateChecksOut, operation_id="ai_assisted_checks_generation")
+def ai_generate_checks(
```
mwojtyczka, line 24: **B1 + B2 (Blocker) — RBAC missing AND no LLM guardrails.**

**B1:** No `require_role(...)` — any VIEWER can trigger LLM calls billed against the SP. Should require RULE_AUTHOR+.

**B2:** No `max_tokens`, no input length cap, no per-user rate limit, no output validation. OWASP LLM04 (DoS/cost exhaustion) + LLM06 (info disclosure). Additionally, `checks` from the LLM are returned to the UI without calling `is_sql_query_safe()` on any `sql_query` field — the preview leaks unsafe SQL even though save-time validation catches it later.

app/src/databricks_labs_dqx_app/backend/routes/v1/quarantine.py
```diff
@@ -0,0 +1,215 @@
...
+        run_id=row.get("run_id", ""),
+        source_table_fqn=row.get("source_table_fqn", ""),
+        requesting_user=row.get("requesting_user"),
+        row_data=row_data,
+        errors=errors,
+        created_at=row.get("created_at"),
+    )
+
+
+@router.get("/runs/{run_id}", operation_id="listQuarantineRecords")
```
mwojtyczka: **B1 (Blocker) — RBAC missing.** Quarantine records contain sampled source-table data (potentially PII). VIEWERs can read or export them as-is. Require RULE_APPROVER+ (or at minimum gate per the permission matrix in `common/authorization.py`).

app/src/databricks_labs_dqx_app/backend/services/ai_rules_service.py
```diff
@@ -0,0 +1,146 @@
...
+
+        Args:
+            user_input: Natural language description of data quality requirements.
+            table_fqn: Optional fully-qualified table name for schema context.
+
+        Returns:
+            List of DQX rule dicts.
+        """
+        system = SystemMessage(content=_SYSTEM_TEMPLATE.format(available_functions=self._get_available_functions()))
+        schema_info = self._get_schema_info(table_fqn) if table_fqn else ""
```
mwojtyczka, line 140: **B2 (Blocker) — no guardrails on LLM call.** `user_input` is embedded in the few-shot chain with no length cap; the `ChatDatabricks` call sets no `max_tokens`. AGENTS.md "LLM denial of service and information disclosure" is the relevant standard. Add:

- `max_tokens` on the chat call
- `len(user_input) <= ~2000` guard at the service boundary
- per-user token/minute rate limiter
- `is_sql_query_safe()` on any `sql_query` in the LLM output before returning

Schema info at `_get_schema_info` already goes to the LLM — ensure returned `filter` / `sql_query` strings are sanitised before constructing `DQRule` instances on save too.

app/src/databricks_labs_dqx_app/backend/services/scheduler_service.py
```diff
@@ -0,0 +1,637 @@
...
+        from databricks_labs_dqx_app.backend.sql_utils import escape_sql_string
+
+        escaped_name = escape_sql_string(name)
+        escaped_status = escape_sql_string(status)
+        last_str = f"'{last_run_at.isoformat()}'" if last_run_at else "NULL"
+        next_str = f"'{next_run_at.isoformat()}'" if next_run_at else "NULL"
+        run_id_str = f"'{escape_sql_string(last_run_id)}'" if last_run_id else "NULL"
+
+        sql = (
+            f"MERGE INTO {self._table} AS target "
```
mwojtyczka, line 282: **B3 (Blocker) — `schedule_name` unvalidated.** `escape_sql_string` neutralises single quotes only. `schedule_name` and `status` come from user-saved configs and flow into this MERGE. Backticks, semicolons, comment tokens (`--`, `/* */`) pass the escape intact. Add a strict regex at the save boundary (`ScheduleConfigService.save`): `^[a-zA-Z0-9_-]{1,64}$`, and whitelist `status` against `VALID_STATUSES`.

app/src/databricks_labs_dqx_app/backend/services/scheduler_service.py
```diff
@@ -0,0 +1,637 @@
...
+
+        for i, table_fqn in enumerate(table_fqns):
+            try:
+                entry = self._get_approved_rule(table_fqn)
+                if entry is None:
+                    continue
+
+                run_id = f"{run_id_prefix}_{i}"
+                is_sql_check = table_fqn.startswith(_SQL_CHECK_PREFIX)
+
```
mwojtyczka, line 323: **H1-adjacent — `run_id` is app-generated but interpolated raw.** `run_id = f"{run_id_prefix}_{i}"` is passed straight into `CREATE TEMP VIEW` in the task runner (line 463 there). `run_id_prefix` is a uuid hex today so it's safe, but this relies on callers — a future refactor that exposes a user-settable `run_id` would break it. Regex-validate at the runner boundary (`[a-fA-F0-9_]+`) to match how `view_fqn` is validated at runner.py:707.

app/src/databricks_labs_dqx_app/backend/migrations/__init__.py
```diff
@@ -0,0 +1,529 @@
...
+
+    def run_all(self) -> int:
+        """Ensure the schema exists and apply all pending migrations.
+
+        Returns:
+            The number of migrations applied in this invocation.
+        """
+        self._ensure_schema()
+        self._ensure_meta_table()
+        applied_versions = self._applied_versions()
```
mwojtyczka, line 390: **H3 (High) — not concurrency-safe.** Check-then-apply spans two separate `execute_statement` calls. Two uvicorn workers booting together both see version N missing, both run the DDL, both INSERT the marker → duplicate rows in `dq_migrations` and, for non-idempotent migrations (v13/29/30 use `ALTER TABLE ADD COLUMNS` without `IF NOT EXISTS`), the second worker crashes. The scheduler file-lock doesn't cover migrations.

Fix: either (a) gate migrations behind the same `_try_acquire_scheduler_lease()` so only the leased worker runs them, or (b) add a PRIMARY KEY on `version`, `INSERT ... WHERE NOT EXISTS`, and make every migration idempotent.

app/src/databricks_labs_dqx_app/backend/run_status_manager.py
```diff
@@ -0,0 +1,148 @@
...
+    em = escape_sql_string(error_message or "")
+
+    set_clause = f"status = '{es}', error_message = '{em}', updated_at = CAST(current_timestamp() AS STRING)"
+    if canceled_by:
+        ec = escape_sql_string(canceled_by)
+        set_clause += f", canceled_by = '{ec}'"
+
+    sql = f"UPDATE {table} SET {set_clause} WHERE run_id = '{er}' AND status = 'RUNNING'"
+    try:
+        resp = job_svc._ws.statement_execution.execute_statement(
```
mwojtyczka: **H2 (Low) — private-attribute access breaks DI.** Helpers here read `job_svc._ws`, `job_svc._warehouse_id`, `job_svc._catalog`, `job_svc._schema` directly. This makes the module untestable without monkey-patching and couples `run_status_manager` to `JobService` internals.

The same `execute_statement` / `_wait_for_completion` pattern is duplicated across `scheduler_service.py`, `view_service.py`, `migrations/__init__.py`, `role_service.py`, and here. Extract a single `SqlExecutor` class, inject it via `Depends(...)`, and let this module accept it as a parameter. That also removes the `# noqa: S608` trail in `quarantine.py`, `metrics.py`, `job_service.py`.

app/src/databricks_labs_dqx_app/backend/routes/v1/dryrun.py
```diff
@@ -0,0 +1,372 @@
...
+@router.post("/runs/{run_id}/cancel", operation_id="cancelDryRun")
+def cancel_dry_run(
+    run_id: str,
+    obo_ws: Annotated[WorkspaceClient, Depends(get_obo_ws)],
+    job_svc: Annotated[JobService, Depends(get_job_service)],
+    app_conf: Annotated[AppConfig, Depends(get_conf)],
+) -> dict[str, str]:
+    """Cancel a running dry-run job."""
+    try:
+        canceling_user = obo_ws.current_user.me().user_name or "unknown"
```
mwojtyczka, line 413: **H4 (High) — role check missing.** The owner check is correct for user-level isolation but a VIEWER can still probe run IDs; the 403 error leaks `meta.requesting_user` (an email) on cross-user attempts. 

Also: ADMIN cannot cancel a stuck run owned by someone else — intentional? Document or change. Need to investigate. If you are not Admin on the workspace, you cannot cancel the run. Investigate, may not be a problem. Simple fix: timeout the job, but check if the clean up works correctly.

Same pattern in `profiler.py` cancel. Both need `require_role(...)` applied at the router.

app/src/databricks_labs_dqx_app/backend/config.py
```diff
@@ -21,13 +22,36 @@ class AppConfig(BaseSettings):
         env_prefix=f"{app_name.upper()}_",
         extra="ignore",
         env_nested_delimiter="__",
+        populate_by_name=True,
     )
     app_name: str = Field(default=app_name)
     api_prefix: str = Field(default="/api")
+    catalog: str = Field(default="dqx")
```
mwojtyczka, line 29: **H5 (High) — deployment-controlled env values flow into raw SQL.** `catalog`, `schema_name`, `tmp_schema_name` are interpolated unchecked into dozens of f-string SQL calls. Not runtime-user-controlled, but a tainted DAB env var would make the app an injection sink. Add a Pydantic `field_validator` running `validate_fqn_part` (same regex as `sql_utils._FQN_PART_RE`) on each of the three.

Nit: `schema_name` shadows a Pydantic reserved-namespace; add `model_config = ConfigDict(protected_namespaces=())` to silence cleanly.

mwojtyczka (commented):
## Review — First App MVP

Comprehensive review against the project's `AGENTS.md`/`CLAUDE.md` standards. Inline comments are attached to specific file:line locations below. Summary here; per-finding details in the inline threads.

### Blockers (must fix before merge)

- **B1.** RBAC missing on `profiler.py`, `dryrun.py`, `generate.py`, `quarantine.py`, `metrics.py`, `discovery.py`. VIEWERs can trigger serverless jobs as the SP, hit the LLM endpoint (cost/DoS), or exfiltrate quarantine rows. The permission matrix exists in `common/authorization.py` but isn't enforced uniformly.
- **B2.** No token-budget / rate limit / length cap on the LLM generate endpoint (OWASP LLM04). LLM-generated `sql_query` is relayed to the UI before `is_sql_query_safe()` is called on it (save-time validation exists, but the preview leaks unsafe queries).
- **B3.** `schedule_name` is unvalidated at save; flows into `MERGE` DDL. `escape_sql_string` handles single quotes only — backticks/semicolons/comment tokens pass through. Add a strict regex at `ScheduleConfigService.save`.
- **B4.** 6 of 9 new services have **zero** unit tests: `SchedulerService` (637 LOC), `RoleService` (authorization-critical), `AiRulesService`, `CommentsService`, `ScheduleConfigService`, `AppSettingsService`. PR body says "manually tested". AGENTS.md requires coverage.

### High-impact

- **H1.** `runner.py:463` — `CREATE OR REPLACE TEMP VIEW _dqx_scheduled_sql_{run_id}` interpolates `run_id` without validation. `view_fqn` on line 707 is validated; apply the same guard to `run_id`.
- **H2.** DRY violation + DI break — `run_status_manager.py`, `quarantine.py`, `metrics.py` reach into `job_svc._ws/_warehouse_id/_catalog/_schema` as private attributes. The `execute_statement`/`_wait_for_completion` pattern is re-implemented 4× across `scheduler_service.py`, `view_service.py`, `migrations/__init__.py`, `role_service.py`. Extract a shared `SqlExecutor` — eliminates the `# noqa: S608` trail and unlocks unit testing.
- **H3.** Migrations not concurrency-safe. `run_all()` does check-then-apply in two separate statements; two workers booting together both INSERT the marker. Non-idempotent migrations (v13/29/30 use `ALTER TABLE ... ADD COLUMNS` without `IF NOT EXISTS`) crash the second worker.
- **H4.** Cancel endpoints owner-check but don't role-check — VIEWER can probe run IDs and see owner emails via 403 errors.
- **H5.** `AppConfig.catalog/schema_name/tmp_schema_name` from env never validated — they flow into dozens of raw SQL f-strings.
- **H6.** `/tmp/.dqx_scheduler.lock` approach: FD stashed in `globals()` to keep it alive is a workaround for a design smell; OOM'd leased worker stalls peers in the same container.

### Medium

- **M1.** `# noqa: S608` silencing bandit across many files — fix via `SqlExecutor` (H2) instead of silencing.
- **M2.** `cache.py:149,157` — unneeded `# type: ignore[return-value]`; constrain the generic.
- **M3.** Bare `except Exception` everywhere; narrow to `DatabricksError`/`RuntimeError` and re-raise metadata errors (`_get_run_fields` silently returns `None` on auth failures).
- **M4.** `SchedulerService._tmp_schema_ensured` declared at class scope, mutated on instance — move to `__init__`.
- **M5.** Log-injection (CWE-117): raw `table_fqn/role/group_name/schedule_name` in f-string logs.
- **M6.** LLM output `filter` strings must be sanitised before reaching `DQRule` construction (OWASP LLM06).
- **M7.** `lifespan` swallows migration failures — partially-migrated schema serves traffic; add a `/health` that reports drift.
- **M8.** Integration tests cover 2 of ~12 new areas.

### Nits

- `runner.py` sleep-loop retries → `tenacity` for back-off consistency.
- `app.py:42` `_find_wheels()` — document subdirectory assumption in a docstring.
- `models.py` 399 LOC — split by domain (rules/runs/schedules/roles).
- `AppConfig.schema_name` shadows a Pydantic reserved; set `model_config = ConfigDict(protected_namespaces=())`.
- `generate.py` eagerly imports `DQEngine` at module top — slows test startup.

### Fixed in working tree (informational)

1. `job_run_id`/`view_fqn` UI→backend mismatch — backend endpoints never declared these as `Query(...)` params; the UI passed them as bogus keys on `AxiosRequestConfig`. Fixed by stripping the keys at `profiler.tsx:291,342`, `rules.generate.tsx:268`, `runs.tsx:1182`. Backend uses `get_run_metadata(..., run_id)` from Delta (authoritative).
2. `scheduler_service.py` pyright `sql_query` possibly-unbound — fixed by pre-binding `sql_query: str | None = None`.
3. `make app-check` target added to run `uv run apx dev check` (TS + pyright) locally and in CI.

### Overall assessment

The security-hardening commit closed the obvious SQL-injection and identity-spoofing gaps, but three things block merge:

1. **RBAC is half-applied** — the routers with the biggest blast radius (LLM spend, job submission, quarantine export) have no guards.
2. **Test coverage below project standard** — 6 services without unit tests for a 36k-line rebuild; a regression in `SchedulerService._compute_next_run` or `RoleService.resolve_role` would be invisible.
3. **SQL-execution architecture debt** — duplicated patterns and private-attribute access across four services; extract `SqlExecutor` now.

Recommend: **Changes Requested.** Address B1–B4 and H1–H3 before approval; H4–H6 and M1–M4 in this or an immediate follow-up PR.

app/src/databricks_labs_dqx_app/backend/routes/v1/generate.py
mwojtyczka: B2 to be fixed in the core

app/src/databricks_labs_dqx_app/backend/services/ai_rules_service.py
mwojtyczka: to be fixed in the core

mwojtyczka: to be fixed in the core

app/src/databricks_labs_dqx_app/backend/services/scheduler_service.py
mwojtyczka: this is fine

app/src/databricks_labs_dqx_app/backend/migrations/__init__.py
mwojtyczka: fixed by using 1 worker, permanent fix to be done later

app/src/databricks_labs_dqx_app/backend/routes/v1/dryrun.py
mwojtyczka: We need to add new role: RUNNER

app/src/databricks_labs_dqx_app/backend/config.py
mwojtyczka: DAB is controlled by the admin.

app/src/databricks_labs_dqx_app/backend/routes/v1/dryrun.py
mwojtyczka: as discussed to be implemented in a separate PR

app/app.yml
```diff
@@ -1,2 +1,18 @@
 command:
   ["uvicorn", "databricks_labs_dqx_app.backend.app:app", "--workers", "2"]
+
+env:
+  - name: DATABRICKS_WAREHOUSE_ID
+    valueFrom: dqx-sql-warehouse
+  - name: DQX_CATALOG
+    value: tsh_dqx_app_demo_catalog
```
mwojtyczka: **Blocker — hardcoded developer catalog committed.**

`DQX_CATALOG: tsh_dqx_app_demo_catalog` and `DQX_WHEELS_VOLUME: /Volumes/tsh_dqx_app_demo_catalog/...` reference what looks like a personal demo catalog. `databricks.yml` already parameterises `${var.catalog_name}` (default `dqx`) via the `app_config` complex variable, so the hardcoded values committed here override the parameterisation for anyone running `databricks bundle deploy` from this repo without `--var catalog_name=...`.

Either:
- regenerate this file from `databricks.yml` at deploy time and add `app/app.yml` to `app/.gitignore`, or
- replace the hardcoded values with `dqx` / generic placeholders so the bundle defaults apply.

app/src/databricks_labs_dqx_app/backend/migrations/__init__.py
```diff
@@ -0,0 +1,472 @@
...
+
+    def _apply(self, migration: Migration) -> None:
+        sql = migration.sql_template.format(catalog=self._catalog, schema=self._schema)
+        self._sql.execute(sql)
+
+        record_sql = (
+            f"INSERT INTO {self._meta_table} (version, description, applied_at) "
+            f"VALUES ({migration.version}, '{migration.description.replace(chr(39), chr(39)*2)}', current_timestamp())"
+        )
+        self._sql.execute(record_sql)
```
mwojtyczka, line 495: **Blocker — migrations are not idempotent on retry.**

`_apply` runs the DDL and the meta-table `INSERT` as two separate statements. If the DDL succeeds but the `INSERT` fails (warehouse hiccup / disconnect), the next startup replays the DDL — and several entries are not re-runnable:

- v5–v8, v10, v12, v16, v18, v21, v24, v26 — `ALTER TABLE … CLUSTER BY (…)` (no `IF NOT EXISTS`)
- v13, v14, v19, v27, v28, v29, v30, v31 — `ALTER TABLE … ADD COLUMNS (…)` without `IF NOT EXISTS`

The first replay throws `COLUMN_ALREADY_EXISTS` / `CLUSTER_BY_ALREADY_DEFINED` and the migration is permanently stuck.

Either wrap DDL + meta-INSERT inside `BEGIN…COMMIT` (Delta supports it on Databricks SQL), or preferably rewrite each entry to be idempotent (e.g., `ADD COLUMN IF NOT EXISTS …`).

Also: line 465 escapes the description with `description.replace(chr(39), chr(39)*2)` instead of calling `escape_sql_string()` from `sql_utils.py` — please use the shared helper for consistency.

app/tasks/src/dqx_task_runner/runner.py
```diff
@@ -0,0 +1,783 @@
...
+        sql_query = config.get("sql_query")
+        if sql_query:
+            from databricks.labs.dqx.utils import is_sql_query_safe
+            from databricks.labs.dqx.errors import UnsafeSqlQueryError
+
+            if not is_sql_query_safe(sql_query):
+                raise UnsafeSqlQueryError("Refusing to execute unsafe sql_query in task runner")
+
+            tmp_name = f"_dqx_scheduled_sql_{run_id}"
+            spark.sql(f"CREATE OR REPLACE TEMP VIEW {tmp_name} AS {sql_query}")
```
mwojtyczka, line 525: **Identifier injection — `run_id` interpolated into a view name without validation.**

```python
tmp_name = f"_dqx_scheduled_sql_{run_id}"
spark.sql(f"CREATE OR REPLACE TEMP VIEW {tmp_name} AS {sql_query}")
```

`run_id` flows in from the `--run_id` job parameter. `is_sql_query_safe` covers `sql_query`, but nothing validates `run_id`. A backtick or whitespace in `run_id` would let it escape the identifier and inject DDL/DML.

Add a defensive regex check (`^[a-zA-Z0-9_-]{1,64}$`) before building the view name, mirroring `sql_utils.validate_schedule_name`. (The same FQN validators are also duplicated in this file at lines 29–49 vs. `backend/sql_utils.py` — separate wheel boundary, but consider publishing them via `databricks.labs.dqx.utils` and importing from there to avoid drift.)

app/databricks.yml
```diff
@@ -24,10 +80,144 @@ resources:
...
+            - ALL_PRIVILEGES
+            - MANAGE
+        - principal: ${var.dqx_service_principal_application_id}
+          privileges:
+            - ALL_PRIVILEGES
+            - MANAGE
+        - principal: "account users"
+          privileges:
+            - USE_SCHEMA
+            - CREATE_TABLE
```
mwojtyczka, line 165: Granting `CREATE_TABLE` to `account users` on the temp schema is intentional (the app needs to create per-user temp views) but worth a YAML comment explaining that

Minor follow-up at line 78: the `user_api_scopes` block lists a partial scope set even though the comment above says `all-apis` is the actual requirement. Either delete the partial list or move it behind a clearly-marked "future" comment so it doesn't read as authoritative.

app/src/databricks_labs_dqx_app/backend/common/authorization.py
```diff
@@ -0,0 +1,100 @@
...
+    """
+    if x_forwarded_email:
+        return x_forwarded_email
+
+    if x_forwarded_access_token:
+        try:
+            from databricks.sdk import WorkspaceClient
+
+            ws = WorkspaceClient(token=x_forwarded_access_token, auth_type="pat")
+            me = ws.current_user.me()
```
mwojtyczka, line 84: `get_user_email` rebuilds a `WorkspaceClient` and calls `current_user.me()` on every request whenever `X-Forwarded-Email` is missing (i.e. local dev). This dependency is injected on every endpoint that requires auth, so the OBO `me()` call dominates request latency in dev.

There's already an `app_cache` with a `cached("auth:obo:{token_hash}", ttl=2700)` example shown in `cache.py:127` — wrap this lookup with it (keyed on a hash of `x_forwarded_access_token`) to avoid hitting SCIM per request.

mwojtyczka (changes requested):

app/src/databricks_labs_dqx_app/backend/sql_executor.py
```diff
@@ -0,0 +1,200 @@
...
+
+from databricks.sdk import WorkspaceClient
+from databricks.sdk.service.sql import Disposition, Format, StatementState
+
+logger = logging.getLogger(__name__)
+
+_TERMINAL_STATES = {StatementState.SUCCEEDED, StatementState.FAILED, StatementState.CANCELED, StatementState.CLOSED}
+
+
+class SqlExecutor:
```
mwojtyczka, line 22: **Hardening — add a parameterised-query path.**

The Databricks Statement Execution API supports `parameters=[StatementParameterListItem(name="foo", value=...)]` for `:named` placeholders. Adding `parameters: list | None = None` to `execute` / `execute_no_schema` / `query` / `query_dicts` and threading it into `execute_statement(...)` would let callers write:

```python
sql.execute(
    "UPDATE :tbl SET status = :status WHERE run_id = :rid AND status = 'RUNNING'",
    parameters=[...],
)
```

instead of the current `f"... '{escape_sql_string(rid)}' ..."` pattern in `run_status_manager.py`, `services/scheduler_service.py`, `services/schedule_config_service.py`, etc.

Benefits: removes every `escape_sql_string()` call and the entire `# noqa: S608` surface, and removes the foot-gun where every new endpoint author has to remember to escape correctly. This is the single biggest hardening lever in the PR — please consider it as a follow-up if not in this one.

app/src/databricks_labs_dqx_app/backend/cache.py
```diff
@@ -0,0 +1,191 @@
...
+                    # Re-check inside lock (another coroutine may have populated it)
+                    hit = await self.get(cache_key)
+                    if hit is not None:
+                        logger.debug("Cache HIT (post-lock) %s", cache_key)
+                        return hit  # type: ignore[return-value]
+
+                    logger.debug("Cache MISS %s", cache_key)
+                    result = await fn(*args, **kwargs)
+
+                    if result is not None:
```
mwojtyczka: `cached` only stores results when `result is not None`. A cached function that legitimately returns `None` (e.g. "no row found", "user has no role mapping yet") will hit the upstream every call — the cache is silently disabled for that case.

Either:
- cache `None` with a sentinel (`_MISS = object()`) and translate on read, or
- document this behaviour explicitly in the decorator's docstring so callers know to wrap nullable results.

app/src/databricks_labs_dqx_app/backend/cache.py
```diff
@@ -0,0 +1,191 @@
...
+        return self._locks[key]
+
+    # -- write modes ----------------------------------------------------
+
+    def set_fire_and_forget(self, key: str, value: Any, ttl: int | None = None) -> None:
+        """Schedule a cache write without blocking the caller."""
+        task = asyncio.create_task(self.set(key, value, ttl=ttl))
+        task.add_done_callback(self._log_task_error)
+
+    async def set_reliable(
```
mwojtyczka, line 89: `set_reliable`'s retry loop is dead code as written — `set()` is a single `dict[k] = v` assignment which cannot raise (no I/O, no contention since asyncio is single-threaded). The `for attempt in range(...)` always succeeds on the first try.

Either delete `set_reliable` (and the `reliable=True` branch in `cached`), or change `set` to something that *can* fail (e.g. write-through to Redis / a Delta table) and document why retries are needed.

app/src/databricks_labs_dqx_app/backend/spa_static.py
```diff
@@ -0,0 +1,28 @@
...
+class SPAStaticFiles(StaticFiles):
+    """Serve static assets with an ``index.html`` fallback for SPA routing."""
+
+    async def get_response(self, path: str, scope: Scope):
+        try:
+            return await super().get_response(path, scope)
+        except HTTPException as exc:
+            if exc.status_code == 404:
+                last_segment = path.rsplit("/", 1)[-1]
+                if "." not in last_segment:
```
mwojtyczka: Edge case — `if "." not in last_segment` falsely treats SPA routes that *contain* a dot as static-asset 404s.

Examples: `/rules/v1.2-preview`, `/runs/2026-04-29.scheduled`, anything with a version-number-ish segment. The user gets a 404 instead of `index.html`, the SPA router never sees the path.

Use a positive allowlist of asset extensions instead:

```python
_ASSET_EXTS = (".js", ".mjs", ".css", ".map", ".png", ".jpg", ".svg", ".ico", ".woff", ".woff2", ".json", ".webmanifest")
...
if not last_segment.endswith(_ASSET_EXTS):
    return await super().get_response("/index.html", scope)
```

app/src/databricks_labs_dqx_app/backend/routes/v1/profiler.py
berrybluecode: resolved:     dependencies=[require_role(*_AUTHORS_AND_ABOVE)],

app/src/databricks_labs_dqx_app/backend/migrations/__init__.py
berrybluecode: 1. ADD COLUMN idempotency (v13, v14, v19, v27–v31):

All ADD COLUMNS (col TYPE) changed to ADD COLUMN IF NOT EXISTS col TYPE
Multi-column adds (v13, v14) split into ;-separated individual ADD COLUMN IF NOT EXISTS statements
2. CLUSTER BY idempotency (v5–v8, v10, v12, v16, v18, v21, v24, v26):

CLUSTER BY has no IF NOT EXISTS syntax, so _apply now catches known-safe errors (COLUMN_ALREADY_EXISTS, FIELDS_ALREADY_EXISTS, already has liquid clustering defined) — if DDL was already applied in a prior interrupted run, it logs a warning and proceeds to record the migration in the meta-table
3. escape_sql_string() consistency:

Replaced the inline description.replace(chr(39), chr(39)*2) with a call to the shared escape_sql_string() helper from sql_utils.py
4. Multi-statement support in _apply:

_apply now splits sql_template on ; and executes each statement individually, enabling the multi-column v13/v14 migrations

app/tasks/src/dqx_task_runner/runner.py
berrybluecode: The fix adds a _validate_run_id() guard using the same ^[a-zA-Z0-9_\-]{1,64}$ pattern the reviewer suggested (mirroring validate_schedule_name in sql_utils.py). It's called at the top of main() before run_id reaches any SQL interpolation — including the _dqx_scheduled_sql_{run_id} temp view name on line 513 and all the Delta write paths that embed run_id in string literals.

app/src/databricks_labs_dqx_app/backend/common/authorization.py
berrybluecode: CREATE_TABLE grant on temp schema — added a comment explaining that account users need it so the app can create per-user temporary views via OBO tokens.

user_api_scopes block — rewrote the comments to make it unambiguous that the partial scope list is just the DABs-managed baseline and does not replace the all-apis scope requirement documented above it.

app/src/databricks_labs_dqx_app/backend/sql_executor.py
berrybluecode: consider for the next PR

app/src/databricks_labs_dqx_app/backend/cache.py
berrybluecode: Added a _MISS sentinel (line 33) to distinguish "not in cache" from a cached None.

get() returns _MISS instead of None (lines 59, 63) when a key is absent or expired.

Cache hit checks use is not _MISS instead of is not None (lines 150, 158), so a cached None is recognized as a valid hit.

Always cache the result (lines 165–168) — removed the if result is not None guard so that None return values are stored, preventing repeated upstream calls for functions like "no row found" or "user has no role mapping yet".

app/src/databricks_labs_dqx_app/backend/spa_static.py
berrybluecode: Replaced the "." not in last_segment heuristic with a positive allowlist of known static-asset extensions. SPA routes containing dots (e.g. /rules/v1.2-preview, /runs/2026-04-29.scheduled) will now correctly fall through to index.html instead of returning a 404.

mwojtyczka (approved):
LGTM

PR #1092, opened 2026-03-20
Allow custom check messages

### LINKED ISSUE #958: [FEATURE]: Enable custom message for all the checks

### Problem statement

Custom messages are currently supported only within sql_expression. However, this capability could be extended to other areas (such as regular expressions or general key-based validations) where the output would benefit from being translated into a more user-friendly field (for example, descriptive text). 
Enabling custom messages in these contexts would allow for more meaningful and tailored feedback for end users when the check results are reported to the end users.
Currently, the only way to implement this is to build a custom check

### Proposed Solution

Having a custom message for all the checks would be the solution. It can be auto-generated if the message is not provided (similar to all default behaviors in checks in the current codebase).
It would also be helpful to provide the facility to have placeholders in the custom message that can be automatically filled in from the record or context

### Additional Context

_No response_

ghanse:
## Changes
Add an optional `message_expr` parameter to `DQRule` that allows users to define custom check failure messages as Spark `Column` objects or SQL expression strings. When `message` is `None` (the default), the default message behavior is preserved. When provided, the custom message replaces the default message for failed rows.

**Skipped violation test helper**
Introduces a `build_skipped_violation` method for testing skipped checks (e.g. due to invalid expressions, column references, or custom messages).

### Linked issues

Resolves #958

### Tests
- [x] manually tested
- [x] added unit tests
- [x] added integration tests


src/databricks/labs/dqx/manager.py
```diff
@@ -167,6 +170,52 @@ def _build_result_struct(self, condition: Column, skipped: bool = False) -> Colu
...
         ).cast(dq_result_item_schema)
 
+    def _build_message_col(self, condition: Column) -> Column:
+        """
+        Builds the message column, using the default message or a custom message callable if specified in the rule
+        definition. Allows use of the following arguments passed at execution time by the DQRuleManager:
+
+        * *rule_name* is the name of the DQX rule
+        * *check_func_name* is the name of the DQX check function
+        * *check_func_args* is a dictionary of DQX check function arguments as key-value pairs
```
mwojtyczka: **MEDIUM - `inspect.signature` called at execution time per row-group:** `_build_message_col` calls `inspect.signature` twice (on `check_func` and `message`) every time a result struct is built. While this runs once per check (not per row), it's still a hot path during DataFrame construction. Consider caching the signatures on the `DQRule` or `DQRuleManager` instance.

Also, `bind_partial` can raise `TypeError` if args don't match — should this be wrapped with error handling that gives a clear message about the check function signature mismatch?

src/databricks/labs/dqx/manager.py
```diff
@@ -167,6 +170,52 @@ def _build_result_struct(self, condition: Column, skipped: bool = False) -> Colu
...
+        * *column_value* is the column value that failed the DQX check
+
+        Args:
+            condition: Default DQX condition message returned by evaluating the DQX check function
```
mwojtyczka, line 188: When `self.check.column` is `None` but `self.check.columns` is set (e.g., a multi-column row rule), `column_value` falls back to `F.lit(None)`. The message function receives no useful value to display. 
Either document this or pass all column values as a struct. Then always pass a struct, even for single-column rules:
```
  if self.check.column:                                                                                                   
      column_value = F.struct(F.col(self.check.column).cast("string").alias(self.check.column))                           
  elif self.check.columns:                                                                                                
      column_value = F.struct(*[F.col(c).cast("string").alias(c) for c in self.check.columns])                            
  else:                                                                                                                   
      column_value = F.lit(None)            
```                              

Passing as struct would be more flexible, but it will add significant complexity for users when defining the message func so maybe just document this.

src/databricks/labs/dqx/manager.py
```diff
@@ -167,6 +170,52 @@ def _build_result_struct(self, condition: Column, skipped: bool = False) -> Colu
...
+        bound = check_func_signature.bind_partial(*args, **kwargs)
+        check_func_args = dict(bound.arguments)
+        column_value = F.col(self.check.column).cast("string") if self.check.column else F.lit(None)
+
+        default_message_func_args = {
+            "rule_name": self.check.name,
+            "check_func_name": self.check.check_func.__name__,
+            "check_func_args": check_func_args,
+            "column_value": column_value,
+        }
```
mwojtyczka: **Good:** The `F.when(condition.isNotNull(), custom_message).otherwise(F.lit(None))` pattern correctly preserves the null/non-null semantics — passing rows get `None` message, failing rows get the custom message. This is the right approach.

src/databricks/labs/dqx/rule.py
```diff
@@ -176,6 +182,7 @@ class DQRule(abc.ABC, DQRuleTypeMixin, SingleColumnMixin, MultipleColumnsMixin):
     check_func_args: list[Any] = field(default_factory=list)
     check_func_kwargs: dict[str, Any] = field(default_factory=dict)
     user_metadata: dict[str, str] | None = None
+    message: Callable[..., Column] | None = None
 
     def __post_init__(self):
         self._validate_rule_type(self.check_func)
```
mwojtyczka, line 188: **MEDIUM - Type annotation:** The field is `message: Callable[..., Column] | None = None` here on `DQRule`, but on `DQForEachColRule` (line 441) it's `message: Callable | None = None` without the return type. These should be consistent — use `Callable[..., Column] | None` in both places.

src/databricks/labs/dqx/rule.py
```diff
@@ -165,6 +165,12 @@ class DQRule(abc.ABC, DQRuleTypeMixin, SingleColumnMixin, MultipleColumnsMixin):
     * *check_func_args* (optional) - Positional arguments for the check function (excluding *column*).
     * *check_func_kwargs* (optional) - Keyword arguments for the check function (excluding *column*).
     * *user_metadata* (optional) - User-defined key-value pairs added to metadata generated by the check.
+    * *custom_message_func* - A user-defined function that returns a message when the check fails. The function should
+        return a Spark Column and can optionally accept the following keyword arguments:
+        - *rule_name* is the name of the DQX rule
```
mwojtyczka: **LOW - Docstring says `custom_message_func` but field is named `message`:** The docstring says `custom_message_func` but the actual field added is `message`. These should match to avoid confusion.

tests/unit/test_custom_messages.py
```diff
@@ -0,0 +1,79 @@
...
+    assert rule.message is _sample_message_callable
+
+
+def test_dq_row_rule_message_defaults_to_none():
+    """DQRowRule without message should default to None."""
+    rule = DQRowRule(check_func=is_not_null, column="id")
+    assert rule.message is None
+
+
+def test_dq_dataset_rule_accepts_message_callable():
```
mwojtyczka: **LOW - Misleading test name:** `test_dq_dataset_rule_accepts_message_callable` creates a `DQRowRule`, not a `DQDatasetRule`. Either rename the test or actually test `DQDatasetRule`.

tests/integration/test_custom_messages.py
```diff
@@ -0,0 +1,326 @@
+"""Integration tests for custom message callable on DQRule."""
+
+from typing import Any
+
+import pyspark.sql.functions as F
```
mwojtyczka, line 3: **Good test coverage.** Tests cover: static message, dynamic message with column_value, default message unchanged, passing rows have no message, DQForEachColRule propagation, **kwargs signature, single-arg signature, check_func_args with column, check_func_args with columns. The null-safety test with `coalesce` is particularly good.

mwojtyczka, line 3: **Missing test:** No test for `DQDatasetRule` with a custom message. Dataset-level rules have different column semantics (no single `column` attribute), so the `column_value` fallback to `F.lit(None)` should be verified.

mwojtyczka (commented):
## Review: Allow custom check messages

Clean feature addition. The inspect-based argument dispatching is flexible and the null-safety handling is correct. A few items below.

src/databricks/labs/dqx/manager.py
```diff
@@ -167,6 +170,52 @@ def _build_result_struct(self, condition: Column, skipped: bool = False) -> Colu
             F.lit(skipped or None).alias("skipped"),
```
mwojtyczka, line 178: We should support custom message for both programmatic and metadata definition.
                                                                                                                      
  **Option 1: Template strings (simplest)**                                                                                   
                                                                                                                          
  Support a message field in YAML as a template string with placeholders:                                                 
                             
```
  - name: email_not_null                                                                                                  
    criticality: error                                                                                                    
    check:                                                                                                                
      function: is_not_null                                                                                               
      arguments:                                                                                                          
        column: email                                                                                                     
    message: "Rule '{rule_name}' failed: {check_func_name} on column '{column}'"               
```                           
                                                                                        
DQX resolves placeholders at runtime using Python's str.format() with the same context args (rule_name, check_func_name, check_func_args, etc.) and wraps the result in F.lit(...).                                                             
                                                                                                                          
This covers static/semi-dynamic messages but can't include column_value (which is a Spark Column expression, not a Python string at template time).
                                                                                                                          
**Option 2: Template string + column_value via Spark concat**                                                               
   
To support column_value in templates, use a special placeholder that gets expanded into a Spark concat expression:  `message: "Rule '{rule_name}' failed for value=${column_value}"   `                                                     
                                                                                                                          
DQX parses the template, splits on ${column_value}, and builds:                                                                                                                                                                                
```
  F.concat(                                                                                                               
      F.lit("Rule 'email_not_null' failed for value="),                                                                   
      F.coalesce(F.col("email").cast("string"), F.lit("null"))                                                            
  )              
```                                                                                                         
                                                                                                                          
This gives dynamic per-row messages from YAML.                                                                          
                             
**Option 3: SQL expression** - I would prefer this one                                                                                                
                             
Allow the message to be a SQL expression string:                                                                     ``message_expr: "concat('Failed: ', coalesce(cast(email as string), 'null'))"   `                                        
                                                                                                                          
  DQX wraps it with F.expr(...). Most flexible, and could unify both approaches

Programmatic                                                                                                          
```
  DQRowRule(                                                                                                              
      check_func=check_funcs.is_not_null,                                                                                 
      column="email",                                                                                                     
      message="concat('Rule failed for value=', coalesce(cast(email as string), 'null'))"                                 
  )                   
```                                                                                                    
                                                                                                                          
YAML                                                                                                                  
```
  - name: email_not_null                                                                                                  
    check:                                                                                                                
      function: is_not_null                                                                                               
      arguments:                                                                                                          
        column: email                                                                                                     
    message: "concat('Rule failed for value=', coalesce(cast(email as string), 'null'))"                    
```              
                                                                                                                          
Both resolve to F.expr(message_str) internally. No inspect.signature, no argument dispatching, no callable serialization problem.

docs/dqx/docs/reference/quality_checks.mdx
```diff
@@ -3897,6 +3897,58 @@ Using DQX classes:
...
+    # create a check with the custom message
+    checks = [
+      DQRowRule(
+        name="email_not_null",
+        criticality="error",
+        check_func=check_funcs.is_not_null,
+        column="email",
+        message=custom_message,
+      )
+    ]
```
mwojtyczka, line 4157: may be worth to showcase how the exact message as defined in the example looks like when it fails

tests/integration/test_custom_messages.py
```diff
@@ -0,0 +1,326 @@
...
+        )
+    ]
+    expected_df = spark.createDataFrame(
+        [[1, None, 3, None, expected_warnings]],
+        EXPECTED_SCHEMA,
+    )
+    assert_df_equality(checked_df, expected_df)
+
+
+def test_apply_checks_without_custom_message_unchanged(ws, spark):
```
mwojtyczka: do we need this test? we have a lot of regression tests

mwojtyczka (changes requested):
Main point is around supporting this with metadata definition

mwojtyczka (commented):
Small, well-isolated change with good happy-path test coverage. Posting 11 inline comments below covering: a security exposure (arbitrary `F.expr` evaluation of strings sourced from YAML/metadata), a UX trap with unquoted strings, several maintainability/polish items, and one test bug.

### One non-inline issue: PR description doesn't match the implementation.

The description says:

> Add an optional `message` callable parameter ... that allows users to define custom check failure messages. The callable receives rule context (`rule_name`, `check_func_name`, `check_func_args`, `column_value`) and returns a Spark Column expression…

But the actual API is `message_expr: str | Column | None` — no callable, no rule context. Either the description should be rewritten to match what shipped, or the implementation should provide the callable form the description promises. Reviewers (and future readers running `git log` / GitHub PR archeology) rely on the description for context.

src/databricks/labs/dqx/manager.py
```diff
@@ -167,6 +169,28 @@ def _build_result_struct(self, condition: Column, skipped: bool = False) -> Colu
...
+
+        Returns:
+            The custom DQX condition message if ``message_expr`` is set on the rule, otherwise the
+            default DQX condition message.
+        """
+        if self.check.message_expr is None:
+            return condition
+
+        custom_message = (
+            self.check.message_expr if isinstance(self.check.message_expr, Column) else F.expr(self.check.message_expr)
```
mwojtyczka: **Arbitrary Spark SQL evaluation of strings from non-trusted sources.**

`F.expr(self.check.message_expr)` runs raw SQL with no validation. When checks come from YAML files, table-stored checks, or any persisted metadata source, this becomes an arbitrary-expression evaluation surface:

```yaml
- check: { function: is_not_null, arguments: { column: id } }
  message_expr: "concat('hi ', (select secret_key from secrets.config limit 1))"
```

Not a classic SQL injection (Spark `expr` doesn't allow DDL/DML), but it does allow:
- Subqueries against any catalog the running session can read.
- Calls to potentially expensive UDFs (`reflect`, `java_method`, custom Spark UDFs).
- Information disclosure via embedded subqueries that surface secrets in the violation message.

The project's CLAUDE.md security policy states:
> User-provided or templated SQL must be validated with `is_sql_query_safe()` from `utils.py` before execution. Raise `UnsafeSqlQueryError` when the query is unsafe.

Options:
- Validate `message_expr` strings via `is_sql_query_safe()` (or a stricter subset that disallows subqueries) before passing to `F.expr`.
- Restrict the string form to literal/`concat`/`coalesce` patterns only.
- At minimum, document `message_expr` strings as **trusted input** so metadata sources validate before storage.

mwojtyczka: **UX trap: a plain string passes through `F.expr` and fails confusingly.**

A new user will write:

```python
DQRowRule(..., message_expr="Email must not be null")
```

`F.expr("Email must not be null")` parses this as `Email` (column ref) `must` (column ref) `not` `be` `null` and fails with a confusing column-not-found error far from the rule definition.

The docs example correctly shows `"'Email must not be null'"` (Python double-quote wrapping a SQL single-quoted literal), but that's an awkward shape and easy to forget.

Options:
- **Auto-detect** plain strings (no parens, no SQL operators) and wrap with `F.lit` automatically.
- **Best-effort validate** at construction time (`__post_init__` tries `F.expr` once and surfaces a clear error mentioning the SQL-quote requirement).
- At minimum, expand the docs Admonition with a "common mistake" callout showing the unquoted form failing.

src/databricks/labs/dqx/manager.py
```diff
@@ -167,6 +169,28 @@ def _build_result_struct(self, condition: Column, skipped: bool = False) -> Colu
...
+        )
+        return F.when(condition.isNotNull(), custom_message).otherwise(F.lit(None).cast("string"))
```
mwojtyczka: **No length cap on rendered message.**

A pathological expression like `concat(repeat('x', 100000))` produces a 100k-char message embedded in every result row's struct — bloating Delta files and breaking downstream consumers that assume short messages. Other parts of DQX cap at 500 chars (`_LLM_FIELD_MAX_LEN` in the anomaly explainer).

Consider wrapping with `substring(message, 1, MAX_MESSAGE_LEN)` here, or applying the cap as a Spark expression on `custom_message` before the `F.when`.

src/databricks/labs/dqx/rule.py
```diff
@@ -176,6 +181,7 @@ class DQRule(abc.ABC, DQRuleTypeMixin, SingleColumnMixin, MultipleColumnsMixin):
     check_func_args: list[Any] = field(default_factory=list)
     check_func_kwargs: dict[str, Any] = field(default_factory=dict)
     user_metadata: dict[str, str] | None = None
+    message_expr: str | Column | None = None
 
     def __post_init__(self):
```
mwojtyczka, line 187: **Add best-effort `message_expr` validation in `__post_init__`.**

A typo'd SQL string (`"concat(invalid"`) currently fails only when the rule is applied to a DataFrame, far from where it was defined. A best-effort validation here surfaces the error at the call site:

```python
if isinstance(self.message_expr, str) and self.message_expr:
    try:
        F.expr(self.message_expr)
    except Exception as exc:
        raise InvalidParameterError(
            f"message_expr is not a valid Spark SQL expression: {exc}. "
            f"Note: literal strings must be wrapped in SQL quotes, e.g. \"'my message'\"."
        ) from exc
```

Cheap (build-only — no DataFrame eval), and the error message hints at the unquoted-string trap in the previous comment.

src/databricks/labs/dqx/rule.py
```diff
@@ -176,6 +181,7 @@ class DQRule(abc.ABC, DQRuleTypeMixin, SingleColumnMixin, MultipleColumnsMixin):
     check_func_args: list[Any] = field(default_factory=list)
     check_func_kwargs: dict[str, Any] = field(default_factory=dict)
     user_metadata: dict[str, str] | None = None
+    message_expr: str | Column | None = None
```
mwojtyczka, line 185: **Overloaded parameter type — consider splitting before this becomes public API.**

`message_expr: str | Column` forces callers (and serialisers, and future API extensions) to runtime-dispatch on type. Cleaner alternatives:

- `message_sql: str | None` and `message_col: Column | None` with mutual-exclusion validation.
- A single `message: Column | None` plus a `message_from_sql(s: str) -> Column` helper that wraps `F.expr` with the validation/sanitisation from the prior comments.

Not a blocker — your current shape works — but worth deciding now since changing the field name later is a breaking change.

src/databricks/labs/dqx/rule.py
```diff
@@ -165,6 +165,11 @@ class DQRule(abc.ABC, DQRuleTypeMixin, SingleColumnMixin, MultipleColumnsMixin):
     * *check_func_args* (optional) - Positional arguments for the check function (excluding *column*).
     * *check_func_kwargs* (optional) - Keyword arguments for the check function (excluding *column*).
     * *user_metadata* (optional) - User-defined key-value pairs added to metadata generated by the check.
+    * *message_expr* (optional) - User-defined expression used as the check failure message. Accepts either
```
mwojtyczka, line 168: **No placeholder support for `rule_name` / `check_func_name` / `column_value` — document the limitation.**

The PR description lists these as expected context, but the implementation evaluates the expression as-is. Users who want a message like `"<rule_name> failed for value <column_value>"` have to hand-type the rule name twice and inline the column reference (and the column reference doesn't generalise across `DQForEachColRule` instances).

This will be a frequent feature request as a follow-up. Worth either:
- Documenting the limitation explicitly in the docstring (`"placeholder substitution is not supported; reference column names directly via Spark SQL"`).
- Considering whether to add `{rule_name}`, `{column_value}` placeholder substitution in this PR or as a tracked follow-up. If you choose to defer, link to a follow-up issue here so users searching the codebase find it.

src/databricks/labs/dqx/rule.py
```diff
@@ -259,6 +265,10 @@ def to_dict(self) -> dict:
 
         if self.user_metadata:
             metadata["user_metadata"] = self.user_metadata
+        # Only string expressions can be round-tripped through metadata; Column objects are
+        # in-process Spark expressions with no canonical YAML/JSON representation.
+        if isinstance(self.message_expr, str) and self.message_expr:
+            metadata["message_expr"] = self.message_expr
```
mwojtyczka, line 274: **`to_dict()` silently drops Column-typed `message_expr` — users won't notice.**

A user who passes `message_expr=F.concat(...)` and calls `to_dict()` to persist their rules to YAML/Delta loses the message with no signal. They'll discover it only when re-loading and seeing the default message instead.

Options:
- Log a `WARNING` when a Column-typed `message_expr` is encountered during serialisation: `"Column message_expr cannot be serialised; falling back to default message on round-trip."`
- Or raise `InvalidParameterError("Column-typed message_expr cannot be serialised; pass a SQL expression string if you need round-trip persistence")`.

Logging is friendlier; raising forces the user to make an explicit choice.

src/databricks/labs/dqx/rule.py
```diff
@@ -428,6 +438,7 @@ class DQForEachColRule(DQRuleTypeMixin):
     check_func_args: list[Any] = field(default_factory=list)
     check_func_kwargs: dict[str, Any] = field(default_factory=dict)
     user_metadata: dict[str, str] | None = None
+    message_expr: str | Column | None = None
```
mwojtyczka, line 464: **`DQForEachColRule.message_expr` is shared across all generated rules — flag in the field comment.**

The integration test correctly notes this *("The same expression is used for every generated rule. To produce a per-column message, reference each column inline")*, but the field itself has no docstring or warning. A user expecting per-column message substitution will be surprised when all generated rules emit identical messages — and won't know the workaround.

Add a one-line comment on the field describing the shared-expression behaviour, ideally with the inline-reference workaround pointer:

```python
# Shared across every generated rule; reference column names inline (e.g. "concat('a=', cast(a as string))")
# if you need per-column messages, or construct rules individually instead of via DQForEachColRule.
message_expr: str | Column | None = None
```

tests/unit/test_custom_messages.py
```diff
@@ -0,0 +1,101 @@
...
+
+def test_dq_row_rule_message_expr_defaults_to_none():
+    """DQRowRule without message_expr should default to None."""
+    rule = DQRowRule(check_func=is_not_null, column="id")
+    assert rule.message_expr is None
+
+
+def test_dq_dataset_rule_message_expr_defaults_to_none():
+    """DQDatasetRule without message_expr should default to None."""
+    rule = DQRowRule(check_func=is_not_null_and_not_empty, column="id")
```
mwojtyczka: **Test bug: this test claims `DQDatasetRule` but constructs `DQRowRule`.**

```python
def test_dq_dataset_rule_message_expr_defaults_to_none():
    """DQDatasetRule without message_expr should default to None."""
    rule = DQRowRule(check_func=is_not_null_and_not_empty, column="id")
    #      ^^^^^^^^^ should be DQDatasetRule with a dataset-level check function
    assert rule.message_expr is None
```

The test name and docstring claim `DQDatasetRule`, the body uses `DQRowRule`. Either rename the test (it's now a duplicate of `test_dq_row_rule_message_expr_defaults_to_none` above), or fix the body to actually construct a `DQDatasetRule` with a `@register_rule("dataset")` check function — at which point the test exercises something new.

tests/unit/test_custom_messages.py
```diff
@@ -0,0 +1,101 @@
...
+def test_dq_rule_to_dict_omits_message_expr_when_column():
+    """Column message_expr is in-process only and is excluded from serialised metadata."""
+    rule = DQRowRule(
+        check_func=is_not_null,
+        column="id",
+        name="id_not_null",
+        message_expr=F.lit("Custom error"),
+    )
+    rule_dict = rule.to_dict()
+    assert "message_expr" not in rule_dict
```
mwojtyczka, line 102: **Add error-path and round-trip tests.**

Current unit tests cover construction + `to_dict()` happy paths. Two missing categories:

**1. Error paths** — what happens when `message_expr` is malformed?
```python
def test_apply_checks_invalid_message_expr_fails_clearly(spark):
    rule = DQRowRule(
        check_func=is_not_null,
        column="id",
        message_expr="concat(broken syntax",
    )
    df = spark.createDataFrame([(None,)], "id: int")
    with pytest.raises(Exception):
        DQEngine(...).apply_checks(df, [rule])
```
Locks in the failure mode and gives users a debugging anchor.

**2. Explicit round-trip** — implicit only today.
```python
def test_string_message_expr_round_trips_through_metadata():
    rule = DQRowRule(..., message_expr="'static msg'")
    serialized = rule.to_dict()
    # ... deserialize, apply, assert message preserved.

def test_column_message_expr_lost_on_round_trip():
    rule = DQRowRule(..., message_expr=F.lit("col msg"))
    serialized = rule.to_dict()
    assert "message_expr" not in serialized  # documented behaviour
```

src/databricks/labs/dqx/manager.py
mwojtyczka: Check if it is a string, to avoid problems with spark Column vs spark connect Column
so use `isinstance(self.check.message_expr, str)` instead

ghanse: Added an `is_sql_query_safe` check

src/databricks/labs/dqx/rule.py
ghanse: Added a warning from the logger to address this

src/databricks/labs/dqx/rule.py
ghanse: This was an intentional choice to simplify the construction of `DQRule`

src/databricks/labs/dqx/rule.py
ghanse: I don't think `F.expr(self.message_expr)` will throw if no action is triggered. We would need to trigger Spark actions for each rule on `__post_init__` if we want to do this validation. Maybe it's best left as a responsibility of the user to specify a valid expression?

mwojtyczka: true, but let's leave a note in docs about this, so that users are aware that validation is done at runtime only.

src/databricks/labs/dqx/manager.py
ghanse: Added a cap on the length and an admonition in the docs

src/databricks/labs/dqx/manager.py
ghanse: It might be messy to validate the message expression. I added an admonition in the docs, do you think that's good enough?

tests/unit/test_custom_messages.py
ghanse: Added

src/databricks/labs/dqx/manager.py
ghanse: Updated the logic to check for a string-valued expression and fallback to Column

tests/unit/test_custom_messages.py
ghanse: Fixed

src/databricks/labs/dqx/rule.py
ghanse: Added a small note in the docstring for `DQRule` that describes the behavior when generating rules with `DQForEachColRule`.

src/databricks/labs/dqx/manager.py
ghanse: Added truncation and added a note in the documentation about the 500 character limit.

src/databricks/labs/dqx/manager.py
ghanse: Added a note in the documentation about this.

ghanse: Added validation with `is_sql_query_safe` for Spark SQL strings.

src/databricks/labs/dqx/rule.py
ghanse: I updated the PR description. Column value is available through column references and the rule name can be passed as a literal. Added an example in the docs for custom message expressions.

src/databricks/labs/dqx/manager.py
```diff
@@ -167,6 +170,37 @@ def _build_result_struct(self, condition: Column, skipped: bool = False) -> Colu
...
+        Returns:
+            The custom DQX condition message if ``message_expr`` is set on the rule, otherwise the
+            default DQX condition message.
+        """
+        _max_message_length = 500
+        if self.check.message_expr is None:
+            return condition
+
+        if isinstance(self.check.message_expr, str):
+            if not is_sql_query_safe(self.check.message_expr):
```
mwojtyczka: **Reusing `is_sql_query_safe` here rejects ordinary English messages.** It does naive word-boundary matching against keywords like `update`, `use`, `create`, `delete`, `alter`, `replace`. Those are common words in human-readable failure messages, so legitimate static messages fail at apply time:

```
'Please update your email address' -> UnsafeSqlQueryError
'Use a valid country code'         -> UnsafeSqlQueryError
'Failed to create record'          -> UnsafeSqlQueryError
```

The keyword sits inside a quoted SQL string literal, so it's harmless — but the check greps rather than parses. A "custom message" feature where you can't write *"Please update your email"* will surprise users, and `UnsafeSqlQueryError` won't point at the offending word.

Suggestions: don't reuse the query-injection guard for message expressions (the risk surface is `F.expr` evaluating a scalar string, not DDL/DML); or at minimum surface which token tripped the check and document the limitation.

src/databricks/labs/dqx/rule.py
```diff
@@ -184,6 +191,8 @@ def __post_init__(self):
         self._validate_attributes()
         check_condition = self.get_check_condition()
         self._initialize_name_if_missing(check_condition)
+        if isinstance(self.message_expr, str):
+            self._validate_message_expression(self.message_expr)
```
mwojtyczka, line 195: Validation is split inconsistently: `_validate_message_expression` (parse via `F.expr`) runs here at **construction** time, but `is_sql_query_safe` runs much later at **apply** time in `manager._build_message_col`. So a malformed expression fails fast, but an "unsafe" one only fails when checks run. The docs say SQL messages are "only validated when checks are run," which contradicts this construction-time validation. `is_sql_query_safe` is a pure function — consider running it here too for fail-fast feedback and aligning the docs.

src/databricks/labs/dqx/manager.py
```diff
@@ -167,6 +170,37 @@ def _build_result_struct(self, condition: Column, skipped: bool = False) -> Colu
...
+                raise UnsafeSqlQueryError(
+                    "Provided message expression is not safe for execution. Please ensure it does not contain any unsafe operations."
+                )
+            return F.when(
```
mwojtyczka: Nit: the two branches of `_build_message_col` differ only by the safety check and the `F.expr(...)` wrap. Could collapse to one path to avoid the duplicated `F.when(...).substr(...).otherwise(...)`:

```python
expr = self.check.message_expr
if isinstance(expr, str):
    if not is_sql_query_safe(expr):
        raise UnsafeSqlQueryError(...)
    expr = F.expr(expr)
return F.when(condition.isNotNull(),
              F.substr(expr, F.lit(1), F.lit(_max_message_length))
     ).otherwise(F.lit(None).cast("string"))
```

Also worth a one-line comment that the `Column` path is intentionally not safety-checked (caller-constructed, trusted), and consider hoisting `_max_message_length = 500` to a constant since the limit is also referenced in the docs.

mwojtyczka (commented):
Re-review of custom check messages. Solid, well-tested feature with clean null-preservation semantics. One higher-impact concern around the safety check producing false positives on ordinary message text, plus a few smaller notes left inline.

src/databricks/labs/dqx/manager.py
mwojtyczka: i think we should documente only

src/databricks/labs/dqx/manager.py
ghanse: Removed the `is_sql_query_safe` check per the previous comment. Refactored this block.

src/databricks/labs/dqx/manager.py
ghanse: I removed the expression validation using `is_sql_query_safe` and added a note in the documentation about expression safety.

mwojtyczka (approved):
LGTM

PR #1097, opened 2026-03-27
Extend summary metrics with per-check-name breakdowns

### LINKED ISSUE #943: [FEATURE]: Display summary metrics grouped by data quality checks.

### Problem statement

The summary metrics show the total number of input lines processed and the respective errors, warnings, and valid counts. It's not possible to drill down this count by data quality check.

### Proposed Solution

It would be nice to have these metrics grouped by quality checks so we can see where things aren't going well. 

### Additional Context

Furthermore, a natural next feature would be to create some kind of limit for each check, in order to interrupt the workflow or send an alert if that limit is exceeded.

- criticality: warn
      check:
        function: is_in_list
        **threshold: 0.9**
        arguments:
          column: col1
          allowed:
            - 1
            - 2
           
In this example, I would accept that I have 10% of col1 out of range [1,2]

### LINKED ISSUE #1065: [FEATURE]: Tests run metrics summary

### Problem statement

Currently the summary metrics https://databrickslabs.github.io/dqx/docs/guide/summary_metrics/#metrics-table-schema are focussed on `row_level` checks. It would be useful to monitor health in dashboarding tools by also being able to look at individual tests.

Example:
mycatalog.myschema.mytable -> run_id: 1234 gave `14/15` tests passed.

- For the failed tests I would obviously still want to see the detailed row-level findings.
- For passed tests it would be still be a useful overview. 

Currently it's possible to check the `rows` that passed all tests, however context is missing on the tests that have run on a certain table/dataframe.

### Proposed Solution

Include a metric/summary for test-level findings. Otherwise users would have to check the tests that ran, what 
rows failed for which tests and create code that generates a pass for tests that don't show up.

### Additional Context

_No response_

ghanse:
## Changes                                                                                          

Extended `DQMetricsObserver` to emit a new `check_metrics` row alongside the existing aggregates (`input_row_count`, `error_row_count`, `warning_row_count`, `valid_row_count`). The value is a JSON array of structs — one entry per check — that fits the existing `metric_name` / `metric_value` schema without widening it:                                                                                                                                                                                             
```json                                                              
  [{"check_name": "id_is_not_null", "error_count": 1, "warning_count": 0}, ...]
```

Backward compatible: existing metrics unchanged, new row is additive and opt-in.  

Key design decisions:
- **Backward compatible**: Existing aggregate metrics are unchanged; per-check metrics are appended after them.
- **Automatically enabled**: no knobs, auto-enabled.
- **Safe aliases**: Check names with special characters are sanitized for use as Spark SQL aliases.
- **No new dependencies**: Uses existing Spark SQL `exists()` higher-order function.

### Linked issues

Closes #943

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

This pull request was AI-assisted by Isaac.


src/databricks/labs/dqx/metrics_observer.py
```diff
@@ -8,6 +9,23 @@
...
+
+    Args:
+        name: The raw check name.
+
+    Returns:
+        A sanitized string safe for use as a Spark SQL alias.
+    """
+    sanitized = re.sub(r"[^a-zA-Z0-9_]", "_", name)
+    sanitized = re.sub(r"_+", "_", sanitized)
+    return sanitized.strip("_")
```
ghanse: **Edge case: empty alias from all-special-character check names.**

If `name` is something like `"---"`, sanitization produces `"___"`, then `strip("_")` returns `""`. This would generate invalid aliases like `_error_count` and `_warning_count`.

Suggestion: add a guard in `_build_per_check_metrics` (or here) to raise a `ValueError` or skip the check if the sanitized alias is empty. A test for this edge case would be good too.

src/databricks/labs/dqx/metrics_observer.py
```diff
@@ -81,24 +100,57 @@ def id(self) -> str:
...
+
+        For each check name, generates a count of rows where that check appears in the errors or warnings array.
+
+        Returns:
+            A list of Spark SQL expressions for per-check metrics.
+        """
+        per_check_metrics: list[str] = []
+        for check_name in self._check_names:
+            safe_alias = _sanitize_metric_alias(check_name)
+            escaped_name = check_name.replace("'", "\\'")
```
ghanse: **Nit: single-quote escaping style.**

`check_name.replace("'", "\\'")`  produces `\'` in the SQL string. While Spark SQL does accept backslash-escaped single quotes, the canonical SQL escape is doubling: `''`. Consider:

```python
escaped_name = check_name.replace("'", "''")
```

This is more portable and avoids potential issues if `spark.sql.ansi.enabled` or other SQL mode settings change escaping behavior.

ghanse (commented):
Good implementation. The per-check metric generation is clean, backward-compatible, and well-tested. Two minor findings below — neither is blocking but both are worth addressing.

src/databricks/labs/dqx/metrics_observer.py
ghanse: We can raise a `DQXError` for this.

ghanse: Changed to `DQXError` and added the `track_extended_metrics` flag (default `False`) to gate per-check metrics.

src/databricks/labs/dqx/metrics_observer.py
ghanse: Fixed — now using doubled single quotes (`''`) for Spark SQL standard escaping.

ghanse (commented):
Addressed duplicate alias collision — `_build_per_check_metrics` now detects when two check names sanitize to the same alias and raises `DQXError`. Added unit test for the edge case (e.g. `check-1` vs `check.1`). Ready for cleanup pass.

ghanse:
**Cleanup pass complete.**

- Renamed `default_metrics` to `all_metrics` in `metrics` property since it accumulates default, per-check, and custom metrics
- Fixed copy-paste docstring in `test_dq_observer_custom_column_names` that incorrectly said "default column names"

Tests pass (16/16). Ready for review.

ghanse (commented):
**Post-cleanup final review: APPROVED**

Reviewed the full diff across `metrics_observer.py`, `engine.py`, `test_observer.py`, and `test_summary_metrics.py`.

**Correctness**: The Spark SQL `exists()` higher-order function is the right approach for checking array-of-struct columns. Single-quote escaping, alias sanitization, and duplicate collision detection are all correctly implemented. The `@cached_property` to `@property` change on `metrics` is necessary and correct since check names are set dynamically.

**Completeness**: Fully addresses #943 with per-check-name error/warning counts. The `track_extended_metrics` flag defaults to `False` for full backward compatibility. Both `apply_checks` and `apply_checks_by_metadata` paths are covered since the latter delegates to the former.

**Test coverage**: 10 new unit tests and 2 parametrized integration tests (4 total runs) cover: default behavior, per-check metrics, metric ordering, custom column names, special character sanitization, check name replacement, empty alias error, and duplicate alias collision.

**Code quality**: Clean naming, proper docstrings, no dead code, pure helper function for sanitization. AGENTS.md compliant — type hints on all signatures, `DQXError` for exceptions, no new `Any` usage.

No changes required. Ready for human approval.

src/databricks/labs/dqx/metrics_observer.py
```diff
@@ -66,10 +88,12 @@ class DQMetricsObserver:
 
     name: str = "dqx"
     custom_metrics: list[str] | None = None
+    track_extended_metrics: bool = False
```
mwojtyczka: do we need parameter for this? shouldn't we have this enabled by default?

src/databricks/labs/dqx/metrics_observer.py
```diff
@@ -81,23 +105,63 @@ def id(self) -> str:
         """
         return self.id_overwrite or str(uuid4())
```
mwojtyczka, line 83: **MEDIUM - `metrics` changed from `cached_property` to `property`:** This means every access to `observer.metrics` rebuilds the full list including per-check SQL expressions. The previous `cached_property` computed once and cached the result.

This change is necessary because `set_check_names` mutates `_check_names` between calls, so caching would return stale results. However, consider documenting this intentional behavioral shift for callers who may access `.metrics` multiple times.

src/databricks/labs/dqx/metrics_observer.py
```diff
@@ -81,23 +105,63 @@ def id(self) -> str:
...
-        default_metrics = [
+        all_metrics = [
             "count(1) as input_row_count",
             f"count(case when {self._error_column_name} is not null then 1 end) as error_row_count",
             f"count(case when {self._warning_column_name} is not null then 1 end) as warning_row_count",
             f"count(case when {self._error_column_name} is null and {self._warning_column_name} is null then 1 end) as valid_row_count",
         ]
+        if self.track_extended_metrics:
+            all_metrics.extend(self._build_per_check_metrics())
         if self.custom_metrics:
```
mwojtyczka, line 103: **Good:** `check_name.replace("'", "''")` is correct ANSI SQL escaping. The `_sanitize_metric_alias` properly strips non-alphanumeric chars for the identifier portion. Both the string literal and identifier are handled safely.

src/databricks/labs/dqx/engine.py
```diff
@@ -165,6 +165,10 @@ def apply_checks(
             ref_dfs,
             rule_set_fingerprint=rule_set_fingerprint,
         )
+
+        if self.observer and self.observer.track_extended_metrics:
```
mwojtyczka: Please test that this works correctly when check.name is not provided but auto-derived.

tests/integration/test_summary_metrics.py
```diff
@@ -2408,3 +2408,115 @@ def test_streaming_observer_metrics_output_and_quarantine_with_empty_checks(
...
+
+
+@pytest.mark.parametrize("apply_checks_method", [DQEngine.apply_checks, DQEngine.apply_checks_by_metadata])
+def test_observer_extended_metrics_checks_change_between_runs(ws, spark, apply_checks_method):
+    """Test that extended metrics reflect the correct checks when the rule set changes between runs."""
+    full_checks_metadata = TEST_CHECKS
+    reduced_checks_metadata = [TEST_CHECKS[0]]
+
+    test_df = spark.createDataFrame(
+        [
```
mwojtyczka, line 2659: **LOW - Test duplication:** The `checks_change_between_runs` test has significant duplication between the `apply_checks` and `apply_checks_by_metadata` branches. Since `apply_checks_by_metadata` delegates to `apply_checks`, they exercise nearly identical paths. Consider whether parametrize is necessary here.

tests/unit/test_observer.py
```diff
@@ -100,3 +102,105 @@ def test_dq_observer_observation_property():
     observation = observer.observation
     assert observation is not None
     assert isinstance(observation, Observation | SparkConnectObservation)
```
mwojtyczka, line 52: Good test coverage - covers: no check names, with check names, ordering, custom column names, special character sanitization, name replacement, empty alias error, and alias collision detection.

mwojtyczka (commented):
## Review: Extend summary metrics with per-check-name breakdowns

Clean feature addition. The approach of generating per-check SQL expressions via the observer is sound. A few items below.

tests/integration/test_summary_metrics.py
```diff
@@ -2408,3 +2408,115 @@ def test_streaming_observer_metrics_output_and_quarantine_with_empty_checks(
...
+            [1, "Alice", 30, 50000],
+            [2, "Bob", 25, 45000],
+            [None, "Charlie", 35, 60000],
+            [4, None, 28, 55000],
+        ],
+        TEST_SCHEMA,
+    )
+
+    if apply_checks_method == DQEngine.apply_checks:
+        checks = deserialize_checks(TEST_CHECKS)
```
mwojtyczka, line 2584: We should have one test that don't have check name specified to ensure it is resolved correctly. TEST_CHECKS contains checks with check.name specified.

mwojtyczka (changes requested):

src/databricks/labs/dqx/metrics_observer.py
```diff
@@ -81,23 +105,63 @@ def id(self) -> str:
...
+    def set_check_names(self, check_names: list[str]) -> None:
+        """
+        Sets the check names used to generate per-check summary metrics.
+
+        Args:
+            check_names: List of check names from the applied quality rules.
+        """
+        self._check_names = check_names
+
+    def _build_per_check_metrics(self) -> list[str]:
```
mwojtyczka: we need more compact way to report check level metrics, e.g.
```
metric name: check_metrics
metric value: array<struct<check_name: string, error_count: int, warning_count: int>>
```
`metric_value` column in the summary metrics table is a string so we can serialize to json

ghanse:
Addressed review feedback — pushed `a1141c1c`:

**Compact check_metrics format**: Replaced individual per-check `{alias}_error_count` / `{alias}_warning_count` metrics with a single `check_metrics` row. The metric_value is a JSON-serialized `array<struct<check_name: string, error_count: int, warning_count: int>>`:

```json
[{"check_name": "id_is_not_null", "error_count": 5, "warning_count": 0}, ...]
```

This uses `to_json(array(named_struct(...)))` in the SQL expression so it fits naturally into the existing `metric_name`/`metric_value` string schema.

**Other changes:**
- Removed `_sanitize_metric_alias` and alias collision logic (no longer needed with single-row format)
- `track_extended_metrics` flag retained as the opt-in gate
- Updated unit tests: 18 tests covering single-use, repeated use with different checks, clearing, custom columns, quote escaping, ordering with custom metrics

mwojtyczka (approved):
LGTM

PR #1098, opened 2026-03-27
Migrated from hatch to uv package manager and hardened GHA workflows

mwojtyczka:
## Changes

* Refactored and simplified workflows to increase infra isolation and avoid azure login
* Moved perf tests from PRs to nightly and update the process for publishing benchmarks for additional isolation:
   - (same) run tests and generate a new perf baseline
   - (same) merge it with the existing one
   - (same) compare against it to detect regressions (>25% mean degradation)
   - (new, commit to current PR before) if the baseline changed and new benchmarks are added or removed, open a PR in the repo. PR commit is unsigned — a maintainer must pull, sign, and push before merging
* Fixed dbt test
* Migrate to `uv` from `hatch` package manager

### Tests

- [x] manually tested
- [ ] added unit tests
- [ ] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests


ghanse (approved):

mwojtyczka (commented):
LGTM

.github/workflows/acceptance.yml
```diff
@@ -144,44 +141,31 @@ jobs:
...
       - name: Install dbt
-        run: |
-          pip install dbt-core==1.10.9 dbt-databricks==1.10.9
+        run: uv pip install --system dbt-core==1.10.9 dbt-databricks==1.10.9
 
+      # Required for DAB (Databricks Asset Bundle) e2e tests
       - name: Install Databricks CLI
         run: |
-          curl -fsSL https://raw.githubusercontent.com/databricks/setup-cli/main/install.sh | sh
+          curl -fsSL https://raw.githubusercontent.com/databricks/setup-cli/e0f69d54c90e95ccad22d5ff17751da665e599a9/install.sh | sh
```
asnare: Use the GHA for this instead:

```suggestion
      # Required for DAB (Databricks Asset Bundle) e2e tests
      - name: Install Databricks CLI
        uses: databricks/setup-cli@v0.296.0
```

(I don't think this needs to be pinned, it's a first-party dependency but not all the tools understand this.)

.github/workflows/release.yml
```diff
@@ -1,9 +1,10 @@
 name: Release
 
 on:
```
asnare: My original instructions had a mistake here, it turns out this needs to be:
```suggestion
on: {}
```

.github/workflows/acceptance.yml
```diff
@@ -206,38 +193,22 @@ jobs:
...
-      - name: Install hatch
-        run: pip install hatch==1.16.5
+          version: "0.11.2"
+          checksum: "7ac2ca0449c8d68dae9b99e635cd3bc9b22a4cb1de64b7c43716398447d42981"
 
+      # Required for dbt e2e tests
       - name: Install dbt
-        run: |
-          pip install dbt-core==1.10.9 dbt-databricks==1.10.9
+        run: uv pip install --system dbt-core==1.10.9 dbt-databricks==1.10.9
```
asnare: These need to be pinned. The simplest way to do this is probably to include them in the `test` dependency group on the project. (The `Makefile` target currently includes all dependency-groups when running tests, but you can adjust if necessary.)

.github/workflows/acceptance.yml
```diff
@@ -206,38 +193,22 @@ jobs:
...
 
+      # Required for DAB (Databricks Asset Bundle) e2e tests
       - name: Install Databricks CLI
         run: |
-          curl -fsSL https://raw.githubusercontent.com/databricks/setup-cli/main/install.sh | sh
+          curl -fsSL https://raw.githubusercontent.com/databricks/setup-cli/e0f69d54c90e95ccad22d5ff17751da665e599a9/install.sh | sh
           databricks --version
```
asnare: Use the GHA, as mentioned above.

.github/workflows/docs-release.yml
```diff
@@ -1,10 +1,13 @@
 name: Docs Release
 
 on:
```
asnare:
```suggestion
on: {}
```

.github/workflows/nightly.yml
```diff
@@ -292,44 +231,31 @@ jobs:
...
-      - name: Install hatch
-        run: pip install hatch==1.16.5
+          version: "0.11.2"
+          checksum: "7ac2ca0449c8d68dae9b99e635cd3bc9b22a4cb1de64b7c43716398447d42981"
 
+      # Required for dbt e2e tests
       - name: Install dbt
-        run: |
-          pip install dbt-core==1.10.9 dbt-databricks==1.10.9
+        run: uv pip install --system dbt-core==1.10.9 dbt-databricks==1.10.9
```
asnare: Noted elsewhere: this should be moved into a dependency group so that it's pinned.

.github/workflows/nightly.yml
```diff
@@ -292,44 +231,31 @@ jobs:
...
 
+      # Required for DAB (Databricks Asset Bundle) e2e tests
       - name: Install Databricks CLI
         run: |
-          curl -fsSL https://raw.githubusercontent.com/databricks/setup-cli/main/install.sh | sh
+          curl -fsSL https://raw.githubusercontent.com/databricks/setup-cli/e0f69d54c90e95ccad22d5ff17751da665e599a9/install.sh | sh
```
asnare: As noted elsewhere, should be the GHA.

.github/workflows/nightly.yml
```diff
@@ -353,38 +282,22 @@ jobs:
...
+          version: "0.11.2"
+          checksum: "7ac2ca0449c8d68dae9b99e635cd3bc9b22a4cb1de64b7c43716398447d42981"
 
+      # Required for dbt e2e tests
       - name: Install dbt
-        run: |
-          pip install dbt-core==1.10.9 dbt-databricks==1.10.9
-      
+        run: uv pip install --system dbt-core==1.10.9 dbt-databricks==1.10.9
+
```
asnare: Just flagging so we don't overlook it.

.github/workflows/nightly.yml
```diff
@@ -353,38 +282,22 @@ jobs:
...
+      # Required for DAB (Databricks Asset Bundle) e2e tests
       - name: Install Databricks CLI
         run: |
-          curl -fsSL https://raw.githubusercontent.com/databricks/setup-cli/main/install.sh | sh
+          curl -fsSL https://raw.githubusercontent.com/databricks/setup-cli/e0f69d54c90e95ccad22d5ff17751da665e599a9/install.sh | sh
```
asnare: (Flagging to avoid overlooking it.)

asnare (commented):
Looks good, although I've noted some things to look at:

 - It turns out the empty workflow triggers need to be `on: {}`.
 - It's probably better to use the GHA for installing the Databricks CLI rather than the curl/install approach.
 - The `dbt` packages needed for e2e tests should be moved into the dependency groups and pinned.

I think there's also a missing file: `.build-constraints.txt`
This contains the hatchling pins that will (eventually) be used for building releases. You can generate it with `make lock-dependencies` and test it with `make build`: I think it will fail because that file is missing. (Although maybe it wasn't checked in?)

.github/workflows/acceptance.yml
ghanse: I pinned it to `acd0e77a1ed7f15f528faca1e1f7f5590bcfdff8` just for safety.

.github/workflows/release.yml
ghanse: Done

.github/workflows/acceptance.yml
ghanse: Added these to `pyproject.toml`. I have updated the versions:
```
# Updated for compatibility with 'databricks-sdk~=0.73'
dbt-core~=1.11.6,
dbt-databricks~=1.11.6,
```

.github/workflows/acceptance.yml
mwojtyczka: When using SHA we should leave version number in the comment

mwojtyczka: i guess if we pin github ations, even if dbx owned, we should follow the same for scripts

.github/actions/jfrog-auth/jfrog-auth
```diff
@@ -0,0 +1,37 @@
+#!/bin/sh
```
alexott: If we use `bash` functionality, it's best explicitly specify it. I had a few cases when `/bin/sh` was pointing to `dash` instead of `bash`

alexott (approved):

.github/actions/jfrog-auth/jfrog-auth
mwojtyczka: updated

.github/workflows/acceptance.yml
asnare: I strongly recommend switching to the GHA here instead of curl/running the install script.

.github/workflows/acceptance.yml
asnare: I strongly advise switching to the GHA here instead of curl/running the install script.

.github/workflows/nightly.yml
```diff
@@ -427,6 +339,172 @@ jobs:
...
+          export DATABRICKS_HOST=$(az keyvault secret show --id "${{ secrets.VAULT_URI }}/secrets/DATABRICKS-HOST" --query value -o tsv)
+          export DATABRICKS_AUTH_TYPE=azure-cli
+          # We don't use the acceptance action here as it may finish before all benchmark files are written.
+          # Tests are run sequentially to reduce variability.
+          # Create baseline benchmarks under: $BENCHMARKS_DIR/<machine-specific-folder>/*_baseline.json
+          # Do at least 5 rounds to get more stable results.
+          UV_LOCKED=1 uv run --all-extras pytest tests/perf -v -n 1 \
+            --benchmark-storage=$BENCHMARKS_DIR \
+            --benchmark-save=baseline \
+            --benchmark-min-rounds=5
```
asnare: I think you want to `--frozen` instead of locking here.

.github/workflows/nightly.yml
```diff
@@ -435,11 +513,15 @@ jobs:
...
+          # remove the new and updated baseline files to avoid confusion, only FINAL_BASELINE is needed for comparison
+          rm -rf $NEW_BASELINE
+          rm -rf $UPDATED_BASELINE
+
           # We are not using acceptance action as it does not show the comparison results.
           # The run fails if performance degrades by more than 25%.
           # Tests are run sequentially to reduce variability.
           # Do at least 5 rounds to get more stable results.
-          hatch run pytest tests/perf -v -n 1 \
+          UV_LOCKED=1 uv run --all-extras pytest tests/perf -v -n 1 \
```
asnare: Here I think you also want `--frozen` instead of locking.

.github/workflows/release.yml
asnare, line 1: I would remove this workflow entirely for now: there's only a downside to it being here at the moment, and once the new release process is available we can restore it if necessary.

asnare (commented):
Looks fine. Some things I would change:

 - Use the GHA for installing the Databricks CLI, instead of the curl/run process.
 - I think there are a few stray `uv run` left over in the workflows which are running locked where I think the intent is that they should be frozen.
 - I'd drop `release.yml` entirely for now.

In the workflows, there are some places where I think JFrog auth is missing and will be needed: that can be dealt with once GHA is enabled.

.github/workflows/nightly.yml
```diff
@@ -292,43 +231,24 @@ jobs:
...
-          client-id: ${{ secrets.ARM_CLIENT_ID }}
-          tenant-id: ${{ secrets.ARM_TENANT_ID }}
-          allow-no-subscriptions: true
-
-      - name: Set env vars for Azure CLI auth
-        run: |
-          val=$(az keyvault secret show --id "${{ secrets.VAULT_URI }}/secrets/DATABRICKS-HOST" --query value -o tsv)
-          echo "DATABRICKS_HOST=$val" >> $GITHUB_ENV
-          echo "DATABRICKS_AUTH_TYPE=azure-cli" >> $GITHUB_ENV
+        run: databricks/setup-cli@acd0e77a1ed7f15f528faca1e1f7f5590bcfdff8 # setup-cli/v0.296.0
```
asnare:
```suggestion
        uses: databricks/setup-cli@acd0e77a1ed7f15f528faca1e1f7f5590bcfdff8 # setup-cli/v0.296.0
```
Oops, I missed this before.

mwojtyczka: updated

.github/workflows/nightly.yml
mwojtyczka: updated to `UV_FROZEN=1`

.github/workflows/nightly.yml
mwojtyczka: updated to `UV_FROZEN=1`

PR #1104, opened 2026-04-01
feat: Improved summary stats report for string datatype columns

### LINKED ISSUE #670: [FEATURE]: Improve summary stats report for string datatype columns

### Problem statement

The "profile" method uses the Databricks built-in "summary()" function to generate the summary statistics report. This function returns some predefined metrics that may not be suitable for text variables, especially ones associated with numeric types (mean,stddev, etc).  For some of them, the summary stats already produce Null values, while for others they still return results, such as the minimum and maximum metrics, as in the example below.

( 'work_order_id': {'count': 6,
  'mean': None,
  'stddev': None,
  'min': 'INVALID',
  '25%': None,
  '50%': None,
  '75%': None,
  'max': 'WO-005',
  'count_non_null': 6,
  'count_null': 0})

### Proposed Solution

One approach is to return null values for "min" and "max" variables too, once it seems a more consistent output for this datatype. Another approach would be to generate/add a set of specific metrics for low cardinality string datatypes, like "count_distinct","max_length","mean_length","median_length","min_length","histogram",etc.

### Additional Context

It would be useful to keep track of changes in the data source profile through time, and this data is available only in the summary_stats variable. I think that the variables metrics could be datatype driven.

Roshan1299:
## Changes

- Nulled out `min` and `max` metrics for string columns in summary stats, since lexicographic min/max values are not meaningful for text data
- Added `count_distinct` metric for all column types

### Linked issues

Resolves #670 

### Tests

- [ ] manually tested
- [ ] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

Both changes are in private methods (`_process_metric`, `_profile`) that are only testable through the public `profile()` API, which requires a live SparkSession. The existing integration tests in `test_profiler.py` exercise these code paths and will validate in CI. Happy to add a dedicated integration test if needed.

- [x] Formatting passes (`make fmt` — pylint 10/10, mypy clean)
- [x] Unit tests pass (`make test` — 939 passed)


ghanse (changes requested):
This looks good to me. Can you add a few small tests to validate that distinct counts are computed and that min and max are not computed for string type columns?

You should be able reuse the existing patterns in `integration/test_profiler.py`.

Roshan1299:
@ghanse Added Integration Tests

ghanse (approved):
LGTM!

src/databricks/labs/dqx/profiler/profiler.py
```diff
@@ -395,6 +395,8 @@ def _profile(
             metrics["count"] = total_count
             metrics["count_null"] = total_count - count_non_null
             metrics["count_non_null"] = count_non_null
+            distinct_row = column_df.select(F.countDistinct(F.col(column_label))).first()
```
mwojtyczka: The `count_distinct` adds a separate Spark action with a shuffle per column. Combine it with the existing `count_non_null` into a single action:
```
stats = column_df.agg(
   F.count(column_label).alias("cnt"),
   F.countDistinct(column_label).alias("cnt_distinct"),
).first()
count_non_null = stats[0]
metrics["count_distinct"] = stats[1]
```
This halves the number of Spark actions in that loop.

mwojtyczka (changes requested):
LGTM - please fix the performance issue as suggested

src/databricks/labs/dqx/profiler/profiler.py
mwojtyczka: fixed

mwojtyczka (approved):
LGTM - i run int and e2e tests from local

PR #1106, opened 2026-04-07
Feature: Foreign key check with null safe support

### LINKED ISSUE #1069: [FEATURE]: Foreign key check with null safe support

### Problem statement

The documentation is currently not clearly stating that `foreign_key` check behaves according to ANSI SQL standard. This is mentioned only in the function docstring. 

Sometimes it is needed to compare combination of fields where source field can be empty or referenced field can be empty (null). While it might be possible to prepare data in a way that they could match with some transformations, I think that:

1. it adds additional complexity to the process
2. when would be followed true medallion architecture (i.e. loading data from file to bronze layer and checking the data before load) you probably don't even want to do any kind of transformation at this point

### Proposed Solution

It would a be a nice quality of life improvement if there would be an option to have null safe foreign key check: [documentation](https://spark.apache.org/docs/3.5.7/api/python/reference/pyspark.sql/api/pyspark.sql.Column.eqNullSafe.html)

### Additional Context

_No response_

IvannKurchenko:
## Changes
Added `null_safe` parameter to `foreign_key` check to explicitly enable `null` foreign key comparison. To distinguish cases when `null` foreign key is matched vs foreign key is not found completely when `null_safe=True` wrapping even single columns comparison into Struct.

**Special behaviour**

Enabling `null_safe=True` on a previously non-null-safe single-column FK changes the message and auto created check name:
   - the rule name from `a_not_exists_in_ref_b` → `struct_a_as_a_not_exists_in_ref_struct_b_as_a`
   - the violation message from `Value '...' in column 'a' not found in reference column 'b'` → `'struct(a AS a)' ... 'struct(b AS a)'`

### Linked issues
https://github.com/databrickslabs/dqx/issues/1069

Resolves #1069

### Tests

- [ ] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests


mwojtyczka:
@IvannKurchenko thank you for the PR. We require all commits to be signed. Please follow the instructions: https://databrickslabs.github.io/dqx/docs/dev/contributing/#first-contribution

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1370,15 +1373,20 @@ def foreign_key(
             - if both *ref_df_name* and *ref_table* are provided.
             - if the number of *columns* and *ref_columns* do not match.
             - if *ref_df_name* is not found in the provided *ref_dfs* dictionary.
+            - if both *negate* and *null_safe* are enabled.
```
mwojtyczka: **Blocker — join condition not updated for null-safe comparison**

The struct wrapping correctly solves the `base_condition` problem (a struct with NULL fields is itself not null, so `col_expr.isNotNull()` returns True). However, the `apply` closure's join at line 1411 is unchanged:

```python
on=(col_expr == F.col(ref_alias)) & col_expr.isNotNull() & filter_expr
```

The `==` operator follows standard SQL NULL semantics: `struct(NULL AS a) == struct(NULL AS a)` evaluates to **NULL** (not True), because the underlying field comparison `NULL == NULL` is NULL. A NULL join condition means no match in the left join, so the ref column will always be NULL for null FK values — indistinguishable from "FK not found".

This affects `test_foreign_key_check_null_safe_foreign_key_present`, `test_foreign_key_check_null_safe_on_composite_keys`, and `test_foreign_key_check_null_safe_yaml` which all expect null-to-null matching to succeed.

The PR description correctly notes that `eqNullSafe` on a plain column won't work because `base_condition` filters out NULLs. But the combination of struct wrapping (for `base_condition`) **plus** per-field `eqNullSafe` (for the join) should give the correct behavior. For example:

```python
# inside apply():
if null_safe:
    field_conditions = [
        _get_column_expr(c).eqNullSafe(F.col(ref_alias)[field_name])
        for c, field_name in zip(columns, columns_names)
    ]
    join_cond = functools.reduce(operator.and_, field_conditions) & filter_expr
else:
    join_cond = (col_expr == F.col(ref_alias)) & col_expr.isNotNull() & filter_expr
```

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1370,15 +1373,20 @@ def foreign_key(
             - if both *ref_df_name* and *ref_table* are provided.
```
mwojtyczka, line 1374: **Questionable restriction** — this docstring says an error is raised "if both *negate* and *null_safe* are enabled", but there's no validation code, and more importantly, there's no clear reason this combination should be invalid. `negate=True` + `null_safe=True` would mean "flag rows where a NULL FK value exists in the reference" — that's a perfectly valid check. I'd suggest removing this claim from the docstring rather than adding validation for it.

mwojtyczka (changes requested):
Thanks for the contribution — the idea of wrapping single columns in a struct to preserve NULL-aware `base_condition` is a sound approach. However, the join condition inside the `apply` closure is not updated for null-safe matching, which means the null-safe tests should fail. See inline comments for details.

**Summary:**

| Severity | Item |
|---|---|
| **Blocker** | Join condition uses `==` which cannot match NULL struct fields — null-safe matching won't work |
| **High** | Missing validation for `negate` + `null_safe` despite being documented as raising an error |
| **Low** | Docstring grammar |

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1354,9 +1355,11 @@ def foreign_key(
         ref_columns: List of column names (str) or Column expressions in the reference dataset.
         ref_df_name: Name of the reference DataFrame (used when passing DataFrames directly).
         ref_table: Name of the reference table (used when reading from catalog).
-        row_filter: Optional SQL expression for filtering rows before checking the foreign key. Auto-injected from the check filter.
         negate: If True, the condition is negated (i.e., the check fails when the foreign key values exist in the
             reference DataFrame/Table). If False, the check fails when the foreign key values do not exist in the reference.
+        row_filter: Optional SQL expression for filtering rows before checking the foreign key. Auto-injected from the check filter.
+        null_safe: If True use check NULL foreign key values to match NULL references values.
```
```suggestion
        null_safe: If True, checks NULL foreign key values to match NULL reference values.
```
correct grammar

mwojtyczka:
@IvannKurchenko you need to repush all commits, the first commit is still not signed

IvannKurchenko:
@mwojtyczka I apologise for this, there was also merge commits from GitHub, now everything is singed.

src/databricks/labs/dqx/check_funcs.py
IvannKurchenko: Thank you!

src/databricks/labs/dqx/check_funcs.py
IvannKurchenko: Sorry, this was left over after previous implementation. Removed, thanks for the comment.

src/databricks/labs/dqx/check_funcs.py
IvannKurchenko: @mwojtyczka Thank you so much for your feedback I do appreciate it a lot. I have a question regarding statement:
 > struct(NULL AS a) == struct(NULL AS a) evaluates to NULL (not True)

I've tried the following query in Databricks (runtime 17.3) and aforementioned comparison evaluated to true:
```
%sql
SELECT named_struct('field', NULL) AS null_struct,
       named_struct('field', NULL) == named_struct('field', NULL) as null_null_struct_eq,
       named_struct('field', NULL) == named_struct('field', 1) as null_not_null_struct_eq,
       named_struct('field', 1) == named_struct('field', 1) as not_null_not_null_struct_eq,
       named_struct('field', 2) == named_struct('field', 1) as not_null_not_null_struct_not_eq
```

Please, find the following screenshot:
image

Could you elaborate on what I missing? Aforementioned tests have passed on own Databricks account.

Thank you so much for your review.

mwojtyczka: you are right, no `eqNullSafe` seems to be needed, on structs spark treats equal NULL bitmaps at the same position as equal, so:
-   named_struct('f', NULL) = named_struct('f', NULL) → TRUE (not NULL)                                                            
-   named_struct('f', NULL) = named_struct('f', 1) → FALSE
-   NULL = NULL (scalar) → NULL (standard SQL 3VL)

tests/integration/test_apply_checks.py
```diff
@@ -1023,6 +1024,291 @@ def test_foreign_key_check_missing_ref_df_key(ws, spark):
...
+                ],
+            ],
+        ],
+        EXPECTED_SCHEMA,
+    )
+
+    assert_check_and_split_results(checked, good_df, bad_df, expected, ["a", "b", "c"], ignore_column_order=True)
+
+
+def test_foreign_key_check_null_safe_on_composite_keys(ws, spark):
```
mwojtyczka, line 1184: Can you add a partial-NULL row to both dataframes so it matches via struct equality?

```
src_df = spark.createDataFrame(                                      
      [                                                                                                                            
          [1, 2, 3],
          [4, 5, 6],                                                                                                               
          [None, None, 7],                                             
          [1, None, 8],    # <-- partial NULL, should match the corresponding ref row
      ],                                                                                                                           
      SCHEMA,
  )                                                                                                                                
  ref_df = spark.createDataFrame(                                      
      [                                                                                                                            
          [1, 2, 3],                                                   
          [None, None, 7],
          [1, None, 8],    # <-- partial NULL, matches src row 4
      ],                                                                                                                           
      "ref_a: int, ref_b: int, e: int",
  )                                                                                                                                
                  
```
Expected: row 4 passes (no violation)

tests/integration/test_apply_checks.py
```diff
@@ -1023,6 +1024,291 @@ def test_foreign_key_check_missing_ref_df_key(ws, spark):
...
+            [None, None, 7, None, None],
+        ],
+        EXPECTED_SCHEMA,
+    )
+
+    assert_check_and_split_results(checked, good_df, bad_df, expected, ["a", "b", "c"], ignore_column_order=True)
+
+
+def test_foreign_key_check_null_safe_on_composite_keys_not_found(ws, spark):
+    dq_engine = DQEngine(workspace_client=ws, extra_params=EXTRA_PARAMS)
```
mwojtyczka, line 1250: Can you pls extend the test to add a partial-NULL src row that has no matching ref row, to verify partial NULLs aren't over-matched?

```
src_df = spark.createDataFrame(                                      
      [
          [1, 2, 3],
          [None, None, 7],
          [1, None, 8],    # <-- partial NULL; ref has (1, 2, ...) but NOT (1, NULL, ...)
      ],                                                                                                                           
      SCHEMA,
  )                                                                                                                                
  ref_df = spark.createDataFrame(                                      
      [
          [1, 2, 3],
      ],
      "ref_a: int, ref_b: int, e: int",
  )
```

Expected: (1, NULL) violates (struct (1, NULL) not equal struct (1, 2)). Proves partial NULL doesn't collapse to wildcard match.

mwojtyczka (changes requested):
Generally looking good. Just small request to extend tests.

tests/integration/test_apply_checks.py
IvannKurchenko: Thank, this has been added.

tests/integration/test_apply_checks.py
IvannKurchenko: Thank, this has been added.

IvannKurchenko:
@mwojtyczka Thank you for the review and feedback. I have added the tests you requested, along with a case for negate enabled.

src/databricks/labs/dqx/check_funcs.py
```suggestion
    By default, NULL values in the foreign key columns are ignored (SQL ANSI behavior).                         
    When *null_safe=True*, NULL foreign-key values are matched against NULL reference values.
```

mwojtyczka, line 1424:
```suggestion

        # col_expr.isNotNull() only filters rows in the single-column non-null-safe path;                               
        # when col_expr is a struct (composite keys or null_safe=True), the struct is never NULL                        
        # so the guard is a no-op but kept for clarity
```

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1374,11 +1378,15 @@ def foreign_key(
     _validate_ref_params(columns, ref_columns, ref_df_name, ref_table)
 
     not_null_condition = F.lit(True)
-    if len(columns) == 1:
+    # Explicitly wrap single columns in a struct in case of null safety to distinguish matched `null` key
+    # and un-matched FK
+    if len(columns) == 1 and not null_safe:
```
mwojtyczka, line 1383: For single column, the message and rule name reads a bit weird now:
- the rule name is constructed from a struct, so instead of `a_not_exists_in_ref_b` → `struct_a_as_a_not_exists_in_ref_struct_b_as_a`
- the violation message instea of being `Value '...' in column 'a' not found in reference column 'b'` → `'struct(a AS a)' ... 'struct(b AS a)'`

From user perspective it is much cleaner if we get rid of the struct for single column comparison. This is a simple fix: 
```
if len(columns) == 1 and not null_safe:                                                                     
    join_col, join_ref_col = columns[0], ref_columns[0]                                                     
else:                                                                                                       
    join_col, join_ref_col, not_null_condition = _handle_fk_composite_keys(
        columns, ref_columns, not_null_condition, null_safe                                                 
    )
                                                                                                            
# Display: for single-column FK, render alias/message from the raw column so                                
# enabling null_safe does not change the rule name or violation message.
display_col = columns[0] if len(columns) == 1 else join_col                                                 
display_ref_col = ref_columns[0] if len(columns) == 1 else join_ref_col                                     

col_str_norm, col_expr_str, display_col_expr = get_normalized_column_and_expr(display_col)                  
ref_col_str_norm, ref_col_expr_str, _ = get_normalized_column_and_expr(display_ref_col)
_, _, col_expr = get_normalized_column_and_expr(join_col)                                                   
_, _, ref_col_expr = get_normalized_column_and_expr(join_ref_col)
```

Then when making the condition use the `display_col`:
```
condition = make_condition(                                                                                 
    condition=F.col(condition_col),
    message=F.concat_ws(
        "",
        F.lit("Value '"),
        display_col_expr.cast("string"),   # ← was col_expr
        F.lit("' in column '"),                                                                             
        F.lit(col_expr_str),
        F.lit(f"' {'' if negate else 'not '}found in reference column '"),                                  
        F.lit(ref_col_expr_str),
        F.lit("'"),                                                                                         
    ),                 
    alias=f"{col_str_norm}_{op_name}_ref_{ref_col_str_norm}",
) 
```

Then the new tests for single column needs to be updated.

@IvannKurchenko can you make this change? the rest looks good now

mwojtyczka: refactored

mwojtyczka (approved):
LGTM

src/databricks/labs/dqx/check_funcs.py
IvannKurchenko: @mwojtyczka Oh, sorry for not handling this in time. I very appreciate your fix.

mwojtyczka:
> @mwojtyczka I apologise for this, there was also merge commits from GitHub, now everything is singed.

No worries, thanks for your awesome contribution! will merge as soon as all tests pass

PR #1111, opened 2026-04-13
Refactor DQX App: add RBAC, API v1 routes, rule management, scheduling, and import functionality

berrybluecode:
## Changes

Major refactor of the DQX App backend and frontend, introducing role-based access control, a versioned API, and expanded rule management capabilities.

### Backend
- **API versioning**: Migrated all routes from `/api/` to `/api/v1/` with a modular route structure (`routes/v1/`) replacing the monolithic `router.py`
- **RBAC**: Added role-based access control (`UserRole.ADMIN`, `VIEWER`) with `require_role` dependency and a `RoleService` backed by Delta tables
- **Rule management**: New services for rule catalog (CRUD with history tracking), comments, metrics, AI-based rule generation, and dry-run validation
- **Scheduling**: Full schedule config service with per-schedule rows in Delta tables (replacing single-row JSON blob), history tracking, and an async scheduler service
- **Import rules**: Support for importing rules from YAML files and Delta tables
- **App settings**: Replaced file-based `ConfigSerializer` with `AppSettingsService` backed by Delta tables via SQL warehouse
- **Database migrations**: Added versioned migration framework for schema changes
- **Profiler & discovery**: New endpoints for data profiling and schema discovery
- **Dependencies**: Refactored FastAPI dependency injection (`dependencies.py`) with proper SP/OBO workspace client separation

### Frontend
- Rebuilt rule creation, editing, and approval workflows
- Added rule import UI (YAML file upload and Delta table browser)
- New scheduling/runs management section
- Role-based UI gating
- Updated all API calls to `/api/v1/`

### Tests
- Updated unit tests to match new service signatures, SQL call patterns (history recording, schema creation), and `ManyError` wrapping
- Updated integration tests: migrated URLs to `/api/v1/`, added dependency overrides for `get_sp_ws`, `get_user_role`, and `get_app_settings_service`
- Fixed integration test fixtures to use `serverless_clusters=True` for serverless-only workspaces
- Made `TEST_CATALOG` configurable via environment variable
- Fixed Pyright/Mypy static analysis errors (import cycles, missing type args, unsafe attribute access)

### Linked issues

Resolves #..

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests


app/src/databricks_labs_dqx_app/backend/dependencies.py
```diff
@@ -208,6 +248,53 @@ async def get_sql_connector(
...
+    role_svc: Annotated[RoleService, Depends(get_role_service)],
+) -> UserRole:
+    """Resolve the user's role based on Databricks group membership.
+
+    Resolution priority:
+    1. Bootstrap admin group from DQX_ADMIN_GROUP env var
+    2. Groups mapped to roles in dq_role_mappings table
+    3. Default to VIEWER if no mappings match
+    """
+    try:
```
mwojtyczka, line 268: **High — fail-open RBAC on error**

If the role service DB query fails (warehouse timeout, table missing, network blip), every user silently gets `VIEWER` access. This is a fail-open pattern.

For an internal app where VIEWER is read-only, this may be an acceptable trade-off (availability over security). But it should be an explicit, documented decision — not a silent `except Exception`. At minimum:

1. Add a comment explaining this is intentional fail-open-to-readonly
2. Consider emitting a metric/alert so ops knows RBAC is degraded
3. If any VIEWER-accessible endpoint can trigger side effects (job runs, exports), reconsider and fail closed (return 503)

app/src/databricks_labs_dqx_app/backend/routes/v1/metrics.py
```diff
@@ -0,0 +1,138 @@
...
+    limit: int = Query(50, ge=1, le=200),
+) -> list[MetricSnapshotOut]:
+    """Return quality metric snapshots for a specific table, ordered by time (newest first)."""
+    from databricks.sdk.service.sql import Disposition, Format, StatementState
+    from databricks_labs_dqx_app.backend.sql_utils import escape_sql_string
+
+    table = f"{app_conf.catalog}.{app_conf.schema_name}.dq_metrics"
+    et = escape_sql_string(table_fqn)
+
+    sql = (
```
mwojtyczka, line 35: **Medium — missing `validate_fqn()` before SQL interpolation**

`table_fqn` comes from a URL path parameter and is escaped with `escape_sql_string()` — that prevents SQL injection. But the project already has `validate_fqn()` in `sql_utils.py` for structural validation (must be `catalog.schema.table`). Other services like `rules_catalog_service.save()` call `validate_fqn()` before building SQL.

For defense-in-depth consistency, add:
```python
from databricks_labs_dqx_app.backend.sql_utils import escape_sql_string, validate_fqn

validate_fqn(table_fqn)
et = escape_sql_string(table_fqn)
```

app/src/databricks_labs_dqx_app/backend/routes/v1/import_rules.py
```diff
@@ -0,0 +1,126 @@
...
+    body: ImportFromTableIn,
+    obo_ws: Annotated[WorkspaceClient, Depends(get_obo_ws)],
+    svc: Annotated[RulesCatalogService, Depends(get_rules_catalog_service)],
+    validate_fn: Annotated[Callable[[list[Any]], ChecksValidationStatus], Depends(get_check_validator)],
+) -> ImportFromTableOut:
+    """Read rules from a Delta table (same schema as dq_quality_rules) and import them."""
+    user = obo_ws.current_user.me()
+    user_email = user.user_name or "unknown"
+
+    try:
```
mwojtyczka, line 95: **High — `source_table_fqn` from user input not validated before use**

`body.source_table_fqn` is passed directly to `svc.read_external_rules_table()` without calling `validate_fqn()`. If the service method builds SQL from this value, it could be a SQL injection vector. Even if it doesn't today, adding validation here is defense-in-depth:

```python
from databricks_labs_dqx_app.backend.sql_utils import validate_fqn

try:
    validate_fqn(body.source_table_fqn)
except ValueError as e:
    raise HTTPException(status_code=400, detail=str(e))

rows = svc.read_external_rules_table(body.source_table_fqn)
```

app/src/databricks_labs_dqx_app/backend/services/role_service.py
```diff
@@ -0,0 +1,202 @@
...
+            )
+            for row in rows
+        ]
+
+    def get_mappings_for_role(self, role: str) -> list[RoleMapping]:
+        """Get all group mappings for a specific role."""
+        escaped_role = role.replace("'", "''")
+        sql = (
+            f"SELECT role, group_name, created_by, "
+            f"CAST(created_at AS STRING), updated_by, CAST(updated_at AS STRING) "
```
mwojtyczka, line 87: **Low — inline `.replace("'", "''")` instead of `escape_sql_string()`**

The new `sql_utils.py` module was created specifically to centralize SQL escaping. But `role_service.py` (and `schedule_config_service.py`) use inline `.replace("'", "''")` throughout. While functionally identical today, centralizing prevents divergence if `escape_sql_string()` ever needs to handle additional characters (e.g., backslash normalization).

Suggestion: import and use `escape_sql_string()` from `sql_utils.py` here and in `schedule_config_service.py`.

app/src/databricks_labs_dqx_app/backend/services/role_service.py
```diff
@@ -0,0 +1,202 @@
...
+            The highest priority role the user has, or VIEWER if no mappings match.
+        """
+        if admin_group and admin_group in user_groups:
+            logger.debug(f"User in bootstrap admin group '{admin_group}', granting ADMIN")
+            return UserRole.ADMIN
+
+        mappings = self.list_mappings()
+        if not mappings:
+            logger.debug("No role mappings configured, defaulting to VIEWER")
+            return UserRole.VIEWER
```
mwojtyczka, line 163: **Medium — `resolve_role()` does a full table scan per request**

`resolve_role()` calls `self.list_mappings()` which runs `SELECT * FROM dq_role_mappings` on every HTTP request that needs authorization. For a small mappings table this is fine, but it means every authenticated API call hits the SQL warehouse.

Consider caching with a short TTL (e.g., 2–5 minutes), similar to how `DiscoveryService` caches results. The `app_cache` is already available in the codebase.

app/src/databricks_labs_dqx_app/backend/routes/v1/roles.py
```diff
@@ -0,0 +1,106 @@
...
+        created_at=mapping.created_at.isoformat() if mapping.created_at else None,
+        updated_by=mapping.updated_by,
+        updated_at=mapping.updated_at.isoformat() if mapping.updated_at else None,
+    )
+
+
+@router.get("", response_model=list[RoleMappingOut], operation_id="listRoleMappings")
+def list_role_mappings(
+    svc: Annotated[RoleService, Depends(get_role_service)],
+) -> list[RoleMappingOut]:
```
mwojtyczka, line 43: **Medium — `group_name` from URL path in DELETE goes to SQL without structural validation**

The `group_name` path parameter flows to `svc.delete_mapping(role, group_name)` which escapes it with `.replace("'", "''")` — so no injection. But unlike `create_mapping` which validates `role` against the `UserRole` enum, `delete_mapping` doesn't validate either parameter's format.

Since this is admin-only, the risk is low. But for consistency, consider validating the role against `UserRole` values here too (the service already does this in `create_mapping`).

mwojtyczka (commented):
## Review: Refactor DQX App

This is a large, ambitious PR (+26k/-7k across ~100 files). The overall architecture is solid — API versioning, RBAC, modular route structure, and Delta-backed services are all good directions. Comments below focus on security, fail-open behavior, and test coverage gaps.

**Summary:**

| Severity | Count | Areas |
|---|---|---|
| **High** | 2 | Fail-open RBAC on error; missing FQN validation on import |
| **Medium** | 3 | Inconsistent SQL escaping; no caching on role resolution; test coverage gaps |
| **Low** | 1 | Inline `.replace` vs centralized `escape_sql_string()` |

mwojtyczka (changes requested):
generally looking good, left some critical issues to address

app/src/databricks_labs_dqx_app/backend/dependencies.py
berrybluecode: Changed except Exception: return UserRole.VIEWER to raise HTTPException(503) so role resolution failures fail closed instead of silently granting access.

app/src/databricks_labs_dqx_app/backend/routes/v1/metrics.py
berrybluecode: Added validate_fqn(table_fqn) before escape_sql_string() for defense-in-depth structural validation on the URL path parameter.

app/src/databricks_labs_dqx_app/backend/routes/v1/import_rules.py
berrybluecode: Added validate_fqn(body.source_table_fqn) with a 400 response before passing user input to read_external_rules_table().

app/src/databricks_labs_dqx_app/backend/services/role_service.py
berrybluecode: Replaced all .replace("'", "''") calls with the centralized escape_sql_string() from sql_utils.py (5 occurrences in role_service.py, 1 in schedule_config_service.py).

app/src/databricks_labs_dqx_app/backend/services/role_service.py
berrybluecode: Added a 2-minute in-memory cache on list_mappings(). resolve_role() now calls list_mappings(use_cache=True). The cache is automatically invalidated on create_mapping() and delete_mapping().

app/src/databricks_labs_dqx_app/backend/routes/v1/roles.py
berrybluecode: Added UserRole enum validation on the role path parameter in delete_role_mapping(), returning 400 for invalid roles (matching what create_mapping already does).

berrybluecode:
Addressed the reviewed issued. please re-review.

PR #1115, opened 2026-04-17
fix: add py.typed marker (PEP 561)

### LINKED ISSUE #1114: [BUG]: py.typed missing

### Current Behavior

The `py.typed` marker to indicate typed Python code is missing from the package, which results, for example, in type checking errors with tools like mypy.

### Expected Behavior

Since the package seems to be typed, I would expect the `py.typed` marker file to be included in the package.

### Steps To Reproduce

1. Run the latest version of mypy on a Python file importing the Databricks DQX package:
  ```python

  from databricks.labs.dqx.check_funcs import make_condition
  from databricks.labs.dqx.rule import register_rule
  from pyspark.sql import Column


  @register_rule("row")
  def is_not_null(column: str) -> Column:
      return make_condition(
          F.col(column).isNull(),
          f"Column '{column}' is null",
          f"is_not_null_{column}",
      )
  ```
2. See the MyPy error: 
  ```
  dqx_mypy.py:10: error: Returning Any from function declared to return "Column"  [no-any-return]
          return make_condition(
          ^~~~~~~~~~~~~~~~~~~~~~
  Found 1 error in 1 file (checked 1 source file)
   ```

### Cloud

AWS

### Operating System

Linux

```shell

```

balgaly:
## Changes

The package ships with type annotations but is missing the `py.typed` marker file. Without this marker, downstream tools like mypy fall back to treating all imports from `databricks.labs.dqx` as untyped (per [PEP 561](https://peps.python.org/pep-0561/)), which is exactly what was reported in the linked issue.

This PR adds an empty `src/databricks/labs/dqx/py.typed` file so consumers can type-check against the package's existing annotations.

### Linked issues

Resolves #1114

### Tests

- [x] manually tested

Verified the marker lands in the built wheel:

```
$ python -m pip wheel --no-deps -w dist .
$ python -m zipfile -l dist/databricks_labs_dqx-0.13.0-py3-none-any.whl | grep py.typed
databricks/labs/dqx/py.typed                   2020-02-02 00:00:00            0
```

Verified mypy now recognizes the package as typed (previously it emitted the "module is installed, but missing library stubs or py.typed marker" message the issue reporter saw):

```
$ python -m mypy consumer.py
Success: no issues found in 1 source file
```

Hatchling's default build already includes the file (the existing `[tool.hatch.build]` uses `include = ["src", ...]`), so no build config changes are needed.


mwojtyczka (approved):
LGTM

PR #1118, opened 2026-04-21
feat: add has_no_aggr_outliers stateless rolling-window sigma outlier check

vpottam-nvidia:
Adds `has_no_aggr_outliers`, a new dataset-level quality check that detects outliers in time-series aggregates   using a stateless rolling-window sigma method.

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [x] added performance tests

### Documentation and Demos

- [x] added/updated docs
- [ ] added/updated demos


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Version changelog
 
+## [Unreleased]
+
+* Added `is_aggr_not_anomalous` dataset-level check for stateless, rolling-window sigma anomaly detection on time-series aggregates. The check computes an aggregate (count, sum, avg, etc.) per time-grain bucket, derives a rolling mean ± N·stddev baseline over a configurable lookback window, and fires when today's value deviates beyond the band. Unlike `has_no_row_anomalies` it requires no MLflow model or training artefacts — logic is pure PySpark and resolves entirely at query time. Supports `group_by` for per-segment bands, `row_filter` for pre-aggregation filtering, configurable `time_truncation` (day/hour/week/month), and warmup suppression for sparse history. Designed as the stateless counterpart to `is_aggr_not_less_than`/`is_aggr_not_greater_than` for dynamic metric-anomaly patterns previously requiring hand-written rolling-window CTEs.
```
mwojtyczka: Please remove this CHANGELOG entry (and the `## [Unreleased]` header above). The changelog is generated automatically during release from PR titles/labels, so manual entries will be overwritten and tend to cause conflicts.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1770,6 +1770,227 @@ def is_aggr_not_equal(
...
+
+
+@register_rule("dataset")
+def is_aggr_not_anomalous(
+    column: str | Column,
+    time_column: str,
+    *,
+    aggr_type: str = "avg",
+    sigma: float = 3.0,
+    lookback_days: int = 14,
```
mwojtyczka: The parameter names `lookback_days` and `warmup_days` are misleading when `time_truncation` is not `"day"` — e.g. with `time_truncation="hour"`, `lookback_days=14` actually means 14 hours. Your own integration test at `test_hour_truncation` even hints at this: `# "lookback_days" works as bucket count regardless of truncation`.

Please rename to something granularity-agnostic (e.g. `lookback_periods` / `warmup_periods`, or `lookback_buckets` / `warmup_buckets`) before this ships. Renaming after release would be a breaking change.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1770,6 +1770,227 @@ def is_aggr_not_equal(
     )
 
 
+_VALID_TIME_TRUNCATIONS: frozenset[str] = frozenset({"day", "hour", "week", "month"})
+
+
+@register_rule("dataset")
+def is_aggr_not_anomalous(
```
mwojtyczka: The YAML example in `docs/dqx/docs/reference/quality_checks.mdx` shows `column: '*'` with `aggr_type: count`, but no unit or integration test exercises that path. Please add explicit coverage (and consider whether you want to allow `column=None` for count, like some internal helpers do — otherwise document that `'*'` is the way to express `count(*)`).

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1770,6 +1770,227 @@ def is_aggr_not_equal(
...
+        the rolling mean/stddev baseline, evaluates the sigma band condition, and
+        adds the condition + auxiliary metric columns to every row of *df*.
+
+        Args:
+            df: The input DataFrame to validate.
+
+        Returns:
+            The DataFrame with additional condition and metric columns appended.
+        """
+        src = df.filter(F.expr(row_filter)) if row_filter else df
```
mwojtyczka: This `row_filter` semantics differs from the sibling `is_aggr_*` checks: `_is_aggr_compare` (around L3072-3073 on `main`) uses `F.when(filter_col, aggr_col_expr)` so filtered-out rows contribute NULL (skipped by the aggregate) rather than being dropped entirely. For `count(*)` the two approaches give different results.

Please either align with the existing pattern (so behavior across all `is_aggr_*` checks is consistent) or add a comment explaining why this deviates.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1770,6 +1770,227 @@ def is_aggr_not_equal(
...
+            F.when(insufficient_history, F.lit(False))
+            .when(F.col("__dq_sigma").isNull() | (F.col("__dq_sigma") == 0), F.lit(False))
+            .when(F.col("__dq_today").isNull(), F.lit(False))
+            .otherwise(delta_expr > F.lit(sigma) * F.col("__dq_sigma"))
+        )
+
+        result = (
+            joined.withColumn(condition_col, violation)
+            .withColumn(today_col, F.col("__dq_today"))
+            .withColumn(mu_col, F.col("__dq_mu"))
```
mwojtyczka: The final `apply()` appends 5 `__dq_anomaly_*` columns (`condition`, `today`, `mu`, `sigma`, `delta`) to every row of the returned DataFrame. Only `condition_col` is needed by `make_condition` directly — the other 4 are referenced by `F.col(...)` inside the `concat_ws` message.

Consider materializing the full message into a single `__dq_anomaly_msg_*` column inside `apply` and referencing only that plus `condition_col` from `make_condition`. This would match the 2-column footprint of `_is_aggr_compare`.

tests/integration/test_has_no_aggr_outliers.py
```diff
@@ -0,0 +1,407 @@
...
+    # Today's bucket: value 50 (> 3σ above baseline of 10)
+    today = BASE_DATE + timedelta(days=14)
+    rows.append((today, 50.0))
+
+    df = spark.createDataFrame(rows, "event_date: date, metric: double")
+
+    condition, apply_fn = is_aggr_not_anomalous(
+        "metric",
+        "event_date",
+        aggr_type="avg",
```
mwojtyczka, line 91: All 10 integration tests use `aggr_type="avg"`. Please add coverage for:

- `count` (or `count` with `column='*'`) — the most common use case in the YAML examples
- `sum` with `group_by`
- `aggr_params` (e.g. `percentile` with `aggr_params={"percentile": 0.95}`)
- The `warmup_days == lookback_days` boundary

mwojtyczka (commented):
Thanks for the contribution @vpottam-nvidia — this is a nicely scoped, stateless counterpart to the existing `is_aggr_*` checks and it's great to see it paired with both unit and integration tests.

A couple of housekeeping items:
1. **Please sign your commits** — DQX requires signed commits per the [First Contribution guide](https://databrickslabs.github.io/dqx/docs/dev/contributing/#first-contribution). You'll need to enable `git config --global commit.gpgsign true` (or equivalent) and force-push the re-signed history.
2. **Please remove the CHANGELOG.md entry** — the changelog is generated automatically during release from PR titles/labels, so manual entries create merge conflicts and get regenerated anyway.
3. **Please address the inline review comments**

mwojtyczka (changes requested):

src/databricks/labs/dqx/check_funcs.py
ghanse: Maybe something like this?

```
aggr_type: str = "avg",
sigma: float = 3.0,
time_interval: str = "day",
lookback_num_intervals: int = 14,
warmup_num_intervals: int = 7,
...
```

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1770,6 +1770,227 @@ def is_aggr_not_equal(
...
+            The DataFrame with additional condition and metric columns appended.
+        """
+        src = df.filter(F.expr(row_filter)) if row_filter else df
+        aggr_expr = _build_aggregate_expression(aggr_type, aggr_col_expr, aggr_params)
+
+        group_cols = [F.col(c) if isinstance(c, str) else c for c in (group_by or [])]
+        grain_col = F.date_trunc(time_truncation, F.col(time_column)).alias("__dq_grain")
+
+        # Step 1: aggregate per time-grain (and group)
+        daily = src.groupBy(*group_cols, grain_col).agg(aggr_expr.alias("__dq_metric"))
```
ghanse: I would rename this variable to something like `aggregate_df` or similar.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1770,6 +1770,227 @@ def is_aggr_not_equal(
...
+            "",
+            F.lit(f"{aggr_type}({aggr_col_str}) today="),
+            F.col(today_col).cast("string"),
+            F.lit(" baseline="),
+            F.col(mu_col).cast("string"),
+            F.lit(" sigma="),
+            F.col(sigma_col).cast("string"),
+            F.lit(" |Δ|="),
+            F.col(delta_col).cast("string"),
+            F.lit(f" > {sigma}σ (lookback={lookback_days}d)"),
```
ghanse: Not sure we should use symbols by default. Can we rewrite this message expression to use alphanumeric characters only?

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1770,6 +1770,227 @@ def is_aggr_not_equal(
...
+    sigma: float = 3.0,
+    lookback_days: int = 14,
+    warmup_days: int = 7,
+    time_truncation: str = "day",
+    group_by: list[str | Column] | None = None,
+    row_filter: str | None = None,
+    aggr_params: dict[str, Any] | None = None,
+) -> tuple[Column, Callable]:
+    """
+    Build a stateless, rolling-window sigma anomaly check for a time-series aggregate.
```
ghanse: I would re-phrase this to something like:

```
Builds a stateless, rolling-window anomaly check for time-series aggregates. Uses mean and standard deviation to identify anomalies.
```

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1770,6 +1770,227 @@ def is_aggr_not_equal(
...
+    2. Derives a rolling baseline (mean and stddev_pop) over the preceding
+       ``lookback_days`` buckets.
+    3. Passes silently when fewer than ``warmup_days`` historical buckets are
+       available, the series is constant (stddev == 0), or today's bucket is
+       missing.
+    4. Fails when ``|today_metric − baseline| > sigma * sigma_val``.
+
+    The check is **stateless** — no MLflow registry, no trained artefacts. It
+    sits alongside the other ``is_aggr_*`` primitives, not the ML-based anomaly
+    module.
```
ghanse: This can be removed IMO

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1770,6 +1770,227 @@ def is_aggr_not_equal(
     )
 
 
+_VALID_TIME_TRUNCATIONS: frozenset[str] = frozenset({"day", "hour", "week", "month"})
```
ghanse: Should we allow minute as well? I think this would be useful.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1770,6 +1770,227 @@ def is_aggr_not_equal(
...
+            join_keys = [c if isinstance(c, str) else get_column_name_or_alias(c) for c in group_by]
+            joined = today.join(stats, on=join_keys, how="left")
+        else:
+            # Use a dummy key left-join so the today row survives even when stats is empty
+            # (e.g. only 1 day of data → hist is empty → stats has 0 rows).
+            joined = today.withColumn("__dq_jkey", F.lit(1)).join(
+                stats.withColumn("__dq_jkey", F.lit(1)),
+                on="__dq_jkey",
+                how="left",
+            ).drop("__dq_jkey")
```
ghanse: Can we use `crossJoin` here instead? We will only have 1 row in this case.

https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.crossJoin.html

ghanse (dismissed):
Thanks for submitting! Requested a few changes.

src/databricks/labs/dqx/check_funcs.py
vpottam-nvidia: done in the latest push - the params are now time_interval, lookback_num_intervals, and warmup_num_intervals, as @ghanse  suggested. The integration test comment has been updated accordingly as well.

src/databricks/labs/dqx/check_funcs.py
vpottam-nvidia: row_filter now uses F.when(filter_col, aggr_col_expr) matching the _is_aggr_compare pattern

src/databricks/labs/dqx/check_funcs.py
vpottam-nvidia: daily renamed to aggregate_df, today → current

src/databricks/labs/dqx/check_funcs.py
vpottam-nvidia: This is not correct. We should not use symbols for sure. Unicode Δ and σ removed; message now says delta= and exceeds N x stddev

src/databricks/labs/dqx/check_funcs.py
vpottam-nvidia: added

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1770,6 +1770,220 @@ def is_aggr_not_equal(
     )
 
 
+_VALID_TIME_TRUNCATIONS: frozenset[str] = frozenset({"minute", "hour", "day", "week", "month"})
+
+
+@register_rule("dataset")
+def is_aggr_not_anomalous(
```
```suggestion
def has_no_aggr_outliers(
```
Please rename it, this is more an outlier detection rather than anomaly detection. Also we have anomaly detection already so calling it the same would be confusing. We have an existing row-level check called `has_no_outliers` so `has_no_aggr_outliers` would align with the existing naming conventions

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1770,6 +1770,220 @@ def is_aggr_not_equal(
...
+    sigma: float = 3.0,
+    lookback_num_intervals: int = 14,
+    warmup_num_intervals: int = 7,
+    time_interval: str = "day",
+    group_by: list[str | Column] | None = None,
+    row_filter: str | None = None,
+    aggr_params: dict[str, Any] | None = None,
+) -> tuple[Column, Callable]:
+    """
+    Rolling-window sigma anomaly check for a time-series aggregate.
```
mwojtyczka: Please use `outlier` instead of `anomaly`. Needs to be replaced in all places
```suggestion
    Rolling-window sigma outlier check for a time-series aggregate.
```

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1770,6 +1770,220 @@ def is_aggr_not_equal(
...
+        # Step 7: broadcast result back to every row in df
+        select_cols = [condition_col, msg_col]
+        if group_by:
+            join_keys = [c if isinstance(c, str) else get_column_name_or_alias(c) for c in group_by]
+            return df.join(result.select(*join_keys, *select_cols), on=join_keys, how="left")
+        return df.crossJoin(result.select(*select_cols))
+
+    # Build alias
+    group_by_str = (
+        "_".join(c if isinstance(c, str) else get_column_name_or_alias(c) for c in group_by) if group_by else None
```
```suggestion
        "_".join(c if isinstance(c, str) else get_column_name_or_alias(c, normalize=True) for c in group_by) if group_by else None
```
Please normalize, otherwise it can produce invalid name like `metric_avg_anomaly_group_by_(a + b)`

tests/integration/test_is_aggr_not_anomalous.py
```diff
@@ -0,0 +1,571 @@
+"""
+Integration tests for is_aggr_not_anomalous.
```
mwojtyczka: Can you please also add 1 test that is using `apply_checks_by_metadata`. The easiest is to add it here: https://github.com/databrickslabs/dqx/blob/2789070dec44f8af3ca52bb3d4252759da76eef8/tests/resources/all_dataset_checks.yaml

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1770,6 +1770,220 @@ def is_aggr_not_equal(
...
+
+        Returns:
+            The DataFrame with additional condition and message columns appended.
+        """
+        filter_col = F.expr(row_filter) if row_filter else F.lit(True)
+        filtered_expr = F.when(filter_col, aggr_col_expr) if row_filter else aggr_col_expr
+        aggr_expr = _build_aggregate_expression(aggr_type, filtered_expr, aggr_params)
+
+        group_cols = [F.col(c) if isinstance(c, str) else c for c in (group_by or [])]
+        grain_col = F.date_trunc(time_interval, F.col(time_column)).alias("__dq_grain")
```
mwojtyczka, line 1918: Please also add validation of time_column similar to has_no_outliers check function. If time_column isn't a date/timestamp, the F.date_trunc fails with a cryptic Spark error at query time.

mwojtyczka (changes requested):

docs/dqx/docs/reference/quality_checks.mdx
```diff
@@ -1808,6 +1808,7 @@ You can also define your own custom dataset-level checks (see [Creating custom c
...
...sing. Complements `is_aggr_not_less_than` / `is_aggr_not_greater_than` for dynamic, data-driven anomaly detection on daily/hourly metrics. | `column`: column name (str) or Column expression to aggregate, e.g. `"revenue"` or `F.col("a") - F.col("b")`; `time_column`: name of the timestamp/date column used to bucket rows into time grains; `aggr_type`: aggregation type applied per bucket (default: `"avg"`); `sigma`: number of standard deviations defining the anomaly band (default: `3.0`, must be > 0); `lookback_num_intervals`: number of preceding buckets used to build the rolling baseline (default: `14`, must be >= 2); `warmup_num_intervals`: minimum historical buckets required before the check fires (default: `7`, must satisfy `1 <= warmup_num_intervals <= lookback_num_intervals`); `time_interval`: granularity for bucketing the `time_column` — one of `"minute"`, `"hour"`, `"day"`, `"week"`, `"month"` (default: `"day"`); `group_by`: (optional) list of columns or column expressions to segment the anomaly band per group; `row_filter`: (optional) SQL expression to filter rows before aggregation; `aggr_params`: (optional) dict of extra parameters for aggregate functions that require them |
```
```suggestion
| `is_aggr_not_anomalous`            | Rolling-window sigma outlier check for a time-series aggregate. Fires when the current bucket's aggregate deviates by more than N standard deviations from the rolling mean of the preceding `lookback_num_intervals` buckets. Passes silently during warmup (fewer than `warmup_num_intervals` historical buckets), when the series is constant (stddev == 0), or when the current bucket is missing. Complements `is_aggr_not_less_than` / `is_aggr_not_greater_than` for dynamic, data-driven outlier detection on daily/hourly metrics. | `column`: column name (str) or Column expression to aggregate, e.g. `"revenue"` or `F.col("a") - F.col("b")`; `time_column`: name of the timestamp/date column used to bucket rows into time grains; `aggr_type`: aggregation type applied per bucket (default: `"avg"`); `sigma`: number of standard deviations defining the outlier band (default: `3.0`, must be > 0); `lookback_num_intervals`: number of preceding buckets used to build the rolling baseline (default: `14`, must be >= 2); `warmup_num_intervals`: minimum historical buckets required before the check fires (default: `7`, must satisfy `1 <= warmup_num_intervals <= lookback_num_intervals`); `time_interval`: granularity for bucketing the `time_column` — one of `"minute"`, `"hour"`, `"day"`, `"week"`, `"month"` (default: `"day"`); `group_by`: (optional) list of columns or column expressions to segment the outlier band per group; `row_filter`: (optional) SQL expression to filter rows before aggregation; `aggr_params`: (optional) dict of extra parameters for aggregate functions that require them |
```

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1770,6 +1770,220 @@ def is_aggr_not_equal(
...
+
+    Args:
+        column: Column name (str) or Column expression to aggregate (e.g. ``"revenue"``
+            or ``F.col("a") - F.col("b")``). Pass ``"*"`` for ``count(*)``.
+        time_column: Name of the timestamp/date column used to bucket rows into
+            time grains.
+        aggr_type: Aggregation type applied per bucket (default: ``"avg"``). All
+            curated DQX aggregate types are supported (count, sum, avg, min, max,
+            count_distinct, stddev, percentile, etc.).
+        sigma: Number of standard deviations that defines the anomaly band
```
```suggestion
        sigma: Number of standard deviations that defines the outlier band
```

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1770,6 +1770,220 @@ def is_aggr_not_equal(
...
+        lookback_num_intervals: Number of preceding time-grain buckets used to
+            build the rolling baseline (default: ``14``). Must be >= 2.
+        warmup_num_intervals: Minimum number of historical buckets required before
+            the check fires (default: ``7``). Must satisfy
+            ``1 <= warmup_num_intervals <= lookback_num_intervals``.
+        time_interval: Granularity at which to bucket the ``time_column``
+            (default: ``"day"``). One of ``"minute"``, ``"hour"``, ``"day"``,
+            ``"week"``, ``"month"``.
+        group_by: Optional list of column names or Column expressions to segment
+            the anomaly band (e.g. ``["csp", "region"]``). The check fires if
```
```suggestion
            the outlier band (e.g. ``["csp", "region"]``). The check fires if
```

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1770,6 +1770,220 @@ def is_aggr_not_equal(
...
+            *any* group exceeds its own band.
+        row_filter: Optional SQL expression to filter rows before aggregation
+            (e.g. ``"status = 'Active'"``). Auto-injected from the check filter.
+        aggr_params: Optional dict of extra parameters for aggregate functions
+            that require them (e.g. ``{"percentile": 0.95}`` for percentile).
+
+    Returns:
+        A tuple of:
+            - A Spark Column representing the anomaly condition (string message on
```
```suggestion
            - A Spark Column representing the outlier condition (string message on
```

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1770,6 +1770,220 @@ def is_aggr_not_equal(
...
+            join_keys = [c if isinstance(c, str) else get_column_name_or_alias(c) for c in group_by]
+            return df.join(result.select(*join_keys, *select_cols), on=join_keys, how="left")
+        return df.crossJoin(result.select(*select_cols))
+
+    # Build alias
+    group_by_str = (
+        "_".join(c if isinstance(c, str) else get_column_name_or_alias(c) for c in group_by) if group_by else None
+    )
+    alias = (
+        f"{aggr_col_str_norm}_{aggr_type}_anomaly_group_by_{group_by_str}".lstrip("_")
```
```suggestion
        f"{aggr_col_str_norm}_{aggr_type}_has_outliers_group_by_{group_by_str}".lstrip("_")
```

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1770,6 +1770,220 @@ def is_aggr_not_equal(
...
+        if group_by_str
+        else f"{aggr_col_str_norm}_{aggr_type}_anomaly".lstrip("_")
```
```suggestion
        else f"{aggr_col_str_norm}_{aggr_type}_has_outliers".lstrip("_")
```

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: Can you pls also add a perf test?

docs/dqx/docs/reference/quality_checks.mdx
```diff
@@ -1812,13 +1812,14 @@ You can also define your own custom dataset-level checks (see [Creating custom c
...
...sing. Complements `is_aggr_not_less_than` / `is_aggr_not_greater_than` for dynamic, data-driven outlier detection on daily/hourly metrics. | `column`: column name (str) or Column expression to aggregate, e.g. `"revenue"` or `F.col("a") - F.col("b")`; `time_column`: name of the timestamp/date column used to bucket rows into time grains; `aggr_type`: aggregation type applied per bucket (default: `"avg"`); `sigma`: number of standard deviations defining the outlier band (default: `3.0`, must be > 0); `lookback_num_intervals`: number of preceding buckets used to build the rolling baseline (default: `14`, must be >= 2); `warmup_num_intervals`: minimum historical buckets required before the check fires (default: `7`, must satisfy `1 <= warmup_num_intervals <= lookback_num_intervals`); `time_interval`: granularity for bucketing the `time_column` — one of `"minute"`, `"hour"`, `"day"`, `"week"`, `"month"` (default: `"day"`); `group_by`: (optional) list of columns or column expressions to segment the outlier band per group; `row_filter`: (optional) SQL expression to filter rows before aggregation; `aggr_params`: (optional) dict of extra parameters for aggregate functions that require them |
```
```suggestion
| `has_no_aggr_outliers`            | Rolling-window sigma outlier check for a time-series aggregate. Fires when the current bucket's aggregate deviates by more than N standard deviations from the rolling mean of the preceding `lookback_num_intervals` buckets. All logic is pure PySpark. Passes silently during warmup (fewer than `warmup_num_intervals` historical buckets), when the series is constant (stddev == 0), or when the current bucket is missing. Complements `is_aggr_not_less_than` / `is_aggr_not_greater_than` for dynamic, data-driven outlier detection on daily/hourly metrics. | `column`: column name (str) or Column expression to aggregate, e.g. `"revenue"` or `F.col("a") - F.col("b")`; `time_column`: name of the timestamp/date column used to bucket rows into time grains; `aggr_type`: aggregation type applied per bucket (default: `"avg"`); `sigma`: number of standard deviations defining the outlier band (default: `3.0`, must be > 0); `lookback_num_intervals`: number of preceding buckets used to build the rolling baseline (default: `14`, must be >= 2); `warmup_num_intervals`: minimum historical buckets required before the check fires (default: `7`, must satisfy `1 <= warmup_num_intervals <= lookback_num_intervals`); `time_interval`: granularity for bucketing the `time_column` — one of `"minute"`, `"hour"`, `"day"`, `"week"`, `"month"` (default: `"day"`); `group_by`: (optional) list of columns or column expressions to segment the outlier band per group; `aggr_params`: (optional) dict of extra parameters for aggregate functions that require them |
```

docs/dqx/docs/reference/quality_checks.mdx
```diff
@@ -2046,6 +2047,46 @@ Complex data types are supported as well.
...
+    function: has_no_aggr_outliers
+    arguments:
+      column: latency_ms
+      time_column: request_time
+      aggr_type: avg
+      sigma: 2.5
+      lookback_num_intervals: 48    # 48 hourly buckets ≈ 2 days of history
+      warmup_num_intervals: 12
+      time_interval: hour           # bucket by hour instead of day
+      row_filter: "status = 'SUCCESS'"
```
```suggestion
```

mwojtyczka (approved):
LGTM

PR #1124, opened 2026-04-24
Fix save results in table when `DQObserver` is present

### LINKED ISSUE #1113: [BUG]: save_results_in_table depends on metrics_config when not saving metrics

### Current Behavior

Given the following:
1. the `DQEngine` was created with a valid `observer`
2. DQX was installed as a library without creating default configuration files on the system

When attempting to save valid and/or quarantined results via `DQEngine.save_results_in_table`, without providing an observation to save (or the relevant configuration), a `databricks.sdk.errors.platform.ResourceDoesNotExist` is raised since `~/.dqx/config.yml` does not exist.

### Expected Behavior

I should be able to use `save_results_in_table` when all relevant configuration is provided within the function call, e.g. `quarantine_df` and `quarantine_config`, whether an `observer` is configured for the engine or otherwise.

### Steps To Reproduce

1. Install DQX "as a Library" using the steps in the documentation
2. Create a `DQEngine` with an `observer` configured
3. Apply some checks and get the results (I used Databricks Connect)
4. Attempt to save the results with `save_results_in_table`

```python
from databricks.labs.dqx.engine import DQEngine, DQMetricsObserver, OutputConfig

def test_dqx_save_results_in_table_without_metrics(spark: SparkSession):
	engine = DQEngine(WorkspaceClient(), observer=DQMetricsObserver())

	data = [
		(1, "Person1", 25),
		(2, "Person2", -5),
		(3, "Person3", 40),
	]
	df = spark.createDataFrame(data, ["id", "name", "age"])

	rules: list[DQRule] = [
		DQRowRule(name="age_positive", column="age", check_func=check_funcs.is_not_less_than, check_func_args=[0]),
	]

	results = engine.apply_checks_and_split(df, rules)

	if len(results) != 3:
		raise ValueError(f"Expected 3 results, got {len(results)}")
	
	valid_df, quarantined_df, observation = results

	engine.save_results_in_table(quarantine_df=quarantined_df, quarantine_config=OutputConfig(location="my_catalog.my_schema.my_table"))
```

When an `observer` is not configured for the engine, the code above works. This is unexpected, since I'm not attempting to save any metrics.

### Cloud

AWS

### Operating System

macOS

```shell
.venv/lib/python3.12/site-packages/databricks/labs/dqx/engine.py:1011: in save_results_in_table
    run_config = self._config_serializer.load_run_config(
.venv/lib/python3.12/site-packages/databricks/labs/dqx/config_serializer.py:68: in load_run_config
    config = self.load_config(product_name, assume_user, install_folder)

databricks.sdk.errors.platform.ResourceDoesNotExist: Path (/Users/<user>/.dqx/config.yml) doesn't exist.
```

ghanse:
## Changes

This PR addresses an issue with `save_results_in_table` when the calling `DQEngine` has an associated observer but no observation or metrics configuration are passed.

I have also updated the asset bundle to use the [direct deployment engine](https://docs.databricks.com/aws/en/dev-tools/bundles/direct).

### Linked issues

Resolves #1113 

### Tests

- [x] manually tested
- [ ] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

### Documentation and Demos

- [x] added/updated demos
- [ ] added/updated docs


demos/dqx_demo_asset_bundle/databricks.yml
```diff
@@ -1,5 +1,6 @@
 bundle:
   name: dqx_demo_bundle
+  engine: direct
```
mwojtyczka: **Unrelated scope?** This line isn't mentioned in the PR description, and it isn't obviously related to the `save_results_in_table` observer fix. `engine: direct` selects the DAB CLI's direct deployment engine, which bypasses Terraform entirely. That's a significant behavior change for the demo — it also happens to dodge the `terraform init`/HashiCorp-registry flake we've been hitting in CI, which is a welcome side effect.

Could you either:
- Split this into its own commit/PR with a description of why direct mode is preferred here (and what happens to existing deployments of this bundle using the default engine), or
- Add a line to the PR body explaining that the switch is intentional and what it changes for the demo?

If this is specifically to make the DAB demo CI test stop flaking on terraform registry connectivity, that's worth calling out — it's useful to know independent of the observer fix.

mwojtyczka (approved):
LGTM

Issue #1030, opened 2026-02-12
[FEATURE]: AI Assisted Tuning of ML-based Anomaly Detection DQ Checks


vb-dbrks:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Problem statement

Related to #957 

Today users must manually tune anomaly settings like threshold, expected_anomaly_rate, segment_by, feature columns, and sampling/training params. This is hard to do consistently across datasets, and teams might end up with either too many false positives or missed anomalies.

### Proposed Solution

**Add an AI-assisted tuning mode that can analyze historical scoring outcomes and recommend tuned configurations.**

Core capabilities:
1. Suggest optimal threshold and expected_anomaly_rate based on desired precision/recall target.
2. Recommend feature/segment choices (columns, segment_by) using profile metadata and drift patterns.
3. Provide “safe” presets (high_recall, balanced, low_noise) and explain tradeoffs.
4. Output a ready-to-apply config snippet (Python and YAML metadata format).

### Additional Context

This should be an optional assistant layer on top of existing APIs (no breaking changes). A first version can be recommendation-only, with no automatic retraining unless explicitly enabled.

PR #1129, opened 2026-04-28
feat(anomaly): Add LLM-generated AI explanations for row anomaly detection

### LINKED ISSUE #1031: [FEATURE]: AI Explained Anomalies for has_no_anomalies Results

### Problem statement

related to #957 
Current anomaly output is technically rich (score, severity_percentile, contributions), but not easy for non-ML users to interpret quickly. Teams need faster root-cause triage and analyst-friendly explanations.

### Proposed Solution

**Add optional AI-generated explanations per anomalous row/group using existing metadata:**

1. Convert contribution maps into short human-readable “why flagged” narratives.
2. Highlight top changed features vs baseline and likely business impact.
3. Group similar anomalies into “patterns” with summary labels.
4. Include confidence/uncertainty indicators and recommended follow-up actions.

### Additional Context

This should respect data governance controls (mask sensitive values, configurable redaction). Could be exposed as an extra column in _dq_info and in report outputs.

fedeflowers:
## Changes

Adds human-readable, LLM-generated explanations to `has_no_row_anomalies` check function. When a row is flagged as anomalous, an `ai_explanation` is attached under _dq_info[].anomaly with a plain-language narrative, business impact, suggested action, the top contributing features, and the group's size / average severity.

  Explanations are generated inside Spark via the SQL `ai_query` function against a Databricks Model Serving endpoint — no extra Python dependencies, no driver-side LLM calls. Anomalous rows are grouped by (segment, top-2 contributing features) and the LLM is called once per group, so cost stays predictable on large datasets (capped by max_groups, ranked by group size × severity).

**What's new**
  - New parameters on `has_no_row_anomalies`: `enable_ai_explanation`, `ai_explanation_llm_model_config`, `redact_columns` (exclude sensitive columns from the prompt), and `max_groups` (LLM-call budget).
  - `redact_columns` keeps named columns out of the prompt, the grouping key, and segment values.
  - Graceful degradation: if the serving endpoint is unreachable (e.g. Foundation Model APIs not enabled), explanations are left null with a warning and scoring still succeeds.
  - `LLMModelConfig` gains `max_tokens`, `temperature`, `timeout`, and `max_retries` (validated) to bound LLM cost/latency.

⚠️  BREAKING CHANGES!
 
- `enable_contributions` param now defaults to `True` (was `False`). SHAP feature contributions are computed by default, adding scoring cost. Requires the shap library (already included in the [anomaly] extra). Set `enable_contributions=False` to restore the old behavior.
- AI explanations are on by default (`enable_ai_explanation=True`). Existing anomaly checks will now make LLM calls against a Databricks Model Serving endpoint (default databricks-claude-sonnet-4-5) and incur cost. This needs Foundation Model APIs available in the workspace; if the endpoint is unreachable, explanations are skipped (null) with a warning rather than failing. Set `enable_ai_explanation=False` to opt out entirely.
- Output schema change: _dq_info[].anomaly now contains an additional `ai_explanation` struct. Downstream consumers that assert on the exact anomaly struct schema should account for the new field.

### Linked issues

Resolves #1031

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

###  Documentation and Demos

- [x]  added/updated demos
- [x]  added/updated docs
- [ ]  added/updated agent skills


src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
```diff
@@ -0,0 +1,438 @@
...
+
+    config: dict = {
+        "model": model,
+        "model_type": "chat",
+        "max_retries": 3,
+    }
+    if api_key:
+        config["api_key"] = api_key
+    if api_base:
+        config["api_base"] = api_base
```
mwojtyczka: **SSRF — `api_base` is unvalidated.**

`api_base` flows from `LLMModelConfig` (user input via the rule's `llm_model_config`) into `dspy.LM(api_base=...)` without any host allowlist or scheme check. This is the exact case CLAUDE.md flags:

> **SSRF (Server-Side Request Forgery):** User-controlled URL fields such as *api_base* in *LLMModelConfig* are passed to outbound HTTP clients. Validate that any user-supplied URL targets an expected, allowlisted host before use. Do not allow arbitrary internal network endpoints to be reached via user config (CWE-918).

A malicious or careless rule definition can point at internal endpoints (e.g. `http://169.254.169.254/...` for cloud-metadata, or any internal service reachable from the cluster) and exfiltrate data via the LLM prompt or just probe the network.

Minimum bar: enforce `https://` only and an allowlist of approved suffixes (e.g. `*.databricks.com`, `*.openai.com`, `*.anthropic.com`, `*.cloud.databricks.com`, plus AI Gateway hosts your team explicitly approves). Reject anything else with `InvalidParameterError`. Add the validation in `_build_lm_config` so it's enforced for both the dataclass and dict paths.

src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
```diff
@@ -0,0 +1,438 @@
...
+
+    Routing rules:
+      - model_name already carries a litellm provider prefix ("provider/model") → pass through as-is.
+      - api_base is provided (explicitly or via OPENAI_API_BASE) → force the "openai/"
+        provider prefix so litellm uses its OpenAI-compatible adapter against that base URL.
+        Without this, a bare "databricks-foo" model name triggers litellm's native Databricks
+        provider and the custom endpoint is ignored (→ ENDPOINT_NOT_FOUND).
+      - Otherwise → pass through; litellm's default auto-detection applies.
+
+    api_key / api_base fall back to OPENAI_API_KEY / OPENAI_API_BASE env vars when
```
mwojtyczka: We should use the existing `LLMModelConfig`

src/databricks/labs/dqx/anomaly/scoring_run.py
```diff
@@ -126,6 +131,15 @@ def score_global_model(
         quantile_points=quantile_points,
     )
 
+    if config.enable_ai_explanation:
+        scored_df = add_explanation_column(
+            scored_df,
+            ExplanationContext.from_scoring_config(config),
+            segment_values=None,
+            is_ensemble=record.identity.is_ensemble,
+            drift_summary=format_drift_summary(drift_result),
```
mwojtyczka: **Redaction leak — `redact_columns` does not scrub `drift_summary`.**

`redact_columns` is documented as "feature names to exclude from the LLM prompt" and the explainer correctly filters them out of the pattern key (`_pattern_spark_expr`) and the per-pattern contributions map (`_aggregate_groups`). But `drift_summary` is built upstream by `format_drift_summary(drift_result)` (drift.py:283-293) without consulting `redact_columns`, and any drifted feature names — possibly the very ones the user asked to redact — are sent verbatim to the LLM:

```
'drift detected: salary=4.12; ssn=3.55'
```

A user redacting `salary` would still see `salary` in the prompt the moment the column also drifts. Same hole at scoring_run.py:254 for the segmented path.

Fix: either (a) filter `drift_result.drifted_columns` and `drift_result.column_scores` against `redact_set` inside `format_drift_summary` (pass `redact_columns` through), or (b) replace the drift_summary input with an opaque indicator like `'drift detected (3 features)'` when redaction is active.

Integration tests should add a case where a redacted column is also drifted, asserting the LLM prompt does not contain the redacted name.

src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
```diff
@@ -0,0 +1,438 @@
...
+
+    model = llm_model_config.model_name
+    if "/" not in model and api_base:
+        model = f"openai/{model}"
+
+    config: dict = {
+        "model": model,
+        "model_type": "chat",
+        "max_retries": 3,
+    }
```
mwojtyczka: **LLM token / budget caps missing.**

`_build_lm_config` produces `{model, model_type, max_retries}` plus optional `api_key` / `api_base` and passes it to `dspy.LM(...)`. There's no `max_tokens`, no `temperature`, no `timeout`, no per-call wall-clock budget. CLAUDE.md explicitly flags this:

> **LLM denial of service and information disclosure**: Malicious or pathological prompts can trigger unbounded or very expensive inference calls (OWASP LLM04) — enforce token/budget limits on all LLM calls.

With the default `max_groups=500`, a runaway model could rack up serious cost / latency on a single anomaly check run.

Add at minimum:
- `max_tokens` (cap the response — narrative/impact/action are documented as ≤ ~85 words combined, so ~300–400 tokens is plenty),
- `temperature=0` for determinism,
- `timeout` (per-call, e.g. 30s).

Expose these on `LLMModelConfig` so users can tune for their endpoint, but enforce sensible defaults here.

src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
```diff
@@ -0,0 +1,438 @@
...
+        result_rows.append(
+            (
+                group[_PATTERN_COL],
+                prediction.narrative,
+                prediction.business_impact,
+                prediction.action,
+                int(group["group_size"]),
+                float(group["group_avg_severity"]),
+            )
+        )
```
mwojtyczka, line 546: **LLM output is written to the DataFrame without validation or length-capping.**

`prediction.narrative` / `business_impact` / `action` come straight from the LLM and are appended into `_dq_info[].anomaly.ai_explanation` with no checks. The DSPy `OutputField(desc="Max 2 sentences, max 40 words total")` is guidance to the model, not enforcement — a misbehaving or jailbroken model can return arbitrary content (multi-KB strings, control characters, prompt-injection payloads echoed back, schema details from the input).

CLAUDE.md:

> **LLM denial of service and information disclosure**: ... LLM responses may inadvertently echo back sensitive inputs such as schema metadata or data samples (OWASP LLM06) — treat LLM output as untrusted data; do not relay raw LLM responses to end users or logs without review.

Minimum hardening before storing in the result struct:
1. `str(...)` cast (DSPy may return non-str on malformed completions).
2. Hard length cap (e.g. `[:500]`) per field — matches the documented intent.
3. Strip newlines / control chars (`re.sub(r'[\x00-\x1f\x7f]', ' ', ...)` or similar) so a malicious response can't forge log entries when the column is logged downstream.
4. Optional: reject responses that contain raw column values from the input — the prompt only sends feature *names* and percentages, so the LLM should never echo data values back.

src/databricks/labs/dqx/anomaly/check_funcs.py
```diff
@@ -4,22 +4,105 @@
...
+        )
+
+    if enable_ai_explanation and not DSPY_AVAILABLE:
+        raise InvalidParameterError(
+            "enable_ai_explanation=True requires the 'dspy' dependency. "
+            "Install: pip install databricks-labs-dqx[anomaly,llm]"
+        )
+
+    if redact_columns is not None and not isinstance(redact_columns, list):
+        raise InvalidParameterError("redact_columns must be a list of column name strings.")
```
mwojtyczka: `redact_columns` is validated only as `isinstance(list)` — element types aren't checked. A dict, number, or `None` smuggled into the list will reach `_pattern_spark_expr` (anomaly_llm_explainer.py:189) where it's f-string-interpolated into a SQL `array(...)` literal, producing a confusing `AnalysisException` at scoring time instead of a clear validation error.

Tighten to:
```python
if redact_columns is not None:
    if not isinstance(redact_columns, list) or not all(isinstance(c, str) and c for c in redact_columns):
        raise InvalidParameterError("redact_columns must be a list of non-empty column name strings.")
```

(Also consider validating each element matches a column-name regex such as `^[a-zA-Z_][a-zA-Z0-9_]*$` — the values flow into Spark SQL string literals so they're escaped today, but the defensive check would harden the whole path against a future refactor that uses them as identifiers.)

src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
```diff
@@ -0,0 +1,438 @@
...
+    kept_groups: list[dict],
+    ctx: ExplanationContext,
+    segment_str: str,
+    is_ensemble: bool,
+    drift_summary: str,
+    predictor,
+) -> list[tuple]:
+    """Invoke the LLM once per retained group. Returns rows for the result DataFrame."""
+    result_rows: list[tuple] = []
+    for group in kept_groups:
```
mwojtyczka: **Sequential LLM calls on the driver — performance cliff at scale.**

`_call_llm_for_groups` invokes the LLM in a tight Python `for group in kept_groups` loop on the driver. With the default `max_groups=500` and a typical p95 LLM latency of 1–3 s per call, a single anomaly check run can spend **8–25 minutes** purely waiting on serial completions, while the cluster sits idle. For segmented scoring this is multiplied across segments.

Two low-friction options:

1. **`databricks.labs.blueprint.parallel.Threads`** — already used elsewhere in this codebase (per AGENTS.md):
   ```python
   from databricks.labs.blueprint.parallel import Threads
   tasks = [partial(_predict_one, group, ctx, segment_str, is_ensemble, drift_summary, predictor) for group in kept_groups]
   results, errors = Threads.gather("ai_explanation", tasks)
   ```
   With ~16-way parallelism this drops 500 calls from ~10 min to <1 min on most endpoints.

2. **Async `dspy.LM`** — DSPy supports `acall` / async predictors; a `gather()` over `kept_groups` works if the surrounding code can be awaited.

Also missing: any per-call timeout (see the related comment on `_build_lm_config`). One stuck completion blocks the entire run today.

Finally, partial failure handling — if one group's LLM call raises, the whole `add_explanation_column` aborts. Consider catching per-group, logging the pattern, and emitting a null `ai_explanation` for that group so a single transient error doesn't kill an otherwise-successful scoring run.

mwojtyczka (changes requested):
Thanks for the PR! I left some comments, mainly AI generated

mwojtyczka:
@fedeflowers, there are some concerns about scalability - lots of LLM calls. Maybe using `ai_query` would be more scalable.

fedeflowers:
> @fedeflowers, there are some concerns about scalability - lots of LLM calls. Maybe using `ai_query` would be more scalable.

yeah, that's why I tried to group similar anomalies in groups and compute llm call just for one group, also capping max groups, I've now pushed also a configuration using ai_query, as having it as option gives more flexibility to which endpoint to use.

Maybe we can add a cache or batch the ai_query to lower LLM calls further

mwojtyczka:
@vb-dbrks can you pls check it?

mwojtyczka (commented):
Architecturally strong PR — the dual `ai_query` / `driver` executor design with shared prompt source-of-truth, the security posture (SSRF allowlist, SQL injection guards, log-injection sanitisation, OWASP LLM04 budget caps), and the per-group failure isolation are all above bar. Test coverage on both executor paths is excellent (~1.2k lines across unit + integration).

Posting 15 inline comments below — a mix of correctness/UX, single-source-of-truth maintainability concerns, and polish. The most important are the `max_groups` cost-surprise (silent over-budget when `max_groups < num_eligible_segments`), the silently-truncated dropped-groups behaviour with no programmatic signal, and the duplicated confidence thresholds across executor paths.

Agent-skill update is intentionally not flagged — leave that to a follow-up.

src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
```diff
@@ -0,0 +1,929 @@
...
+except ImportError:
+    dspy = None
+    DSPY_AVAILABLE = False
+
+if TYPE_CHECKING:
+    from databricks.labs.dqx.anomaly.scoring_config import ScoringConfig
+
+logger = logging.getLogger(__name__)
+
+_TOP_N = 5
```
mwojtyczka, line 121: **Hardcoded prompt knobs limit user control.**

`_TOP_N = 5` (contributions in prompt), the top-2 pattern slice (in `_pattern_spark_expr`), and `_LLM_FIELD_MAX_LEN = 500` are all module-level constants. These materially affect prompt quality and group cardinality — high-cardinality SHAP outputs may want top-3 patterns or top-10 contributions for a richer narrative.

Suggest exposing on `LLMModelConfig`:

```python
explanation_top_features: int = 5     # in-prompt SHAP feature count
explanation_pattern_size: int = 2     # group key cardinality
explanation_field_max_len: int = 500  # output truncation
```

Validate in `__post_init__`. The `ai_query` path's `_format_contributions_sql(_TOP_N)` reads from the same config, keeping both paths aligned.

src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
```diff
@@ -0,0 +1,929 @@
...
+_PATTERN_COL = "__dqx_pattern"
+_LLM_FIELD_MAX_LEN = 500
```
mwojtyczka, line 127: **`_LLM_FIELD_MAX_LEN` truncates mid-word.**

500 chars is generous for narratives capped at "max 40 words" in the prompt, but the slice `[:_LLM_FIELD_MAX_LEN]` truncates mid-word, hurting readability when the model overruns. Either cap at a word boundary (regex-based `r'\s'` rfind) or expose as a config knob (see prior comment).

src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
```diff
@@ -0,0 +1,929 @@
...
+AnomalyGroupExplanationSignature: type | None = None
+if DSPY_AVAILABLE:
+    AnomalyGroupExplanationSignature = _build_dspy_signature()
+
+
+def _derive_confidence(mean_std: float | None, is_ensemble: bool) -> str:
+    """Map aggregated score_std + is_ensemble flag to a human-readable confidence label."""
+    if not is_ensemble or mean_std is None:
+        return "n/a"
+    if mean_std < 0.05:
```
mwojtyczka: **Confidence thresholds duplicated across executor paths — single-source-of-truth violation.**

`_derive_confidence` (driver) hardcodes `0.05` / `0.15`. `_build_ai_query_prompt_column` at line ~581 hardcodes the same values inline as Spark expressions. Anyone tuning one without the other will silently produce different `confidence` labels for the same input across the two executor paths.

Extract:

```python
_CONFIDENCE_HIGH_BELOW = 0.05
_CONFIDENCE_MIXED_BELOW = 0.15
```

Reference both from Python and from the SQL `F.when(...)` chain.

src/databricks/labs/dqx/anomaly/scoring_run.py
```diff
@@ -32,6 +39,23 @@
...
+
+    Bound analysis: with N eligible segments and budget B, the total LLM-call cap is
+    ``N * (B // N) <= B`` when ``B >= N``. When ``B < N`` the floor of 1 kicks in and the
+    cap becomes ``N`` — wider than B but still finite and proportional to the input.
+    Documented as a deliberate tradeoff so the feature remains useful when users
+    drastically under-provision the budget.
+    """
+    if num_eligible_segments <= 0:
+        raise InvalidParameterError("num_eligible_segments must be positive")
+    return max(1, max_groups // num_eligible_segments)
```
mwojtyczka, line 60: **`max_groups` cost surprise when `max_groups < num_eligible_segments`.**

The `max(1, ...)` floor means a user setting `max_groups=10` against 100 segments gets up to **100 LLM calls**, not 10. The docstring mentions this as "a deliberate tradeoff" but the user-facing parameter description on `has_no_row_anomalies` doesn't. At Anthropic/OpenAI prices this is a real cost surprise.

Options:
- Reject at validation time when `max_groups < num_eligible_segments`, asking the user to bump it.
- Or log a `WARNING` at scoring time with the actual derived cap (e.g. `"max_groups=10 < 100 segments; effective cap is 100 LLM calls"`).
- At minimum, document in the `max_groups` parameter docstring with a concrete example.

src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
```diff
@@ -0,0 +1,929 @@
...
+    # Lazy import: *llm_core* eagerly imports dspy at module load. Pulling it in only when the
+    # driver path actually constructs an LM keeps users on ``executor='ai_query'`` free of the
+    # dspy dependency, matching the documented install matrix.
+    from databricks.labs.dqx.llm.llm_core import LLMModelConfigurator  # pylint: disable=import-outside-toplevel
+
+    llm_cfg = ctx.llm_model_config or LLMModelConfig()
+    return LLMModelConfigurator(llm_cfg).create_lm()
+
+
+def _log_dropped_groups(dropped_groups_count: int, dropped_rows_count: int, max_groups: int) -> None:
```
mwojtyczka, line 592: **Dropped groups are silent to users.**

When `max_groups` is exceeded, those rows get `ai_explanation = NULL`. The warning log goes to the driver — Databricks Job users running scheduled scoring won't see it without digging through driver logs. There's no programmatic signal in the result schema.

Consider adding a small struct alongside `_dq_info[0].anomaly.ai_explanation`:

```python
ai_explanation_truncation: {
  groups_total: long,
  groups_kept: long,
  rows_dropped: long,
}
```

Then users can detect truncation in SQL: `WHERE _dq_info[0].anomaly.ai_explanation_truncation.rows_dropped > 0`.

src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
```diff
@@ -0,0 +1,929 @@
...
+        "action",
+        "One sentence, max 20 words. What a data analyst should investigate for this group.",
+    ),
+)
+# JSON schema the ai_query path sends as ``responseFormat`` so the endpoint returns
+# {narrative, business_impact, action} as a structured object — no string parsing needed.
+# NB: Databricks ai_query rejects ``maxLength`` on string types (BAD_REQUEST). Length-capping
+# is enforced post-parse via *_sanitize* in *_call_llm_for_groups_ai_query* (matches the driver
+# path's *_sanitize_llm_field*) so we don't need it on the schema here.
+_AI_QUERY_RESPONSE_FORMAT = (
```
mwojtyczka: **`_AI_QUERY_RESPONSE_FORMAT` is hand-rolled JSON — drift risk.**

The schema string at lines 98–106 enumerates `narrative / business_impact / action` literally, while `_PROMPT_OUTPUT_FIELDS` enumerates the same names elsewhere. Adding a fourth output field requires updating both, and the hand-rolled JSON is easy to typo (missing comma, wrong nesting).

Build it from `_PROMPT_OUTPUT_FIELDS`:

```python
def _build_ai_query_response_format() -> str:
    fields = [name for name, _ in _PROMPT_OUTPUT_FIELDS]
    return json.dumps({
        "type": "json_schema",
        "json_schema": {
            "name": "explanation",
            "strict": True,
            "schema": {
                "type": "object",
                "additionalProperties": False,
                "required": fields,
                "properties": {f: {"type": "string"} for f in fields},
            },
        },
    })
```

One source of truth; impossible to drift.

src/databricks/labs/dqx/anomaly/check_funcs.py
```diff
@@ -72,6 +177,52 @@ def has_no_row_anomalies(
...
+                    "model_name": "databricks-qwen3-next-80b-a3b-instruct",
+                    "api_key": "<pat_token>",
+                    "api_base": "https://<gateway-id>.ai-gateway.cloud.databricks.com/mlflow/v1",
+                }
+
+            When ``api_base`` is set, the model is routed through litellm's OpenAI-compatible
+            adapter (``openai/<model>``) — use this for AI Gateway and other OpenAI-compatible
+            endpoints. ``api_key`` / ``api_base`` also fall back to ``OPENAI_API_KEY`` /
+            ``OPENAI_API_BASE`` env vars when unset on the config.
+        redact_columns: Column names to exclude from the LLM prompt. Filters SHAP contribution
```
mwojtyczka, line 202: **Redaction does not cover segment values — flag more prominently.**

The docstring correctly notes that segment `key=value` pairs reach the LLM prompt verbatim. For PII-sensitive segmentation columns (e.g. `customer_id`, `email_domain`), this is a real privacy exposure that's easy to miss.

Suggestions, in order of effort:
1. **Easiest**: log a `WARNING` once when `enable_ai_explanation=True` against a segmented model, listing the segment keys that will appear in prompts.
2. **Better**: extend redaction so any segment key in `redact_columns` is filtered from `_format_segment` output (e.g. emit `customer_id=<redacted>` or skip the entry).
3. **Best**: an explicit `redact_segment_values: bool = False` flag.

At minimum, escalate the docstring note to an `Admonition type="warning"` in the user-facing reference docs.

src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
```diff
@@ -0,0 +1,929 @@
...
+
+    Drops null and redacted entries, takes the top-2 features by |value| desc,
+    sorts their names asc, and joins with '+'. Empty or null maps yield 'unknown'.
+    Ranking uses absolute value so signed SHAP contributions pick the same top-2
+    as `format_contributions_map`. Implemented in SQL so Databricks Connect /
+    serverless workers don't need the dqx package installed.
+    """
+    col = f"`{contributions_col}`"
+    if redact_set:
+        escaped = [r.replace("'", "''") for r in redact_set]
```
mwojtyczka: **Hand-rolled SQL escaping for `redact_columns` is fragile.**

```python
escaped = [r.replace("'", "''") for r in redact_set]
redact_arr = "array(" + ", ".join(f"'{r}'" for r in escaped) + ")"
```

This is correct for single-quoted SQL string literals but brittle — `redact_columns` comes from user input (per-rule kwargs), so an exotic value (newline, control chars, escaped quotes) could either break the SQL or alter the filter semantics. Better to skip string interpolation entirely:

```python
entries = (
    f"filter(map_entries({col}), e -> e.value is not null "
    "and not array_contains(:redact_keys, e.key))"
)
return F.expr(entries, {"redact_keys": F.array(*[F.lit(r) for r in redact_set])})
```

Or build the entire expression with `F.col` / `F.array` instead of an `F.expr` SQL string. Eliminates the manual escape and the trust boundary.

src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
```diff
@@ -0,0 +1,929 @@
...
+            )
+    except Exception as exc:
+        # Per-group LLM-call failure isolation: DSPy's *AdapterParseError*, litellm errors,
+        # and network errors don't share a clean base class, and a single bad call must not
+        # abort the scoring run — every other group should still get its explanation.
+        # Narrowing the catch risks regressing this contract when DSPy/litellm grow new
+        # exception types.
+        # Sanitise the pattern key for the log line — it derives from column names that may
+        # contain newlines or control chars (CWE-117). Strip both before interpolating.
+        safe_pattern = pattern.replace("\n", "_").replace("\r", "_") if isinstance(pattern, str) else "<non-str>"
```
mwojtyczka: **Partial log-injection sanitisation — use the existing regex.**

```python
safe_pattern = pattern.replace("\n", "_").replace("\r", "_") if isinstance(pattern, str) else "<non-str>"
```

Only strips `\n` and `\r`. Other control chars (`\t`, `\x1b` ESC, `\b` BS) can also forge log entries (CWE-117) or hide content from log viewers. You already have `_CONTROL_CHAR_RE` two functions above — reuse it:

```python
safe_pattern = _CONTROL_CHAR_RE.sub("_", pattern) if isinstance(pattern, str) else "<non-str>"
```

Matches the sanitiser used elsewhere in this module.

src/databricks/labs/dqx/config.py
```diff
@@ -190,16 +193,121 @@ class RunConfig:
     lakebase_port: str | None = None
 
 
+_DEFAULT_API_BASE_ALLOWED_HOSTS: tuple[str, ...] = (
```
mwojtyczka: **Default allowed-hosts list misses Azure OpenAI and AWS Bedrock.**

`_DEFAULT_API_BASE_ALLOWED_HOSTS` covers Databricks, OpenAI, Anthropic. Users on common production endpoints will be silently rejected:

- Azure OpenAI: `<resource>.openai.azure.com`
- AWS Bedrock: `bedrock-runtime.<region>.amazonaws.com`
- Vertex AI: `<region>-aiplatform.googleapis.com`

Users have to extend `api_base_allowed_hosts` themselves on every `LLMModelConfig` instance. Either expand the default to cover the major providers, or — at minimum — document the additions users will need in the param docstring with copy-pasteable examples.

src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
```diff
@@ -0,0 +1,929 @@
...
+    if not is_ensemble or mean_std is None:
+        return "n/a"
+    if mean_std < 0.05:
+        return "high"
+    if mean_std < 0.15:
+        return "mixed"
+    return "low"
+
+
+def _pattern_spark_expr(contributions_col: str, redact_set: frozenset[str]) -> Column:
```
mwojtyczka, line 219: **Pattern collisions across feature sets — document the tradeoff.**

Two groups with different feature distributions but the same top-2 contributors collide into one group key. The LLM gets contributions averaged across the colliding groups, producing a less-precise narrative.

E.g. group A `[amount, quantity, discount]` and group B `[amount, quantity, region]` both yield pattern `amount+quantity` and share one explanation. The PR description says this is intentional ("keeps group cardinality low") — agreed it's the right call — but users won't know unless it's documented in a user-facing place. Add a note to the `add_explanation_column` docstring or the `has_no_row_anomalies` parameter docs.

src/databricks/labs/dqx/anomaly/check_funcs.py
```diff
@@ -4,22 +4,123 @@
...
+    enable_contributions: bool,
+    enable_ai_explanation: bool,
+    llm_model_config: LLMModelConfig | None,
+) -> None:
+    if enable_contributions and not SHAP_AVAILABLE:
+        raise InvalidParameterError(
+            "enable_contributions=True requires the 'shap' dependency. "
+            "Install anomaly extras: pip install databricks-labs-dqx[anomaly]"
+        )
+    if enable_ai_explanation and not enable_contributions:
```
mwojtyczka, line 87: **`enable_ai_explanation=True` requires `enable_contributions=True` — friction.**

Validated correctly, but the two-step opt-in is a UX trap: users who want explanations will repeatedly forget the prerequisite. Two options:

1. **Auto-enable** `enable_contributions` when `enable_ai_explanation=True`, with an `INFO` log noting the implicit enablement.
2. **Strengthen the error message** to include the fix verbatim (e.g. add `"…set enable_contributions=True"` — done already 👍).

Option 1 is friendlier; option 2 is purer. Either way, the docstring on `enable_ai_explanation` should state the dependency in the *first sentence* (currently buried).

src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
```diff
@@ -0,0 +1,929 @@
...
+    # instead of aborting the whole job, matching the driver path's per-group failure isolation
+    # via *Threads.gather*. The platform's retry count itself is not exposed, so *max_retries*
+    # on *LLMModelConfig* is documented as driver-only.
+    raw = enriched.withColumn(
+        "__raw_response",
+        F.expr(
+            f"ai_query('{endpoint}', __prompt, "
+            f"modelParameters => named_struct('max_tokens', {int(llm_cfg.max_tokens)}, "
+            f"'temperature', {float(llm_cfg.temperature)}), "
+            f"responseFormat => '{_AI_QUERY_RESPONSE_FORMAT}', "
```
mwojtyczka: **`failOnError=false` swallows endpoint failures silently.**

If the entire serving endpoint is down, every row gets `NULL` and there's no log signal — users see an empty `ai_explanation` column and can't tell whether the model saw nothing anomalous, or the LLM call failed wholesale.

The driver path logs per-group failures via `_explain_one_group`. Mirror that: after the `ai_query` call, count nulls in the result and log a `WARNING` if `nulls == kept_groups_count` (i.e. nothing came back). Cheap to implement (`F.sum(F.col("narrative").isNull().cast("int"))` over the result, one collect).

src/databricks/labs/dqx/anomaly/check_funcs.py
```diff
@@ -30,6 +131,10 @@ def has_no_row_anomalies(
     drift_threshold: float | None = None,
     enable_contributions: bool = False,
     enable_confidence_std: bool = False,
+    enable_ai_explanation: bool = False,
+    llm_model_config: LLMModelConfig | dict | None = None,
+    redact_columns: list[str] | None = None,
+    max_groups: int = 500,
```
mwojtyczka, line 130: **`max_groups=500` default — surface cost expectations in docs.**

At today's mid-tier-model pricing, 500 LLM calls per scoring run is roughly $1–3, and some users will run scoring hourly (≈ $700–2,200/month per pipeline). The default is reasonable for power users but quietly expensive for casual ones.

Either lower the default (100? — most exploration cases need fewer) or add a one-line *cost budget* note in the `max_groups` parameter docstring: e.g. `"each group uses one LLM call; with 500 groups expect ~$1-3/run on mid-tier models"`. Combine with the cost-surprise note from the `_split_max_groups_budget` comment.

src/databricks/labs/dqx/anomaly/check_funcs.py
```diff
@@ -30,6 +131,10 @@ def has_no_row_anomalies(
     drift_threshold: float | None = None,
     enable_contributions: bool = False,
     enable_confidence_std: bool = False,
+    enable_ai_explanation: bool = False,
```
mwojtyczka: **Add performance tests before merge.**

PR task list shows perf tests as unchecked. This feature adds LLM calls per group with measurable Spark-side group-aggregation cost; the `_aggregate_groups`/`_aggregate_groups_spark` paths involve windowed sorts, exploded contributions maps, and joins.

Please add a perf test under `tests/perf/` covering at least:
- small-anomaly run (a few groups, exercises orchestration overhead)
- large-anomaly run (cap-saturating groups, exercises ranking + collect)
- segmented path (exercises the budget split + multi-segment loop)

The LLM call itself can be mocked at the predictor level — the perf signal you want is the Spark-side aggregation cost.

mwojtyczka (changes requested):
thanks for the update! left some new comments. Once code changes are finilized, the demo and docs should also be updated accordingly.

src/databricks/labs/dqx/config.py
```diff
@@ -190,16 +193,121 @@ class RunConfig:
...
+    # to permit explicitly approved AI Gateway hosts. Each entry is matched as a case-insensitive
+    # suffix of the URL host (e.g. ".gateway.corp.example"). IP literals must match exactly.
+    api_base_allowed_hosts: tuple[str, ...] = ()
+    # Where the LLM call is executed for AI explanations:
+    #   "ai_query" — Spark SQL ``ai_query`` on executors against a Databricks Model Serving endpoint
+    #                whose name is taken from *model_name* (provider prefix stripped). Scales with the
+    #                cluster, no driver collect — recommended default on Databricks.
+    #   "driver"   — DSPy on the driver, threaded fan-out. Use for non-Databricks endpoints
+    #                (any provider DSPy supports via *api_base*).
+    executor: str = "ai_query"
```
vb-dbrks: I'd reduce the feature's scope to only work with ai_query. I really don't see the flexibility issue; ai_query covers most needs for this without having us to add more plumbing on our end, keeping this feature simple. We can consider more options in the future if we get some feedback.

src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
```diff
@@ -0,0 +1,929 @@
...
+    dspy = None
+    DSPY_AVAILABLE = False
+
+if TYPE_CHECKING:
+    from databricks.labs.dqx.anomaly.scoring_config import ScoringConfig
+
+logger = logging.getLogger(__name__)
+
+_TOP_N = 5
+_PATTERN_COL = "__dqx_pattern"
```
vb-dbrks: Internal columns added during scoring use a UUID-suffixed name (see ScoringOutputColumns in check_funcs.py:111-118)... __dqx_pattern is a fixed module-level constant. A user with a pre-existing column literally named  __dqx_pattern will have it silently overwritten, and there's no check_reserved_row_id_columns-style guard for it either (scoring_utils.py:205-211 only reserves the _dqx_row_id prefixes). I'd add a pattern field to ScoringOutputColumns and thread a UUID-suffixed name through ExplanationContext like the other internal columns,

src/databricks/labs/dqx/anomaly/anomaly_info_schema.py
```diff
@@ -3,12 +3,27 @@
...
 )
 
+# Schema for the AI explanation sub-struct inside the anomaly info struct.
+# narrative / business_impact / action are LLM-generated; pattern is deterministic.
+# group_size / group_avg_severity describe the (segment, pattern) group this row belongs to.
+ai_explanation_struct_schema = StructType(
+    [
+        StructField("narrative", StringType(), True),
+        StructField("business_impact", StringType(), True),
+        StructField("pattern", StringType(), True),
```
vb-dbrks: The pattern is a sorted top-2 SHAP feature key (e.g., "amount+quantity"), but the field name doesn't make that obvious. Suggest renaming to one of: 

- top_features (matches what's in the  #string)
- contributing_features (pairs naturally with the sibling contributions field one level up in anomaly_info_struct_schema)

I'd prefer top_features as the field.

Whatever the rename, _PATTERN_COL in anomaly_llm_explainer.py:122 and the pattern references in _attach_explanation_struct (anomaly_llm_explainer.py:786-792) should be updated together.

src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
```diff
@@ -0,0 +1,929 @@
...
+        entries = (
+            f"filter(map_entries({col}), e -> e.value is not null " f"and not array_contains({redact_arr}, e.key))"
+        )
+    else:
+        entries = f"filter(map_entries({col}), e -> e.value is not null)"
+    sql = (
+        f"case when {col} is null or size({entries}) = 0 then 'unknown' "
+        f"else concat_ws('+', array_sort(transform(slice(array_sort({entries}, "
+        f"(a, b) -> case when abs(b.value) > abs(a.value) then 1 "
+        f"when abs(b.value) < abs(a.value) then -1 else 0 end), 1, 2), e -> e.key))) end"
```
vb-dbrks, line 240: The pattern key and mean_contributions both strip sign via abs(value). This will lead the LLM to see the amount (61%), quantity (22%), but it has no way to know whether the amount was anomalously high or low for the group. I'd advise preserving signed SHAP in the per-pattern mean (_aggregate_groups_spark lines 294-300, the F.avg("__v") aggregate), then passing signed percentages to the prompt: amount (+61%), quantity (-22%). Rank by |value| but display with sign.

src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
```diff
@@ -0,0 +1,929 @@
...
+    (
+        "business_impact",
+        "One sentence, max 25 words. Likely downstream business impact if this group of rows is "
+        "processed unchanged. Concrete, tied to the contributing features.",
+    ),
+    (
+        "action",
+        "One sentence, max 20 words. What a data analyst should investigate for this group.",
+    ),
+)
```
vb-dbrks, line 86: The prompt tells the model what to write, but not what to avoid, and there's no exemplar to pin the style. Consider: 
1. Explicit anti-patterns in _PROMPT_INSTRUCTIONS. Negative examples are often more effective than positive ones with small models: e.g.,"Avoid: phrases like 'The data shows...', 'It appears that...', 'might indicate...'. Be direct. Do not restate the input fields back to the user."
2. One or two few-shot examples in the instructions block (one with drift, one without).

Also, the prompt is hard-coded, with no A/B framework or evaluation harness... which is fine for now in favour of a simple approach. We can consider an evaluation framework in the future before shipping; however, was this prompt iterated on? this isn't a blocker just trying to understand the iteration history

tests/integration_anomaly/test_anomaly_ai_explanation.py
vb-dbrks, line 1: **No prompt-quality test exists**. I'd suggest at least one integration test that's a low-bar guard... won't catch quality regressions, but will catch "the LLM is producing structurally bogus output" regressions: 
1. Scores a deterministic outlier (e.g., amount=1_000_000 in a uniform [0, 100] distribution). 
2. Calls the explainer with ai_query.
3. Asserts the narrative contains the dominant feature name ("amount"), the simplest factual-correctness check.
4. asserts narrative word count ≤ 50, business_impact ≤ 30, action ≤ 25 (matches the prompt's documented constraints).

vb-dbrks (commented):
Hi @fedeflowers - Awesome PR! Love to see the detailed thought process in the design. Some simplification and hardening are required. My advice is to keep this PR focused on the ai_query-based option and remove the DSPy-based executor.

src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
vb-dbrks: Additionally, I don't know how much the Model Name will help or hinder in the prompt, so consider removing it unless you have other ideas.

fedeflowers: Yes, I agree on the changes, no specific iteration was made, just manual tested against simple datasets, how would you go for the evaluation framework? dpsy golden dataset + metric?

vb-dbrks: Just an iterative manual evaluation with some integration tests for now should be good @fedeflowers - There's a bigger story that will address the scaffolding #1030  which will have to address this.

tests/integration_anomaly/test_anomaly_ai_explanation.py
fedeflowers: I'm having little time to contribute in this period, sorry if the PR is stagnant I'll try to get back at it ASAP!

vb-dbrks: nw @fedeflowers  - I'll pick up some of these changes in the mean time.

mwojtyczka:
### Review recap after the `ai_query` refactor — resolved vs. still-open

Re-reviewed against the current head. The switch from DSPy to SQL `ai_query` (running the LLM inside Spark) plus several targeted fixes addressed a large cluster of earlier comments. I've **resolved the threads below** after verifying each in the current code; the remaining open threads are listed so the handoff is clear.

**✅ Resolved (verified fixed):**
- Use existing `LLMModelConfig` — now imported and used.
- `drift_summary` redaction leak — `format_drift_summary(..., redact_columns)` now scrubs it.
- 4× Copilot DSPy concerns (ImportError handling, `build_language_model`, conditional `AnomalyGroupExplanationSignature`, `find_spec` false positive) — **DSPy removed entirely**, all moot.
- LLM token/budget caps — `max_tokens`/temperature now passed via `modelParameters`.
- LLM output validation/length-capping — `_sanitize` strips control chars and caps to `_LLM_FIELD_MAX_LEN` on all output fields.
- `redact_columns` element-type validation — now requires a list of non-empty strings.
- Sequential LLM calls on the driver — replaced by distributed `ai_query` (no per-group driver loop).
- Confidence thresholds duplicated — single-source module constants, one `ai_query` path.
- `_AI_QUERY_RESPONSE_FORMAT` hand-rolled drift — built from `_PROMPT_OUTPUT_FIELDS` + parsed with `from_json(schema)`.
- Fragile SQL escaping — `_sql_string_literal` now escapes both `\` and `'`.
- "Reduce scope to `ai_query` only" — done.
- `__dqx_pattern` fixed name → UUID-suffixed `__dq_anomaly_pattern_{uuid}`.
- Rename `pattern` field → `top_features` in the info schema.

**⏳ Still open (not addressed / judgment calls):**
- _Code/UX_: `_TOP_N=5` hardcoded; `_LLM_FIELD_MAX_LEN` mid-word truncation; `enable_ai_explanation` requires `enable_contributions`; `failOnError=false` still silent when the endpoint is down; `abs()` strips SHAP sign so the LLM can't tell direction.
- _Design/docs_: `max_groups` cost surprise & `=500` default cost expectations; dropped-groups are silent; pattern-collision tradeoff; segment-values-in-prompt prominence.
- _Tests_: performance test and prompt-quality guard test still missing.
- _Couldn't verify confidently, left open_: allowed-hosts (Azure/Bedrock) — the `api_base` allowlist appears gone with the refactor; partial log-injection sanitisation; Copilot's docstring-vs-segment accuracy note.

Resolutions reflect code state only — the open items remain for follow-up.

demos/dqx_row_anomaly_detection_demo.py
mwojtyczka:
```suggestion
# MAGIC ## Section 5a: (Optional) Review Results to understand why some records are anomalous
```

src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
```diff
@@ -0,0 +1,643 @@
...
+    # validated config fields. responseFormat is built from *_PROMPT_OUTPUT_FIELDS* (no user
+    # input). failOnError => false: per-row failures (after the platform's internal retries)
+    # yield NULL instead of aborting the whole job, so one bad completion doesn't kill an
+    # otherwise-successful scoring run. The platform's retry count is not exposed, so
+    # *max_retries* on *LLMModelConfig* has no effect on this path.
+    raw = enriched.withColumn(
+        "__raw_response",
+        F.expr(
+            f"ai_query('{endpoint}', __prompt, "
+            f"modelParameters => named_struct('max_tokens', {int(llm_cfg.max_tokens)}, "
```
mwojtyczka, line 512: Minor robustness/consistency: `int(llm_cfg.max_tokens)` / `float(llm_cfg.temperature)` aren't validated. `LLMModelConfig.__post_init__` validates `max_retries` but not these, so `LLMModelConfig(max_tokens=None)` raises a raw `TypeError` here at SQL-build time (instead of a clean `InvalidParameterError`), and a negative/zero `max_tokens` or out-of-range `temperature` is passed straight to the endpoint. Suggest adding a check to `LLMModelConfig.__post_init__` (e.g. `max_tokens` positive int, `temperature >= 0`) to match the existing `max_retries` validation.

src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
```diff
@@ -0,0 +1,643 @@
...
+            f"'temperature', {float(llm_cfg.temperature)}), "
+            f"responseFormat => '{_AI_QUERY_RESPONSE_FORMAT}', "
```
mwojtyczka: Latent fragility: `responseFormat => '{_AI_QUERY_RESPONSE_FORMAT}'` interpolates the JSON schema into the `ai_query` SQL as an unescaped single-quoted literal. Safe today because `_AI_QUERY_RESPONSE_FORMAT` is built from internal field names (`narrative`/`business_impact`/`action`) with no quotes — but if a future field name/description added to `_PROMPT_OUTPUT_FIELDS` ever contains a `'`, this SQL string breaks. Worth a comment noting the constraint, or escaping it the way `_sql_string_literal` does, to harden against future edits.

mwojtyczka (commented):
Two minor, non-blocking findings from re-reviewing the refactored `ai_query` code (the rewrite itself is solid — injection-safe, output sanitized/capped, driver collect bounded by `max_groups`). Everything else I'd flag is already in the open threads.

docs/dqx/docs/reference/quality_checks.mdx
```diff
@@ -3397,6 +3397,10 @@ checks = [
...
 - `enable_contributions`: Include per-feature contributions in `_dq_info[0].anomaly.contributions` (default `False`). Set `True` for explainability; adds significant scoring cost. See the section below and [Schema of _dq_info](/docs/guide/row_anomaly_detection#schema-of-the-info-column-_dq_info) for field details.
 - `enable_confidence_std`: Include `confidence_std` for ensembles (default `False`). Useful when using ensemble training.
+- `enable_ai_explanation`: Add an LLM-generated plain-language explanation in `_dq_info[0].anomaly.ai_explanation` (default `False`). Requires `enable_contributions=True` (the SHAP contributions feed the prompt). The LLM call runs inside Spark via the SQL `ai_query` function against a Databricks Model Serving endpoint — no extra dependency. See the **AI Explanations (LLM)** section below.
+- `llm_model_config`: `LLMModelConfig` (or a dict with keys `model_name`, `api_key`, `api_base`) used by `enable_ai_explanation`. `model_name` must resolve to a Databricks Model Serving endpoint (the `databricks/` prefix, if present, is stripped). Defaults to `databricks/databricks-claude-sonnet-4-5`.
```
mwojtyczka: we need to give it a more meaningful name, otherwise it may be confusing with `model_name` param
```suggestion
- `ai_explanation_llm_model_config`: `LLMModelConfig` (or a dict with keys `model_name`, `api_key`, `api_base`) used by `enable_ai_explanation`. `model_name` must resolve to a Databricks Model Serving endpoint (the `databricks/` prefix, if present, is stripped). Defaults to `databricks/databricks-claude-sonnet-4-5`.
```
also

docs/dqx/docs/guide/row_anomaly_detection/index.mdx
```diff
@@ -324,9 +324,52 @@ Each element is a **struct** with a shared “wide” schema. Currently, the onl
...
+| `narrative` | string | Plain-language description of why the group was flagged. |
+| `business_impact` | string | Likely downstream impact if the rows are processed unchanged. |
+| `action` | string | What an analyst should investigate. |
+| `top_features` | string | Deterministic top-2 contributing features (e.g. `amount+quantity`) — the group's pattern key. |
+| `group_size` | long | Number of anomalous rows in this `(segment, pattern)` group. |
+| `group_avg_severity` | double | Mean `severity_percentile` across the group. |
 
 **Access in PySpark:** use `F.element_at(F.col("_dq_info"), 1)` for the first element (1-based), then `.getField("anomaly").getField("severity_percentile")` etc. Alternatively `F.col("_dq_info").getItem(0)` for 0-based index (see [Troubleshooting](/docs/guide/row_anomaly_detection/troubleshooting) for Spark Connect–friendly patterns).
 
+### AI explanations (LLM)
```
mwojtyczka: a bit redundant
```suggestion
### AI explanations
```

src/databricks/labs/dqx/anomaly/check_funcs.py
mwojtyczka: I would go with option 1 - auto-enable

docs/dqx/docs/reference/quality_checks.mdx
```diff
@@ -3461,6 +3465,55 @@ display(
 )
 ```
 
+**AI Explanations (LLM)**
```
```suggestion
**AI Explanations**
```

demos/dqx_row_anomaly_detection_demo.py
vb-dbrks: Done — renamed to "Section 5a" in 018f3f1d. Thanks.

demos/dqx_row_anomaly_detection_demo.py
```diff
@@ -485,6 +485,63 @@ def inject_anomalies_and_dq_issues(
...
+checks_with_ai = [
+    DQDatasetRule(
+        check_func=has_no_row_anomalies,
+        check_func_kwargs={
+            "model_name": model_name_auto,
+            "registry_table": registry_table,
+            "enable_contributions": True,  # required for AI explanations
+            "enable_ai_explanation": True,
+            # model_name must resolve to a Databricks Model Serving endpoint
+            "llm_model_config": {"model_name": "databricks-claude-sonnet-4-5"},
```
mwojtyczka: we need to be specific to avoid confusion
```suggestion
            "ai_explanation_llm_model_config": {"model_name": "databricks-claude-sonnet-4-5"},
```

src/databricks/labs/dqx/anomaly/check_funcs.py
```diff
@@ -30,6 +114,10 @@ def has_no_row_anomalies(
...
```
mwojtyczka: to reduce default knobs for users
```suggestion
    enable_ai_explanation: bool = True,
```

src/databricks/labs/dqx/anomaly/check_funcs.py
```diff
@@ -30,6 +114,10 @@ def has_no_row_anomalies(
     drift_threshold: float | None = None,
     enable_contributions: bool = False,
```
mwojtyczka: to reduce default knobs for users
```suggestion
    enable_contributions: bool = True,
```

src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
vb-dbrks: I'd rather not add prompt-tuning knobs (top-N features, pattern slice, field length) to the public config in this PR — they're tuned for the small Databricks-hosted models and exposing them invites tuning/support questions without a concrete use case yet. Happy to surface them on `LLMModelConfig` in a follow-up if demand appears.

src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
vb-dbrks: Kept the WARNING log + docs note for now; held off on adding an `ai_explanation_truncation` struct to the result schema — that's a schema change I'd prefer not to bundle into this PR. Good follow-up if users need a programmatic signal rather than a log.

src/databricks/labs/dqx/anomaly/check_funcs.py
vb-dbrks: Skipping the perf test here — the cluster-side work is a groupBy + bounded top-N collect + one ai_query per group, and ai_query runs on the serving endpoint (not the cluster), so there's no real cluster-side hot path to benchmark, and a perf test bound to a live endpoint would be flaky. Added a prompt snapshot test for change-control instead.

src/databricks/labs/dqx/anomaly/check_funcs.py
```diff
@@ -72,6 +159,39 @@ def has_no_row_anomalies(
...
+                    "api_key": "<pat_token>",
+                    "api_base": "https://<gateway-id>.ai-gateway.cloud.databricks.com/mlflow/v1",
+                }
+
+            When ``api_base`` is set, the model is routed through litellm's OpenAI-compatible
+            adapter (``openai/<model>``) — use this for AI Gateway and other OpenAI-compatible
+            endpoints. ``api_key`` / ``api_base`` also fall back to ``OPENAI_API_KEY`` /
+            ``OPENAI_API_BASE`` env vars when unset on the config.
+        redact_columns: Feature names to exclude from the LLM prompt. Only feature names and SHAP
+            percentages are ever sent — raw data values are never included.
```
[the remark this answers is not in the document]
vb-dbrks: Fixed — the docstring no longer claims only feature names + SHAP are sent. It now states that segment `key=value` pairs reach the prompt verbatim, and `redact_columns` filters matching segment keys as well (emitted as `key=<redacted>`). The docs carry the same caveat.

src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
vb-dbrks: Done — the sanitiser now trims back to a word boundary when a field overruns the cap instead of slicing mid-word.

src/databricks/labs/dqx/anomaly/scoring_run.py
vb-dbrks: Went with the documentation option: the `max_groups` reference entry now states the cap is split across eligible segments with a one-call-per-segment floor, and a WARNING is logged when `max_groups < eligible segments`. I'd rather emit one explanation per segment (with a warning) than hard-error and block scoring.

src/databricks/labs/dqx/anomaly/check_funcs.py
vb-dbrks: This is your option 2 — `redact_columns` filters matching segment keys (not just SHAP feature names), so a sensitive segment column can be kept out of the prompt entirely. It's flagged in the docstring and in the Cost and privacy callout in the docs. I skipped the automatic per-run WARNING to avoid noise on every segmented run.

src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
vb-dbrks: Moot now — that `safe_pattern` snippet lived in the driver-path `_explain_one_group`, which the ai_query pivot removed. In the ai_query path the pattern key is constrained by the strict character class (which doubles as the anti-injection filter), and LLM output fields get the control-char strip as Spark expressions.

src/databricks/labs/dqx/config.py
vb-dbrks: Moot — the `api_base` allowlist plumbing was dropped entirely in the ai_query pivot. Explanations go through Databricks Model Serving via `ai_query` (no client-side HTTP from our code), so there's no `api_base_allowed_hosts` to extend anymore.

src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
vb-dbrks: Done — the guide documents the top-2 collision tradeoff: two anomaly types sharing the same top-2 contributors get one shared, slightly more general explanation, intentionally, to keep group cardinality (and LLM cost) down.

src/databricks/labs/dqx/anomaly/check_funcs.py
vb-dbrks: Ended up going default-on instead, which I think gets the same outcome with less magic: both flags now default to `True`, so explanations work out of the box and nobody hits the two-step opt-in. The only remaining case is someone *explicitly* setting `enable_contributions=False` — auto-enabling contributions there would override an explicit user choice, so instead explanations are downgraded with a warning (no error). Happy to revisit if you feel strongly about option 1.

src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
vb-dbrks: Addressed with an endpoint reachability probe — before issuing `ai_query` we probe the endpoint once with a 1-token call; if it's unreachable we log a WARNING ("endpoint not reachable; skipping AI explanations") and degrade to null explanations instead of failing or going silent. A down endpoint is no longer indistinguishable from "nothing anomalous".

src/databricks/labs/dqx/anomaly/check_funcs.py
vb-dbrks: Done — the Cost and privacy callout and the `max_groups` reference entry both state that each group is one LLM call and 500 is the per-run cap. Kept the default at 500 with the cost now documented. Also worth noting: the LLM results are now pinned once per scoring run, so repeated actions on the result no longer multiply the cost.

src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
vb-dbrks: Deferring to #1030 — preserving sign end-to-end needs the SHAP layer to stop taking abs() at compute time, which is a scoring-schema change bigger than this PR.

src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
vb-dbrks: Done — the instructions now include an explicit anti-pattern list (no hedging phrases, no restating input fields) and two few-shot exemplars (one with drift, one without) pinning the style and JSON shape.

tests/integration_anomaly/test_anomaly_ai_explanation.py
vb-dbrks: Done — added `test_ai_query_explanation_references_dominant_feature_within_word_caps` (deterministic outlier; asserts the narrative names the dominant feature and respects the word caps) plus a prompt snapshot test for change control.

src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
vb-dbrks: Done — `LLMModelConfig.__post_init__` now validates `max_tokens` (positive int) and `temperature` (non-negative), so bad values raise a clean `InvalidParameterError` at construction time instead of a raw `TypeError` at SQL-build time.

src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
vb-dbrks: Done — the response-format JSON now goes through `_sql_string_literal()` when interpolated into the `ai_query` SQL, so a future field name or description containing a quote can't break the statement.

docs/dqx/docs/reference/quality_checks.mdx
vb-dbrks: Done — renamed to `ai_explanation_llm_model_config` everywhere (param, docs, demo); the reference entry matches your suggested wording.

docs/dqx/docs/guide/row_anomaly_detection/index.mdx
vb-dbrks: Done — heading is now `### AI explanations`.

docs/dqx/docs/reference/quality_checks.mdx
vb-dbrks: Done.

demos/dqx_row_anomaly_detection_demo.py
vb-dbrks: Done — the demo uses `ai_explanation_llm_model_config` with the explicit endpoint name.

src/databricks/labs/dqx/anomaly/check_funcs.py
vb-dbrks: Done — defaults to `True` now.

src/databricks/labs/dqx/anomaly/check_funcs.py
vb-dbrks: Done — defaults to `True` now.

src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
mwojtyczka: sounds good, let's leave it for now

src/databricks/labs/dqx/anomaly/anomaly_llm_explainer.py
mwojtyczka: sounds good

src/databricks/labs/dqx/anomaly/check_funcs.py
mwojtyczka: yes, we are good with this. I think it's better to have both enabled by default, and keep the settings explicit, rather than auto-enabling

pyproject.toml
```diff
@@ -883,6 +884,10 @@ redefining-builtins-modules = ["six.moves", "past.builtins", "future.builtins",
 "tests/integration/test_app_backend.py" = "import-outside-toplevel"
 "src/databricks/labs/dqx/anomaly/anomaly_workflow.py" = "import-outside-toplevel"
 "src/databricks/labs/dqx/check_funcs.py" = "protected-access"
+"tests/unit/test_anomaly_llm_explainer.py" = "protected-access"
```
mwojtyczka, line 887: this is technical debt - should be addressed as a follow up

mwojtyczka (approved):
LGTM

PR #1158, opened 2026-05-05
feat: add is_valid_email row check function

### LINKED ISSUE #1156: [FEATURE]: Add email validation check function

### Problem statement

It's a common task to validate emails. Today, this can be implemented either as regex or sql expression checks but it's cumbersome.

### Proposed Solution

Add new check function to validate email.

### Additional Context

Follow RFC 5322

aarushisingh04:
### changes
added a new `is_valid_email` row-level check function that validates email addresses following a pragmatic subset of RFC 5321/5322. 

implementation:
- adds a ReDoS-safe `EMAIL_ADDRESS` pattern to the `DQPattern` registry.
- behaves identically to `is_valid_ipv4_address` by ignoring `null` values (no violation reported).
- can be used in programmatic `DQRowRule` / `DQForEachColRule` creation and declarative YAML metadata setups (`function: is_valid_email`).

### linked issues
resolves #1156

### tests

- [ ] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

### documentation and demos

- [ ] added/updated demos
- [ ] added/updated docs


tests/unit/test_row_checks.py
```diff
@@ -141,3 +142,16 @@ def test_sql_expression_complex_exists_negate_auto_name():
...
+    result = is_valid_email("email")
+    assert get_column_name_or_alias(result) == "email_does_not_match_pattern_email_address"
+
+
+def test_is_valid_email_with_column_expression():
+    import pyspark.sql.functions as F
+
+    col_expr = F.col("user_email")
+    result = is_valid_email(col_expr)
+    assert get_column_name_or_alias(result) == "user_email_does_not_match_pattern_email_address"
```
ghanse: I don't think these tests are needed. They are only testing the default alias creation which is tested elsewhere. We can omit them.

ghanse (changes requested):
Looks good overall. You will need to sign your commits (see the [contributor docs](https://databrickslabs.github.io/dqx/docs/dev/contributing/#first-contribution) for more info).

Can you add documentation for this function to `/docs/reference/quality_checks.mdx`? You will need to update the row-level checks table, Python examples, and YAML examples.

aarushisingh04:
@ghanse thanks for the review, i have addressed all of the above mentioned points

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -69,6 +75,24 @@ class DQPattern(Enum):
...
     IPV4_CIDR_BLOCK = rf"{IPV4_ADDRESS[:-1]}/{_IPV4_CIDR_SUFFIX}$"
+    # RFC 5322 pragmatic subset: dot-atom or quoted-string local part, dot-atom or
+    # IP-literal domain, with RFC 5321 length caps (local ≤ 64, total ≤ 254).
+    # Excludes CFWS, obsolete grammar, and SMTPUTF8/IDN. ReDoS-safe.
+    EMAIL_ADDRESS = (
+        rf"^(?=.{{1,254}}$)"
+        # Local part: dot-atom (with 64-char cap) or quoted-string
+        rf"(?:"
+        rf"(?=[^@]{{1,64}}@){_EMAIL_ATEXT}+(?:\.{_EMAIL_ATEXT}+)*"
+        rf'|"(?:{_EMAIL_QTEXT}|{_EMAIL_QPAIR})*"'
```
mwojtyczka, line 87: **Quoted-string branch bypasses the 64-char local-part cap.**

The `(?=[^@]{1,64}@)` lookahead on line 85 only guards the dot-atom branch. A 200-char quoted local with a short domain still passes the regex (the total-length cap of 254 only partially mitigates this), violating RFC 5321 §4.5.3.1.

Suggest hoisting the lookahead out of the alternation:

```python
rf"(?=[^@]{{1,64}}@)"
rf"(?:"
rf"{_EMAIL_ATEXT}+(?:\.{_EMAIL_ATEXT}+)*"
rf'|"(?:{_EMAIL_QTEXT}|{_EMAIL_QPAIR})*"'
rf")"
```

And add a test asserting a 65-char quoted local (`"aaaa..."@example.com` with 65 a's inside the quotes) is rejected.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -69,6 +75,24 @@ class DQPattern(Enum):
...
+        rf")"
+        rf"@"
+        # Domain: LDH hostname with alphabetic TLD, or IP-literal
+        rf"(?:"
+        rf"(?:{_EMAIL_DOMAIN_LABEL}\.)+[A-Za-z]{{2,63}}"
+        rf"|\[(?:{_IPV4_OCTET}(?:\.{_IPV4_OCTET}){{3}}|IPv6:[A-Fa-f0-9:]+)\]"
```
mwojtyczka, line 93: **IPv6 IP-literal is matched too loosely.**

`IPv6:[A-Fa-f0-9:]+` accepts malformed strings like `[IPv6:]`, `[IPv6:::::::::]`, or `[IPv6::::]`. The docstring already calls this out and recommends `ipaddress.IPv6Address` for semantic validation, so this is documented behaviour — but please add at least one negative test case (e.g. `user@[IPv6:::::::::]`) so the documented behaviour is locked in. Long-term, consider tightening the pattern to enforce at least colon-separated hex groups.

mwojtyczka (commented):
Solid contribution overall — well-tested, well-documented, and follows the existing `_matches_pattern` shape. Posting 6 inline comments (1 correctness gap, several polish items). One additional ask below.

### Add a performance test
Regex-based row checks like this can have non-obvious cost on adversarial inputs. Could you add a perf test under `tests/perf/` benchmarking `is_valid_email` against a representative mix (valid dot-atom, quoted-local, IP-literal, long-but-rejected strings) so we have a baseline before merge? The existing IPv4/IPv6 perf cases are a good template.

src/databricks/labs/dqx/check_funcs.py
mwojtyczka, line 93: **(Cosmetic) Captured-group count grows.**

`_IPV4_OCTET = r"(25[0-5]|2[0-4]\d|1\d{2}|[1-9]?\d)"` is a *capturing* group. Each reuse here adds 4 more capture groups beyond what's needed (the existing `IPV4_ADDRESS` does the same). No correctness impact, but switching `_IPV4_OCTET` to non-capturing `(?:…)` once at its definition would clean this up across all callers. Optional; can be a follow-up.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -29,6 +29,12 @@
 IPV4_MAX_OCTET_COUNT = 4
 IPV4_BIT_LENGTH = 32
 
+# Email helpers (RFC 5322 §3.2.3, §3.2.4 + RFC 5321 §4.1.3, §4.5.3.1).
+_EMAIL_ATEXT = r"[A-Za-z0-9!#$%&'*+/=?^_`{|}~-]"
+_EMAIL_DOMAIN_LABEL = r"[A-Za-z0-9](?:[A-Za-z0-9-]{0,61}[A-Za-z0-9])?"
+_EMAIL_QTEXT = r"[\x21\x23-\x5B\x5D-\x7E]"  # printable ASCII except '"' (0x22) and '\' (0x5C)
+_EMAIL_QPAIR = r"\\[\x09\x20-\x7E]"  # quoted-pair: '\' + VCHAR or WSP
```
mwojtyczka: **Nit: clarify scope of `_EMAIL_QPAIR`.**

A one-line comment noting that `_EMAIL_QPAIR` is only valid *inside* the quoted-string production (using it bare in dot-atom would be malformed) would help future readers who copy-paste this constant. Same applies to `_EMAIL_QTEXT`.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -971,6 +995,38 @@ def is_valid_ipv4_address(column: str | Column) -> Column:
...
+      alphabetic TLD, or an IP-literal (RFC 5321 §4.1.3) — bracketed IPv4
+      with octet validation, or *[IPv6:...]* matched syntactically only
+      (callers requiring semantic IPv6 validation should re-check via
+      *ipaddress.IPv6Address*).
+    * Length caps per RFC 5321 §4.5.3.1: local-part max 64 octets, total
+      address max 254 octets.
+
+    Comments and folding whitespace (CFWS), obsolete grammar
+    (*obs-local-part*, *obs-domain*, *obs-qtext*, *obs-qp*), and
+    internationalized addresses (RFC 6531 / SMTPUTF8) are not supported;
```
mwojtyczka, line 1017: **Worth noting punycode/IDN homograph behaviour.**

Since the domain regex gates on ASCII LDH labels, punycode-encoded IDNs (`xn--…`) are accepted as ASCII without any check that they decode to a valid Unicode label. A short note in the docstring next to the SMTPUTF8 line — "punycode-encoded IDN labels (`xn--…`) pass syntactically; no homograph protection" — would set correct expectations for users dealing with internationalised data.

tests/integration/test_row_checks.py
```diff
@@ -1842,6 +1843,108 @@ def test_col_is_valid_ipv4_address(spark):
...
+
+    test_df = spark.createDataFrame(
+        [
+            ["user@example.com"],
+            ["User.Name+tag@sub.domain.org"],
+            ["a@b.cd"],  # minimum valid
+            ["1234567890@numbers.io"],  # all-numeric local
+            ["mixed.case@MixedCase.Org"],  # case variations
+            [local_part_at_char_limit + "@b.com"],  # exactly at 64-character limit for local parts
+            [full_text_at_char_limit],  # exactly at 268-character limit for the full text
```
mwojtyczka: **Stale comment: 268 → 254.**

`full_text_at_char_limit = "a" * 64 + "@" + "b" * 186 + ".co"` is 64 + 1 + 186 + 3 = **254** characters, matching the RFC 5321 §4.5.3.1 cap. The `# exactly at 268-character limit` comment is wrong — please update to `254` to avoid confusing future readers.

mwojtyczka (changes requested):
Thanks for your contribution. The PR is generally close to be ready. I left some comments. Please add also one performance test and add the check here: https://github.com/databrickslabs/dqx/blob/main/tests/resources/all_row_checks.yaml

aarushisingh04:
@ghanse thanks for the ongoing fixes. to avoid potential duplication of efforts, shall i address the comments above that are to be resolved? i am ready to do it, please let me know if you have a different plan.

tests/integration/test_row_checks.py
ghanse: Updated the comment

src/databricks/labs/dqx/check_funcs.py
ghanse: Updated the docstring notes

src/databricks/labs/dqx/check_funcs.py
ghanse: Added a note in the docstring for this

src/databricks/labs/dqx/check_funcs.py
ghanse: Updated the regex and added a test case

ghanse:
> @ghanse thanks for the ongoing fixes. to avoid potential duplication of efforts, shall i address the comments above that are to be resolved? i am ready to do it, please let me know if you have a different plan.

I think this looks good now. Made a few small changes to the pattern and added a few test cases.

tests/perf/test_apply_checks.py
```diff
@@ -2091,3 +2091,29 @@ def test_benchmark_has_no_aggr_outliers(benchmark, ws, generated_df):
...
+    ],
+)
+@pytest.mark.benchmark(group="test_benchmark_is_valid_email_address")
+def test_benchmark_is_valid_email_address(benchmark, ws, generated_email_df, column):
+    dq_engine = DQEngine(workspace_client=ws, extra_params=EXTRA_PARAMS)
+    checks = [
+        DQRowRule(
+            name=f"{column}_is_valid_email_address",
+            criticality="warn",
+            check_func=check_funcs.is_valid_email_address,
```
mwojtyczka: The name is wrong. It should be is_valid_email
```suggestion
            check_func=check_funcs.is_valid_email,
```

tests/perf/test_apply_checks.py
```diff
@@ -2091,3 +2091,29 @@ def test_benchmark_has_no_aggr_outliers(benchmark, ws, generated_df):
     checked = dq_engine.apply_checks(generated_df, checks)
     actual_count = benchmark(lambda: checked.count())
     assert actual_count == EXPECTED_ROWS
+
+
+@pytest.mark.parametrize(
+    "column",
+    [
+        "col1_ipv4_standard",
```
mwojtyczka: The generated_email_df fixture produces `col1_email_standard`, `col2_email_with_quoted_local_part`, `col3_email_with_ip_domain`, `col4_email_with_multi_part_domain`. The benchmark targets columns that don't exist in the fixture

tests/perf/conftest.py
```diff
@@ -285,3 +285,28 @@ def generated_timestamp_df(request, spark):
...
+        "col2_email_with_quoted_local_part: string, "
+        "col3_email_with_ip_domain: string, "
+        "col4_email_with_multi_part_domain: string"
+    )
+    schema = _parse_datatype_string(email_schema_str)
+
+    email_templates = {
+        "col1_email_standard": r"kkkkkkkkkkkk@kkkkkkkk.org",
+        "col2_email_with_quoted_local_part": r"\"kkkkkkkkkkkk\"@kkkkkkkk.org",
+        "col3_email_with_ip_domain": r"kkkkkkkkkkkk@\[\n.\n.\n.\n\]",
```
mwojtyczka: I think  dbldatagen template grammar uses `\\n `(escaped) for "digit", not  `\n` which is a literal newline character

mwojtyczka (changes requested):

mwojtyczka:
@ghanse @aarushisingh04 perf tests have some small issues, left comments. Otherwise LGTM

tests/perf/test_apply_checks.py
ghanse: Updated

tests/perf/test_apply_checks.py
ghanse: Fixed the column names in `@pytest.mark.parametrize`

tests/perf/conftest.py
ghanse: Updated the templates

mwojtyczka (approved):
LGTM

PR #1164, opened 2026-05-07
[FEATURE]: Add geofencing checks

### LINKED ISSUE #604: [FEATURE]: Add geofencing checks

### Problem statement

I want to add more precise geofencing checks to my pipelines. I have the following use cases in mind:
1) Given a reference polygon, check if the geom/geog is contained by the reference polygon
2) Given a reference point, check if the geog/geog is within a specified distance of the reference point

### Proposed Solution

Add new checks that cover these use cases

### Additional Context

[geofence](https://en.wikipedia.org/wiki/Geofence)

IvannKurchenko:
## Changes
As per proposal #604 added geo-fencing checks based on `ST_*` family of functions (`is_within_polygon_precise`) and `H3_*` functions (`is_within_polygon_approximate`).

Original discussion suggested differentiate approaches by specifying argument for a check. Please, see the following comment: https://github.com/databrickslabs/dqx/issues/604#issuecomment-3400468470

Although, I found the way of two geo-fence approaches too different to align in a single method. For example, H3 index operations with `resolution` notion which is no applicable for WGS-84.

### Linked issues

Resolves #604

### Tests

- [ ] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [x] added performance tests

### Documentation and Demos

- [x] added/updated demos
- [x] added/updated docs


src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -918,3 +931,189 @@ def apply(df: DataFrame) -> DataFrame:
...
+
+    return condition, apply
+
+
+@register_rule("row")
+def is_within_polygon_approximate(
+    column: str | Column,
+    reference_polygon: str | Column,
+    resolution: int | Column,
+):
```
ghanse: Should we include `topological_relationship` for approximate polygon checks?

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -17,6 +19,17 @@
 GEOMETRYCOLLECTION_TYPE = "ST_GeometryCollection"
 DEFAULT_SRID = 4326
 
+_TOPOLOGICAL_RELATIONSHIPS = frozenset({"CONTAINS", "COVERS", "WITHIN"})
```
ghanse: What about other relationships e.g. `ST_Intersects()` and `ST_Touches()`?

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -918,3 +931,189 @@ def apply(df: DataFrame) -> DataFrame:
     )
 
     return condition, apply
+
+
+@register_rule("dataset")
```
ghanse: I think the intent should be to implement these as row-level checks where the input `column` contains geometry WKB/WKT/GeoJSON. Similar to the other row-level spatial checks.

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -918,3 +931,189 @@ def apply(df: DataFrame) -> DataFrame:
...
+    InvalidParameterError.require(
+        topological_relationship in _TOPOLOGICAL_RELATIONSHIPS,
+        f"'topological_relationship' must be one of {sorted(_TOPOLOGICAL_RELATIONSHIPS)}, got '{topological_relationship}'.",
+    )
+
+    col_str_norm, col_expr_str, col_expr = get_normalized_column_and_expr(column)
+    unique_str = uuid.uuid4().hex
+    condition_col = f"__is_within_polygon_precise_{col_str_norm}_{unique_str}"
+
+    def apply(df: DataFrame) -> DataFrame:
```
ghanse: Given these are row-level checks, we should not need to use an `apply` method and should be able to use `make_condition` directly. The other row level checks in this module have good examples of this.

src/databricks/labs/dqx/errors.py
```diff
@@ -13,6 +13,12 @@ class InvalidConfigError(DQXError):
 class InvalidParameterError(ParameterError):
     """Raised when a parameter is invalid (malformed, wrong type, not supported, ambiguous, or incompatible with other inputs)."""
 
+    @classmethod
+    def require(cls, condition: bool, error: str) -> None:
+        """Raise this error type if condition is False."""
+        if not condition:
+            raise cls(error)
+
```
ghanse: Not sure this is needed. I can see how it's useful, but I think our error conditions are simple enough to write without a helper.

ghanse (changes requested):
Thanks for working on this, it will be a good addition. Left a few comments.

src/databricks/labs/dqx/geo/check_funcs.py
IvannKurchenko: Thanks for the feedback, these have been added.

src/databricks/labs/dqx/geo/check_funcs.py
IvannKurchenko: Thank you for spotting, sorry this was silly mistake. Fixed.

src/databricks/labs/dqx/geo/check_funcs.py
IvannKurchenko: Thanks, this was refactored

src/databricks/labs/dqx/errors.py
IvannKurchenko: Thanks, removed.

src/databricks/labs/dqx/geo/check_funcs.py
IvannKurchenko: Thanks for the suggestion, it make sense generally. Although I need to think a bit, because this is less trivial with H3 index.

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -824,7 +836,7 @@ def _compare_spatial_sql_function_result(
     )
 
 
-@register_rule("dataset")
+@register_rule("row")
```
mwojtyczka: pls revert this change

tests/integration/test_dataset_checks_geo.py
```diff
@@ -526,3 +531,249 @@ def test_are_polygons_mutually_disjoint_row_filter_with_duplicates(skip_if_runti
...
+            False,
+        ),
+        ("POINT(4.90 52.37)", "POLYGON((4.73 52.28, 5.05 52.28, 5.05 52.43, 4.73 52.43, 4.73 52.28))", "COVERS", False),
+        (
+            "POINT(4.90 52.37)",
+            "POLYGON((4.73 52.28, 5.05 52.28, 5.05 52.43, 4.73 52.43, 4.73 52.28))",
+            "INTERSECTS",
+            False,
+        ),
+        ("POINT(4.90 52.37)", "POLYGON((4.73 52.28, 5.05 52.28, 5.05 52.43, 4.73 52.43, 4.73 52.28))", "TOUCHES", True),
```
mwojtyczka: **WITHIN — the default value of `topological_relationship` — is not tested.**

All three parametrize blocks (`test_is_within_polygon_precise_inside`, `_edge`, `_outside`) cover `CONTAINS`, `COVERS`, `INTERSECTS`, `TOUCHES`, but skip `WITHIN`. The most-used code path is unverified.

Please add a `WITHIN` row to each of the three parametrize lists:
- inside — `expect_violation=False`
- edge   — `expect_violation=True` (boundary excluded for WITHIN)
- outside — implicitly via the parametrize on line 655

Line 655's parametrize also needs `"WITHIN"` added to the relationship list.

docs/dqx/docs/reference/quality_checks.mdx
```diff
@@ -1821,6 +1821,8 @@ You can also define your own custom dataset-level checks (see [Creating custom c
...
...                                                                                                                                                                                                                                                                                                                                | `column`: column to check (can be a string column name or a column expression); `reference_polygon`: reference polygon as a literal WKT/WKB/EWKT/EWKB string or a column name; `convert_column`: when `True`, applies `try_to_geometry` to convert the column values to GEOMETRY (default `False`); `convert_reference_polygon`: when `True`, applies `try_to_geometry` to convert the reference polygon to GEOMETRY (default `False`); `topological_relationship`: spatial predicate — `WITHIN` (default), `CONTAINS`, or `COVERS`                                                                                                                                                                                                                                                                                                                                                                          |
```
mwojtyczka: **Docs lists only 3 of the 5 supported predicates.**

The table reads `topological_relationship: spatial predicate — \`WITHIN\` (default), \`CONTAINS\`, or \`COVERS\``. The implementation also accepts `INTERSECTS` and `TOUCHES` (see `_TOPOLOGICAL_RELATIONSHIPS` and the function docstring). Update the reference to list all five so users discover them.

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -918,3 +930,133 @@ def apply(df: DataFrame) -> DataFrame:
...
+    is_inside = F.call_function(_TOPOLOGICAL_FUNCS[topological_relationship], ref_geom, col_geom)
+
+    condition = F.when(col_expr.isNull(), F.lit(None)).otherwise(~is_inside)
+
+    return make_condition(
+        condition,
+        F.concat_ws(
+            "",
+            F.lit("value `"),
+            col_expr.cast("string"),
```
mwojtyczka: **Casting GEOMETRY to string for the error message produces unreadable output.**

When `convert_column=False`, `col_expr` holds a native `GEOMETRY` value. `cast("string")` on a geometry doesn't produce WKT — users will see something like `value \`[B@1f3d...\` in column...`, which defeats the purpose of including the value in the violation message.

Suggest using `st_astext` so both code paths produce readable WKT:

```python
F.call_function("st_astext", col_expr) if not convert_column else col_expr.cast("string")
```

…or call `st_astext` unconditionally — it's a no-op cost-wise after `try_to_geometry`, and the output is consistently WKT regardless of input shape.

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -918,3 +930,133 @@ def apply(df: DataFrame) -> DataFrame:
...
     return condition, apply
+
+
+@register_rule("row")
+def is_within_polygon_precise(
+    column: str | Column,
+    reference_polygon: str | bytearray | Column,
+    convert_column: bool = False,
+    convert_reference_polygon: bool = False,
+    topological_relationship: (
```
mwojtyczka: **Verbose `Literal` union syntax — collapse into a single `Literal`.**

```python
topological_relationship: Literal["CONTAINS", "COVERS", "INTERSECTS", "TOUCHES", "WITHIN"] = "WITHIN",
```

Same semantics, easier to read, and matches the conventional `typing.Literal` usage. The current `Literal["X"] | Literal["Y"] | …` form forces type-checkers into extra union narrowing for no benefit.

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -918,3 +930,133 @@ def apply(df: DataFrame) -> DataFrame:
     )
 
     return condition, apply
+
+
+@register_rule("row")
+def is_within_polygon_precise(
+    column: str | Column,
+    reference_polygon: str | bytearray | Column,
```
mwojtyczka: **Nit: `bytearray` is an unusual type for binary input.**

The conventional binary type in Python is `bytes` — it's immutable and what most callers will pass when constructing a WKB literal. `bytearray` is mutable and rarely used as a typed parameter. Consider `str | bytes | Column` (or include both `bytes | bytearray` if you genuinely want to accept either).

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -918,3 +930,133 @@ def apply(df: DataFrame) -> DataFrame:
     )
 
     return condition, apply
+
+
+@register_rule("row")
+def is_within_polygon_precise(
```
mwojtyczka: **Add a performance test before merge.**

The PR's task list shows perf tests as unchecked. Geo functions on large datasets are exactly where regressions hurt — `ST_*` predicates over polygons can be expensive, and H3 `array_contains` cost depends heavily on resolution. Please add a perf test under `tests/perf/` covering both checks at representative dataset sizes; the existing geo perf cases (`are_polygons_mutually_disjoint` etc.) are a template.

mwojtyczka (commented):
Solid implementation overall — good docstrings, clean parameter validation, sensible reuse of `make_condition` and the H3/ST_* function families. Posting inline comments below covering one critical test gap, a docs/code mismatch, and several polish items.

Note: my earlier draft flagged the error-message format (`` ``value `X` in column `Y` `` ``) as inconsistent with the rest of DQX, but on re-reading I see this is the existing convention throughout `geo/check_funcs.py` (see `is_latitude` etc.) — so the PR is consistent with its own module. Dropping that comment.

The most critical issue I raised (the `are_polygons_mutually_disjoint` decorator change from `dataset` → `row` while the function still returns `(condition, apply)` — looks like an accidental breakage) is **not** posted inline at your request. Please double-check that change before merge — it appears unrelated to the geofencing work in this PR.

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -918,3 +930,133 @@ def apply(df: DataFrame) -> DataFrame:
...
+            F.lit(f"` in column `{col_expr_str}` is outside the reference polygon"),
+        ),
+        f"{col_str_norm}_outside_reference_polygon_precise",
+    )
+
+
+@register_rule("row")
+def is_within_polygon_approximate(
+    column: str | Column,
+    reference_polygon: str | Column,
```
mwojtyczka: **Approximate check is silently WITHIN-only — worth documenting.**

The precise version takes a `topological_relationship` parameter; the approximate version uses `array_contains` and is implicitly WITHIN-only. The PR description acknowledges that the two approaches don't fit a single signature, which is reasonable, but a one-line note in the docstring would help users who go looking for the parameter:

> Unlike *is_within_polygon_precise*, this function is WITHIN-only by design — H3 cells don't preserve the full set of topological predicates.

tests/integration/test_dataset_checks_geo.py
```diff
@@ -526,3 +531,249 @@ def test_are_polygons_mutually_disjoint_row_filter_with_duplicates(skip_if_runti
...
+
+
+@pytest.mark.parametrize(
+    "column_value,reference_polygon_value,resolution",
+    [
+        ("POINT(4.90 52.37)", "POLYGON((4.73 52.28, 5.05 52.28, 5.05 52.43, 4.73 52.43, 4.73 52.28))", 5),
+        ("POINT(4.90 52.37)", "POLYGON((4.73 52.28, 5.05 52.28, 5.05 52.43, 4.73 52.43, 4.73 52.28))", 7),
+    ],
+)
+def test_is_within_polygon_approximate_inside(
```
mwojtyczka: **Missing test for the precision divergence the docstring promises.**

The `is_within_polygon_approximate` docstring says it's *"less precise than `is_within_polygon_precise` near polygon boundaries"*. That contract is currently unverified — there's no test demonstrating a boundary-adjacent point where the two checks disagree.

Please add at least one test case showing:
- a point just outside the polygon at low H3 resolution (e.g. resolution 5) → approximate accepts (false positive), precise rejects;
- the same point at high resolution (e.g. 12) → both reject.

This locks in the documented tradeoff and helps users understand when to pick which check.

mwojtyczka (changes requested):
Thanks for the PR!

I left some comments. Please also add performance test and an example here https://github.com/databrickslabs/dqx/blob/main/tests/resources/all_row_geo_checks.yaml

src/databricks/labs/dqx/geo/check_funcs.py
IvannKurchenko: Thank you for the comment once again. Since H3 index functions behave somewhat differently from ST_ I've added support for `INTERSECTS` and `COVERS`. Although they do not precisely correspond to ST behaviour, they imitate it to a certain extent.

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -918,3 +932,171 @@ def apply(df: DataFrame) -> DataFrame:
     )
 
     return condition, apply
+
+
+@register_rule("row")
+def has_topological_relationship_precise(
```
ghanse: It might be best for each topological relationship to have its own check and delegate to `_has_topological_relationship_...` methods as internal helpers. Something like:

```
@register_rule("row")
def is_within(
  column: str | Column,
  precise: bool = False,
  reference_geometry: str | bytearray | Column,
  convert_column: bool = False,
  convert_reference_geometry: bool = False
) -> Column:
  if precise:
    return _has_topological_relationship_precise(...)
  return _has_topological_relationship_approximate(...)
```

I'm thinking this would offer a few benefits:
- Check functions are more discoverable and aligned to industry-standard `ST_...` functions
- Check functions are scoped to a single topological relationship and can be easily tested
- Fewer keyword arguments required for users

ghanse (dismissed):

tests/integration/test_dataset_checks_geo.py
IvannKurchenko: Thanks, this test was added.

docs/dqx/docs/reference/quality_checks.mdx
IvannKurchenko: Thanks, documented.

src/databricks/labs/dqx/geo/check_funcs.py
IvannKurchenko: Thanks for the comment, replaced.

src/databricks/labs/dqx/geo/check_funcs.py
IvannKurchenko: Thanks, fixed.

src/databricks/labs/dqx/geo/check_funcs.py
IvannKurchenko: Done, thanks

src/databricks/labs/dqx/geo/check_funcs.py
IvannKurchenko: Added, thanks

tests/integration/test_dataset_checks_geo.py
IvannKurchenko: Thanks, added.

src/databricks/labs/dqx/geo/check_funcs.py
IvannKurchenko: Thanks for the comment, this is done with small difference - I've added `geo` to this functions, because I thought that using more general names might lead to potential conflicts. So names are `is_geo_within`, `is_geo_covers` etc.

IvannKurchenko:
@ghanse @mwojtyczka Thank you so much for your review and comments, this is very valuable input. I apologise for some silly mistake  I made or things missed. Mind to review latests changes, please? Thanks.

tests/perf/test_apply_checks.py
```diff
@@ -2091,3 +2091,58 @@ def test_benchmark_has_no_aggr_outliers(benchmark, ws, generated_df):
     checked = dq_engine.apply_checks(generated_df, checks)
     actual_count = benchmark(lambda: checked.count())
     assert actual_count == EXPECTED_ROWS
+
+
+def test_benchmark_is_geo_covers_precise(benchmark, ws, generated_df):
+    """Benchmark is_geo_covers check on.
```
mwojtyczka: **Bug — copy-paste docstring.** The docstring describes col6/timestamp/45k day-buckets/groupBy-Window/broadcast-join behaviour — none of which this benchmark does. It only calls `is_geo_covers` on `col_geo_point`. Looks lifted verbatim from an aggregate benchmark.

Also the opening sentence `"Benchmark is_geo_covers check on."` reads like a truncated copy-paste. Same issue in `test_benchmark_is_geo_covers_approximate` at line ~2124.

Replace with a one-line description of what this benchmark actually measures (e.g., per-row `st_covers` evaluation cost on a fixed-size geometry input).

tests/perf/conftest.py
```diff
@@ -110,6 +110,7 @@ def generated_df(spark, rows=DEFAULT_ROWS):
         .withColumnSpec("col_ipv4", template=r"\n.\n.\n.\n")
         .withColumnSpec("col_ipv6", template="XXXX:XXXX:XXXX:XXXX:XXXX:XXXX:XXXX:XXXX")
         .withColumnSpec("col_json_str", template=r"{'key1': '\w', 'key2': 'd\w'}")
+        .withColumnSpec("col_geo_point", values=["POINT(4.48 51.92)"])
```
mwojtyczka: **Bug — every row is the same value.** `values=["POINT(4.48 51.92)"]` forces every row to the identical exterior point, which the Spark/Photon optimiser can constant-fold. The benchmark then measures one geometry evaluation amortised over N identical inputs — not realistic perf.

Mix interior/boundary/exterior locations (and ideally a few polygon shapes too) so the optimiser can't short-circuit:
```python
.withColumnSpec(
    "col_geo_point",
    values=[
        "POINT(4.90 52.37)",   # inside
        "POINT(4.73 52.28)",   # boundary
        "POINT(4.48 51.92)",   # outside
        "POINT(5.20 52.35)",   # near-boundary
    ],
)
```

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -918,3 +932,317 @@ def apply(df: DataFrame) -> DataFrame:
...
+
+    col_str_norm, col_expr_str, col_expr = get_normalized_column_and_expr(column)
+
+    col_geom = F.call_function("try_to_geometry", col_expr) if convert_column else col_expr
+    ref_col = reference_geometry if isinstance(reference_geometry, Column) else F.lit(reference_geometry)
+    ref_geom = F.call_function("try_to_geometry", ref_col) if convert_reference_geometry else ref_col
+
+    has_relationship = F.call_function(_PRECISE_TOPOLOGICAL_FUNCS[topological_relationship], ref_geom, col_geom)
+    condition = F.when(col_expr.isNull(), F.lit(None)).otherwise(~has_relationship)
+    text_value_col = F.call_function("st_astext", col_expr) if not convert_column else col_expr.cast("string")
```
mwojtyczka: **Bug — error message is garbage for WKB bytes input.** When `convert_column=True` the caller can pass WKT (string), EWKT, WKB (bytes), or EWKB. The branch `col_expr.cast("string")` on bytes produces something like `[B@1a2b3c4d` or raw binary in the error message — unreadable for operators triaging quarantine.

Use `st_astext(try_to_geometry(col_expr))` unconditionally — it produces a human-readable WKT regardless of input format:
```python
text_value_col = F.call_function("st_astext", F.call_function("try_to_geometry", col_expr))
```

Minor secondary win: this also normalises EWKT/EWKB-prefixed inputs into plain WKT in the error log, which is friendlier to grep.

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -918,3 +932,317 @@ def apply(df: DataFrame) -> DataFrame:
...
+) -> Column:
+    if topological_relationship not in _PRECISE_TOPOLOGICAL_RELATIONSHIPS:
+        raise InvalidParameterError(
+            f"'topological_relationship' must be one of {sorted(_PRECISE_TOPOLOGICAL_RELATIONSHIPS)}, got '{topological_relationship}'."
+        )
+
+    col_str_norm, col_expr_str, col_expr = get_normalized_column_and_expr(column)
+
+    col_geom = F.call_function("try_to_geometry", col_expr) if convert_column else col_expr
+    ref_col = reference_geometry if isinstance(reference_geometry, Column) else F.lit(reference_geometry)
```
mwojtyczka, line 978: **Doc mismatch — `reference_geometry` as a column name doesn't work.** Every check's docstring (and the .mdx tables) says:
> `reference_geometry`: reference geometry as a literal WKT/WKB/EWKT/EWKB string **or a column name**

But this line wraps any non-`Column` value with `F.lit(...)`, so a `str` is always interpreted as a literal — never as a column reference. Callers must explicitly pass `F.col("ref_col")`.

Two options:
1. Fix the docstrings everywhere: drop "or a column name" — clarify that string = literal, `Column` = column reference (matches actual behaviour).
2. Add a heuristic: `isinstance(ref, str) and ref in df.columns` → treat as column. Brittle (requires DataFrame context not available here), so option 1 is the realistic fix.

Same mismatch in `_has_topological_relationship_approximate` at line 985.

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -918,3 +932,317 @@ def apply(df: DataFrame) -> DataFrame:
...
+    Note:
+        This function requires Databricks serverless compute or runtime 17.1 or above.
+    """
+    return _has_topological_relationship_precise(
+        column, reference_geometry, convert_column, convert_reference_geometry, "TOUCHES"
+    )
+
+
+@register_rule("row")
+def is_geo_within(
```
mwojtyczka, line 1272: **API confusion — `is_geo_contains` vs `is_geo_within` argument-direction asymmetry.**
- `is_geo_contains(col, ref)` → "ref contains col" — reference is the outer container.
- `is_geo_within(col, ref)` → "ref is within col" — column is the outer container.

These read like English duals ("A contains B" ↔ "B within A"), but the function signatures share the same `(col, ref)` shape. A user who knows `is_geo_contains(point_col, polygon)` flags points outside the polygon will reasonably assume `is_geo_within(point_col, polygon)` does the same — but it doesn't (it asks whether the polygon literal sits inside the column geometry).

The integration test name `test_is_geo_within_interior_reference_no_violation` hints at this, but the function-level docstring needs a prominent callout:
```
Note: the column geometry is the OUTER container in this check; the
reference is the inner test geometry. Contrast with is_geo_contains
where the reference is the outer container.
```
Or (better but breaking) rename: `is_geo_contains_reference` / `is_geo_within_reference` to make the direction explicit.

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -918,3 +932,317 @@ def apply(df: DataFrame) -> DataFrame:
...
+
+    return make_condition(
+        condition,
+        F.concat_ws(
+            "",
+            F.lit("value `"),
+            text_value_col,
+            F.lit(f"` in column `{col_expr_str}` has no relationship with the reference geometry"),
+        ),
+        f"{col_str_norm}_outside_reference_geometry",
```
mwojtyczka: **Error alias `_outside_reference_geometry` is wrong for `touches` and `within`.**
- For `is_geo_touches`: a failing value is *not* "outside" — it's "doesn't share a boundary point with".
- For `is_geo_within`: a failing value is a column geometry that doesn't contain the reference — also not "outside".

Specialise per relationship — both for the alias and the message at line 965:
```python
_ALIAS_SUFFIX = {
    "CONTAINS":   "not_in_reference_geometry",
    "COVERS":     "not_covered_by_reference_geometry",
    "INTERSECTS": "does_not_intersect_reference_geometry",
    "TOUCHES":    "does_not_touch_reference_geometry",
    "WITHIN":     "does_not_contain_reference_geometry",
}
_MESSAGE_SUFFIX = {...}
```

This matters for users grepping quarantine columns by name — `_outside_reference_geometry` is misleading for 2 of the 5 checks.

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -918,3 +932,317 @@ def apply(df: DataFrame) -> DataFrame:
...
+
+    return make_condition(
+        condition,
+        F.concat_ws(
+            "",
+            F.lit("value `"),
+            col_expr.cast("string"),
+            F.lit(f"` in column `{col_expr_str}` is approximately outside the reference geometry"),
+        ),
+        f"{col_str_norm}_outside_reference_geometry_approximate",
```
mwojtyczka: **Approximate-mode message has the same issue.** `"is approximately outside the reference geometry"` is correct for COVERS-fail but wrong for INTERSECTS-fail (which means "shares no H3 cells with" — not the same as "outside"). Specialise like the precise version (see comment above).

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -918,3 +932,317 @@ def apply(df: DataFrame) -> DataFrame:
...
+        ),
+        f"{col_str_norm}_outside_reference_geometry",
+    )
+
+
+def _has_topological_relationship_approximate(
+    column: str | Column,
+    reference_geometry: str | Column,
+    resolution: int | Column,
+    topological_relationship: Literal["COVERS"] | Literal["INTERSECTS"] = "COVERS",
```
mwojtyczka: **Style — `Literal["COVERS"] | Literal["INTERSECTS"]` should be `Literal["COVERS", "INTERSECTS"]`.** `Literal` accepts multiple args directly; the union form is just noisier and Mypy treats them as equivalent.

Line 942 already uses the multi-arg form correctly — keep it consistent.

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -918,3 +932,317 @@ def apply(df: DataFrame) -> DataFrame:
...
+        F.call_function("h3_try_coverash3", col_expr, resolution_col),
+        F.array(F.call_function("h3_pointash3", col_expr, resolution_col)),
+    )
+    ref_h3_array = F.coalesce(
+        F.call_function("h3_try_coverash3", ref_col, resolution_col),
+        F.array(F.call_function("h3_pointash3", ref_col, resolution_col)),
+    )
+    intersection_size = F.size(F.array_intersect(col_h3_array, ref_h3_array))
+
+    match topological_relationship:
```
mwojtyczka, line 1038: **`match` without default branch.** If `topological_relationship` is somehow neither `"INTERSECTS"` nor `"COVERS"`, `is_inside` is unbound and line 1004 raises `UnboundLocalError` rather than DQX's `InvalidParameterError`. The earlier validation at line 979 guards this today, but it's a fragile invariant — a future refactor that splits validation could expose the gap.

Add a default:
```python
case _:
    raise InvalidParameterError(
        f"Unsupported approximate relationship: {topological_relationship!r}"
    )
```

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -918,3 +932,317 @@ def apply(df: DataFrame) -> DataFrame:
...
+        case "INTERSECTS":
+            is_inside = intersection_size > 0
+        case "COVERS":
+            is_inside = intersection_size == F.size(col_h3_array)
```
mwojtyczka, line 1042: **`array_intersect` deduplicates — equality check is fragile.** This line compares `size(intersect)` to `size(col_h3_array)`. `array_intersect` returns distinct elements; `F.size(col_h3_array)` counts with duplicates. If `h3_try_coverash3` ever returns duplicate cell IDs for complex multi-polygon geometries (the H3 spec doesn't forbid it, even if the implementation doesn't), `intersection_size == size(col_h3_array)` fails spuriously.

Apply `array_distinct` first to make set semantics explicit:
```python
col_h3_array = F.array_distinct(F.coalesce(...))
ref_h3_array = F.array_distinct(F.coalesce(...))
intersection_size = F.size(F.array_intersect(col_h3_array, ref_h3_array))
is_inside = intersection_size == F.size(col_h3_array)
```

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -918,3 +932,317 @@ def apply(df: DataFrame) -> DataFrame:
...
+            F.lit("value `"),
+            col_expr.cast("string"),
+            F.lit(f"` in column `{col_expr_str}` is approximately outside the reference geometry"),
+        ),
+        f"{col_str_norm}_outside_reference_geometry_approximate",
+    )
+
+
+@register_rule("row")
+def is_geo_contains(
```
mwojtyczka, line 1068: **No runtime gate for DBR 17.1+.** Every check docstring says *"Requires Databricks serverless compute or runtime 17.1 or above"*. On older runtimes, callers get a cryptic `AnalysisException: function 'st_contains' does not exist` from Spark at apply-time — not a clean DQX error.

Consider adding a one-time runtime check at engine apply time (e.g., probe `SELECT typeof(try_to_geometry('POINT(0 0)'))` on first geo-check construction, raise a clear error if it fails). Doing it once-per-session keeps the per-row cost zero.

Not a blocker for this PR if you'd rather land the runtime gate as a follow-up — but please file an issue and link it from the docstrings.

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -824,7 +838,7 @@ def _compare_spatial_sql_function_result(
     )
 
 
-@register_rule("dataset")
+@register_rule("row")
```
mwojtyczka: **Behavioural change: `are_polygons_mutually_disjoint` re-registered as `row` (was `dataset`).** This changes the public rule type. If any existing user has classified rules by type (e.g., filtering metadata by `check_func_type == 'dataset'`), this is a breaking change.

Looking at the function body, it does appear to be row-level (returns a per-row condition with no aggregation), so the rename is probably correct — but:

1. Call this out explicitly in the PR description (changelog-worthy).
2. Consider splitting into a dedicated commit with rationale, to make `git log -p` archeology easier.
3. Verify no tests/docs still describe it as dataset-level.

tests/unit/test_geo_check_funcs.py
```diff
@@ -0,0 +1,100 @@
+import pytest
```
mwojtyczka, line 1: **Unit tests are smoke tests only — could verify alias structure.** All the "does_not_raise" tests just confirm the function builds a Column without erroring. You can still verify:

1. **Alias suffix** — `assert str(condition).endswith('_outside_reference_geometry')` (or whichever per-relationship suffix you adopt after the alias fix). This would automatically catch a regression where the alias names drift apart from the integration tests' expected column names.

The `InvalidParameterError` tests are good as-is. Consider grouping the smoke tests under a single parametrised test to shrink the file:
```python
@pytest.mark.parametrize("func, extra_kwargs", [...])
def test_call_does_not_raise(func, extra_kwargs):
    func("location", _REFERENCE_GEOMETRY_WKT, **extra_kwargs)
```

tests/integration/test_dataset_checks_geo.py
```diff
@@ -526,3 +550,320 @@ def test_are_polygons_mutually_disjoint_row_filter_with_duplicates(skip_if_runti
...
+def test_is_geo_covers_precise_exterior_point_wkt_violation(skip_if_runtime_not_geo_compatible, spark):
+    """Exterior point is not covered — violation (WKT input)."""
+    test_df = spark.createDataFrame([[_POINT_OUTSIDE], [None]], _GEO_SCHEMA)
+    condition = is_geo_covers("geom", _REF_POLYGON, precise=True, convert_column=True, convert_reference_geometry=True)
+    actual = test_df.select("geom", condition)
+    expected = spark.createDataFrame(
+        [[_POINT_OUTSIDE, _precise_violation(_POINT_OUTSIDE)], [None, None]], _PRECISE_SCHEMA
+    )
+    assertDataFrameEqual(actual, expected, checkRowOrder=False)
+
```
mwojtyczka: **Integration test repetition.** 20+ near-identical tests for interior/boundary/exterior across 5 functions × 2 modes. Each is ~10 lines of boilerplate.

Not a blocker — current form is more grep-friendly. But a parametrised matrix would shrink ~70% and make the test surface visible at a glance:
```python
@pytest.mark.parametrize(
    "check_func, mode, point, should_violate",
    [
        (is_geo_contains, "precise", _POINT_INSIDE,  False),
        (is_geo_contains, "precise", _POINT_EDGE,    True),  # boundary excluded
        (is_geo_contains, "precise", _POINT_OUTSIDE, True),
        (is_geo_covers,   "precise", _POINT_EDGE,    False), # boundary included
        ...
    ],
)
```
Up to you. Worth considering when the next geo PR adds more checks (this is part 1).

src/databricks/labs/dqx/geo/check_funcs.py
mwojtyczka: @IvannKurchenko the change is still there, pls revert

tests/integration/test_dataset_checks_geo.py
```diff
@@ -1,7 +1,31 @@
 from pyspark.testing.utils import assertDataFrameEqual
 from databricks.sdk import WorkspaceClient
 from databricks.labs.dqx.engine import DQEngine
-from databricks.labs.dqx.geo.check_funcs import are_polygons_mutually_disjoint
+from databricks.labs.dqx.geo.check_funcs import (
```
mwojtyczka: These are row-level checks and should be moved under `test_row_checks.py`

mwojtyczka (changes requested):
@IvannKurchenko did a second round of review. Please address the comments.

src/databricks/labs/dqx/geo/check_funcs.py
IvannKurchenko: Thank you, this makes total sense. Logged the following issue as I believe such verification is applicable in other   checks: https://github.com/databrickslabs/dqx/issues/1225.

tests/integration/test_dataset_checks_geo.py
IvannKurchenko: Yes, please, I'd consider this for the next PR. Thank you.

mwojtyczka (approved):
LGTM

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -918,3 +958,350 @@ def apply(df: DataFrame) -> DataFrame:
...
+        )
+
+    col_str_norm, col_expr_str, col_expr = get_normalized_column_and_expr(column)
+    ref_col = reference_geometry if isinstance(reference_geometry, Column) else F.lit(reference_geometry)
+    resolution_col = resolution if isinstance(resolution, Column) else F.lit(resolution)
+
+    col_h3_array = F.array_distinct(
+        F.coalesce(
+            F.call_function("h3_try_coverash3", col_expr, resolution_col),
+            F.array(F.call_function("h3_pointash3", col_expr, resolution_col)),
```
mwojtyczka: **Approximate mode can raise on an invalid (non-null) geometry rather than skip it.** The fallback is `F.coalesce(h3_try_coverash3(...), F.array(h3_pointash3(...)))` — but `h3_pointash3` is the **non-`try`** variant (same on line 1029 for the reference). For a non-null value that fails to parse as a point geometry, `h3_pointash3` errors at Spark-action time instead of yielding null, so the row neither passes nor is skipped — it blows up the job. The precise path routes everything through `try_to_geometry`, so this is an asymmetry. The docstrings promise "Null values are skipped," but invalid-non-null values aren't handled as gracefully. Consider a `try`-variant / guarding the fallback, or documenting that approximate mode assumes parseable geometries. (Null *column* values are fine — handled by the outer `F.when(col_expr.isNull(), ...)`.)

tests/perf/test_apply_checks.py
```diff
@@ -2093,27 +2093,49 @@ def test_benchmark_has_no_aggr_outliers(benchmark, ws, generated_df):
...
-        "col1_email_standard",
-        "col2_email_with_quoted_local_part",
-        "col3_email_with_ip_domain",
-        "col4_email_with_multi_part_domain",
-    ],
-)
-@pytest.mark.benchmark(group="test_benchmark_is_valid_email")
-def test_benchmark_is_valid_email(benchmark, ws, generated_email_df, column):
+def test_benchmark_is_geo_covers_precise(benchmark, ws, generated_df):
+    """Benchmark is_geo_covers check on.
```
mwojtyczka: Docstring is truncated: `"""Benchmark is_geo_covers check on."""` — looks like the sentence got cut off (the copy-paste aggregate docstring was correctly removed, this is just leftover polish). Same truncated line appears on the sibling benchmark at line 2122. One-line fix each.

mwojtyczka (commented):
Re-reviewed against head `8c71b90`. Verified the prior review feedback is addressed in code and resolved the 13 stale threads (decorator revert, WKB-bytes error message via `st_astext`, per-relationship alias/message maps, collapsed `Literal`s, `array_distinct` dedup, the `reference_geometry` literal-vs-column doc fix, plus the moved row tests and the new WITHIN + precision-divergence integration tests — `test_is_geo_covers_precise_flags_near_boundary_point_approximate_does_not` is a nice touch). Two small remaining items inline; neither blocking.

src/databricks/labs/dqx/geo/check_funcs.py
mwojtyczka: @IvannKurchenko pls double check this, and maybe add test for this. Also most of the integraiton tests fail.

mwojtyczka:
> LGTM

@IvannKurchenko seems like most of the new integration tests are failing. Can you pls check this?
https://github.com/databrickslabs/dqx/actions/runs/27132299451

src/databricks/labs/dqx/geo/check_funcs.py
IvannKurchenko: Thank you so much for pointing out. I've checked and both `h3_try_coverash3` and `h3_pointash3` raise error on invalid WKT values such as empty string. Additionally there is no `try_h3_pointash3` function. (Please see the list in the [docs](https://docs.databricks.com/aws/en/sql/language-manual/functions/h3_pointash3)). 

I added additional note to documentation about this behaviour.

If it is still desired goal to skip invalid WKT/WKB/EWKT/EWKB values, I could suggest:
- Wrap `h3_try_coverash3` and `h3_pointash3` into UDF with try-except block 
- Validate the value with regexp and call either function only in case if the value is valid.

Happy to hear your thoughts.

tests/perf/test_apply_checks.py
IvannKurchenko: Thanks, fixed.

IvannKurchenko:
@mwojtyczka I apologize for failed tests. I updated the PR, May I ask to re-run them?

src/databricks/labs/dqx/geo/check_funcs.py
mwojtyczka: **We should not trip the execution due to invalid values**. Instead, only call `h3_pointash3` when the value actually parses to a point, otherwise yield NULL and skip the row; replace lines from 1020-1032 with:
```
def _h3_cells(geom_expr: Column) -> Column:
    # h3_try_coverash3 returns NULL for point geometries (no areal coverage) and
    # for invalid/unsupported input. For points we fall back to h3_pointash3, but
    # that is the non-`try` variant and raises on non-point/invalid input, so gate
    # it on the value actually parsing to a point. coalesce and CASE WHEN both
    # short-circuit, so h3_pointash3 only runs on a confirmed point; anything
    # instead of failing the job.
    cover = F.call_function("h3_try_coverash3", geom_expr, resolution_col)
    is_point_geom = F.call_function(
        "st_geometrytype", F.call_function("try_to_geometry", geom_expr)
    ) == F.lit(POINT_TYPE)
    point_cell = F.when(is_point_geom, F.array(F.call_function("h3_pointash3", geom_expr, resolution_col)))
    return F.array_distinct(F.coalesce(cover, point_cell))
    
col_h3_array = _h3_cells(col_expr)
ref_h3_array = _h3_cells(ref_col)
intersection_size = F.size(F.array_intersect(col_h3_array, ref_h3_array))
```

Suggested next steps:
  1. Apply the guarded _h3_cells fix + remove the docstring NOTE.
  2. Add a regression test: a non-null invalid geometry (e.g. "NOT WKT") plus a valid point/polygon → job completes, invalid row skipped (no violation).

```
def test_is_geo_covers_approximate_invalid_geometry_is_skipped(skip_if_runtime_not_geo_compatible, spark):
      """A non-null but invalid geometry must be skipped (no violation), not raise."""
      test_df = spark.createDataFrame([[_POINT_INSIDE], ["NOT A GEOMETRY"], [None]], _GEO_SCHEMA)
      condition = is_geo_covers("geom", _REF_POLYGON, resolution=7)
      actual = test_df.select("geom", condition)
          [[_POINT_INSIDE, None], ["NOT A GEOMETRY", None], [None, None]], _COVERS_APPROXIMATE_SCHEMA
      )
      assertDataFrameEqual(actual, expected, checkRowOrder=False)
```

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -918,3 +958,354 @@ def apply(df: DataFrame) -> DataFrame:
...
+    is_inside = None
+    match topological_relationship:
+        case "INTERSECTS":
+            is_inside = intersection_size > 0
+        case "COVERS":
+            is_inside = intersection_size == F.size(col_h3_array)
+        case _:
+            raise InvalidParameterError(f"Unsupported approximate relationship: {topological_relationship!r}")
+
+    condition = F.when(col_expr.isNull(), F.lit(None)).otherwise(~is_inside)
```
mwojtyczka: to complete the fix to skip invalid geometry
```suggestion
    condition = F.when(col_expr.isNull() | col_h3_array.isNull(), F.lit(None)).otherwise(~is_inside)
```

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -918,3 +958,354 @@ def apply(df: DataFrame) -> DataFrame:
...
+    When *precise* is False (default), approximates coverage using H3 cell indexing: all hexagonal
+    cells that represent the column geometry must also be present in the H3 cells of the reference
+    geometry. Higher *resolution* values give finer precision at the cost of more cells. Because H3
+    cells are discrete approximations, boundary behaviour differs from the exact `st_covers` predicate
+    and geometries near the boundary may be misclassified.
+
+    Both the target column and the reference geometry are always handled as `GEOMETRY` in precise mode.
+    When conversion is requested (*convert_column* or *convert_reference_geometry* set to True),
+    *try_to_geometry* is applied. These flags are ignored in approximate mode.
+    **NOTE**: If `precise=False` and `column` or `reference_geometry` contains an invalid WKT/WKB/EWKT/EWKB value,
```
mwojtyczka:
```suggestion
```

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -918,3 +958,354 @@ def apply(df: DataFrame) -> DataFrame:
...
+
+    When *precise* is False (default), approximates intersection using H3 cell indexing: at least one
+    hexagonal cell must be shared between the H3 representations of the column and reference geometries.
+    Higher *resolution* values give finer precision at the cost of more cells. Because H3 cells are
+    discrete approximations, boundary behaviour differs from the exact `st_intersects` predicate.
+
+    Both the target column and the reference geometry are always handled as `GEOMETRY` in precise mode.
+    When conversion is requested (*convert_column* or *convert_reference_geometry* set to True),
+    *try_to_geometry* is applied. These flags are ignored in approximate mode.
+    **NOTE**: If `precise=False` and `column` or `reference_geometry` contains an invalid WKT/WKB/EWKT/EWKB value,
```
mwojtyczka:
```suggestion
```

mwojtyczka (changes requested):

mwojtyczka:
> @mwojtyczka I apologize for failed tests. I updated the PR, May I ask to re-run them?

No worries, I triggered the tests but we should update the implementation as suggested in the comments

src/databricks/labs/dqx/geo/check_funcs.py
mwojtyczka: I applied the fix

mwojtyczka (approved):
LGTM

Issue #1171, opened 2026-05-11
[FEATURE]: DQX Studio UI with per-user language switcher


lfbraz:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Problem statement

DQX Studio is currently English-only. Hard-coded labels, descriptions, validation messages, and empty-state copy live inside each component, which creates two problems:

- **Accessibility gap for non-English speakers.**: Data stewards, analysts, and approvers who are more comfortable in Portuguese, Italian, Spanish, or other languages have to mentally translate the UI on every page, slowing down rule authoring and approval workflows. For a no-code tool aimed explicitly at lowering the barrier for non-technical users, the English-only barrier directly undermines that goal.

- **No extension point for translations**:  Even contributors who want to help cannot add a language without first refactoring every component to externalize its strings, which is a large, all-or-nothing change.

- **Inconsistent UX when embedded in a localized host app**: DQX Studio can be surfaced inside other Databricks Apps (as a tab, iframe, or linked module). When the host app is localized — for example, a Portuguese- or Spanish-speaking team's internal data platform — the surrounding chrome, menus, and copy are in the user's language, but the embedded DQX Studio panel suddenly switches back to English, causing a weird experience for the user.

There is also no per-user way to override the language — even if some strings were translated, every user in the workspace would see the same locale.

### Proposed Solution

Introduce an i18n layer in the studio frontend and expose a language switcher on the Profile page.

- **i18n framework**: Adopt i18next + react-i18next with i18next-browser-languagedetector. Move all user-facing strings into JSON locale files and replace hard-coded strings with t("...") keys throughout components and routes.

- **Supported languages (initial set)**: English (en, default and fallback), Portuguese — Brazil (pt-BR), Italian (it), Spanish (es). The list is centralized in SUPPORTED_LANGUAGES, so future languages plug in by dropping a new locale file and adding one entry.

- **Per-user preference**: Add a LanguageSelector component to a new Preferences card on the Profile page. Changing it applies immediately across the studio (no reload) and updates document.documentElement.lang for accessibility.

<img width="601" height="285" alt="Image" src="https://github.com/user-attachments/assets/d00aad17-5e54-4713-bf87-95542225c57c" />

### Additional Context

I forked the repo at https://github.com/lfbraz/dqx and will submit a PR to review the proposed solution.

mwojtyczka:
Completed as part of https://github.com/databrickslabs/dqx/pull/1172

PR #1172, opened 2026-05-11
Feat/profile switch language

lfbraz:
### Linked issues
https://github.com/databrickslabs/dqx/issues/1171

### Tests

- [x] manually tested
- [ ] added unit tests
- [ ] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

### Documentation and Demos

Adds in-app language switching to DQX Studio, surfaced from the user's Profile page. The studio now ships with four locales (English, Portuguese-BR, Italian, Spanish) and persists the user's choice per-browser without touching server-side state.

**Frontend — i18n infrastructure**

- Add i18next + react-i18next + i18next-browser-languagedetector to app/package.json.
- New app/src/databricks_labs_dqx_app/ui/lib/i18n/index.ts bootstraps i18next with fallbackLng: "en", supportedLngs whitelist, and a detector chain of localStorage → navigator (cached under dqx.language).
- Wire the bootstrap into app/src/databricks_labs_dqx_app/ui/main.tsx so translations are initialized before the router mounts.
- Add four locale files under app/src/databricks_labs_dqx_app/ui/lib/i18n/locales/ (en.json, pt-BR.json, it.json, es.json) — keys mirror the existing UI copy across header, sidebar, profile, runs, rules, profiler, discovery, config, and labels surfaces.

**Frontend — UI**

- New app/src/databricks_labs_dqx_app/ui/components/apx/LanguageSelector.tsx: shadcn Select showing each language by native label (with English label as secondary hint); on change it calls i18n.changeLanguage and updates document.documentElement.lang.
- New Preferences card on app/src/databricks_labs_dqx_app/ui/routes/_sidebar/profile.tsx hosts the selector.
- Replace hard-coded strings with t(...) calls across the sidebar routes (config.tsx, discovery.tsx, home.tsx, profile.tsx, profiler.tsx, rules.active.tsx, rules.create.tsx, rules.create-sql.tsx, rules.drafts.tsx, rules.import.tsx, rules.single-table.tsx, runs.tsx, runs-history.tsx, route.tsx) and shared components (HeaderUserMenu, PageBreadcrumb, SidebarLayout, SidebarUserFooter, mode-toggle, AICheckGenerator, AuthGuard, CatalogBrowser, ColumnDiscoveryPanel, CommentThread, DryRunResults, Labels, RoleManagement, RulesReview).
- Behaviour preserved: this is a string-extraction refactor — no logic, routes, RBAC, or API contracts changed.

**Docs**

- Add a **Profile preferences → Language** section to docs/dqx/docs/guide/dqx_studio.mdx covering supported locales, persistence behavior, scope (per-user, UI only), and a caution note that non-English translations are AI-generated and not yet reviewed by native speakers.

**Notes for reviewers**

- Persistence is browser-local (localStorage); no backend or table changes.
- No new dependencies in the Python wheel — frontend-only.
- Non-English copy was AI-assisted; native-speaker review welcome (as I'm a Portuguese native speaker, I already reviewed all the PT-BR sections).

**Images**

profile-language-switch <br> <br>

main-page-dqx <br> <br>

- [ ] added/updated demos
- [x] added/updated docs


ghanse (commented):
Solid i18n landing — clean key parity across en/es/it/pt-BR, JSON validates, and every route I spot-checked routes user-facing strings through `t()`. Two functional gotchas to address before merge: (1) `<html lang>` is never synchronized when `LanguageDetector` picks a non-English locale on first load (a11y / screen-reader regression), and (2) `rulesSingleTable.summary` / `duplicateChecksExist_*` interpolate a hard-coded English `"s"` via `{{tablesPlural}}`/`{{checksPlural}}`/`{{rulesPlural}}` placeholders, which the non-English translators correctly dropped — producing a placeholder-count divergence and defeating proper pluralization.

Smaller items addressed inline: `t` in `AuthGuard`'s effect deps, all four locales bundled eagerly (~262 KB raw), ~50 unused `common.*` keys. Recommend adding a `make app-check` script that asserts placeholder-set parity across locales to prevent this class of bug going forward.

Two follow-ups on pre-existing code that this PR makes newly relevant (not anchorable to the diff, hence here in the summary):

- **`app/src/databricks_labs_dqx_app/ui/routes/_sidebar/config.tsx` ~ lines 60 and 66** — `Intl.DateTimeFormat("en-US", …)` is hard-coded for two timezone-label formatters. Every user (including pt-BR / es / it) sees timezone short names (`PST`, `CET`) and offset formatting in English. Pass `undefined` (browser default) or `i18n.language` so this matches the rest of the UI. `format-utils.ts` already uses `undefined`, so this is the only outlier.
- **`app/src/databricks_labs_dqx_app/ui/routes/_sidebar/rules.drafts.tsx:7`** (also `rules.active.tsx:54,148`, `runs-history.tsx:243`) — the synthetic catalog label `"Cross-table rules"` is used both as an internal filter id and rendered to the user. With i18n in this PR, it should be translated at render time while keeping the internal id stable, e.g. `cat === CROSS_TABLE_CATALOG ? t("common.crossTableRules") : cat`. Most visible untranslated remnant.

app/src/databricks_labs_dqx_app/ui/lib/i18n/index.ts
```diff
@@ -0,0 +1,44 @@
...
+      escapeValue: false,
+    },
+    detection: {
+      order: ["localStorage", "navigator"],
+      lookupLocalStorage: LANGUAGE_STORAGE_KEY,
+      caches: ["localStorage"],
+    },
+  });
+
+export default i18n;
```
ghanse, line 85: **a11y bug — `<html lang>` is never synchronized when `LanguageDetector` resolves a non-English locale on first load.**

`index.html` ships `<html lang="en">` and `LanguageSelector.handleChange` (line 30) is the only code that updates `document.documentElement.lang`. If a user lands on the page with browser locale `es`/`it`/`pt-BR`, the UI text switches but `<html lang>` still reports `en` — silently breaking screen readers, browser auto-translate, hyphenation, and Lighthouse a11y scoring.

Wire it here in the bootstrap (and an initial sync after init) so it's correct everywhere, not just when the user opens the selector:

```suggestion
  })
  .then(() => {
    document.documentElement.lang = i18n.resolvedLanguage ?? "en";
  });

i18n.on("languageChanged", (lng) => {
  document.documentElement.lang = lng;
});

export default i18n;
```

Then the corresponding line in `LanguageSelector.tsx:30` becomes redundant and should be removed.

app/src/databricks_labs_dqx_app/ui/routes/_sidebar/rules.single-table.tsx
```diff
@@ -1319,14 +1315,19 @@ function UnifiedRulesPage() {
...
-                  <strong>{totalTargetPairs}</strong> rule{totalTargetPairs !== 1 ? "s" : ""} across{" "}
-                  <strong>{allTargetTables.length}</strong> table{allTargetTables.length !== 1 ? "s" : ""}
+                  {t("rulesSingleTable.summary", {
+                    checks: checks.filter((c) => c.fn !== "").length,
+                    checksPlural: checks.filter((c) => c.fn !== "").length !== 1 ? "s" : "",
+                    rules: totalTargetPairs,
+                    rulesPlural: totalTargetPairs !== 1 ? "s" : "",
+                    tables: allTargetTables.length,
+                    tablesPlural: allTargetTables.length !== 1 ? "s" : "",
+                  })}
```
ghanse: **i18n anti-pattern — `{{tablesPlural}}` / `{{checksPlural}}` / `{{rulesPlural}}` carry a hard-coded English `"s"` through the translation layer.**

For English this works, but for the other three locales it's wrong:
- The Spanish / Italian / Portuguese translators correctly dropped these placeholders, so the suffix never appears — and now `en.json` has placeholders that the other locales don't. i18next will silently log a `MissingInterpolation` warning at runtime and the rendered text shape diverges across locales.
- Native i18next pluralization (`_one`/`_other`) is already in use elsewhere (e.g. `rulesActive.rulesCount_*`), so it's the consistent fix.

Recommend restructuring as three pluralized noun fragments composed into a sentence:

```jsonc
// en.json (and parallel keys in es/it/pt-BR)
"summaryChecks_one": "{{count}} check",
"summaryChecks_other": "{{count}} checks",
"summaryRules_one": "{{count}} rule",
"summaryRules_other": "{{count}} rules",
"summaryTables_one": "{{count}} table",
"summaryTables_other": "{{count}} tables",
"summary": "{{checks}} → {{rules}} across {{tables}}"
```

```ts
t("rulesSingleTable.summary", {
  checks: t("rulesSingleTable.summaryChecks", { count: nChecks }),
  rules:  t("rulesSingleTable.summaryRules",  { count: nRules }),
  tables: t("rulesSingleTable.summaryTables", { count: nTables }),
})
```

Same treatment for `duplicateChecksExist_one` / `_other` at line 1318.

app/src/databricks_labs_dqx_app/ui/components/AuthGuard.tsx
```diff
@@ -80,22 +85,22 @@ export function AuthGuard({ children }: AuthGuardProps) {
         clearTimeout(timeoutId);
       }
     };
-  }, [retryCount, isAuthReady, error]);
+  }, [retryCount, isAuthReady, error, t]);
```
ghanse: **`t` from `useTranslation()` is unstable and adding it to this effect's deps can interrupt the auth-polling loop.**

`AuthGuard` runs before any UI is mounted and polls `/current-user` with exponential backoff (line 40). If `t`'s identity changes (i18n finishes initialising between the first attempt and a retry, or `react-i18next` re-renders on resource load), this effect's cleanup tears down the pending `setTimeout` (lines 84-86) and re-enters from scratch. In a Databricks App where OBO tokens take a few seconds to flow, this can manifest as a stuck loading state.

`t` is only consumed in the terminal `setError` branch (lines 50-60). Read it lazily from a ref so the dep can be removed:

```ts
const tRef = useRef(t);
useEffect(() => { tRef.current = t; }, [t]);

// inside the catch branch, use tRef.current(...) instead of t(...)
}, [retryCount, isAuthReady, error]);
```

app/src/databricks_labs_dqx_app/ui/components/apx/LanguageSelector.tsx
```diff
@@ -0,0 +1,65 @@
...
+}: LanguageSelectorProps) {
+  const { t, i18n } = useTranslation();
+  const current =
+    SUPPORTED_LANGUAGES.find((l) => l.code === i18n.resolvedLanguage)?.code ??
+    SUPPORTED_LANGUAGES.find((l) => i18n.resolvedLanguage?.startsWith(l.code))?.code ??
+    "en";
+
+  const handleChange = (value: string) => {
+    void i18n.changeLanguage(value);
+    document.documentElement.lang = value;
```
ghanse: Once the `languageChanged` listener suggested in `lib/i18n/index.ts` is in place, this line becomes redundant — and skips the case where i18next normalises the requested code (e.g. `en-US` → `en`). Drop it:

```suggestion
  const handleChange = (value: string) => {
    void i18n.changeLanguage(value);
  };
```

app/src/databricks_labs_dqx_app/ui/components/apx/LanguageSelector.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+export default function LanguageSelector({
+  className,
+  hideLabel = false,
+}: LanguageSelectorProps) {
+  const { t, i18n } = useTranslation();
+  const current =
+    SUPPORTED_LANGUAGES.find((l) => l.code === i18n.resolvedLanguage)?.code ??
+    SUPPORTED_LANGUAGES.find((l) => i18n.resolvedLanguage?.startsWith(l.code))?.code ??
+    "en";
```
ghanse: Minor — the fallback chain handles `en-US` → `en` but would miss a future country variant like `pt-PT` (it would fall through to `"en"`, because `"pt-PT".startsWith("pt-BR")` is false). Since `pt-BR` is the only supported `pt-*` variant today, this is fine for now, but consider matching on the primary tag (`l.code.split("-")[0]`) for forward compatibility. Not blocking.

app/src/databricks_labs_dqx_app/ui/lib/i18n/index.ts
```diff
@@ -0,0 +1,44 @@
+import i18n from "i18next";
+import LanguageDetector from "i18next-browser-languagedetector";
+import { initReactI18next } from "react-i18next";
+
+import en from "./locales/en.json";
+import ptBR from "./locales/pt-BR.json";
+import it from "./locales/it.json";
+import es from "./locales/es.json";
```
ghanse: **Bundle-size note** — all four locale files (~262 KB raw JSON: 60 / 67 / 67 / 67) are imported statically and land in the initial JS chunk. For the 95% of users who only ever see one locale, that's pure overhead. Consider lazy-loading via `import()` / `i18n.addResourceBundle`, or via `i18next-http-backend` with the JSONs served as static assets:

```ts
i18n.use(LanguageDetector).use(initReactI18next).init({
  resources: { en: { translation: en } },  // bundle default only
  // ...
});

i18n.on("languageChanged", async (lng) => {
  if (lng === "en" || i18n.hasResourceBundle(lng, "translation")) return;
  const mod = await import(`./locales/${lng}.json`);
  i18n.addResourceBundle(lng, "translation", mod.default);
});
```

Not blocking, but worth doing before more locales land.

app/src/databricks_labs_dqx_app/ui/lib/i18n/locales/en.json
```diff
@@ -0,0 +1,1253 @@
+{
```
ghanse, line 1: **~50 `common.*` keys are defined but never referenced by any `t("…")` call** — spot-checked `common.discard`, `common.successLower`, `common.ago`, `common.warn`, `common.exportYaml`, `common.failedToAddComment`; all six return 0 hits across the UI codebase. Others I noticed in the same shape: `common.save`, `common.submit`, `common.submitting`, `common.discardLeave`, `common.success`, `common.page`, `common.rows`, `common.min`, `common.max`, `common.preview`, `common.validate`, `common.noComments`, `common.loadingComments`, `common.addComment`, `common.showComments`, `common.hideComments`, `common.failedToDeleteComment`, …

These add ship weight × 4 locales and invite translation drift — each one has to be kept in sync across four locales when nothing actually renders it. Either remove them, or replace inline literals in the migrated routes with these `common.*` keys so they're actually used.

ghanse, line 1: **Suggest a CI check for placeholder-set parity across locales.**

The `rulesSingleTable.summary` / `duplicateChecksExist_*` divergence flagged above would have been caught by a build-time lint. A short Node script that flattens each locale, regex-extracts `{{…}}` placeholders, and asserts every key has the same placeholder set across locales would prevent this class of bug going forward. Easy to add to `make app-check` per `app/DEVELOPMENT.md` — happy to follow up with a draft if useful.

app/src/databricks_labs_dqx_app/ui/lib/i18n/index.ts
lfbraz: Good finding — I applied it.

Notes about implementation:

- Now only 'en' is included in the initial bundle. pt-BR / it / es are loaded on demand using dynamic import().
- We save around 54 KB gzip on the first JS chunk for users who use only English.
- I added an exported helper ensureLocaleLoaded(lng) and also an i18nReady promise. The Selector is waiting for the bundle before calling changeLanguage, and main.tsx waits for i18nReady before root.render. Because of this, there is no flashing of untranslated keys on the first paint or when the user changes the language.
- One small detail to notice: when you reload, and a non-English locale is saved in localStorage, i18n.resolvedLanguage will return "en" (because only the 'en' bundle exists at init time). So, the init .then() reads i18n.language instead and runs changeLanguage again after addResourceBundle, so that the resolvedLanguage will match the user's choice.

app/src/databricks_labs_dqx_app/ui/lib/i18n/locales/en.json
lfbraz: I applied the first option, and removed all the 52 unused common.* keys from the four locale files.

app/src/databricks_labs_dqx_app/ui/routes/_sidebar/rules.single-table.tsx
lfbraz: Done. Please let me know if it is OK.

app/src/databricks_labs_dqx_app/ui/components/AuthGuard.tsx
lfbraz: Done. Please let me know if it is OK.

app/src/databricks_labs_dqx_app/ui/components/apx/LanguageSelector.tsx
lfbraz: Done. Please let me know if it is OK.

app/src/databricks_labs_dqx_app/ui/components/apx/LanguageSelector.tsx
lfbraz: Done. Please let me know if it is OK.

app/src/databricks_labs_dqx_app/ui/lib/i18n/locales/en.json
lfbraz: It would be great to have your draft so I can learn the best way to apply this. I appreciate your help.

app/src/databricks_labs_dqx_app/ui/lib/i18n/index.ts
lfbraz: Done. Please let me know if it is OK.

lfbraz:
Hi @ghanse . Thanks for your review. I've tried to apply all the suggestions. Please let me know if they are OK or if there are any missing issues. I really appreciate your support!

app/src/databricks_labs_dqx_app/ui/main.tsx
```diff
@@ -55,13 +56,15 @@ const rootElement = document.getElementById("root")!;
 
 if (!rootElement.innerHTML) {
   const root = createRoot(rootElement);
```
mwojtyczka, line 58: **Init failure produces a blank page.** This `void i18nReady.then(() => root.render(...))` has no `.catch`. If i18next init throws — network failure on a lazy locale fetch, JSON parse error, anything in `ensureLocaleLoaded` — `root.render` never runs and users see a blank screen with no error.

Add a fallback so the app still renders with the bundled English locale if init fails:
```ts
void i18nReady
  .then(() => root.render(/* tree */))
  .catch((err) => {
    console.error("i18n init failed; rendering with English fallback", err);
    root.render(/* same tree — i18next falls back to embedded `en` */);
  });
```

Also flagging: first-paint is delayed by a dynamic-import round-trip for any returning user with `pt-BR`/`it`/`es` in localStorage. Acceptable for ~50KB locales, but consider rendering an English skeleton immediately and switching on resolve if you see real-world latency complaints.

app/src/databricks_labs_dqx_app/ui/lib/i18n/index.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export const i18nReady = i18n
+  .use(LanguageDetector)
+  .use(initReactI18next)
+  .init({
+    resources: {
+      en: { translation: en },
+    },
+    fallbackLng: "en",
+    supportedLngs: SUPPORTED_LANGUAGES.map((l) => l.code),
```
mwojtyczka, line 55: **`en-US` won't match `en` cleanly.** With `supportedLngs: ["en", "pt-BR", "it", "es"]` and `load: "currentOnly"`, a typical browser-default `en-US` from the detector fails the whitelist → falls back to `en` via `fallbackLng`. That works for first-visit users, but bypasses any localStorage value the user explicitly set (which gets the same treatment if it has a region tag).

Add `nonExplicitSupportedLngs: true` to the init config so region-tagged values resolve to their base code:
```ts
supportedLngs: SUPPORTED_LANGUAGES.map((l) => l.code),
nonExplicitSupportedLngs: true,
load: "currentOnly",
```
This is a no-op for `pt-BR` (already region-tagged), but means `en-US`, `en-GB`, `es-MX`, `it-CH` all resolve cleanly to the closest supported base locale.

app/src/databricks_labs_dqx_app/ui/components/apx/LanguageSelector.tsx
```diff
@@ -0,0 +1,65 @@
...
+  return (
+    <div className={className}>
+      {!hideLabel && (
+        <Label htmlFor="language-selector" className="flex items-center gap-2 mb-2">
+          <Languages className="h-4 w-4" />
+          {t("language.label")}
+        </Label>
+      )}
+      <Select value={current} onValueChange={handleChange}>
+        <SelectTrigger id="language-selector" className="w-full sm:w-64">
```
mwojtyczka: **`hideLabel=true` makes the select inaccessible.** When `hideLabel` is true, the `<Label htmlFor="language-selector">` block at line 36 is omitted, but the `<SelectTrigger id="language-selector">` here stays. Screen readers can't announce what the field is for.

Add a fallback `aria-label` so the trigger is always labelled:
```tsx
<SelectTrigger
  id="language-selector"
  aria-label={hideLabel ? t("language.label") : undefined}
  className="w-full sm:w-64"
>
```

Separately: hard-coded `id="language-selector"` breaks if this component ever renders twice on the same page (e.g., header switcher + profile page). Use `useId()` for a unique id.

app/src/databricks_labs_dqx_app/ui/components/apx/LanguageSelector.tsx
```diff
@@ -0,0 +1,65 @@
...
+}: LanguageSelectorProps) {
+  const { t, i18n } = useTranslation();
+  const primaryTag = i18n.resolvedLanguage?.split("-")[0];
+  const current =
+    SUPPORTED_LANGUAGES.find((l) => l.code === i18n.resolvedLanguage)?.code ??
+    SUPPORTED_LANGUAGES.find((l) => l.code.split("-")[0] === primaryTag)?.code ??
+    "en";
+
+  const handleChange = (value: string) => {
+    void ensureLocaleLoaded(value).then(() => i18n.changeLanguage(value));
```
mwojtyczka: **Silent failure on locale switch.** `void ensureLocaleLoaded(value).then(() => i18n.changeLanguage(value))` swallows rejection from either promise. If the dynamic import fails (offline, CDN hiccup, corrupted JSON), the user clicks the new language, nothing happens, no feedback.

Add a `.catch` that surfaces a toast and reverts the visible selection:
```ts
const handleChange = (value: string) => {
  ensureLocaleLoaded(value)
    .then(() => i18n.changeLanguage(value))
    .catch((err) => {
      console.error("Failed to switch language", err);
      toast.error(t("language.switchFailed"));
    });
};
```

app/src/databricks_labs_dqx_app/ui/lib/i18n/locales/it.json
```diff
@@ -0,0 +1,1209 @@
...
+    "allowCustomValues": "Consenti valori personalizzati",
+    "noValuesHint": "Nessun valore — gli autori alterneranno questa etichetta come tag booleano (true/false).",
+    "removeValueAria": "Rimuovi valore {{value}}",
+    "weightValuePlaceholder": "aggiungi valore peso (es. 1)",
+    "addValuePlaceholder": "aggiungi valore… (premi Invio)",
+    "keyPlaceholder": "es. team"
+  },
+  "discovery": {
+    "breadcrumb": "Discovery",
+    "title": "Discovery",
```
mwojtyczka: **Missed Italian translation.** `discovery.breadcrumb` and `discovery.title` (line 162) both read `"Discovery"` — the English word, untranslated. Every other route in `it.json` translates its title/breadcrumb (`profiler.title = "Profila e genera"`, `rulesActive.title = "Regole Attive"`, etc.), and the parallel `es.json` and `pt-BR.json` both translate this one (`"Descubrimiento"` / `"Descoberta"`).

Fix both keys to e.g. `"Esplorazione"` (more natural) or `"Scoperta"` (literal). Native-speaker review preferred.

mwojtyczka (changes requested):

mwojtyczka:
> Hi @ghanse . Thanks for your review. I've tried to apply all the suggestions. Please let me know if they are OK or if there are any missing issues. I really appreciate your support!

@lfbraz thank you for the PR. I posted additional comments.

app/src/databricks_labs_dqx_app/ui/main.tsx
lfbraz: Good catch — fixed.

Wrapped the render in a `renderApp()` helper and added the missing `.catch`, so a failed init (lazy locale fetch, JSON parse error, anything in ensureLocaleLoaded) now still renders with the bundled English locale instead of a blank page.

I used a shared helper rather than duplicating the tree in both branches, but the behavior matches your suggestion exactly. Verified by serving the production build and blocking the pt-BR chunk in  DevTools — the app loads in English and logs the error, and removing the block restores Portuguese:

english-fallback


On the first-paint delay for returning non-English users: agreed it's acceptable for ~50 KB locales, so I'll leave it as-is for now and revisit with an English skeleton if we get real latency reports.

app/src/databricks_labs_dqx_app/ui/lib/i18n/index.ts
lfbraz: Thanks for the suggestion! I tried adding `nonExplicitSupportedLngs: true`, but when I tested it locally it broke pt-BR — the UI fell back to English even though the bundle loaded.
 
Digging in, I think the reason is that the flag makes i18next strip the region tag before the supportedLngs check (`isSupportedCode, i18next.js:947`): "pt-BR" → "pt". Since our whitelist has "pt-BR" and not "pt", pt-BR ends up rejected and falls back to en. It works for en-US/es-MX/it-CH because their base (en/es/it) is in the list — pt-BR just happens to be the one entry whose base isn't.

It also looks like i18next already resolves region variants to the closest supported base on its own (en-US→en, es-MX→es, pt-PT→pt-BR), so I went with a small `toSupportedCode()` helper to normalize
  the code for the lazy-loader lookup (mapping back through supportedLngs so pt-BR isn't split to pt) instead of the flag. Happy to revisit if you think there's a cleaner approach.

I tested en, pt-BR, it, es, plus en-US, es-MX, pt-PT, and an unsupported fr-FR (clean English fallback) and they all resolve correctly now.

app/package.json
```diff
@@ -25,12 +25,15 @@
...
     "clsx": "^2.1.1",
+    "i18next": "^26.0.8",
+    "i18next-browser-languagedetector": "^8.2.1",
     "js-yaml": "^4.1.1",
     "lucide-react": "^0.548.0",
     "motion": "^12.23.24",
     "react": "^19.2.0",
     "react-dom": "^19.2.0",
     "react-error-boundary": "^6.0.0",
+    "react-i18next": "^17.0.6",
```
laurencewells, line 36: Should we also update CLAUDE.md? They need to know that any additional text needs storing in the Jsons now with Language translations ?

app/src/databricks_labs_dqx_app/ui/components/apx/LanguageSelector.tsx
lfbraz: Good catch on both — fixed.

- Added aria-label={hideLabel ? t("language.label") : undefined} so the trigger always has an accessible name, while avoiding a redundant label when the visible <Label> is shown.
- Switched the hard-coded id to useId(), wired through both the <Label htmlFor> and the trigger, so it stays unique if the component is ever rendered more than once (e.g. header switcher + profile
  page).

Both reuse the existing language.label key, so no new strings. Thanks!

app/src/databricks_labs_dqx_app/ui/components/apx/LanguageSelector.tsx
lfbraz: Good catch — done. Added the `.catch` with `console.error` + `toast.error(t("language.switchFailed"))`, and added the language.switchFailed key to all four locales (en, pt-BR, it, es) so non-English users don't get the English fallback string.

The visible selection reverts on its own since the `<Select>` is controlled by current (derived from i18n.resolvedLanguage) — when the switch fails, changeLanguage never runs, so the dropdown stays on the previous language. Thanks!

app/src/databricks_labs_dqx_app/ui/lib/i18n/locales/it.json
lfbraz: Translated `discovery.breadcrumb` and `discovery.title` to "Esplorazione". I wasn't able to use a native-speaker review since I don't have any good contacts for this. Happy to change in the future when we have a native-speaker support on this.

app/package.json
lfbraz: Good point, thanks.

I went ahead and added an Internationalization (i18n) section to the frontend guide (app/src/databricks_labs_dqx_app/ui/CLAUDE.md). I put it there rather than the root AGENTS.md/CLAUDE.md, since those seem to be scoped to the Python library — happy to move it if you think it belongs somewhere more visible. It tries to capture:

- all user-facing text should go through t() — no hard-coded literals (incl. toasts, aria-labels, errors);
- every new key needs to be added to all four locales (en, pt-BR, it, es), with en as the source of truth, to avoid silent English fallbacks;
- pluralize with native i18next _one/_other + {{count}} rather than a hard-coded "s" or {{...Plural}} placeholder;
- a short note on how to add a new language.
  
I also added react-i18next to the Stack list and lib/i18n/ to the architecture tree. Let me know if you'd word any of it differently.

berrybluecode:
Hi @lfbraz , could you please give me permission to push code to your folk as I would like to resolve the conflicts. 
Also, are there only three language added? We need to add more languages for the next step
image

mwojtyczka:
> Hi @lfbraz , could you please give me permission to push code to your folk as I would like to resolve the conflicts. Also, are there only three language added? We need to add more languages for the next step image

I think this is fine. We can have follow up PRs to add more languages

app/src/databricks_labs_dqx_app/ui/lib/i18n/locales/en.json
```diff
@@ -0,0 +1,1210 @@
...
+    "unknownBadge": "Unknown"
+  },
+  "rulesActive": {
+    "breadcrumb": "Active Rules",
+    "title": "Active rules",
+    "subtitle": "Approved rules currently enforced on your tables.",
+    "exportYaml": "Export YAML",
+    "createRules": "Create rules",
+    "approvedRuleSets": "Approved rule sets",
+    "tablesAndChecks": "{{tables}} table{{tablesPlural}} · {{checks}} total checks",
```
mwojtyczka: **Residual `{{...Plural}}` anti-pattern** — the same issue fixed in `rules.single-table.tsx` still lives here in `rulesActive.tablesAndChecks` (and `rulesActive.checksAcrossTables` at line 328), called from `rules.active.tsx:337` and `:746` with a hard-coded English `"s"`:
```ts
t("rulesActive.tablesAndChecks", { tablesPlural: filteredRules.length !== 1 ? "s" : "", ... })
```
Unlike the single-table case, all four locales kept the placeholder so there's no `MissingInterpolation` crash — but it bakes English grammar into the other locales. Spanish renders `verificación` + `s` = `verificacións` (should be `verificaciones`); Italian/Portuguese have the same problem. This also contradicts the pluralization rule you just added to `ui/CLAUDE.md`. Recommend the same `_one`/`_other` + `{{count}}` treatment you applied to `rulesSingleTable`.

mwojtyczka (commented):
Re-reviewed after the latest round — thanks for the thorough fixes. Verified and resolved 13 of the 14 threads (a11y `<html lang>` sync, AuthGuard `t` ref, lazy-loading, `useId`/aria-label, locale-switch `.catch`, the blank-page `.catch`, the `common.*` cleanup, the Italian `discovery` translation, and the CLAUDE.md i18n section all check out; locale key sets are now at full parity — 1142 keys in each of the 4 files). Left the CI placeholder-parity-check thread open since that's a pending follow-up rather than a code fix.

One residual instance of the same plural anti-pattern from the (now-fixed) `rules.single-table.tsx` thread — flagged inline.

berrybluecode:
resolved the conflicts

app/src/databricks_labs_dqx_app/ui/lib/i18n/locales/en.json
mwojtyczka: Fixed in 9830447: replaced the `{{tablesPlural}}`/`{{checksPlural}}` placeholders with native i18next `_one`/`_other` fragments (`rulesActive.tablesCount` / `rulesActive.checksCount`) composed via nested `t()`, matching the existing `rulesSingleTable.summary` pattern. Added to all four locales (parity preserved — 1146 keys each); es/it/pt-BR now render proper plurals (e.g. `verificaciones` instead of `verificacións`).

app/src/databricks_labs_dqx_app/ui/lib/i18n/locales/en.json
mwojtyczka: Added in 0404527 as a pytest (`app/tests/test_i18n_locale_parity.py`) so it runs under the existing `make app-test` with no extra JS toolchain. It asserts three things against `en.json` as source of truth: (1) identical key sets per locale (catches silent English fallback), (2) identical `{{placeholder}}` sets per key (catches the `missingInterpolation` divergence you originally flagged), and (3) no hard-coded `{{*Plural}}` placeholders (use native `_one`/`_other` + `{{count}}`).

Worth noting: placeholder-set *parity alone* would not have caught the residual `rulesActive` plural bug, because all four locales shared the same `{{tablesPlural}}` — parity held. The explicit `{{*Plural}}` rule (3) is what flags it (12 hits pre-fix, 0 after). All 10 cases pass on the current locales.

mwojtyczka: I added tests for this

mwojtyczka (approved):
LGTM

lfbraz:
Thanks for the reviews, guys. Please let me know if you need anything else from my side at this time.

PR #1173, opened 2026-05-11
DQX Studio: Support for Lakebase storage backend, declarative storage with destroy protection, observer-aligned error/warning metrics

berrybluecode:
… label filtering, error/warning metrics

Storage: move schemas, wheels volume, and Lakebase instance + logical database into databricks.yml with lifecycle.prevent_destroy; replace bootstrap_storage.sh with 'make app-bind' for adopting existing resources in workspaces created by the prior flow.

Backend: introduce PgExecutor + Postgres migration runner for OLTP tables (rules, settings, RBAC, comments, schedules, scheduler bookkeeping); keep analytical tables (validation runs, profiling, quarantine, metrics) on Delta.

Metrics: persist DQX observer's error_row_count / warning_row_count / input_row_count via dq_validation_runs.error_rows / warning_rows and fix Spark Connect Observation.get mutability bug that was overwriting total_rows with limit-pushed values.

UI: rename 'Invalid' -> 'Errors', add 'Warnings' column to quarantine detail, replace 'Has invalid' filter with 'Has failures', surface label badges next to table names in rule selection and schedule editor, add label filtering to Execute Rules and Schedule settings.

Migrations: fix FIELD_ALREADY_EXISTS idempotency in MigrationRunner so v3/v4/v5 ADD COLUMN migrations no-op on fresh deploys whose v1 baseline already includes warning_rows / warnings / error_rows.

Misc: retention test coverage, custom metrics service, post-deploy grants script reuse, docs rewrite (DEPLOYMENT.md, README.md, CLAUDE.md, installation.mdx) for the new declarative storage model.

### Linked issues

Resolves #..

### Tests

- [x] manually tested
- [x] added unit tests
- [ ] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

### Documentation and Demos

- [ ] added/updated demos
- [x] added/updated docs


laurencewells (commented):
## Review

Large, well-structured PR. The architectural split (Delta for append/analytical tables, Lakebase for OLTP) is sound, `prevent_destroy` guards are the right default, and the documentation is thorough. A few issues below — two of them (partial migration commits, service-layer dialect branching) worth addressing before merge.

**Summary of items:**
1. `OltpExecutor` type alias is a dead string — not a real type
2. Dialect branching is leaking into service code — the abstraction isn't holding
3. `PgExecutor.execute` commits per statement — migrations can leave partial committed state
4. No unit tests for `PgExecutor` or `PgMigrationRunner`
5. `cast(SqlExecutor, pg_executor)` — no structural check enforces the parity contract
6. `bind_resources.sh` swallows bind failures as warnings instead of errors
7. Schema rename (`dqx_app` → `dqx_studio`) needs a one-liner in the migration guide for users with existing data

app/src/databricks_labs_dqx_app/backend/dependencies.py
```diff
@@ -31,6 +31,43 @@
...
+if TYPE_CHECKING:
+    from .pg_executor import PgExecutor
+
+# Type alias used by every OLTP-touching service: either the legacy
+# Delta-backed :class:`SqlExecutor` or, when Lakebase is enabled, the
+# :class:`PgExecutor`.  The two classes share the public surface
+# (``execute``, ``query``, ``query_dicts``, ``upsert``, ``q``,
+# ``json_literal_expr``, ``ts_text``, ``dialect``) so service code can
+# stay backend-agnostic.
+OltpExecutor = "SqlExecutor | PgExecutor"
```
laurencewells: This is a plain string, not a type alias — mypy/pyright treats it as `str`, not as a union type. It will not catch usages that pass incompatible types.

```python
# Current — a dead string
OltpExecutor = "SqlExecutor | PgExecutor"

# Fix — an actual type alias (works with or without from __future__ import annotations)
from typing import Union
OltpExecutor = Union["SqlExecutor", "PgExecutor"]
```

The alias also appears to be unused as an annotation anywhere in this diff, so if it is intended purely as documentation it can just be a comment.

app/src/databricks_labs_dqx_app/backend/pg_executor.py
```diff
@@ -0,0 +1,429 @@
...
+    def execute(self, sql: str, *, timeout_seconds: int = 120) -> None:  # noqa: ARG002 - parity with SqlExecutor
+        """Run a non-result-returning statement."""
+        with self._pool.connection() as conn:
+            with conn.cursor() as cur:
+                # psycopg accepts ``str`` at runtime (``Query`` is a
+                # union of bytes/str/Composable/SQL); the published
+                # stubs only declare the ``Template`` overload so we
+                # silence basedpyright here.
+                cur.execute(sql)  # pyright: ignore[reportCallIssue, reportArgumentType]
+            conn.commit()
```
laurencewells, line 424: `execute` commits after every single statement. Combined with `PgMigrationRunner._apply` splitting a migration on `;` and calling `execute` per statement, a migration that fails halfway leaves committed DDL with no version recorded in `dq_migrations`. On the next restart the runner retries the whole migration from scratch — idempotent DDL (`CREATE TABLE IF NOT EXISTS`, `CREATE INDEX IF NOT EXISTS`) survives this fine, but it is a fragile contract.

Two cleaner options:

**Option A** — wrap the whole migration in a transaction inside `_apply`:
```python
def _apply(self, migration: PgMigration) -> None:
    formatted = migration.sql.format(schema=self._schema)
    with self._pool.connection() as conn:
        with conn.cursor() as cur:
            cur.execute(formatted)  # Postgres accepts multi-statement strings
        # Record version atomically in the same transaction
        escaped = migration.description.replace("'", "''")
        cur.execute(
            f"INSERT INTO {self._meta_table} (version, description, applied_at) "
            f"VALUES ({migration.version}, '{escaped}\', CURRENT_TIMESTAMP)"
        )
        conn.commit()
```

**Option B** — keep per-statement splitting but move the `conn.commit()` out of `execute` into an explicit transaction in `_apply` (requires exposing a `connection()` context manager from the executor).

Either way the INSERT into `dq_migrations` must be in the same transaction as the DDL.

app/src/databricks_labs_dqx_app/backend/services/schedule_config_service.py
```diff
@@ -35,13 +35,18 @@ class ScheduleConfigService:
 
     def __init__(self, sql: SqlExecutor) -> None:
         self._sql = sql
-        self._table = f"{sql.catalog}.{sql.schema}.dq_schedule_configs"
-        self._history_table = f"{sql.catalog}.{sql.schema}.dq_schedule_configs_history"
+        if getattr(sql, "dialect", "delta") == "postgres":
+            self._table = f"{sql.schema}.dq_schedule_configs"
```
laurencewells: `getattr(sql, "dialect", "delta") == "postgres"` branches appear here in `__init__` and again in `save` — and similarly in `SchedulerService._qualify_oltp` and `RulesCatalogService`. This means the stated goal of `PgExecutor` ("services can target either backend with the same call signatures") is not holding: adding a third executor type would require updating every service that branches on dialect.

The per-row `version` increment in `save` is the legitimate exception (it cannot be expressed as a pure upsert). But the table-qualification branch here (`schema.table` vs `catalog.schema.table`) should be absorbed by the executor. Both `SqlExecutor` and `PgExecutor` already expose `catalog` and `schema` properties — the service could just write:

```python
self._table = f"{sql.catalog}.{sql.schema}.dq_schedule_configs"
```

and `PgExecutor.catalog` could return `""` (empty) so the qualification collapses to `".schema.table"`, or better: add a `fqn(table)` helper to both executors:

```python
# SqlExecutor
def fqn(self, table: str) -> str:
    return f"{self.catalog}.{self.schema}.{table}"

# PgExecutor
def fqn(self, table: str) -> str:
    return f"{self.schema}.{table}"
```

Then every service just calls `self._sql.fqn("dq_schedule_configs")` and the dialect branch disappears from service code entirely.

app/scripts/bind_resources.sh
```diff
@@ -0,0 +1,114 @@
...
+# "apply bundle resource updates to the existing remote resource on the
+# next deploy" — which is exactly what we want, so auto-approving is
+# safe.
+bind() {
+  local key="$1"
+  local id="$2"
+  echo "   binding ${key} -> ${id}"
+  if ! $CLI bundle deployment bind "$key" "$id" --auto-approve "${BUNDLE_FLAGS[@]}"; then
+    echo "   WARNING: bind for ${key} failed. It may already be bound, or the remote resource may not exist yet." >&2
+  fi
```
laurencewells, line 138: A bind failure here is not a safe no-op in all cases. The comment says "It may already be bound, or the remote resource may not exist yet" — but those two situations have completely different implications:

- **Already bound**: The CLI still exits 0 (or gives a clear "already bound" message). In practice this branch would not be hit.
- **Remote resource does not exist**: The subsequent `bundle deploy` will try to CREATE the resource, which is exactly the scenario this script is trying to avoid.
- **Permission error / auth failure**: Silent continuation means the user gets a confusing failure on the next `make app-deploy` with no pointer to the real cause.

Suggest changing to `exit 1` on bind failure and letting the user re-run with `--auto-approve` manually if they know the resource is truly already bound. The current warning-and-continue makes debugging harder, not easier.

app/src/databricks_labs_dqx_app/backend/migrations/postgres.py
```diff
@@ -0,0 +1,294 @@
...
+
+        # ANSI-escape the description rather than parameterising the
+        # INSERT — the executor surface intentionally doesn't expose
+        # parameter binding to keep parity with SqlExecutor.
+        escaped_desc = migration.description.replace("'", "''")
+        self._exec.execute(
+            f"INSERT INTO {self._meta_table} (version, description, applied_at) "
+            f"VALUES ({migration.version}, '{escaped_desc}', CURRENT_TIMESTAMP)"
+        )
+        logger.info("Postgres migration v%d applied", migration.version)
```
laurencewells, line 375: No unit tests for this class or the `PgMigration` dataclass. The `_Executor` Protocol is specifically designed to accept a mock, so the test setup is almost free:

```python
from unittest.mock import MagicMock
from databricks_labs_dqx_app.backend.migrations.postgres import PgMigrationRunner, PgMigration

def make_executor(schema="public"):
    m = MagicMock()
    m.schema = schema
    m.database = "test_db"
    return m

def test_run_all_applies_pending():
    exec = make_executor()
    exec.query.return_value = []  # no applied versions
    runner = PgMigrationRunner(exec)
    count = runner.run_all()
    assert count == len(PG_MIGRATIONS)
    # schema CREATE + meta-table CREATE + all migration statements + INSERT per migration
    assert exec.execute.called

def test_run_all_skips_applied():
    exec = make_executor()
    exec.query.return_value = [[str(m.version)] for m in PG_MIGRATIONS]
    runner = PgMigrationRunner(exec)
    assert runner.run_all() == 0
```

At minimum: skips already-applied, applies pending, handles partial failure (raises, version not recorded).

laurencewells (commented):
LGTM, code review, deploy tested, smoke tested.

Makefile
```diff
@@ -123,14 +123,55 @@ app-grant-permissions:
...
+# ``bundle deploy`` and ``bundle run``. The common use case is
+# overriding ``lakebase_instance_name`` when the original name is
+# locked by Lakebase's ~7-day soft-delete retention after a manual
+# delete (the deploy errors out with "Instance name is not unique"
+# otherwise). Per-deploy CLI overrides keep ``databricks.yml`` clean.
 app-deploy: app-build
 	@test -n "$(PROFILE)" || (echo "Usage: make app-deploy PROFILE=<databricks-profile> TARGET=<bundle-target>"; exit 1)
 	@test -n "$(TARGET)" || (echo "Usage: make app-deploy PROFILE=<databricks-profile> TARGET=<bundle-target>"; exit 1)
-	cd app && databricks bundle deploy -p $(PROFILE) -t $(TARGET)
+	cd app && databricks bundle deploy -p $(PROFILE) -t $(TARGET) $(BUNDLE_VARS)
```
mwojtyczka, line 189: **Must-fix (#1)** — `app-deploy` does not forward `$(BUNDLE_VARS)` to `post_deploy_grants.sh`. If a deployer runs `make app-deploy BUNDLE_VARS='--var=catalog_name=foo --var=schema_name=bar'`, the bundle deploys with the overrides but the grants script (which reads `bundle validate` *without* `--var`) will issue `GRANT`s on the default catalog/schema instead. The `app-bind` target at line 150 correctly forwards `BUNDLE_VARS` — copy that pattern here.

Suggested:
```
app/scripts/post_deploy_grants.sh -p $(PROFILE) -t $(TARGET) $(BUNDLE_VARS)
```

app/src/databricks_labs_dqx_app/backend/config.py
```diff
@@ -42,10 +42,57 @@ class AppConfig(BaseSettings):
...
+    # Leaving these empty keeps the legacy "everything on Delta"
+    # behaviour, so existing deployments continue to work without
+    # changes.  See ``app/databricks.yml`` for the deploy-time toggle.
+    lakebase_instance_name: str = Field(
+        default="",
+        validation_alias="DQX_LAKEBASE_INSTANCE_NAME",
+        description="Lakebase instance name. Empty disables Lakebase routing.",
+    )
+    lakebase_database_name: str = Field(
+        default="dqx_studio",
```
mwojtyczka: **Must-fix (#2)** — `lakebase_database_name` defaults to `"dqx_studio"` here, but `app/databricks.yml:67` declares the bundle var default as `"databricks_postgres"` (the always-present admin DB) and the env-var injection in `databricks.yml` propagates that. Result: when env vars *aren't* propagated (local dev, tests, debug shells), the app tries to connect to a non-existent `dqx_studio` logical DB. Align the code default with the bundle default (`"databricks_postgres"`).

app/databricks.yml
```diff
@@ -27,6 +27,47 @@ variables:
...
+  #
+  # Connection target: ``databricks_postgres`` — the always-present
+  # admin database every Lakebase instance ships with. We don't create
+  # a per-app logical Postgres database because the only DAB resource
+  # that creates one (``database_catalogs``) also creates a Unity
+  # Catalog catalog and therefore requires ``CREATE CATALOG`` on the
+  # metastore — a permission most app deployers don't hold. Using
+  # ``databricks_postgres`` keeps the bundle fully declarative with no
+  # external bootstrap steps, and we still get per-app isolation via a
+  # dedicated Postgres schema (``DQX_LAKEBASE_SCHEMA_NAME`` in the app
```
mwojtyczka, line 54: **Must-fix (#3)** — Comment references `DQX_LAKEBASE_SCHEMA_NAME`, but the actual env var (per `config.py:69`) is `DQX_LAKEBASE_SCHEMA`. Operators copy-pasting this name into local `.env` files will see no effect. Fix the comment to match the real env var name.

docs/dqx/docs/installation.mdx
```diff
@@ -464,26 +465,45 @@ Databricks CLI will confirm a few options:
...
 
-5. Open the deployed app from the **Apps** page in your Databricks workspace.
+6. Open the deployed app from the **Apps** page in your Databricks workspace.
 
-For the full walkthrough — including step-by-step commands, manual `GRANT` statements, troubleshooting, and target-specific configuration — see the [DQX Studio deployment guide](https://github.com/databrickslabs/dqx/blob/v0.14.0/app/DEPLOYMENT.md).
+<Admonition type="note" title="Stateful storage is destroy-protected">
+The studio's schemas (`dqx_studio`, `dqx_studio_tmp`), wheels volume, Lakebase Postgres instance, and the logical Postgres database inside it are all declared as bundle resources with `lifecycle.prevent_destroy: true` (Databricks CLI 0.268+). `databricks bundle destroy` is blocked from dropping them, so production data survives accidental destroy/replace operations. To intentionally tear something down, remove the flag, `databricks bundle deployment unbind <key>`, then destroy manually.
```
mwojtyczka: **Must-fix (#4)** — Documentation overclaim. "the logical Postgres database inside it" does NOT carry `lifecycle.prevent_destroy: true` — it isn't a bundle resource. The app connects to the pre-existing `databricks_postgres` admin database and creates its own schema there at startup; only the *instance* is bundle-managed (and protected). Data safety is preserved via the instance guard, but the doc should accurately describe what's protected.

Same overclaim at `app/README.md:80`. Rewrite to: "The schemas, wheels volume, and Lakebase instance are declared as bundle resources with `lifecycle.prevent_destroy: true`. The app's Postgres schema inside the Lakebase database is created at startup and is protected transitively by the instance-level guard."

app/src/databricks_labs_dqx_app/backend/app.py
```diff
@@ -173,12 +174,65 @@ async def lifespan(app: FastAPI):
...
+                logger.info("Lakebase schema is up to date")
+            set_oltp_executor(pg_executor)
+            logger.info(
+                "Lakebase OLTP routing enabled (instance=%s, database=%s, schema=%s)",
+                conf.lakebase_instance_name,
+                conf.lakebase_database_name,
+                conf.lakebase_schema_name,
+            )
+        except Exception:
+            logger.exception(
```
mwojtyczka, line 275: **Must-fix (#5)** — Silent Lakebase→Delta fallback. `logger.exception` + continue means an existing Lakebase-enabled deployment can flip backends on a transient init failure and start writing OLTP rules to a different physical store than prior reads expect (via `get_sp_oltp_executor`). Operators only notice via log grep.

Suggested: when `lakebase_instance_name` is configured non-empty, init failure should be **fatal** (raise) rather than silent fallback. Fallback should only fire when Lakebase was never configured. At minimum, surface a startup-health flag the UI can display.

app/src/databricks_labs_dqx_app/ui/components/DryRunResults.tsx
```diff
@@ -345,6 +417,9 @@ export function DryRunResults({ result }: DryRunResultsProps) {
                       </th>
```
mwojtyczka, line 417: **Must-fix (#6)** — Stale terminology at line 408 (just above this hunk): `<div>No invalid rows to display</div>`. The surrounding heading is "Failed rows" and the filter elsewhere is "Has failures". Should read "No failed rows to display" to match the rename. Other "Invalid" residues in `runs.tsx:1972,2382` correctly refer to "invalid YAML/schedule config" (not the DQ term) — those are fine.

app/src/databricks_labs_dqx_app/backend/migrations/__init__.py
```diff
@@ -373,10 +559,27 @@ def _applied_at_map(self) -> dict[int, str]:
...
+        # next run sees the constraint and emits this error — safe to
+        # swallow because the desired state is already in place. The
+        # second fragment guards against future error-message tweaks
+        # since Databricks has used both the SQLSTATE-prefixed code and
+        # plain English wording at different times.
+        "DELTA_CONSTRAINT_ALREADY_EXISTS",
+        "constraint already exists",
     )
 
     def _apply(self, migration: Migration) -> None:
```
mwojtyczka, line 761: **Correctness** — Delta `_apply` (just past this hunk, around line 591) splits the migration body on `;` and calls `self._sql.execute(stmt)` per statement. The Statement Execution API auto-commits per call, so a mid-migration failure leaves earlier statements committed but the version row uninserted; recovery depends entirely on `_IDEMPOTENT_ERROR_FRAGMENTS`. The Postgres path (`migrations/postgres.py:298-313`) correctly uses one transactional cursor — Delta side did not get the same fix. Consider documenting the asymmetry or restructuring v1 baselines to be re-runnable.

**Future risk**: any future DDL with `;` inside a string literal (CHECK constraint, default value) will be silently split mid-statement.

app/src/databricks_labs_dqx_app/backend/migrations/__init__.py
```diff
@@ -373,10 +559,27 @@ def _applied_at_map(self) -> dict[int, str]:
     # the whole migration to abort.
     _IDEMPOTENT_ERROR_FRAGMENTS = (
         "COLUMN_ALREADY_EXISTS",
+        # Databricks Delta surfaces an ``ADD COLUMN`` that targets an
+        # already-present column as ``FIELD_ALREADY_EXISTS`` (singular).
+        # Older versions used the plural ``FIELDS_ALREADY_EXISTS``;
+        # keep both to defend against runtime wording drift across
+        # workspaces / DBR versions.
+        "FIELD_ALREADY_EXISTS",
```
mwojtyczka, line 743: **Correctness** — The `FIELD_ALREADY_EXISTS` idempotency fix uses substring match against `str(exc)` (line 593: `if any(frag in str(exc) for frag in self._IDEMPOTENT_ERROR_FRAGMENTS)`). Brittle to message-wording drift across Databricks runtime versions, and could silently mask a real schema-drift problem if a future migration adds a column whose name collides with an existing column of a *different type*.

Suggested: when the SDK exposes an error code, match on `exc.error_code` instead of substrings. Or gate the swallow by version number (only swallow on v3/v4/v5).

app/src/databricks_labs_dqx_app/backend/migrations/postgres.py
```diff
@@ -0,0 +1,313 @@
...
+        Statements are still split on ``;`` and executed one at a time
+        through a single cursor so an error message pinpoints the exact
+        failing DDL rather than a position inside a multi-kilobyte
+        compound string.
+        """
+        formatted = migration.sql.format(schema=self._schema)
+        escaped_desc = migration.description.replace("'", "''")
+        with self._exec.connection() as conn:
+            with conn.cursor() as cur:
+                for stmt in formatted.split(";"):
```
mwojtyczka, line 351: **Correctness** — Splitting the migration body on `;` is brittle. Any future DDL containing `;` inside a string literal (e.g., a CHECK constraint with `';'`, or a default value containing a semicolon) will be silently split mid-statement and produce confusing errors. psycopg accepts multi-statement `cursor.execute()` calls — consider passing the formatted body whole, or use a real SQL splitter that respects quoting.

Low priority right now (current migrations don't have this), but worth tracking.

app/src/databricks_labs_dqx_app/backend/migrations/postgres.py
```diff
@@ -0,0 +1,313 @@
...
+                    stmt = stmt.strip()
+                    if stmt:
+                        cur.execute(stmt)
+                # Same transaction: either the whole migration lands
+                # *and* its version row is recorded, or nothing does.
+                cur.execute(
+                    f"INSERT INTO {self._meta_table} (version, description, applied_at) "
+                    f"VALUES ({migration.version}, '{escaped_desc}', CURRENT_TIMESTAMP)"
```
mwojtyczka: **Correctness / code-style** — `f"VALUES ({migration.version}, '{escaped_desc}', CURRENT_TIMESTAMP)"` builds SQL via f-string with only `'` doubling. `migration.description` is a constant today so this is safe, but the project has a centralized helper (`escape_sql_string` in `utils.py`) — using it here would be consistent with the rest of the codebase and remove the latent injection vector if descriptions ever become user-influenced.

Better still: use psycopg's parameter binding (`cur.execute("INSERT ... VALUES (%s, %s, CURRENT_TIMESTAMP)", (version, description))`) — Postgres has real bind parameters, no reason to build SQL strings here.

app/src/databricks_labs_dqx_app/backend/pg_executor.py
```diff
@@ -0,0 +1,456 @@
...
+        caller is responsible for ``conn.commit()``; if the ``with``
+        block exits without committing (exception or otherwise) the
+        connection is rolled back and returned to the pool by psycopg-
+        pool. Single one-shot statements should keep using
+        :meth:`execute`, which commits per call.
+        """
+        with self._pool.connection() as conn:
+            yield conn
+
+    def execute(self, sql: str, *, timeout_seconds: int = 120) -> None:  # noqa: ARG002 - parity with SqlExecutor
```
mwojtyczka: **Correctness** — `timeout_seconds` is declared (`# noqa: ARG002 - parity with SqlExecutor`) but never used. A misbehaving Postgres query (accidental cross join, missing predicate) will block a pool connection indefinitely. The legacy `SqlExecutor` enforces a deadline via warehouse polling — Postgres has no equivalent here.

Fix: `cur.execute(f"SET LOCAL statement_timeout = {timeout_seconds * 1000}")` before each statement, or `connect_timeout`/per-cursor timeout via psycopg. Don't ship a parameter that lies about what it does.

app/src/databricks_labs_dqx_app/backend/pg_executor.py
```diff
@@ -0,0 +1,456 @@
...
+                self._connect_kwargs["password"] = fresh
+                logger.info("Lakebase OAuth token refreshed")
+            except Exception:  # noqa: BLE001
+                # Don't crash the app on a transient SDK failure — the
+                # next iteration retries and existing connections keep
+                # working until the previous token expires.
+                logger.warning("Failed to refresh Lakebase token; will retry", exc_info=True)
+                # Back off a bit so we don't tight-loop on persistent
+                # failures.
+                time.sleep(60)
```
mwojtyczka: **Risk** — Hard-coded 60s sleep in the token-refresh daemon. Combined with the pool's 50-minute `max_lifetime` recycling, sustained token-refresh failures will drain the pool: every in-pool connection dies at 50min, no valid replacement gets minted, and the app stops accepting work. Operators see this only through downstream connection errors.

Suggested: shorter back-off (5–10s with jitter), surface a `last_successful_refresh_at` metric, and make the loop exit (re-raise) after N consecutive failures so the supervisor restarts the process.

app/src/databricks_labs_dqx_app/backend/pg_executor.py
```diff
@@ -0,0 +1,456 @@
...
+
+    def execute(self, sql: str, *, timeout_seconds: int = 120) -> None:  # noqa: ARG002 - parity with SqlExecutor
+        """Run a single non-result-returning statement and commit it."""
+        with self._pool.connection() as conn:
+            with conn.cursor() as cur:
+                # psycopg accepts ``str`` at runtime (``Query`` is a
+                # union of bytes/str/Composable/SQL); the published
+                # stubs only declare the ``Template`` overload so we
+                # silence basedpyright here.
+                cur.execute(sql)  # pyright: ignore[reportCallIssue, reportArgumentType]
```
mwojtyczka: **Convention** — `# pyright: ignore[reportCallIssue, reportArgumentType]` repeats on lines 296, 319, 327. AGENTS.md: "Fix the code instead of adding `# pylint: disable`, `# type: ignore`, `# noqa`." If psycopg's stubs reject `cur.execute(sql)` where `sql: str`, the fix is to cast the argument once with a properly typed wrapper (`cast(LiteralString, sql)` or a `_Query` newtype) — not to suppress at every call site.

app/src/databricks_labs_dqx_app/backend/pg_executor.py
```diff
@@ -0,0 +1,456 @@
...
+
+    # ------------------------------------------------------------------
+    # Lifecycle
+    # ------------------------------------------------------------------
+
+    def close(self) -> None:
+        self._stop.set()
+        try:
+            self._pool.close()
+        except Exception:  # noqa: BLE001 - best-effort shutdown
```
mwojtyczka: **Convention** — `# noqa: BLE001` here and at line 403. AGENTS.md prohibits silencing lint to avoid fixes. Broad `except Exception` IS appropriate for best-effort shutdown / token-refresh resilience, but the right move is either:
  - narrow to `except (psycopg.Error, OSError):`, or
  - add a project-wide exemption in `pyproject.toml` documented for shutdown handlers,

not per-line `# noqa` that future readers have to inspect each time.

app/src/databricks_labs_dqx_app/backend/dependencies.py
```diff
@@ -31,6 +31,40 @@
...
 
+if TYPE_CHECKING:
+    from .pg_executor import PgExecutor
+
+# Union of the two executors that share the OLTP service surface
+# (``execute``, ``query``, ``query_dicts``, ``upsert``, ``q``,
+# ``json_literal_expr``, ``ts_text``, ``dialect``). String forward refs
+# because ``PgExecutor`` is TYPE_CHECKING-only to avoid importing psycopg
+# at module load when Lakebase isn't configured.
+OltpExecutor = Union["SqlExecutor", "PgExecutor"]
```
mwojtyczka: **Convention** — `OltpExecutor = Union["SqlExecutor", "PgExecutor"]` should be PEP 604: `OltpExecutor = "SqlExecutor | PgExecutor"`. AGENTS.md mandates `str | None` over `Optional[str]`, `list[str]` over `List[str]` — same rule applies to `Union`.

Separately, `get_sp_oltp_executor` (lines 153–176) does `cast(SqlExecutor, pg)` to satisfy downstream service signatures. That defeats static typing for every new Postgres code path. The `_Executor` Protocol you already declared at `migrations/postgres.py:51-68` is the right shape — promote it to an `OltpExecutorProtocol` in this module and have services depend on the protocol instead of the concrete `SqlExecutor` class.

app/src/databricks_labs_dqx_app/backend/app.py
```diff
@@ -10,9 +10,10 @@
...
 from ._scheduler_registry import get_scheduler, set_scheduler
 from .config import conf
-from .dependencies import get_sp_ws
+from .dependencies import get_sp_ws, set_oltp_executor
 from .logger import logger
 from .migrations import MigrationRunner
+from .migrations.postgres import PgMigrationRunner
 from .routes import api_router
 from .services.scheduler_service import SchedulerService
 from .services.view_service import mark_tmp_schema_ready
```
mwojtyczka, line 28: **Convention** — at line 39 (just below this import hunk): `globals()["_scheduler_lock_fd"] = fd`. This is an awkward way to keep an fd alive across function returns. Cleaner:

```python
_scheduler_lock_fd: int | None = None  # module-level, at top of file

def _acquire_scheduler_lock() -> bool:
    global _scheduler_lock_fd
    ...
    _scheduler_lock_fd = fd
```

Avoids the `globals()` magic-string lookup.

app/src/databricks_labs_dqx_app/backend/services/app_settings_service.py
```diff
@@ -55,25 +56,10 @@ def get_config(self) -> WorkspaceConfig:
             return WorkspaceConfig(run_configs=[])
```
mwojtyczka, line 93: **Convention** — `Installation._unmarshal_type(...)` is a private blueprint API (leading underscore). Calling it directly couples DQX to blueprint internals; the next blueprint release can break this without warning.

Options: (1) inline the deserialization with `dataclasses.fields` + `getattr` since `WorkspaceConfig` is a well-known shape, or (2) file an upstream issue to expose a public unmarshal helper.

app/src/databricks_labs_dqx_app/backend/services/role_service.py
```diff
@@ -96,25 +98,42 @@ def get_mappings_for_role(self, role: str) -> list[RoleMapping]:
...
-            f"USING (SELECT '{escaped_role}' AS role, '{escaped_group}' AS group_name) AS source "
-            "ON target.role = source.role AND target.group_name = source.group_name "
-            "WHEN MATCHED THEN UPDATE SET "
-            f"  updated_by = '{escaped_user}', "
-            "  updated_at = current_timestamp() "
-            "WHEN NOT MATCHED THEN INSERT (role, group_name, created_by, created_at, updated_by, updated_at) "
-            f"VALUES ('{escaped_role}', '{escaped_group}', '{escaped_user}', current_timestamp(), "
-            f"'{escaped_user}', current_timestamp())"
-        )
+        if getattr(self._sql, "dialect", "delta") == "postgres":
```
mwojtyczka: **Convention / DRY** — `if getattr(self._sql, "dialect", "delta") == "postgres":` is duplicated across `role_service.py:115`, `schedule_config_service.py:84`, `rules_catalog_service.py:94`, `scheduler_service.py:615`, `comments_service.py:73`. Every service emits dialect-specific SQL for the same logical operation (upsert with audit columns, select JSON column).

Lift these into the executor protocol — e.g.
```python
def upsert_audit(self, table, key_cols, value_cols, *, preserve_created=True): ...
def select_json_text(self, col: str) -> str: ...
```
Services stay dialect-agnostic; new backends only need to implement the protocol, not be scattered through `if` branches.

app/src/databricks_labs_dqx_app/backend/services/scheduler_service.py
```diff
@@ -77,21 +119,44 @@ def __init__(
...
         self._catalog = catalog
         self._schema = schema
         self._tmp_schema = tmp_schema
         self._sql = SqlExecutor(ws=ws, warehouse_id=warehouse_id, catalog=catalog, schema=schema)
         self._tmp_sql = SqlExecutor(ws=ws, warehouse_id=warehouse_id, catalog=catalog, schema=tmp_schema)
+        # OLTP executor — either a PgExecutor (Lakebase) or the same
+        # Delta executor (legacy mode).  All schedule / settings /
+        # rule access goes through this; only analytical table
+        # operations (retention sweep, orphan view GC) use ``self._sql``.
+        self._oltp_sql: SqlExecutor = cast(SqlExecutor, oltp_sql) if oltp_sql is not None else self._sql
```
mwojtyczka: **Design** — `self._oltp_sql: SqlExecutor = cast(SqlExecutor, oltp_sql)`. Casting a `PgExecutor` to `SqlExecutor` so existing service signatures still typecheck — but this silently disables type-checker help for every Postgres path inside this class. Once a shared `OltpExecutorProtocol` exists (see comment on `dependencies.py:42`), this annotation becomes `OltpExecutorProtocol` and the cast disappears.

app/src/databricks_labs_dqx_app/backend/migrations/__init__.py
```diff
@@ -300,6 +477,15 @@ def run_all(self) -> int:
...
+                    "Skipping Delta OLTP fallback migration v%d "
+                    "(Lakebase enabled — these tables live in Postgres): %s",
+                    migration.version,
+                    migration.description,
+                )
+                continue
+
             logger.info(
                 "Applying migration v%d: %s",
                 migration.version,
```
mwojtyczka, line 662: **Convention / consistency** — at line 530 (just past this hunk): `sql = f"CREATE SCHEMA IF NOT EXISTS {self._catalog}.{self._schema}"` interpolates catalog/schema identifiers raw. The Postgres counterpart at `migrations/postgres.py:265` correctly double-quotes (`f'CREATE SCHEMA IF NOT EXISTS "{self._schema}"'`). Trusted-input today (values come from `databricks.yml` vars), but the inconsistency is a foot-gun the next time someone adds catalog names with dashes.

Suggested: a shared `q_identifier(name: str) -> str` helper on the executor.

app/tests/test_pg_migration_runner.py
```diff
@@ -0,0 +1,304 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _make_executor(*, applied_versions: tuple[int, ...] = (), schema: str = "public") -> MagicMock:
+    """Build a mock ``_Executor`` pre-seeded with the given applied-versions list."""
+    exec_mock = MagicMock(name="PgExecutor")
```
mwojtyczka: **Test quality** — `MagicMock(name="PgExecutor")` should be `create_autospec(PgExecutor, instance=True)`. AGENTS.md explicitly: "Use dependency injection to enable testing: construct dependencies with `create_autospec` rather than patching internal module state."

The footgun you hand-fix at lines 50–55 (auto-`__exit__` returning truthy swallowing exceptions) is the *exact* case `create_autospec` prevents — it would give you a typed spec where `__exit__` returns `None`/`False` by default.

Same fix applies to the `monkeypatch.setattr(... PG_MIGRATIONS, ...)` at lines 155, 261, 290 — accept a migrations list as a constructor arg on `PgMigrationRunner` so tests can inject directly without monkeypatching the module-level constant.

app/tests/test_retention.py
```diff
@@ -0,0 +1,248 @@
...
+
+class TestSchedulerResolveRetention:
+    """Scheduler must never crash on a missing / corrupt retention setting."""
+
+    @pytest.fixture
+    def scheduler(self):
+        from databricks_labs_dqx_app.backend.services.scheduler_service import SchedulerService
+
+        # Bypass __init__ — only the attributes the resolver touches.
+        svc = SchedulerService.__new__(SchedulerService)
```
mwojtyczka: **Test quality** — `SchedulerService.__new__(SchedulerService)` + hand-attaching private attributes (`_oltp_sql`, `_sql`, `_settings_table`, `_catalog`) at lines 138-144 and again at 207-219, plus once in `test_custom_metrics.py:166`. This couples tests to internal field names: if the service refactors `_oltp_sql` → `_oltp_executor`, every test silently breaks with `AttributeError` at access time rather than `TypeError` at construction.

Fix: construct with the real `__init__` and inject dependencies via constructor args (the constructor already accepts `sql`, `oltp_sql`, `installation`, `settings_table`, `catalog`).

app/src/databricks_labs_dqx_app/backend/pg_executor.py
```diff
@@ -0,0 +1,456 @@
+"""Lakebase Postgres executor.
```
mwojtyczka, line 1: **Test coverage gap** — `PgExecutor` has no direct unit tests in this PR. Specifically uncovered:

- `_to_text` coercion (datetimes, Decimal, bytes, bool-vs-int ordering)
- `_pg_render_value` `RawSql` translation (only handles `current_timestamp()`/`now()`; what does it do with anything else?)
- `upsert` SQL shape — both `DO NOTHING` (empty value_cols) and `DO UPDATE`
- `q()` quoting of reserved words
- `connection()` context-manager: commit-on-success, rollback-on-exception
- `_token_refresh_loop` happy and sad paths (no test exists for the 60s back-off, no test for what happens when refresh fails permanently)
- `build_pg_executor`: empty-username and missing-DNS error paths (lines 436-444)

`test_pg_migration_runner.py` only mocks the executor protocol — it doesn't exercise `PgExecutor` itself.

app/src/databricks_labs_dqx_app/backend/services/scheduler_service.py
```diff
@@ -628,6 +796,136 @@ def _gc_orphan_views(self) -> None:
...
+                interval = f"INTERVAL '{days} days'"
+            else:
+                interval = f"INTERVAL {days} DAY"
+            stmt = f"DELETE FROM {table} " f"WHERE {time_col} < CURRENT_TIMESTAMP - {interval}"
+            try:
+                self._oltp_sql.execute(stmt)
+                logger.info("Retention sweep (OLTP): cleaned %s (cutoff=%dd)", table_name, days)
+                total_deleted += 1
+            except Exception as exc:
+                logger.warning("Retention sweep: %s failed (%s); continuing", table_name, exc)
```
mwojtyczka, line 919: **Test coverage gap** — `_run_retention`'s Postgres branch (around lines 912-919) is uncovered. `test_retention.py:TestRunRetentionUsesQuarantineCutoff` sets `dialect = "delta"` only. Add a parallel `dialect = "postgres"` test asserting that retention uses `INTERVAL '<N> days'` Postgres syntax (not the Delta `INTERVAL N DAY` syntax), and that quarantine vs. trend cutoffs are routed correctly.

mwojtyczka (changes requested):

Makefile
berrybluecode: The suggested ... $(BUNDLE_VARS) (no separator) would crash the script, because post_deploy_grants.sh uses getopts "p:t:" and would error out on the first --var=... it sees. The pattern that already works on app-bind (line 158) is $(if $(BUNDLE_VARS),-- $(BUNDLE_VARS)) — a literal -- then the vars — which terminates getopts cleanly. I applied that same pattern.

Changes
app/scripts/post_deploy_grants.sh — mirror bind_resources.sh:

Added shift $((OPTIND - 1)) after the getopts loop.
Captured the rest into EXTRA_VARS=("$@").
Threaded "${EXTRA_VARS[@]}" into the bundle validate call (line 73).
Updated the usage string and the script header comment to document the -- separator and warn that the overrides must match deploy-time.
Makefile — two targets, two doc blocks:

app-deploy: app/scripts/post_deploy_grants.sh -p $(PROFILE) -t $(TARGET) $(if $(BUNDLE_VARS),-- $(BUNDLE_VARS)) — same $(if ...) idiom already used on app-bind.
app-grant-permissions: applied the same forwarding (otherwise a user re-running grants manually after a BUNDLE_VARS deploy would still hit the original bug).
Expanded the app-deploy header comment to call out that the overrides flow through bundle deploy → post_deploy_grants.sh → bundle run, and to explain why threading them through the grants script is load-bearing.

app/src/databricks_labs_dqx_app/backend/config.py
berrybluecode: app/src/databricks_labs_dqx_app/backend/config.py — changed default from "dqx_studio" to "databricks_postgres" so the Python default matches the bundle default. Added a comment explaining the bundle invariant, why these two defaults must stay in sync, and that schema-level isolation (lakebase_schema_name, still "dqx_studio") is unaffected.
app/DEVELOPMENT.md:50 — fixed the .env template line that had the same stale value. Without this, anyone copy-pasting the documented .env block would reintroduce the exact bug the must-fix is closing.

app/databricks.yml
berrybluecode: app/databricks.yml:54 — corrected DQX_LAKEBASE_SCHEMA_NAME → DQX_LAKEBASE_SCHEMA to match the actual validation_alias in app/src/databricks_labs_dqx_app/backend/config.py:78.

app/src/databricks_labs_dqx_app/backend/app.py
berrybluecode: Fix
app/src/databricks_labs_dqx_app/backend/app.py (the core of the change):

When conf.lakebase_enabled is true and Lakebase init throws, the handler now: (a) closes any partially-built psycopg pool so a restart loop doesn't accumulate orphan server-side connections, (b) resets set_oltp_executor(None) for hygiene, (c) logs a loud, actionable message, and (d) re-raises so the lifespan handler aborts and the Databricks Apps platform restarts the container.
Rewrote the comment block that previously defended silent fallback. The old comment said "partial functionality beats a hard crash loop" — true for many subsystems (scheduler, volume sync), wrong for an OLTP backend selector. New comment mirrors the existing fail-loud reasoning at lines 170–173 (which was already correctly applied to sp_ws and migrations).
New error message lists the four realistic causes (provisioning, permission, OAuth, network reachability) and tells the operator the exact env var to clear (DQX_LAKEBASE_INSTANCE_NAME="") if they want to intentionally run on Delta only.
Two doc updates (the old silent-fallback behavior was advertised as a feature in three places — leaving them would point operators at obsolete troubleshooting):

docs/dqx/docs/installation.mdx:506 — updated the first-start note.
app/DEPLOYMENT.md:158 — updated the first-start callout with the new log message.
app/DEPLOYMENT.md:336 — rewrote the entire troubleshooting entry. It now matches the new log message, explains why the app refuses to start (split-brain risk), and provides four numbered diagnostic steps including the Lakebase instance state machine (AVAILABLE / STARTING / UPDATING).
Untouched (left alone deliberately)
The legitimate "var unset → Delta only" path (else branch at app.py:253-255). That's still the documented opt-out, now also surfaced in the new error message.
app/DEVELOPMENT.md:62 and docs/dqx/docs/dev/contributing.mdx:206 — both describe local dev with the var intentionally empty. Correct behavior, unchanged.
app/src/databricks_labs_dqx_app/backend/CLAUDE.md:184 — describes the legitimate Delta-only path. Correct, unchanged.
On the reviewer's "at minimum, surface a startup-health flag the UI can display"
Since we went with the stronger fix (fatal failure), the health flag becomes redundant — there's no degraded runtime state for the UI to surface. The app either starts on Lakebase or it doesn't start at all, and the Databricks Apps platform makes the "not started" state visible via its app-health UI / restart-loop alerting. If you specifically want an in-app health flag (e.g. for an Admin → System Status panel), say the word and I'll add one — but it would only ever report "healthy" under the new behavior, so it'd be window dressing.

app/src/databricks_labs_dqx_app/ui/components/DryRunResults.tsx
berrybluecode: app/src/databricks_labs_dqx_app/ui/components/DryRunResults.tsx:408 — changed "No invalid rows to display" → "No failed rows to display" so the empty-state message matches the section heading "Failed rows" (line 339) and the global filter "Has failures" (runs-history.tsx:496).

app/src/databricks_labs_dqx_app/backend/migrations/__init__.py
berrybluecode: app/src/databricks_labs_dqx_app/backend/migrations/__init__.py
New _validate_template_split_safe(template: str) helper — small state-machine that walks the template, tracks single-quoted-literal state (correctly handling the SQL '' apostrophe escape), and raises AssertionError with an offset + 80-char excerpt the moment it sees ; inside a literal.
Module-level loop at the bottom of the MIGRATIONS block runs the validator against every template at app import. Future authors who introduce a literal-embedded ; will see an AssertionError on python -c "from … import migrations" rather than a partially-applied migration in production.
Expanded module docstring with a new "Atomicity model — Delta vs Postgres asymmetry" section. Spells out why the asymmetry exists (platform constraint), what the recovery contract is, and gives the next migration author a concrete three-point checklist (idempotent DDL kinds, error fragments to maintain, ;-in-literal ban).
New _apply docstring mirroring the Postgres one — explains why this method can't be transactional, what failure modes look like, and pointers to the contract.
app/tests/test_migration_runner.py (new file)
13 tests in two classes:

TestValidateTemplateSplitSafe — parametrized positive cases (empty, multi-statement, literals without ;, '' apostrophe escape, adjacent literals) and negative cases (the exact reviewer-flagged class: ; in a CHECK literal, ; in a default value, ; after '' escape, trailing ; in literal). Plus an assertion-message quality test (offset + excerpt must be present).
TestLiveMigrationsAreSplitSafe — regression that runs the validator against every entry in MIGRATIONS. Fast PR-time feedback for any future author who edits a template into a no-longer-safe shape.
What I did NOT do (and why)
Did not attempt to make Delta _apply transactional. Not possible — the Databricks Statement Execution API doesn't expose BEGIN/COMMIT for DDL on a SQL warehouse. The reviewer knew this; the explicit ask was "document the asymmetry or restructure to be re-runnable", not "force atomicity."
Did not swap str.split(";") for sqlparse.split (or similar real splitter). Today every template is split-safe, and sqlparse would be a new dependency on a hot startup path. The validator + module-import assertion gets us the same safety guarantee for free. The migration author docstring tells future authors what to do if they ever genuinely need ; in a literal: replace the splitter with sqlparse.split then.

app/src/databricks_labs_dqx_app/backend/migrations/__init__.py
berrybluecode: Changes
app/src/databricks_labs_dqx_app/backend/sql_executor.py
Hardened SqlExecutor.q() to double internal backticks, mirroring PgExecutor.q()'s symmetric handling of internal double-quotes. Old form f"{identifier}" would silently break for any identifier containing a backtick.
Expanded the docstring to spell out when you must quote (reserved words and any non-[A-Za-z0-9_] characters, with hyphenated catalogs as the concrete example).
app/src/databricks_labs_dqx_app/backend/migrations/__init__.py
MigrationRunner.__init__ now pre-computes self._catalog_q = sql.q(sql.catalog) and self._schema_q = sql.q(sql.schema) and builds _meta_table from the quoted forms.
_ensure_schema uses the quoted forms in the CREATE SCHEMA DDL.
_apply substitutes the quoted forms into the migration template's {catalog} / {schema} placeholders.
Extended the import-time validator: renamed _validate_template_split_safe → _validate_template_safe and extracted the assertion-raising into a helper. It now checks two invariants instead of one: (1) no ; inside a string literal (existing), and (2) no {catalog}/{schema} placeholder inside a string literal (new — because substituting the quoted form into literal context would emit nonsense). Both produce actionable error messages with offset, excerpt, and a remediation hint specific to which invariant was broken.
Updated the module docstring's "Atomicity model" section to reflect the broadened invariants.
app/src/databricks_labs_dqx_app/backend/migrations/postgres.py
Mirror-symmetric treatment: PgMigrationRunner.__init__ pre-computes self._schema_q = executor.q(self._schema) and uses it everywhere (meta-table reference, _ensure_schema, _ensure_meta_table, _apply). Previously _ensure_schema was the only quoted site while _meta_table and the migration templates raw-interpolated — that asymmetry is gone now.
Added q(identifier: str) -> str to the _Executor Protocol so the dialect contract is part of the runner's typed dependency surface, not implicit.
app/tests/test_migration_runner.py
27 tests across four classes (was 13):

TestValidateTemplateSafe — positive cases (including the new "placeholder in object-name position is fine"), parametrized negative cases for both invariants with per-case message assertions.
TestLiveMigrationsAreTemplateSafe — regression that runs the validator against the live MIGRATIONS list.
TestSqlExecutorQuoting — direct tests of the hardened q() covering the reviewer-mentioned prod-east / team-data-platform, reserved-word check, internal-backtick doubling (including pre-doubled), and Unity Catalog unicode.
TestMigrationRunnerUsesQuotedIdentifiers — end-to-end regression with a real SqlExecutor over catalog="prod-east". Asserts _meta_table, _ensure_schema's emitted DDL, and _apply's template substitution all reference the quoted form, plus an explicit "raw prod-east.dqx_studio must NOT appear anywhere in the captured SQL" check that would have caught the original bug.
app/tests/test_pg_migration_runner.py
Updated _make_executor to expose q() on the mock with the same behavior as PgExecutor.q so the now-required Protocol method is satisfied. No behavioral changes to existing tests; all 16 pre-existing PG tests still pass.
What I did NOT do (and why)
Did not move q() into sql_utils.py as a free function. It's intentionally dialect-specific — sql_utils.quote_fqn already exists for the multi-part Delta FQN case, but per-identifier quoting belongs on the executor where it can vary per backend without callers branching on dialect.
Did not re-quote the literal table names in migration templates (dq_app_settings, dq_quality_rules, etc.). Those are hardcoded ASCII identifiers with no risk of containing reserved characters; quoting them would be noise and would also change the captured-SQL assertion shape in integration tests.
Did not touch other call sites that interpolate sql.catalog / sql.schema into f-strings outside the migration runners. Out of scope for this review item, but worth a follow-up: a quick grep -n '{sql\\.catalog}\\|{sql\\.schema}' over the services would surface the same class of bug elsewhere.

app/src/databricks_labs_dqx_app/backend/pg_executor.py
berrybluecode: (PgExecutor test coverage gap) is closed.

Result: 82 new tests in app/tests/test_pg_executor.py, full suite green (511/511), no lints, 0% → 90% line coverage on pg_executor.py. The remaining 10% is lines 159-218 — the __init__ body that opens the real ConnectionPool and spawns the refresher thread, which is integration-only by nature.

What's covered, mapped to each bullet in the review comment:

_to_text coercion — TestToText (12 cases): None, dict/list → compact JSON, aware + naive datetime, date, bool True/False (with an explicit note that the bool branch MUST come before int because bool is an int subclass — the regression the reviewer was guarding against), Decimal, ints, floats, bytes / bytearray / memoryview → hex, plain str, custom object falling back to __str__.
_pg_render_value RawSql translation — TestPgRenderValue: the 4 Spark-idiom variants (case + whitespace) collapse to CURRENT_TIMESTAMP; everything else (CURRENT_TIMESTAMP, uuid_generate_v4(), arbitrary subexpressions, EXCLUDED.col references) passes through verbatim. Plus bool → TRUE/FALSE, None → NULL, numerics, and ANSI-escaped strings.
upsert SQL shape — TestUpsertSqlShape: both branches (DO UPDATE when value cols populated, DO NOTHING when empty) with full SQL-shape assertions; ValueError on empty key_cols; RawSql("current_timestamp()") in value cols is translated; reserved-word column names (check, order) are quoted in both the INSERT list and the ON CONFLICT target.
q() quoting of reserved words — TestPgQuoting: parametrised over user, order, check, internal " doubling, pre-doubled input, Unicode, empty string.
connection() context manager — TestConnectionContextManager: yielded object is the pool's connection; our wrapper never implicitly commits or rolls back (proving the documented caller-managed contract); exceptions inside the block propagate.
_token_refresh_loop happy/sad/permanent paths — TestTokenRefreshLoop: happy path updates both _token_holder and _connect_kwargs["password"]; stop-event-during-wait terminates without calling _generate_token; sad path warns + sleeps exactly 60s + recovers on next iteration with the recovered token; permanent failure spins indefinitely without crashing, leaving the previous token intact so existing connections keep working.
build_pg_executor error paths — TestBuildPgExecutor: missing read_write_dns raises before asking for the calling identity; empty user_name AND id raises; user_name=None with id="sp-…" falls through to use the id (proving the or-fallback isn't accidentally short-circuiting).
Plus bonus coverage of the surface area I needed mocks for anyway: dialect, schema, database, catalog, warehouse_id, fqn, ts_text, json_literal_expr, _TokenHolder (including a concurrency smoke test), _generate_token (success + missing-token), execute, execute_no_schema, query, query_dicts, and close (including the documented best-effort behaviour when pool close itself raises).

Key strategic decision worth flagging: tests build a half-initialised PgExecutor via __new__ rather than mocking ConnectionPool + threading.Thread + _generate_token at the patch boundary. The helper sets only the attributes a given test needs, which keeps each test focused on the behaviour-under-test and avoids the test becoming a tautology against three layers of patches. The module docstring explains this strategy so the next contributor sees why.

app/src/databricks_labs_dqx_app/backend/services/scheduler_service.py
berrybluecode: Postgres retention dialect coverage gap) closed with 6 new tests in app/tests/test_retention.py, fully parallel to the existing TestRunRetentionUsesQuarantineCutoff Delta-dialect class.

app/tests/test_pg_migration_runner.py
berrybluecode: Test-side changes (4 new tests in TestMigrationsInjection):

test_default_is_the_module_level_catalogue — verifies the no-arg default still wires to PG_MIGRATIONS.
test_injected_list_is_used_in_place_of_default — proves the injected list shadows production catalogue, and explicitly asserts no leak from PG_MIGRATIONS (would catch a regression where the runner accidentally chained both lists).
test_injected_list_is_snapshotted_defensively — proves the tuple(...) snapshot survives mutation of the caller's list.
test_empty_injected_list_runs_zero_migrations_but_still_bootstraps — proves an empty list is a valid input (zero migrations applied, but schema + meta-table still bootstrap).
Net result: 521/521 tests pass (4 new), zero monkeypatch.setattr remaining in test_pg_migration_runner.py, zero MagicMock(name=...) for PgExecutor. A future rename like PgExecutor.query → PgExecutor.run_query would now produce loud AttributeErrors in 12+ test cases simultaneously instead of silently passing against stale untyped mocks.

app/src/databricks_labs_dqx_app/backend/services/scheduler_service.py
berrybluecode: Five OLTP services to update. Three keep SqlExecutor (scheduler — uses tmp/analytical SqlExecutor too; view_service — both sql and sp_sql are SqlExecutor; job_service — analytical SqlExecutor). Apply the change to the five OLTP-only ones.

app/src/databricks_labs_dqx_app/backend/services/app_settings_service.py
berrybluecode: Concrete diff:

app_settings_service.py — three edits:

Removed from databricks.labs.blueprint.installation import Installation. DQX no longer touches blueprint internals from this file.

Added a small module-private unmarshaller (~40 lines) that replaces Installation._unmarshal_type. Handles exactly the typing constructs that show up in the WorkspaceConfig tree — nested dataclasses, list[T], dict[str, T], T | None, primitives — and nothing else. Unknown keys in the stored JSON are ignored, missing keys fall back to dataclass defaults, and __post_init__ validators (OutputConfig, the storage configs) still run because we go through cls(**kwargs).

get_config now calls _unmarshal_dataclass(data, WorkspaceConfig) and catches (TypeError, ValueError) to keep the existing soft-fallback behavior (logger.warning → return default). The old isinstance(result, WorkspaceConfig) defensive check is gone because the helper either returns a WorkspaceConfig or raises — isinstance was only there to paper over blueprint's typing.

app/src/databricks_labs_dqx_app/backend/sql_executor.py
```diff
@@ -187,6 +504,113 @@ def query_dicts(self, sql: str, *, timeout_seconds: int = 120) -> list[dict[str,
...
+
+        on_clause = " AND ".join(f"target.{k} = source.{k}" for k in key_cols)
+        source_select = ", ".join(f"{_render_value(v)} AS {k}" for k, v in key_cols.items())
+
+        # UPDATE SET excludes created_* columns when preserve_created;
+        # the increment column (if any) gets the dialect-specific
+        # self-reference form instead of the literal value.
+        update_pairs: list[str] = []
+        for col, val in value_cols.items():
+            if preserve_created and col.startswith("created_"):
```
mwojtyczka, line 599: **Regression introduced by the refactor** — Delta `upsert_with_audit` path does NOT quote column names via `self.q(col)` here or in the `insert_cols` build below at line 596. The Postgres mirror at `pg_executor.py:1109,1120` correctly quotes via `self.q(c)`. Result: a future audit column named `check`, `order`, `group`, or any other reserved word would produce valid Postgres SQL but a Delta parse error.

Current callers (`role`, `group_name`, `schedule_name`, `version`, etc.) all use safe names, so this is latent. Suggested fix here and at the `insert_cols` join:
```python
update_pairs.append(f"{self.q(col)} = target.{self.q(col)} + 1")
...
insert_cols = ", ".join(self.q(c) for c in all_cols)
```

app/src/databricks_labs_dqx_app/backend/pg_executor.py
```diff
@@ -0,0 +1,783 @@
...
+from psycopg import Connection, Cursor
+from psycopg_pool import ConnectionPool
+
+from databricks_labs_dqx_app.backend.sql_executor import RawSql, _render_value
+from databricks_labs_dqx_app.backend.sql_utils import escape_sql_string
+
+logger = logging.getLogger(__name__)
+
+
+def run_trusted_sql(cur: Cursor[Any], sql: str) -> None:
```
mwojtyczka: **Security-contract concern** — The docstring lists "hand-written DDL / migration strings vendored alongside the code" as a trusted input. But this helper's callers include `migrations/postgres.py:355` which passes a SQL string containing `'{escaped_desc}'` (manual `replace("'", "''")`) for the migration description. The `LiteralString` cast here silently rubber-stamps any caller that builds SQL with f-strings, regardless of whether values are properly bound.

Tighten the contract: "static literals + identifier-validated via `q()` + value literals rendered via `_render_value`/`_pg_render_value`". Value literals from non-static sources should go through `cur.execute(sql, params)` directly (psycopg native parameter binding) rather than this helper. Otherwise the helper provides type-checker silence without actual injection protection.

app/src/databricks_labs_dqx_app/backend/pg_executor.py
```diff
@@ -0,0 +1,783 @@
...
+                # policy block in ``pyproject.toml``.
+                self._consecutive_refresh_failures += 1
+                logger.warning(
+                    "Failed to refresh Lakebase token (attempt %d/%d); will retry",
+                    self._consecutive_refresh_failures,
+                    self._token_refresh_max_failures,
+                    exc_info=True,
+                )
+                if self._consecutive_refresh_failures >= self._token_refresh_max_failures:
+                    self._escalate_refresh_failure()
```
mwojtyczka, line 714: **Test coverage gap** — `self._escalate_refresh_failure()` is the load-bearing safety net for the original 60s-sleep/pool-drain finding. Comment at this line suggests a test patches `_escalate_refresh_failure` ("return only reached if _escalate is patched in tests"), but I don't see one in `test_pg_executor.py`'s `TestRefreshFailureEscalation` that actually verifies the `os._exit(1)` path fires after exactly `max_failures` consecutive failures.

Also missing: "fail N-1 → succeed → fail N" sequence — `TestTokenRefreshLoop::test_failure_then_success_increments_then_resets_counter` proves "success resets to 0", but not "the next escalation needs a fresh N consecutive failures, not 1". Add a test that interleaves failures with one success, then drives N more failures, and asserts escalation fires on the Nth, not the 2nd.

app/src/databricks_labs_dqx_app/backend/app.py
```diff
@@ -10,9 +10,20 @@
...
 from ._scheduler_registry import get_scheduler, set_scheduler
 from .config import conf
-from .dependencies import get_sp_ws
+from .dependencies import get_sp_ws, set_oltp_executor
 from .logger import logger
 from .migrations import MigrationRunner
+
+# ``PgMigrationRunner`` is intentionally NOT imported here. Its
+# module (`migrations.postgres`) eagerly imports ``pg_executor``,
+# which in turn imports ``psycopg`` — a Lakebase-only optional
```
mwojtyczka: **Workaround acknowledged in comment** — The lazy `PgMigrationRunner` import is necessary, but `migrations/postgres.py:38` itself imports `run_trusted_sql` from `pg_executor`, which imports `psycopg` at module top. So the lazy import here just defers the failure; it doesn't decouple psycopg from the migration runner.

Cleaner fix for a follow-up: move `run_trusted_sql` to a psycopg-free helper module (it only needs `cast(LiteralString, sql)` — no psycopg types at runtime). Then `migrations/postgres.py` can be imported even without psycopg installed, and the lazy-import dance becomes unnecessary.

app/tests/test_pg_executor.py
```diff
@@ -0,0 +1,1492 @@
...
+        assert _to_text(dt.date(2026, 5, 28)) == "2026-05-28"
+
+    def test_bool_true_renders_lowercase(self) -> None:
+        """Critical: bool must match BEFORE int (bool is an int subclass)."""
+        assert _to_text(True) == "true"
+
+    def test_bool_false_renders_lowercase(self) -> None:
+        assert _to_text(False) == "false"
+
+    def test_decimal_renders_as_str_no_scientific(self) -> None:
```
mwojtyczka: **Misleading test name** — `test_decimal_renders_as_str_no_scientific` then asserts `_to_text(Decimal("0.0000001")) == "1E-7"`. `1E-7` IS scientific notation; the test passes because that's exactly what `Decimal.__str__` returns. Either:
  - rename to `test_decimal_uses_decimal_str_repr` (documents the actual behaviour), or
  - if "no scientific" is the intended contract, the test is wrong — use `format(d, 'f')` in `_to_text` to force plain notation.

Given downstream Postgres and CSV consumers handle `1E-7` fine, the rename is the right move.

app/tests/test_pg_migration_runner.py
```diff
@@ -0,0 +1,455 @@
...
+                "CREATE TABLE {schema}.t1 (id int);"
+                "CREATE TABLE {schema}.t2 (id int);"
+                "CREATE TABLE {schema}.t3 (id int);"
+            ),
+        )
+
+        exec_mock = _make_executor(applied_versions=())
+        cur = _cursor_of(exec_mock)
+        # Pass: t1 → succeed, t2 → succeed, t3 → boom. INSERT never reached.
+        cur.execute.side_effect = [None, None, RuntimeError("boom"), None, None]
```
mwojtyczka: **Brittle positional coupling, unchanged from original review** — `cur.execute.side_effect = [None, None, RuntimeError("boom"), None, None]` is positionally coupled to the implementation's SQL-split logic. If the runner changes how it batches or splits statements, the failure index moves and the test passes for the wrong reason.

Suggested: assert on a specific SQL substring before raising, e.g.
```python
def _side_effect(sql):
    if "CREATE TABLE dq_app_settings" in sql:
        raise RuntimeError("boom")
cur.execute.side_effect = _side_effect
```
Now the test fails meaningfully when the migration changes, not because the offset drifted.

mwojtyczka (changes requested):

mwojtyczka (approved):
LGTM

PR #1205, opened 2026-05-28
Added data drift detection guide to profiling docs

### LINKED ISSUE #1101: [FEATURE]: Document data drift detection for profiling purposes

### Problem statement

Row anomaly detection contains functionality to detect data drift. This info is also useful, when running profiler to generate quality rule candidates. 

### Proposed Solution

Extend documentation on how to identify data drift, for the purpose of profiling.

### Additional Context

_No response_

Vsatyam013:
### Linked issues

Resolves #..

### Tests

- [x] manually tested
- [ ] added unit tests
- [ ] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

### Documentation and Demos

- [ ] added/updated demos
- [x] added/updated docs


mwojtyczka:
Thanks for adding this — useful guide for #1101. I pushed a commit to this branch with a few corrections so the section aligns with what the profiler actually emits and connects to the existing drift functionality:

**Metric names** — `summary_stats` doesn't include some of the referenced fields:
- `distinct_count` → the emitted key is **`count_distinct`**.
- `completeness` isn't a metric — it should be derived from `count_non_null / count` (the table exposes `count_null` / `count_non_null`).
- "top-frequency values" aren't produced; the profiler exposes the allowed-value set via the `is_in` check instead.
- `mean`, `stddev`, `min`, `max` are correct.

**Cross-link to existing drift detection** — issue #1101 specifically notes that row anomaly detection already has drift functionality. The commit adds a tip linking to the [Row Anomaly Detection](/docs/guide/row_anomaly_detection) guide, which ships an automated, scored drift detector (`compute_drift_score`: z-score / std-change vs. a trained baseline, flagged past a sigma threshold). The profiling section now frames itself as the lightweight, manual approach and points to that for the automated path.

**Code example** — added a small `profiler.profile(...)` baseline-vs-current comparison so the section is actionable.

Optional follow-up: `count_distinct` and `empty_count` are emitted but missing from the "Summary Statistics Reference" table above — worth adding there for consistency.

mwojtyczka (approved):
LGTM - I updated in places

PR #1223, opened 2026-06-05
DQX Studio: Replace apx framework with first-party build/dev scripts

berrybluecode:
The app previously relied on the unpublished ``apx`` package (github.com/databricks-solutions/apx) to drive its build and dev workflows. Drop the dependency and reimplement what we need ourselves:

- scripts/build_app.py: replaces ``apx build``. Generates the FastAPI OpenAPI schema, runs orval (now configured via orval.config.ts) to regenerate ui/lib/api.ts, runs ``vite build``, and produces the application wheel + ``app.yml`` + ``requirements.txt`` under ``.build/``. The wheel carries a build-tag local-version segment (e.g. ``.b20260530t012345``) baked into both the filename and the internal METADATA, so successive deploys at the same git commit always reinstall fresh code in the Databricks Apps persistent venv. This obsoletes ``scripts/_align_wheel_version.py`` (now removed), which used to repair apx's filename-only mangling after the fact.

- scripts/dev.py: replaces ``apx dev start``. Runs uvicorn with ``--reload`` on a fixed internal port and serves the SPA via Vite on :9001, proxying ``/api`` and ``/docs`` to the backend. SIGINT is forwarded to both children and either child crashing tears the other down, so no orphan processes are left behind.

- pyproject.toml: drop the ``apx`` dependency-group entry and the ``[tool.uv.sources].apx`` git source. Rename the metadata table from ``[tool.apx.metadata]`` to ``[tool.dqx_app.metadata]`` (same schema, new namespace) and document its consumers (build_app.py, vite.config.ts). Add ``basedpyright`` to the dev group.

- databricks.yml: rewrite ``artifacts.default.build`` to call ``scripts/build_app.py`` instead of ``apx build``. Add explicit Mode A (bundle-managed) vs Mode B (external/reuse) documentation on the ``sql_warehouse_id`` variable; default to empty so a target that forgets to set it fails fast.

- post_deploy_grants.sh: extend the warehouse grant path so external warehouses (Mode B) get ``CAN_USE`` granted to the app SP, job SP, and the ``users`` group via the warehouses permissions API — the bundle doesn't own the warehouse in that mode, so the bundle's ``permissions:`` block can't do it.

- Documentation refresh across DEPLOYMENT.md, DEVELOPMENT.md, README.md, AGENTS.md, the UI/backend CLAUDE.md files, and docs/dqx/docs/{installation,dev/contributing}.mdx to reflect the new build/dev entry points and warehouse modes.

No runtime behaviour change in the app itself: app/src/...backend/app.py is touched only to pick up the new metadata namespace.

### Linked issues

Resolves #..

### Tests

- [x] manually tested
- [x] added unit tests
- [ ] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

### Documentation and Demos

- [ ] added/updated demos
- [x] added/updated docs


app/scripts/post_deploy_grants.sh
```diff
@@ -158,9 +164,55 @@ run_sql "GRANT ALL PRIVILEGES ON SCHEMA \`$CATALOG\`.\`$SCHEMA\` TO \`$JOB_SP\`"
...
+  # app SP, the job SP, and the workspace ``users`` group so the OBO-
+  # token dry-run path also works.
+  # Note: the API expects the warehouse ID as a path segment and the
+  # principal as either ``user_name`` (for users), ``group_name`` (for
+  # workspace groups), or ``service_principal_name`` (for SPs, identified
+  # by their application ID UUID, NOT the numeric workspace ID). The UC
+  # ``account users`` group is account-scoped and rejected here — we use
+  # workspace ``users`` (which every workspace member is in by default)
+  # to match the original bundle-managed warehouse's permissions block.
+  $CLI api patch "/api/2.0/permissions/warehouses/$EXTERNAL_WH_ID" --json "$(cat <<EOF
```
mwojtyczka: **Hand-built JSON payload with shell interpolation.** This heredoc (and `run_sql` above, which does `--json "{\"statement\": \"$stmt\", ...}"` with `$CATALOG`/`$SCHEMA`/`$VOLUME` interpolated into both the JSON and the backtick-quoted SQL) will break or escape its quoting if any value contains a `"`, `\`, or backtick. These are operator-controlled deploy-time values so real risk is low, but this is exactly the safe-construction pattern AGENTS.md calls out, and it runs with deploy privileges. Prefer building the payload with `jq -n --arg ...` so values are always correctly escaped, and validate the identifier variables before interpolating them into SQL.

app/scripts/post_deploy_grants.sh
```diff
@@ -158,9 +164,55 @@ run_sql "GRANT ALL PRIVILEGES ON SCHEMA \`$CATALOG\`.\`$SCHEMA\` TO \`$JOB_SP\`"
...
+# so the user's source-table read permissions are enforced. The view
+# itself lives in $CATALOG.$TMP_SCHEMA, so end users need CREATE TABLE
+# and USE SCHEMA there in addition to USE CATALOG on the parent catalog.
+# Granting only USE CATALOG (the historical behavior) made every dry-run
+# fail with PERMISSION_DENIED on the CREATE OR REPLACE VIEW.
 run_sql "GRANT USE CATALOG ON CATALOG \`$CATALOG\` TO \`account users\`"
+run_sql "GRANT USE SCHEMA, CREATE TABLE ON SCHEMA \`$CATALOG\`.\`$TMP_SCHEMA\` TO \`account users\`"
 
 echo ""
 echo "==> Done. All grants applied."
```
mwojtyczka, line 309: **Grant failures are swallowed.** `run_sql` only prints `WARNING: $STATUS — ...` on a non-`SUCCEEDED` result and never propagates failure, and the warehouse PATCH branch likewise falls through with a `|| { echo WARNING ... }`. So a partial failure (e.g. one `GRANT` denied) still reaches this unconditional `==> Done. All grants applied.` and the script exits 0 under `set -e` — leaving the app silently half-provisioned. Consider tracking a failure count and `exit 1` at the end (or making this final message conditional) so CI/operators actually notice.

app/scripts/build_app.py
```diff
@@ -0,0 +1,250 @@
...
+  includes a build-tag local-version segment (e.g. ``+abc1234.b20260530T012345``)
+  so successive deploys at the same git commit reinstall fresh code in
+  Databricks Apps' persistent venv — the same problem the legacy
+  ``scripts/_align_wheel_version.py`` post-processor was solving on top
+  of apx's filename-mangling hack.
+
+Side effects on the source tree:
+
+* ``src/databricks_labs_dqx_app/_metadata.py`` is regenerated from
+  ``[tool.apx.metadata]`` in ``pyproject.toml``.
```
mwojtyczka: Stale namespace reference: this whole PR renames `apx`→`dqx_app`, but the docstring still says `_metadata.py` is regenerated from ``[tool.apx.metadata]``. The code correctly reads `[tool.dqx_app.metadata]` (`_read_app_metadata`). Update to `[tool.dqx_app.metadata]`.

Minor while here: the module docstring / `pyproject.toml` comment show the build tag as `b20260530T012345` (capital `T`), but `time.strftime("b%Y%m%dt%H%M%S")` emits lowercase `t`. PEP 440 lowercases anyway, but aligning the docs avoids confusion.

mwojtyczka (commented):
Strong refactor — dropping the unpublished/un-pinned `apx` git dependency is a clear supply-chain win, and the scripts are unusually well-documented (the process-group signal handling in `dev.py` and the Mode A/B warehouse docs in `databricks.yml` are excellent). A few notes inline, plus two non-anchored points:

- **PR claims "added unit tests" but the diff contains none.** `build_app.py` (`_read_app_metadata` missing-key path, `_write_metadata_module` formatting, newest-wheel/stale-wheel logic) and `dev.py`'s shutdown logic are testable. Please add a couple or correct the checklist.
- **Removed `apxDevProxyGuard` from `vite.config.ts`** looks correct given the new topology (Vite is now itself the public-facing proxy rather than sitting behind apx), so the origin guard no longer applies — just flagging it as a conscious drop of a (dev-only) control rather than an accident.

Nothing blocking on correctness; I'd want the grant-failure propagation and the JSON-construction hardening addressed in this deploy-privileged script before merge.

mwojtyczka (changes requested):
left some comments

app/scripts/post_deploy_grants.sh
berrybluecode: Added three validation helpers at the top of the script:

validate_uc_identifier — rejects anything outside [A-Za-z0-9_-] and requires a leading alphanumeric/underscore.
validate_uuid — strict 36-char UUID regex.
validate_warehouse_id — alphanumeric only (also stops URL path-segment smuggling).
Validate every parsed value before it touches a SQL statement or URL: CATALOG, SCHEMA, TMP_SCHEMA, VOLUME, APP_SP_ID, JOB_SP, WH_ID, and (inside the Mode-B branch) EXTERNAL_WH_ID. Failures abort the script with a clear error.

Replaced both hand-built JSON payloads with jq -n --arg:

run_sql now builds {warehouse_id, statement, wait_timeout} via jq -n, so any quote/backslash/newline in the statement is correctly escaped.
The warehouse permissions PATCH heredoc is gone — replaced with a jq -n invocation that takes $APP_SP_ID and $JOB_SP as --arg values.

app/scripts/post_deploy_grants.sh
berrybluecode: Good catch — fixed in the latest commit:

Added a top-level FAILURES counter that run_sql and the warehouse-PATCH branch both bump on a non-SUCCEEDED result (now logged as FAILED: on stderr, not WARNING:).
The warehouse PATCH branch snapshots PIPESTATUS into a regular array so a failure in either the API call or the jq formatter is detected and counted.
The final ==> Done. All grants applied. is now reached only when FAILURES == 0. Otherwise the script writes ==> FAILED: N grant(s) did not succeed — see warnings above. to stderr and exit 1s, so make app-deploy (and CI) actually fail.
Kept the apply-all-then-fail-at-end behaviour rather than abort-on-first because the GRANTs are independent and operators benefit from seeing every failure in one run, but happy to switch to fail-fast if you'd prefer.

Makefile
ghanse, line 1: The number of makefile commands is growing quite large, should we consider a `make help` that echoes some info about each command?

```
help:
	@echo "Available commands:"
	@echo "  dev          		        Install dependencies"
	@echo "  test         		        Run unit tests"
	@echo "  fmt          		        Format and lint code"
         ...
```

app/DEVELOPMENT.md
```diff
@@ -87,61 +93,73 @@ From the **project root**:
...
-uv run apx dev status        # check server status
-make app-stop-dev            # stop all servers
+make app-start-dev    # all logs stream to stdout (foreground)
+make app-stop-dev     # stop dev servers started in another shell
+```
+
+To background the dev loop and tail logs separately, redirect to a file:
+
+```bash
+nohup make app-start-dev > /tmp/dqx-dev.log 2>&1 &
```
mwojtyczka, line 137: we should add this to the contributing guide as well so that people know how to troubleshoot

docs/dqx/docs/dev/contributing.mdx
mwojtyczka, line 221: Can you please add the following, it usually trips the app:
DQX_CATALOG=dqx                             # optional, Unity Catalog catalog name
DQX_SCHEMA=dqx_studio                       # optional, schema inside the catalog

app/src/databricks_labs_dqx_app/ui/CLAUDE.md
```diff
@@ -22,7 +22,7 @@ ui/
 │       └── profile.tsx         # User profile
 ├── components/
 │   ├── ui/                     # shadcn/ui primitives (button, card, dialog, select, etc.)
-│   ├── apx/                    # App framework (Navbar, SidebarLayout, ThemeProvider, Logo)
+│   ├── apx/                    # Shell components (Navbar, SidebarLayout, ThemeProvider, Logo) — folder name is a historical artifact from the apx scaffolding, no runtime dependency on apx
```
mwojtyczka: Please remove. No point of explaining legacy structure. The apx folder should also be removed

app/src/databricks_labs_dqx_app/ui/lib/i18n/locales/de.json
```diff
@@ -0,0 +1,1214 @@
+{
```
mwojtyczka: Move to a saparate PR

mwojtyczka (approved):
LGTM

PR #1232, opened 2026-06-09
fix: allow special chars in catalog and schema

GewoonMaarten:
## Changes
Unfortunately our company included a `-` in the schema name, but DQX doesn't allow that. I've updated the regex:
image

I didn't find any unit tests for this. Let me know if you'd like me to add some.

### Linked issues

Resolves #..

### Tests

- [x] manually tested
- [ ] added unit tests
- [ ] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

### Documentation and Demos

- [ ] added/updated demos
- [ ] added/updated docs


src/databricks/labs/dqx/io.py
```diff
@@ -13,8 +13,8 @@
 logger = logging.getLogger(__name__)
 
 STORAGE_PATH_PATTERN = re.compile(r"^(/|s3:/|abfss:/|gs:/)")
-# catalog.schema.table or schema.table or database.table (backticks allow special chars like hyphens for table)
-TABLE_PATTERN = re.compile(r"^(?:[a-zA-Z0-9_]+\.)?[a-zA-Z0-9_]+\.(?:`[^`]+`|[a-zA-Z0-9_]+)$")
+# catalog.schema.table or schema.table or database.table (backticks allow special chars like hyphens for schema and table)
+TABLE_PATTERN = re.compile(r"^(?:[a-zA-Z0-9_]+\.)?(?:`[^`]+`|[a-zA-Z0-9_]+)\.(?:`[^`]+`|[a-zA-Z0-9_]+)$")
```
mwojtyczka: I would make all three namespace segments symmetric and make this change for a catalog as well:
```suggestion
TABLE_PATTERN = re.compile(r"^(?:(?:`[^`]+`|[a-zA-Z0-9_]+)\.)?(?:`[^`]+`|[a-zA-Z0-9_]+)\.(?:`[^`]+`|[a-zA-Z0-9_]+)$")
```

mwojtyczka (changes requested):

mwojtyczka:
@GewoonMaarten thanks for the PR. There is one suggestion, otherwise LGTM

src/databricks/labs/dqx/io.py
GewoonMaarten: Thanks! That makes sense, otherwise some else might run into that issue. I updated the description with the new test cases

mwojtyczka (approved):
LGTM - I added some unit tests to cover the change

PR #1236, opened 2026-06-10
feat: support metrics-only writes without output or quarantine tables

### LINKED ISSUE #1216: [FEATURE]: Allow writing summary metrics only

### Problem statement

The output or quarantine table must always be written. There are scenarios where user want to persist summary metrics only.

### Proposed Solution

Allow providing metrics config only and write summary metrics only (no output or quarantine), in methods that apple checks and save in tables

### Additional Context

Follow up PR for https://github.com/databrickslabs/dqx/pull/1183

mvanhorn:
## Changes

`save_results_in_table()` (and the corresponding workflow path) can now persist summary metrics without requiring an output or quarantine table. Previously a metrics-only configuration raised because the writer insisted on at least one results table; observability-focused pipelines that only want the metrics table had no supported path. Streaming and no-observer cases are validated with explicit errors, and batch observations are triggered before metrics are saved so the metrics table is populated correctly. Existing behavior with output or quarantine configs present is untouched; metrics-only is additive.

## Linked issues

Resolves #1216

## Tests

- [ ] manually tested
- [x] added unit tests
- [x] added integration tests

Integration tests cover the new metrics-only path (with and without an explicit metrics config) and the updated error message for the no-argument call. The suite needs a Databricks workspace and runs in CI.

## Documentation and Demos

- [x] updated documentation

`quality_checks_apply.mdx` and `summary_metrics.mdx` now describe the metrics-only mode.


src/databricks/labs/dqx/engine.py
```diff
@@ -1177,6 +1238,14 @@ def save_results_in_table(
             )
             metrics_config = run_config.metrics_config
 
+        if output_df is None and quarantine_df is None and metrics_config is None:
+            raise InvalidConfigError(
+                "At least one of 'output_config', 'quarantine_config' or 'metrics_config' must be provided"
+            )
+
+        if output_df is None and quarantine_df is None and metrics_config is not None and not self._engine.observer:
```
mwojtyczka: **This guard checks the wrong object.** The metrics-only save below uses the passed-in `observation`, not `self._engine.observer`:
```python
if observation is not None and metrics_config is not None and target_query is None:
    self.save_summary_metrics(observed_metrics=observation.get, metrics_config=metrics_config, ...)
```
So a caller with a valid `observation` (from an observer-enabled run) who invokes `save_results_in_table(observation=obs, metrics_config=cfg)` on a **different `DQEngine` instance that has no observer** would be wrongly rejected — even though the metrics are fully available from `obs`. The real precondition ("an observation must be passed") is already enforced by the earlier `output_df is None and quarantine_df is None and observation is None` guard. I think we should drop this guard.

src/databricks/labs/dqx/engine.py
```diff
@@ -898,15 +941,20 @@ def apply_checks_by_metadata_and_save_in_table(
...
+                metrics_only_df = checked_df
             target_streaming_query = output_streaming_query
 
         assert checks is not None  # guaranteed: either provided or loaded from checks_location above
         rule_set_fingerprint = (
             compute_rule_set_fingerprint_by_metadata(checks, custom_check_functions) if checks else None
         )
 
+        if metrics_config is not None and batch_observation is not None and metrics_only_df is not None:
+            metrics_only_df.count()
```
mwojtyczka: Same as the `count()` in `apply_checks_and_save_in_table` — please add the explanatory comment here too (it's necessary to populate the Observation; without it `batch_observation.get` blocks).

mwojtyczka (commented):
Solid, well-scoped feature — and the tricky part (forcing an action so the Spark `Observation` is populated before `Observation.get`, which would otherwise block in metrics-only mode) is handled correctly. Three notes inline, plus two test gaps:

- No test for the `save_results_in_table` observer-guard case flagged inline (an engine without an observer but given a valid `observation` + `metrics_config`).
- `apply_checks_and_save_in_tables_for_patterns` metrics-only: the `write_output`/`metrics_only` branch changed but I don't see a for-patterns metrics-only test — worth confirming the per-matched-table metrics table is named/written correctly when `output_table_suffix` is absent.

Not blocking beyond Finding 1.

src/databricks/labs/dqx/engine.py
```diff
@@ -773,13 +807,18 @@ def apply_checks_and_save_in_table(
...
+                output_streaming_query = save_dataframe_as_table(checked_df, output_config)
+            else:
+                metrics_only_df = checked_df
             target_streaming_query = output_streaming_query
 
         assert checks is not None  # guaranteed: either provided or loaded from checks_location above
         rule_set_fingerprint = compute_rule_set_fingerprint(checks) if checks else None
 
+        if metrics_config is not None and batch_observation is not None and metrics_only_df is not None:
+            metrics_only_df.count()
```
mwojtyczka: Please add a comment so a future reader doesn't "optimize" it away, e.g.:
```suggestion
            # Force an action so the Observation is populated; otherwise batch_observation.get below blocks
            metrics_only_df.count()
```

src/databricks/labs/dqx/engine.py
```diff
@@ -898,15 +942,20 @@ def apply_checks_by_metadata_and_save_in_table(
...
+                metrics_only_df = checked_df
             target_streaming_query = output_streaming_query
 
         assert checks is not None  # guaranteed: either provided or loaded from checks_location above
         rule_set_fingerprint = (
             compute_rule_set_fingerprint_by_metadata(checks, custom_check_functions) if checks else None
         )
 
+        if metrics_config is not None and batch_observation is not None and metrics_only_df is not None:
+            metrics_only_df.count()
```
```suggestion
            # Force an action so the Observation is populated; otherwise batch_observation.get below blocks
            metrics_only_df.count()
```

mwojtyczka (changes requested):
Thanks for your contribution. The PR generally looks good to me. I left small comment.

mvanhorn:
Thanks for applying the comment suggestions directly. The rest is addressed in 8674a01: dropped the observer guard in save_results_in_table per Finding 1 (save_summary_metrics now builds the metrics frame via the static builder, so a valid passed-in observation works on an engine without an observer), and added the two tests you flagged - observer-less save_results_in_table, and for-patterns metrics-only table naming when output_table_suffix is absent. One heads-up: the fmt CI failure was pylint R1260 (the new three-way guard pushed both apply_checks_*_save_in_table to McCabe 11), so I extracted it into a _populate_batch_observation helper - your comment text lives in its docstring. make test passes locally (1159 unit tests), pylint back at 10.00.

mwojtyczka:
@mvanhorn can you please sing the [Contributor License Agreement](https://cla-assistant.io/databrickslabs/dqx?pullRequest=1236)

mwojtyczka (approved):
LGTM

ghanse (approved):
Overall LGTM, should we add another issue to allow users to configure metrics-only jobs via the installation wizard?

mvanhorn:
Thanks for merging, @mwojtyczka. Metrics-only writes without output or quarantine tables make pure-observability runs much cheaper.

PR #1252, opened 2026-06-15
MCP Server for DQX

### LINKED ISSUE #1045: [FEATURE]: Implement MCP Server for DQX

### Problem statement

Currently, DQX offers powerful data quality checking capabilities, but its usage is tightly coupled to Spark environments. Integrating DQX into agentic systems requires custom implementation today.

### Proposed Solution

We want to make DQX available as a reusable, agent-agnostic MCP Server that can be easily integrated into multi-agent systems.

The MCP Server should be able to:
* Load and apply appropriate DQX checks for a given catalog, schema, or table
* Automatically determine the source of checks
* Fallback to profiling if no checks are found
* Run the checks as a Databricks job

### Additional Context

_No response_

souravg-db2:
## Changes

Adds an **MCP (Model Context Protocol) server for DQX**, exposing DQX's data-quality
capabilities as tools that any MCP-compatible AI agent (Claude, Genie Code, Cursor, Mosaic AI)
can discover and orchestrate. It runs as a **Databricks App** with on-behalf-of (OBO)
authentication, so all data access is governed by the calling user's Unity Catalog permissions.

### Architecture

- **Databricks App + FastMCP** server, exposed over Streamable HTTP. The server itself carries
  no PySpark/DQX dependency.
- **OBO governance:** for any data access the server creates a temporary **definer's-rights view**
  over the source table using the user's forwarded token (`X-Forwarded-Access-Token`), so the
  service principal reads data *as the user*, never directly.
- **Async job pattern:** long-running operations are submitted to a pre-deployed **runner notebook
  job** and return a `run_id` immediately; the client polls `get_run_result`. `get_run_status` does
  a single non-blocking poll (the client drives cadence) — this matches MCP's canonical
  long-running-tool model and avoids holding the HTTP connection / saturating the worker pool.
- **Genie Code compatibility:** `stateless_http` + `json_response` + CORS preflight scoped to
  Databricks domains.

### Tools

| Tool | Purpose |
|---|---|
| `get_workflow` | Recommended sequence of tool calls (call first) |
| `get_table_schema` | Table columns/types (direct, OBO) |
| `profile_table` | Profile data → summary stats + profiles |
| `generate_rules` | Generate checks from profiler output |
| `generate_rules_from_contract` | Generate checks from an ODCS data contract (deterministic) |
| `validate_checks` | Validate check definitions |
| `run_checks` | Execute checks → counts + per-rule summary + failing sample |
| `apply_checks_and_save_to_table` | Operationalized run: write valid/quarantine rows to Delta tables |
| `save_checks` / `load_checks` | Persist / retrieve a rule set (table, UC volume, or workspace file) |
| `list_available_checks` | Discover all built-in DQX check functions |
| `get_run_result` | Poll an async job for status/results |

### Temp-view lifecycle (stateless, restart/replica-safe)

- The runner job **drops its own input view in a `finally`** (it runs as the SP, which **owns the
  temp schema** — see setup), so cleanup happens in the guaranteed job execution regardless of
  whether/where the user polls.
- A **throttled TTL sweeper** (timestamped `v_<epoch>_<uuid>` view names) reaps any orphans whose
  job never ran. No per-request server state is kept, so app restarts / multiple replicas don't
  leak views or lose context.

### Security & governance

- Data-access boundary enforced by OBO definer's-rights views; the app SP owns only the scratch
  `tmp` schema (object lifecycle), not the underlying data.
- CORS restricted to Databricks workspace/app domains with credentials; SQL identifiers validated
  and backtick-quoted; log values sanitized (CWE-117); catalog name sourced from a Databricks secret.

### Deployment

- Databricks Asset Bundle (`databricks.yml`): runner job, one-time setup job (UC grants + schema
  ownership), and the app. `requirements.txt` (the App's runtime manifest) and `pyproject.toml` are
  kept in sync; the runner installs `databricks-labs-dqx[datacontract]`.
- New `make` targets and `fmt`/CI wiring for the sub-project.

### Linked issues

Resolves #1045

### Tests

- [x] added unit tests

Layered MCP-server test suite (62 passing, deterministic, no workspace needed for CI):
- **Handler unit tests** (`test_tools.py`)
- **Protocol tests** via FastMCP's in-memory `Client` over the real MCP protocol (`test_mcp_protocol.py`)
- **HTTP integration** over the real ASGI app — health, `initialize`/`tools-list`, and OBO token
  propagation (`test_app_http.py`); **CORS** policy (`test_cors.py`)
- **Agent-in-the-loop integration** (gated/skipped without a deployed server + workspace LLM
  endpoint): a tool-calling model is handed the tool schemas and an instruction, and we assert it
  discovers and invokes the right tools (`test_integration_agent.py`)

### Documentation and Demos

- [x] added/updated docs (`docs/dqx/docs/guide/dqx_mcp_server.mdx`)

---
This description was written by Isaac.


mcp-server/databricks.yml
```diff
@@ -0,0 +1,73 @@
...
+        - task_key: dqx_run
+          notebook_task:
+            notebook_path: "./notebooks/runner.py"
+          environment_key: default
+      environments:
+        - environment_key: default
+          spec:
+            environment_version: "2"
+            dependencies:
+              - databricks-labs-dqx
```
mwojtyczka: pls leave a comment that this should be pinned in produtction

mwojtyczka (commented):
## Review: MCP Server for DQX

Strong PR — the OBO + definer's-rights-view + SP-job architecture is the right pattern, and it neatly avoids needing a Jobs scope at the user auth level. Nice touches: the 4.5 MB `notebook.exit()` guard, pure-ASGI OBO middleware, idempotent setup grants, and a fixed `OPERATIONS` dispatch (no arbitrary code exec).

A few things to address before merge — inline comments below. Two summary-level points:

**Rebase on `main` (stale base).** The root `pyproject.toml` mypy-comment change reintroduces a reference to `apx dev check`, but PR #1223 (already merged) removed `apx` and replaced it with first-party scripts (`bun run tsc -b` + `basedpyright`). The branch is behind `main`; please rebase so this drift is resolved.

**Add an integration test with Genie (see inline on the tests).** `app.py` is explicitly tuned for "Genie Code compatibility" (`stateless_http`, `json_response`, CORS preflight) — but nothing exercises that path. An end-to-end test that drives the MCP server through Databricks Genie (Genie Code as the MCP client) over the deployed app would protect exactly the behaviour these settings exist for.

Priorities: the rebase are blockers; CORS and the in-memory run-state are important follow-ups.

mcp-server/server/utils.py
```diff
@@ -0,0 +1,394 @@
...
+    if len(parts) != 3:
+        raise ValueError(f"Table name '{table_name}' must be fully qualified (catalog.schema.table)")
+
+    view_id = uuid.uuid4().hex[:12]
+    view_fqn = f"{catalog}.{schema}.v_{view_id}"
+
+    logger.info(f"Creating temp view {view_fqn} over {table_name}")
+    execute_sql(
+        ws,
+        f"CREATE VIEW {view_fqn} AS SELECT * FROM {table_name}",
```
mwojtyczka: **SQL identifier injection (must fix — AGENTS.md security policy).** `table_name` is interpolated raw into `CREATE VIEW ... AS SELECT * FROM {table_name}`. The `len(parts) != 3` check doesn't prevent injection — e.g. `cat.sch.t WHERE 1=1 UNION SELECT ...` still splits into 3 parts. It runs under the user's OBO token (so no privilege escalation), but (a) the policy requires validating user/templated SQL via `is_sql_query_safe()`, (b) the caller is an LLM agent (prompt-injection target), and (c) `setup.py` in this same PR *does* backtick-quote identifiers — inconsistent. Validate each part against `^[A-Za-z0-9_]+$` and/or backtick-quote with escaping, and add a negative test (`test_create_temp_view_validates_table_name` only checks part count).

mcp-server/server/tools.py
```diff
@@ -0,0 +1,295 @@
...
+            - 'table_name': the input table name
+            - 'columns': list of {name, type, comment} for each column
+        """
+        logger.info(f"Getting schema for table: {table_name}")
+        obo_ws = utils.get_obo_client()
+        warehouse_id = utils.get_warehouse_id(obo_ws)
+
+        rows = utils.execute_sql(
+            obo_ws,
+            f"DESCRIBE TABLE {table_name}",
```
mwojtyczka: Same SQL-injection concern as `utils.create_temp_view`: `table_name` is interpolated raw into `DESCRIBE TABLE {table_name}`. Validate/quote the identifier before interpolation.

mcp-server/server/app.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+combined_app.routes.insert(0, Route("/", health_check))
+
+# Add OBO middleware (pure ASGI — no BaseHTTPMiddleware to avoid streaming timeouts)
+combined_app.add_middleware(OBOAuthMiddleware)
+
+# Add CORS middleware — Genie Code sends OPTIONS preflight from workspace origin
+combined_app.add_middleware(
+    CORSMiddleware,
```
mwojtyczka: `allow_origins=["*"]` together with `allow_credentials=True` is rejected by browsers and is a security smell. With OBO via header (not cookies) the blast radius is limited, but please scope `allow_origins` to the workspace origin and drop `allow_credentials` unless it's actually required.

mcp-server/server/utils.py
```diff
@@ -0,0 +1,394 @@
...
+    try:
+        execute_sql(ws, f"DROP VIEW IF EXISTS {view_fqn}", warehouse_id=warehouse_id)
+    except Exception:
+        logger.warning(f"Failed to drop temp view {view_fqn}", exc_info=True)
+
+
+# ── Jobs API — async submit + poll ───────────────────────────────────
+
+# Track pending runs: run_id → metadata (view to clean up, table name, etc.)
+_pending_runs: dict[int, dict[str, Any]] = {}
```
mwojtyczka: `_pending_runs` is process-local in-memory state. On a Databricks App that can restart or run multiple replicas, this means: (1) **orphaned temp views** leak in the shared `tmp` schema if the user never polls or the app restarts, and (2) a `get_run_result` landing on a different replica won't find the metadata (no view cleanup, can't re-attach `table_name`). Consider TTL-based view cleanup (the view name has a UUID — adding a timestamp would let a sweeper drop stale views) or reconstructing cleanup from job tags.

mcp-server/server/tools.py
```diff
@@ -0,0 +1,295 @@
...
+        """Retrieve the schema and basic metadata for a Databricks table.
+
+        Args:
+            table_name: Fully qualified table name (e.g. 'catalog.schema.table').
+
+        Returns a dict with:
+            - 'table_name': the input table name
+            - 'columns': list of {name, type, comment} for each column
+        """
+        logger.info(f"Getting schema for table: {table_name}")
```
mwojtyczka: Log injection (CWE-117): `table_name` is untrusted and interpolated into the log message via f-string. Strip newlines/control chars or use structured logging fields — AGENTS.md calls this out explicitly. (Applies to the other `logger.info(f"... {table_name}")` / `{view_fqn}` lines too.)

mcp-server/server/main.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def main():
+    # Add parent dir to path so 'server' package is importable
+    parent = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+    if parent not in sys.path:
+        sys.path.insert(0, parent)
+    print(f"sys.path[0]={parent}", flush=True)
+    print(f"cwd={os.getcwd()}", flush=True)
+    print(f"files in parent: {os.listdir(parent)}", flush=True)
```
mwojtyczka: These `print(os.listdir(...))` debug lines run on every boot and will spam production app logs. Drop them or move to `logger.debug`.

mcp-server/pyproject.toml
```diff
@@ -0,0 +1,28 @@
...
+description = "DQX MCP Server - AI agent tools for data quality checks"
+requires-python = ">=3.11"
+dependencies = [
+    "fastapi>=0.115.12",
+    "mcp[cli]>=1.14.0",
+    "uvicorn>=0.34.2",
+    "databricks-sdk>=0.60.0",
+    "pydantic>=2",
+    "fastmcp>=2.12.5",
+    "databricks-labs-dqx",
```
mwojtyczka: `databricks-labs-dqx` (and `fastapi`) appear unused by the server itself — only the runner notebook imports DQX, and the server imports `fastmcp`/`starlette`/`uvicorn`. 

The real concern is **drift** between the two files: `requirements.txt` is the runtime manifest the App installs, while `pyproject.toml`/`uv.lock` drive local dev and packaging. A runtime dep added in one but not the other will work in one place and fail in the other. Consider generating `requirements.txt` from the lockfile (e.g. `uv export`) as a single source of truth, and pin versions consistently across both. And drop `databricks-labs-dqx`/`fastapi` here if they're genuinely unused by the server.

mcp-server/databricks.yml
```diff
@@ -0,0 +1,73 @@
...
+            permission: "CAN_MANAGE_RUN"
+        - name: dqx-catalog-secret
+          secret:
+            scope: dqx-config
+            key: catalog_name
+            permission: "READ"
+
+      permissions:
+        - level: CAN_USE
+          group_name: users
```
mwojtyczka: `group_name: users` is hardcoded here, but the setup job uses `${var.users_group}` (default `account users`). These should likely reference the same variable, otherwise the app is granted to a different group than the one that gets the UC grants.

mcp-server/tests/test_tools.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def test_list_available_checks(self):
+        tools = _register_tools()
+
+        with patch("server.tools.utils.submit_notebook_job", return_value={"checks": [], "count": 0}) as mock_job:
+            tools["list_available_checks"]()
+
+        mock_job.assert_called_once_with("list_available_checks", {})
+
+    def test_get_workflow_returns_steps(self):
```
mwojtyczka, line 229: **Add an integration test with Genie.** The server is explicitly built for "Genie Code compatibility" (`stateless_http`/`json_response`/CORS preflight in `app.py`), but no test exercises that end-to-end path. An integration test that registers the deployed MCP app with Databricks Genie and drives the documented workflow (get_table_schema → profile_table → generate_rules → run_checks, polling get_run_result) would protect the OBO header propagation and the async submit/poll contract that these settings exist for. Unit coverage of the server layer here is otherwise good.

mcp-server/server/tools.py
```diff
@@ -0,0 +1,295 @@
...
+        )
+
+        return {
+            "status": "submitted",
+            "run_id": run_id,
+            "message": "Profiling job submitted. Call get_run_result with this run_id to get results.",
+        }
+
+    @mcp_server.tool
+    def generate_rules(profiles: list[dict], criticality: str = "error"):
```
mwojtyczka, line 127: it would be good to provide option to generate rules based on user input as well, both are supported today in DQX Core. I would leave one method and make user input optional

mwojtyczka (changes requested):
Going in the right direction, left some comments to address

mcp-server/server/utils.py
souravg-db2: Made fix. please check

mcp-server/server/tools.py
souravg-db2: Made fix. please check

mcp-server/server/main.py
souravg-db2: Made fix. please check

mcp-server/server/tools.py
vb-dbrks: good shout. Let's think about this @souravg-db2 - there are two approaches for this - 
1. we already have a purpose built DQX Agent which has DQX specific instructions and validation we wire that as a tool? 
2. another is the calling Agent itself analyses the user need and only leverages granular tools that we provide via this MCP. I just think perhaps the output would be wildly different on basis of the quality of the calling agent.

mcp-server/tests/test_tools.py
vb-dbrks: Will evaluate feasibility of this, I doubt if we have any APIs yet for Genie Code. The UCode integration is using just API Gateway https://github.com/databricks/ucode

mcp-server/pyproject.toml
vb-dbrks: will look into this.

mcp-server/databricks.yml
souravg-db2: Fixed. please check

mcp-server/server/tools.py
souravg-db2: Fixed. please check

mcp-server/server/app.py
souravg-db2: Added change to keep allow origins to db domains

mcp-server/server/utils.py
mwojtyczka: Verified on the latest head — `create_temp_view` now routes through `validate_and_quote_table_name()` / `_validate_sql_identifier()` (`^[A-Za-z0-9_]+$` + backtick quoting), so the 3-part-split bypass is closed. Resolving.

mcp-server/server/tools.py
mwojtyczka: Verified — `get_table_schema` now calls `validate_and_quote_table_name()` before `DESCRIBE TABLE`. Resolving.

mcp-server/server/app.py
mwojtyczka: Verified — now uses `allow_origin_regex` scoped to Databricks domains instead of `allow_origins=["*"]`, which keeps `allow_credentials=True` spec-compliant (Starlette reflects the matched origin via fullmatch). Resolving.

mcp-server/server/tools.py
mwojtyczka: Verified — `sanitize_for_log()` strips CR/LF and is applied at every `table_name`/`view_fqn` log site. Resolving.

mcp-server/server/main.py
mwojtyczka: Verified — the `print(os.listdir(...))` lines are now `logger.debug(...)`. Resolving.

mcp-server/databricks.yml
mwojtyczka: Verified — `group_name: ${var.users_group}` now matches the setup job's grant group. Resolving.

mcp-server/server/utils.py
```diff
@@ -0,0 +1,466 @@
...
+    while elapsed < max_wait:
+        run = ws.jobs.get_run(run_id)
+        life_cycle = run.state.life_cycle_state.value if run.state and run.state.life_cycle_state else "UNKNOWN"
+
+        if life_cycle not in ("PENDING", "RUNNING", "QUEUED", "BLOCKED"):
+            break
+
+        elapsed += poll_interval
+        if elapsed < max_wait:
+            time.sleep(poll_interval)
```
mwojtyczka: `get_run_status` blocks here for up to 90s (`time.sleep(10)` × 9) before returning. The MCP tools are sync `def`, so FastMCP runs each call in an anyio worker thread — the blocking sleep won't stall the event loop, but two costs remain: (1) it holds the HTTP connection open for up to 90s per `get_run_result` poll, which can trip client/proxy timeouts (Genie Code / Databricks Apps front-end), and (2) it occupies one of the limited anyio threadpool tokens for the whole wait, so a burst of concurrent polls can saturate the pool and stall unrelated tool calls. Since the tool already returns `status: running` for the caller to retry, consider a much shorter internal wait (a single short poll) and let the client drive the cadence.

mcp-server/server/utils.py
```diff
@@ -0,0 +1,466 @@
...
+            _user_token_var.set((host, user_token))
+        else:
+            _user_token_var.set(None)
+
+        _user_email_var.set(user_email)
+
+        await self.app(scope, receive, send)
+
+
+def get_workspace_client():
```
mwojtyczka: `get_workspace_client` is never called anywhere in the server package — `tools.py` uses `get_obo_client()` / `_get_sp_client()` instead. Same story for `make_json_safe` at L453 (only referenced by itself; the real JSON coercion happens in the runner notebook's own `_make_json_safe`). Both look like dead code that just widens the surface area — worth removing. (`get_workspace_client` also has no return-type annotation.)

mcp-server/pyproject.toml
mwojtyczka: still not addressed

mwojtyczka (changes requested):
Closed some previous comments, there are still open one and a few new.

mcp-server/pyproject.toml
vb-dbrks: Done. Trimmed pyproject.toml to the actual server runtime deps (fastmcp, starlette, uvicorn, databricks-sdk) and made requirements.txt match exactly...  mcp/pydantic come in transitively via fastmcp, and starlette (imported directly) was previously only in requirements.txt. Dropped fastapi and databricks-labs-dqx (is only used by the runner notebook, installed via the job env spec in databricks.yml). Regenerated uv.lock via make lock-mcp-dependencies.

mcp-server/server/utils.py
vb-dbrks: Good catch removed both.

mcp-server/databricks.yml
vb-dbrks: Added an inline comment on the dependency noting it's unpinned for dev and should be pinned to an explicit version in production so runner environments stay reproducible. (The line now also carries the [datacontract] extra for the new contract-based rule generation.)

mcp-server/server/utils.py
vb-dbrks: So this is an interesting one, The App/MCP SP is not the schema owner, and has no MANAGE. The schema is owned by whoever ran databricks bundle run dqx_setup (the comment shows a human admin via --profile). Each temp view is owned by the OBO user who created it. A sweeper running as the SP cannot drop user-created views. UC will reject the DROP. our best option is to give the app SP manage on the tmp schema.

vb-dbrks: FYI @souravg-db2

mcp-server/server/tools.py
vb-dbrks: Hi @mwojtyczka  @souravg-db2  - I tried this out I think I am leaning on leaving the MCP granular and deterministic with composable primitives. seems to be what Anthropic also advises https://www.anthropic.com/engineering/writing-tools-for-agents

mcp-server/server/utils.py
souravg-db2: I will make this change.

mcp-server/server/utils.py
mwojtyczka: Verified on the latest head — `get_run_status` now does a single non-blocking poll: no `time.sleep`, returns `status: running` immediately for the caller to retry (docstring spells out the HTTP-connection / anyio-threadpool rationale). Resolving.

mcp-server/server/utils.py
mwojtyczka: Verified — the in-memory `_pending_runs` state is gone and cleanup is now stateless/replica-safe: the runner drops its own view in a `finally` (runner.py), a throttled TTL sweeper reaps orphans by `v_<epoch>_<uuid>` age (utils.py `sweep_stale_views`), and `setup.py` transfers `tmp`-schema ownership to the app SP (`ALTER SCHEMA ... OWNER TO`) so the SP can drop user-created views — which addresses the MANAGE concern raised here. Resolving.

mcp-server/notebooks/runner.py
```diff
@@ -0,0 +1,403 @@
...
+        result = {"error": f"Unknown operation: {operation}. Valid: {list(OPERATIONS.keys())}"}
+    else:
+        result = OPERATIONS[operation](params)
+        # Echo the source table name so the client knows which table the result is for.
+        # Done here (instead of re-attaching in the server) so the server needs no per-run state.
+        if isinstance(result, dict) and params.get("table_name") and "table_name" not in result:
+            result["table_name"] = params["table_name"]
+except Exception as e:
+    logger.error(f"Operation '{operation}' failed: {e}", exc_info=True)
+    result = {"error": f"{type(e).__name__}: {str(e)}"}
```
mwojtyczka: **Operation failures are reported as `completed`, not `failed`.** When an operation raises, this `except` sets `result = {"error": ...}` and the notebook still exits normally, so the job's `result_state` is `SUCCESS`. `get_run_status` then returns `{"status": "completed", "result": {"error": ...}}` — so an agent driving the workflow treats a failed op as success and proceeds (e.g. feeds the error dict into `generate_rules`). The documented `status: failed` + `error` contract is never hit for in-operation errors. Consider `raise` (or re-raise after logging) so the run fails, or surface a distinct failure status the client can branch on.

mcp-server/server/tools.py
```diff
@@ -0,0 +1,501 @@
...
+        rows = utils.execute_sql(
+            obo_ws,
+            f"DESCRIBE TABLE {safe_table}",
+            warehouse_id=warehouse_id,
+        )
+
+        columns = [
+            {"name": row["col_name"], "type": row["data_type"], "comment": row.get("comment", "")}
+            for row in rows
+            if row.get("col_name") and not row["col_name"].startswith("#")
```
mwojtyczka: **Partitioned tables yield duplicate columns.** This filter only drops rows whose `col_name` starts with `#`. For a partitioned table, `DESCRIBE TABLE` re-lists each partition column under a `# Partition Information` section *without* a `#` prefix, so every partition column ends up in `columns` twice. Either stop reading at the first `# Partition Information`/blank separator row, or de-duplicate by `col_name`.

mcp-server/server/app.py
```diff
@@ -0,0 +1,60 @@
...
+
+# Add CORS middleware — Genie Code sends OPTIONS preflight from workspace origin.
+# Use allow_origin_regex (not allow_origins=["*"]) so that allow_credentials=True stays
+# spec-compliant: Starlette reflects the matched origin instead of the forbidden "*".
+combined_app.add_middleware(
+    CORSMiddleware,
+    allow_origin_regex=CORS_ALLOWED_ORIGIN_REGEX,
+    allow_methods=["*"],
+    allow_headers=["*"],
+    allow_credentials=True,
```
mwojtyczka: **Credentialed CORS over a multi-tenant wildcard.** The earlier `allow_origins=["*"]` fix is good, but the regex matches *any* `*.databricksapps.com` host, and that domain is multi-tenant — every customer's app lives under it. With `allow_credentials=True` + `allow_methods=["*"]` + `allow_headers=["*"]`, any other tenant's app page is an allowed credentialed origin and could replay JSON-RPC tool calls if the Apps proxy forwards the OBO token cross-origin.

Is it possible to narrow the regex to the specific workspace/app host(s) rather than the whole `databricksapps.com` domain? (and confirm whether the proxy actually attaches the token on cross-origin requests).

mcp-server/notebooks/runner.py
```diff
@@ -0,0 +1,403 @@
...
+    checks = params["checks"]
+    output_table = params["output_table"]
+    quarantine_table = params.get("quarantine_table")
+    mode = params.get("mode", "append")
+
+    ws = WorkspaceClient()
+    engine = DQEngine(workspace_client=ws, spark=spark)
+
+    input_config = InputConfig(location=view_name)
+    output_config = OutputConfig(location=output_table, mode=mode)
```
mwojtyczka: **Write-target identifiers aren't validated.** `output_table`/`quarantine_table` go straight into `OutputConfig(location=...)` and `spark.table(...)` without `validate_and_quote_table_name`, unlike the source `table_name` read path. Injection risk is low (`spark.table` parses an identifier rather than raw SQL), but routing the write targets through the same validator keeps identifier handling consistent and guards against any future string interpolation.

mwojtyczka: removed, low risk

mcp-server/server/tools.py
mwojtyczka: sounds good, we can make a follow up for this. I still think this would be very important way on how people can interact with the agent, make it more interactive

mcp-server/server/tools.py
souravg-db2: Fixed . please check

mcp-server/notebooks/runner.py
souravg-db2: Fixed. please check

mwojtyczka: Fixed in 38ab4ff — the runner now re-raises on failure (the `finally` still drops the temp view), so the job's `result_state` becomes `FAILED` and `get_run_status` reports `failed` instead of `completed`. Resolving.

mcp-server/server/tools.py
mwojtyczka: Fixed in 38ab4ff — `get_table_schema` now breaks at the first blank/`#`-prefixed row, with a `test_partition_columns_not_duplicated` regression test. Resolving.

mcp-server/runner/src/dqx_mcp_runner/runner.py
```diff
@@ -0,0 +1,462 @@
...
+    from databricks.labs.dqx.engine import DQEngine
+    from databricks.labs.dqx.checks_storage import ChecksStorageHandlerFactory
+
+    location = params["location"]
+    run_config_name = params.get("run_config_name", "default")
+
+    ws = WorkspaceClient()
+    engine = DQEngine(workspace_client=ws, spark=spark)
+    _handler, config = ChecksStorageHandlerFactory(ws, spark).create_for_location(location, run_config_name)
+    checks = engine.load_checks(config)
```
mwojtyczka, line 142: **OBO governance gap (read): `load_checks` runs as the app SP, not the calling user.** `ws = WorkspaceClient()` here is the service principal, so `engine.load_checks(config)` reads any user-supplied 3-part table name with the SP's permissions, not the caller's. `generate_rules_from_contract` (~line 211) has the same shape for the contract file path. The OBO definer's-rights-view pattern only protects the *source-table* read; these paths don't go through it — so the "all data access is governed by the calling user's UC permissions" claim doesn't hold here. At minimum correct that claim; ideally enforce the caller's permissions on these paths.

mcp-server/notebooks/runner.py
```diff
@@ -0,0 +1,462 @@
...
+    from databricks.labs.dqx.engine import DQEngine
+    from databricks.labs.dqx.config import InputConfig, OutputConfig
+
+    view_name = params["view_name"]
+    checks = params["checks"]
+    output_table = params["output_table"]
+    quarantine_table = params.get("quarantine_table")
+    mode = params.get("mode", "append")
+
+    ws = WorkspaceClient()
```
mwojtyczka: **OBO governance gap (write): `apply_checks_and_save_to_table` (and `save_checks`) write as the app SP to caller-supplied destinations.** `output_table`/`quarantine_table` are written with the SP's permissions and no check that the *caller* can write them, so a user who reaches this tool can have the SP overwrite a table they couldn't modify directly (`mode=overwrite`). Same root cause as the read path — governance here is the SP's grants, not the caller's.

mcp-server/notebooks/setup.py
```diff
@@ -0,0 +1,130 @@
...
+# Create schema if it doesn't exist
+spark.sql(f"CREATE SCHEMA IF NOT EXISTS `{catalog_name}`.`{schema_name}`")
+logger.info(f"Schema `{catalog_name}`.`{schema_name}` ready")
+
+# COMMAND ----------
+
+# All grants — single source of truth
+grants = [
+    # Catalog-level
+    f"GRANT USE CATALOG ON CATALOG `{catalog_name}` TO `{users_group}`",
```
mwojtyczka, line 194: **`users_group` (and `app_name`/`schema_name`) are interpolated into `GRANT … TO \`{users_group}\`` without `_validate_identifier`** — only `catalog_name` is validated (line 70). A value containing a backtick breaks out of the quoted identifier in a statement run by the setup principal. One-time operator-run, so reach is limited, but the validator already exists — apply it to these too.

mcp-server/databricks.yml
```diff
@@ -0,0 +1,103 @@
...
+        - environment_key: default
+          spec:
+            environment_version: "2"
+            dependencies:
+              # Install the DQX library wheel built from THIS repo (see artifacts.dqx_library_wheel),
+              # not a published PyPI release, so the runner exercises in-repo (possibly unreleased)
+              # source and never drifts from what the PR actually changes. The [datacontract] extra's
+              # dependency is listed explicitly because pip extras can't be applied to a wheel path.
+              - ./.build/databricks_labs_dqx-*.whl
+              - datacontract-cli>=0.11.1,<1.0
```
mwojtyczka: **Runner env installs the bare wheel + `datacontract-cli` but not the `[llm]` extra (dspy)**, yet `generate_rules_from_contract` exposes `process_text_rules: bool` and passes it through to DQX. Calling it with `process_text_rules=True` fails with an `ImportError` on dspy at runtime. Either add the `[llm]` extra here, or have the tool reject `process_text_rules=True` with a clear message.

mcp-server/notebooks/runner.py
```diff
@@ -0,0 +1,462 @@
...
+    checks = params["checks"]
+    sample_size = params.get("sample_size", 50)
+
+    ws = WorkspaceClient()
+    engine = DQEngine(workspace_client=ws, spark=spark)
+
+    df = spark.table(view_name)
+    valid_df, invalid_df = engine.apply_checks_by_metadata_and_split(df, checks)
+
+    total_rows = df.count()
```
mwojtyczka: **Minor (perf): `run_checks` scans the source ~6×.** `df.count()`, `valid_df.count()`, `invalid_df.count()`, `invalid_df.limit().collect()`, and `_compute_rule_summary`'s explode+groupBy on `_errors` and again on `_warnings` each recompute the lazy split over the view. Caching `invalid_df` (and `df`) before the count/collect/summary actions cuts this to ~1–2 scans — it runs on every invocation, so it matters on large tables.

mcp-server/notebooks/runner.py
souravg-db2: Do we need apply_checks_and_Save_to_table tool ?

mwojtyczka: yes, the server should provide persistence

mcp-server/server/app.py
vb-dbrks: Fixed. Replaced the wildcard regex with an exact allow-list built from DATABRICKS_HOST (+ optional DQX_MCP_EXTRA_CORS_ORIGINS), so only this workspace's origin is a credentialed origin — not the shared domain. Tests assert other tenants' app hosts are rejected.

mcp-server/runner/src/dqx_mcp_runner/runner.py
vb-dbrks: The issue has been resolved and is now in effect. The `load_checks` table reads from a definer's-rights view with the caller's permissions. Additionally, file reads (specifically for `load_checks` and `generate_rules_from_contract`) are now pre-checked to ensure they align with the caller's permissions before the job executes. Furthermore, the runner job now runs under a dedicated least-privilege stored procedure rather than the deployer.

mcp-server/notebooks/runner.py
vb-dbrks: done.  Before submitting, we verify via the caller's OBO client that they may write the destination (own an existing table, or can create in the target schema); the SP writes only after that passes. Kept the tool (persistence) as you confirmed.

mcp-server/databricks.yml
vb-dbrks: Fixed. Made generate_rules_from_contract deterministic-only... dropped the process_text_rules parameter (the runner has no [llm] extra, matching the Studio task runner). Documented how to enable it as a follow-up.

mcp-server/notebooks/setup.py
vb-dbrks: Validation has been completed for the parameters that reach SQL: `users_group` and `schema_name` are now checked with `_validate_identifier` before interpolation.

I didn't include `app_name` in this validation because it’s only used in `ws.apps.get(app_name)` and can contain a hyphen (e.g., "mcp-dqx"), which would be rejected by the validator's character set. If you prefer, I can add a separate hyphen-allowing check, although there's currently no SQL endpoint for `app_name`.

`Schema_name` is set to the constant "tmp" and is validated as a precaution since it's used in schema DDL and grants.

mcp-server/runner/src/dqx_mcp_runner/runner.py
mwojtyczka: Verified on the latest head (f79922d) — `load_checks` routes table reads through a definer's-rights OBO temp view, and `verify_obo_read_access` pre-checks volume/workspace-file paths as the caller; `generate_rules_from_contract` verifies the contract file under the caller's OBO client before submission. Resolving.

mcp-server/notebooks/runner.py
mwojtyczka: Verified — both `save_checks` and `apply_checks_and_save_to_table` now call `verify_obo_write_access` before the SP writes: owner check on existing tables, a CREATE-probe for new ones, and a file-write probe for volume/workspace destinations. The SP no longer writes a destination the caller couldn't write themselves. Resolving.

mcp-server/notebooks/setup.py
mwojtyczka: Verified — `_validate_identifier` now guards `users_group` and `schema_name` before they are interpolated into the GRANT/DDL statements. `app_name` is fine to skip: it's only passed to `ws.apps.get()` and never reaches SQL (and legitimately contains a hyphen). Resolving.

mcp-server/databricks.yml
mwojtyczka: Verified — the server tool no longer exposes `process_text_rules` (deterministic-only), and the runner guards `process_text_rules=True` with a clear `ValueError` instead of an opaque dspy ImportError. Resolving.

docs/dqx/docs/guide/dqx_mcp_server.mdx
```diff
@@ -0,0 +1,480 @@
...
+### Runner service principal (one-time)
+
+The runner job runs as a dedicated, least-privilege **workspace service principal** — *not* the app's auto-created SP (which is app-scoped and cannot be a job's `run_as`) and *not* the deploying user. Create it once, then pass its application id as `runner_service_principal_id` at deploy:
+
+1. Create a workspace service principal (UI: **Settings → Identity and access → Service principals → Add**, or the SCIM API). Note its **Application ID**.
+2. Grant your deploying identity the **User** role on it (Settings → that SP → **Permissions**), or `bundle deploy` fails with *"User is not authorized to use this service principal"*. (Workspace admins usually have this implicitly, but confirm.)
+3. Give the SP the **workspace-access** entitlement so it can run serverless jobs.
+4. The setup job grants the SP the UC privileges it needs on the temp schema + results volume automatically. For `save_checks` / `apply_checks_and_save_to_table`, the SP also needs **write access wherever you point outputs** — keep outputs in a schema the SP can write, or grant it there (see [Write access](#write-access)).
+
+## Deploy
```
mwojtyczka: I think this should be moved under installation page and linked here

mcp-server/tests/test_tools.py
mwojtyczka: sounds good, there's no public API to make Genie Code itself drive the server as a test client. Closing given the e2e coverage now exists (MCP-client-driven against the deployed app), and that Genie-Code-agent-driven testing isn't automatable

mcp-server/server/utils.py
```diff
@@ -0,0 +1,874 @@
...
+            raise PermissionError(f"Could not verify your write access to {sanitize_for_log(table_name)}: {e}") from e
+
+    if rows is not None:
+        owner = None
+        for row in rows:
+            if (row.get("col_name") or "").strip() == "Owner":
+                owner = (row.get("data_type") or "").strip() or None
+                break
+        caller = get_user_email()
+        if caller and owner and owner.lower() == caller.lower():
```
mwojtyczka: **Write pre-check rejects legitimate non-owner writers.** A write is allowed only when `owner.lower() == caller.lower()`, so a caller with `MODIFY` via a group grant, a group-owned table (owner is never an email), or a table owned by an SP/display-name is refused even though they can write it — `save_checks` / `apply_checks_and_save_to_table` return an error for a destination the caller is actually authorized for. The docstring promises rejection only when *you* couldn't write it, but this also rejects the strictly-weaker MODIFY-without-ownership case. Consider checking effective privileges (`information_schema` / `SHOW GRANTS`) instead of owner-equality.

mcp-server/notebooks/setup.py
```diff
@@ -0,0 +1,234 @@
...
+    f"GRANT USE SCHEMA ON SCHEMA `{catalog_name}`.`{schema_name}` TO `{users_group}`",
+    f"GRANT CREATE TABLE ON SCHEMA `{catalog_name}`.`{schema_name}` TO `{users_group}`",
+    # Schema-level: app SP reads through views and drops stale views (the in-app sweeper). MANAGE
+    # (not ownership) is enough to DROP views created by any user, so the deployer keeps ownership.
+    f"GRANT USE SCHEMA ON SCHEMA `{catalog_name}`.`{schema_name}` TO `{sp_principal}`",
+    f"GRANT SELECT ON SCHEMA `{catalog_name}`.`{schema_name}` TO `{sp_principal}`",
+    f"GRANT MANAGE ON SCHEMA `{catalog_name}`.`{schema_name}` TO `{sp_principal}`",
+]
+
+if runner_sp_configured:
```
mwojtyczka, line 206: **Runner SP isn't granted on caller destinations, so passing `verify_obo_write_access` still fails at runtime.** The pre-check validates the *caller's* permissions, but the runner SP performs the write and setup grants it only the `tmp` schema/catalog + results volume. A user who legitimately owns `cat.sch.out` passes every app-side check, then the job fails `PERMISSION_DENIED` because the SP has no MODIFY there. The note at L26-28 acknowledges destinations must be granted per deployment, but nothing enforces it — so the pre-check gives false assurance. Either write as the caller (OBO) or model the SP's grants in the check.

mcp-server/server/utils.py
```diff
@@ -0,0 +1,874 @@
...
+    except Exception as e:
+        text = str(e)
+        if _looks_like_permission_error(text):
+            # The table EXISTS but the caller can't access it — never let the SP write it for them.
+            raise PermissionError(
+                f"You do not have access to existing table {sanitize_for_log(table_name)}, so the "
+                f"service principal will not write to it on your behalf. Choose a destination you "
+                f"own or one that does not yet exist."
+            ) from e
+        if _looks_like_not_found(text):
```
mwojtyczka: **UC "not found" for permission-hidden tables is misclassified as "table absent."** UC returns `TABLE_OR_VIEW_NOT_FOUND` for tables the caller can't see, which matches `_NOT_FOUND_MARKERS` ("not found") → `rows=None` → falls through to the create-probe. If the caller has `CREATE TABLE` on the schema, verify *passes* for a table owned by someone else (practically gated by the SP's own grants — see the setup.py comment). More broadly, this security branch is driven by substring-matching free-text error strings (`_looks_like_not_found`), which breaks on any SDK/localization wording change — classify from a structured `error_code`/SQLSTATE instead.

mcp-server/server/utils.py
```diff
@@ -0,0 +1,874 @@
...
+    )
+
+    state = str(result.status.state.value if hasattr(result.status.state, "value") else result.status.state)
+    if state != "SUCCEEDED":
+        error_msg = getattr(result.status.error, "message", str(result.status.error)) if result.status.error else state
+        raise RuntimeError(f"SQL query failed: {error_msg}")
+
+    columns = [col.name for col in result.manifest.schema.columns]
+    rows = []
+    if result.result and result.result.data_array:
```
mwojtyczka: **`execute_sql` reads only the first result chunk (no pagination).** It consumes `result.result.data_array` and never follows chunk links, so: `sweep_stale_views` only sees the first chunk of `SHOW VIEWS` and never reaps orphan temp views beyond it (unbounded leak in the shared `tmp` schema); and `get_table_schema` silently truncates the column list for a very wide table, so the agent generates checks against an incomplete schema. Follow `next_chunk`/external links, or use a paginating helper.

mcp-server/runner/src/dqx_mcp_runner/runner.py
```diff
@@ -0,0 +1,386 @@
...
+    from databricks.labs.dqx.engine import DQEngine
+
+    checks = params["checks"]
+    location = params["location"]
+    run_config_name = params.get("run_config_name", "default")
+    mode = params.get("mode", "append")
+
+    engine = DQEngine(workspace_client=ws, spark=spark)
+    _handler, config = ChecksStorageHandlerFactory(ws, spark).create_for_location(location, run_config_name)
+    if hasattr(config, "mode"):
```
mwojtyczka: **`save_checks` mode is both silently dropped and validation-bypassed.** `if hasattr(config, "mode"): config.mode = mode` discards a caller's `mode="overwrite"` for any backend whose config lacks a `mode` attribute (still returns `saved: True` while appending), and the direct attribute set bypasses the Pydantic validator, so `mode="upsert"` fails deep in the Delta write with an opaque error instead of a clear rejection. The tool forwards `mode` with no whitelist — validate it up front and build the config with the mode set (via `.replace()` so the validator runs).

mcp-server/runner/src/dqx_mcp_runner/runner.py
```diff
@@ -0,0 +1,386 @@
...
+    from databricks.labs.dqx.profiler.profiler import DQProfile
+
+    profiles = params["profiles"]
+    criticality = params.get("criticality", "error")
+
+    dq_profiles = [
+        DQProfile(
+            name=p["name"],
+            column=p["column"],
+            description=p.get("description"),
```
mwojtyczka: **Hard `p["column"]` on caller-supplied profiles.** The workflow invites editing profiles/rules between steps and `tools.py` forwards them unvalidated, so a hand-edited profile missing `column` (or a future `DQProfile` field rename) raises `KeyError` inside the wheel task, surfacing only as a generic `failed` run with an opaque message. Validate required keys (or use `.get`) and raise a clear `InvalidParameterError`.

pyproject.toml
```diff
@@ -191,18 +191,18 @@ markdown = { header_level_by_type = { Method = 3, Function = 3, Class = 2, Modul
...
-# black uses the same ``line-length = 120`` setting in both, so the
-# library-level ``make fmt`` / ``make lint`` formats and lints the app
-# alongside the library.
-exclude = ['venv', '.venv', 'demos/*', 'tests/e2e/*', 'app/*']
+# ``basedpyright`` directly) instead. Black and ruff still traverse
+# into ``app/`` and ``mcp-server/`` from the root — ruff picks up each
+# project's own ``[tool.ruff]`` config automatically, and black uses
+# the same ``line-length = 120`` setting, so the library-level
+# ``make fmt`` / ``make lint`` formats and lints them alongside the library.
+exclude = ['venv', '.venv', 'demos/*', 'tests/e2e/*', 'app/*', 'mcp-server/*']
```
mwojtyczka, line 205: **mcp-server ships with no static type gating.** Adding `mcp-server/*` to the mypy exclude here (the comment implies parity with `app/`), but there's no `mcp-check`/basedpyright target and the `mcp` CI job in `push.yml` (L162) runs only `make mcp-test`. ~800 lines including the SQL-identifier validators and OBO auth handling merge with zero type checking — strictly weaker than the `app` job it's modeled on (which runs `app-check` + `app-test`). Add an `mcp-check` target and wire it into the mcp CI job.

.github/workflows/mcp.yml
```diff
@@ -0,0 +1,93 @@
...
+        uses: ./.github/actions/setup-env
+
+      - name: Pre-build DQX wheel
+        uses: ./.github/actions/prebuild-wheel
+
+      - name: Prepare code coverage configuration for MCP tests
+        run: |
+          cat > tests/integration_mcp/.coveragerc << EOF
+          [run]
+          source = ../../src
```
mwojtyczka: **MCP integration coverage measures the wrong code.** `source = ../../src` makes the `mcp`-flagged coverage reflect the DQX library, not `mcp-server/server` or `runner/`. New untested server branches show no coverage drop, giving false assurance that the server code is exercised. Point `source` at the mcp-server packages (and adjust the working dir / `pythonpath` accordingly).

mwojtyczka (commented):
Re-reviewed the whole PR from scratch

mcp-server/server/utils.py
```diff
@@ -0,0 +1,838 @@
...
+            "status": "failed",
+            "run_id": run_id,
+            "error": f"Job failed: {detail or 'Unknown error'}. Debug at: {run_url}",
+        }
+
+    # Read the result the runner (wheel task) wrote to the results volume, keyed by run id.
+    # (Wheel tasks have no notebook_output; the runner writes <run_id>.json via the Files API.)
+    result_path = f"{_get_results_volume()}/{run_id}.json"
+    try:
+        download = ws.files.download(result_path)
```
mwojtyczka, line 854: **Cross-user result disclosure (IDOR / missing authorization)**

`get_run_status(run_id)` (via the `get_run_result` tool) reads the result file `mcp_results/<run_id>.json` **as the app SP** here and returns it to whoever called the tool. The only isolation check is `run.job_id == runner_job_id` (above) — that the run belongs to the MCP runner job, **not** that the calling user owns it. There is no binding between the per-request user (the `X-Forwarded-Email` contextvar) and `run_id`.

Databricks job run-ids are sequential integers, and every workspace user has `CAN_USE` on the app + `CAN_MANAGE_RUN` on the runner job.

*Attack:* user B calls `get_run_result(<run_id belonging to user A>)` (e.g. their own id ±1) and receives A's result payload. For `run_checks` / `profile_table` that payload contains **sampled rows of A's governed source table** (`error_sample`, up to `sample_size`) plus profiling stats — bypassing the OBO definer's-rights-view governance the whole design rests on. The `mcp_results` volume itself is correctly locked to the SPs, so this SP-mediated read is the sole disclosure path — which is exactly why the missing per-caller check matters.

*Fix:* bind the caller identity to the run at submit (e.g. echo `owner_email` into the result, or a `job_parameters` value) and verify it equals `get_user_email()` before returning the result; otherwise return the structured `not_found`.

mcp-server/server/tools.py
```diff
@@ -0,0 +1,583 @@
...
+            - 'run_id': job run ID to pass to get_run_result
+        """
+        logger.info(f"Profiling table: {utils.sanitize_for_log(table_name)}")
+        obo_ws = utils.get_obo_client()
+        warehouse_id = utils.get_warehouse_id(obo_ws)
+        catalog, schema = _get_tmp_view_config()
+
+        view_fqn = utils.create_temp_view(obo_ws, table_name, catalog, schema, warehouse_id)
+
+        run_id = utils.submit_job_async(
```
mwojtyczka: **Temp-view leak if `submit_job_async` throws after `create_temp_view`.**

Here (and in `run_checks`, `load_checks`, `apply_checks_and_save_to_table`) the OBO temp view is created first, then `submit_job_async` is called. Cleanup relies on the runner's `finally` — but if submission fails *after* the view is created (e.g. `DQX_RUNNER_JOB_ID` unset → `RuntimeError`, `run_now` RPC error, throttled sweep raising), the runner never runs and nothing drops the view.

*Failure:* each failed submission leaves a persistent `v_<epoch>_<uuid>` view in the shared `tmp` schema until the TTL sweeper reaps it — up to the 1h TTL plus the 10-min sweep interval, and the sweep only fires on a *later successful* submit. Under a flapping runner job these accumulate.

*Fix:* wrap `submit_job_async` in try/except and `drop_view` (as the OBO client) on failure, or create the view only after a successful submit.

mcp-server/server/utils.py
```diff
@@ -0,0 +1,838 @@
...
+    Returns:
+        List of row dicts.
+
+    Raises:
+        RuntimeError: If the query fails.
+    """
+    result = ws.statement_execution.execute_statement(
+        statement=query,
+        warehouse_id=warehouse_id,
+        wait_timeout="30s",
```
mwojtyczka, line 292: **Minor / low-trigger: `wait_timeout="30s"` can misreport a slow-but-healthy statement as failed.**

With the default `on_wait_timeout` (CONTINUE), a statement still `PENDING`/`RUNNING` after 30s returns in a non-terminal state; the code below then sees `state != "SUCCEEDED"` and raises `RuntimeError("SQL query failed: PENDING")`. All current callers are fast metadata ops (`CREATE VIEW`, `DESCRIBE`, `SHOW VIEWS`, `DROP VIEW`), so the realistic trigger is a cold-start warehouse or a very wide/partitioned table. Consider polling to completion or handling the pending state as "keep waiting" rather than a hard failure.

mwojtyczka (commented):
Fresh review of the current head (medium effort). The earlier security items are verified fixed. Three net-new findings below — one is a genuine cross-user data-disclosure that defeats the OBO governance model, so worth blocking on.

mcp-server/runner/src/dqx_mcp_runner/runner.py
mwojtyczka: Verified on the latest head \u2014  now applies the write mode via  guarded by  (raising  otherwise), so the mode is no longer silently dropped nor set via validation-bypassing attribute assignment. Resolving.

mcp-server/runner/src/dqx_mcp_runner/runner.py
mwojtyczka: Verified on the latest head — profiles now go through _to_dq_profile, which raises a clear InvalidParameterError when name/column is missing instead of an opaque KeyError inside the wheel task. Resolving.

mcp-server/server/utils.py
mwojtyczka: Verified on the latest head — execute_sql now follows the chunk chain (while chunk is not None ... get_statement_result_chunk_n), so SHOW VIEWS sweeps and wide-table DESCRIBEs see the full result set. Resolving.

mcp-server/server/utils.py
mwojtyczka: Verified on the latest head — the caller-supplied write destination and the verify_obo_write_access owner==caller pre-check were removed entirely; outputs now go to the caller's SP-owned per-user schema (dqx_mcp_<user>), so there's no non-owner-writer rejection. Resolving.

mcp-server/server/utils.py
mwojtyczka: Verified on the latest head — the write pre-check and its _NOT_FOUND_MARKERS string-matching classification were removed with the move to per-user SP-owned output schemas, so the UC not-found misclassification path no longer exists. Resolving.

mcp-server/notebooks/setup.py
mwojtyczka: Verified on the latest head — writes no longer target caller-supplied destinations; the runner SP creates and owns the per-user schema (it has CREATE SCHEMA per setup.py) and writes there, so the SP-grant / PERMISSION_DENIED-at-runtime gap is gone. Resolving.

.github/workflows/mcp.yml
mwojtyczka: Verified on the latest head — mcp.yml's coverage source now points at mcp-server/server and mcp-server/runner/src/dqx_mcp_runner (with relative_files = true), so the mcp-flagged coverage measures the server/runner code, not the DQX library. Resolving.

pyproject.toml
mwojtyczka: Verified on the latest head — this is now addressed: there's a `make mcp-check` target (`basedpyright --level error server`, standard mode config in mcp-server/pyproject.toml, parity with `app-check`), and the `mcp` CI job ("Build and Check MCP Server" in push.yml) runs both `make mcp-check` and `make mcp-test`. The server's SQL-identifier validators and OBO handling are no longer merging untyped. Resolving.

One residual (non-blocking, optional follow-up): `mcp-check` type-checks only `server`, not `runner/src/dqx_mcp_runner` — the runner wheel (~470 lines, incl. the per-user schema/grant SQL construction) still isn't statically checked. Consider adding it to the basedpyright invocation.

mcp-server/server/utils.py
mwojtyczka: Verified fixed on the latest head (bdc8f4b1). `submit_job_async` now records `requesting_user` (from the OBO `X-Forwarded-Email`) as a declared job parameter, and `get_run_status` reads it back from `run.job_parameters` and denies the read (returns `not_found`, without downloading the result) unless the caller's identity matches the submitter — with an explicit empty-either-side guard so an unowned run or an identity-less caller can never match. Confirmed against the SDK (`Run.job_parameters` / `JobParameter.name`/`.value` round-trip) and covered by regression tests in `test_utils.py` (`test_denies_run_submitted_by_another_user`, `test_denies_when_caller_has_no_identity`, `test_denies_when_run_has_no_submitter`, `test_allows_run_submitted_by_same_user`, all asserting `ws.files.download.assert_not_called()` on deny). Since submit and get_run_result both traverse the same OBOAuthMiddleware transport (Databricks Apps always injects `X-Forwarded-Email` for an authenticated user), the guard is symmetric — no legitimate-caller lockout. Resolving.

mcp-server/server/tools.py
mwojtyczka: Verified fixed on the latest head — a new `_submit_or_drop_view` wrapper drops the OBO temp view (as the creating OBO client) if `submit_job_async` raises after `create_temp_view`, and it's applied to all view-backed tools (profile_table, run_checks, apply_checks_and_save_to_table, and the table branch of load_checks; the file branch correctly uses plain submit with no view). Double-drop with the runner's `finally` is harmless (`DROP VIEW IF EXISTS`). Covered by `test_tools.py::test_drops_temp_view_if_submit_fails`. Resolving.

mcp-server/server/utils.py
mwojtyczka: Verified fixed on the latest head — `execute_sql` now polls `get_statement(statement_id)` (2s interval, 120s cap) while the statement is PENDING/RUNNING instead of misreporting a slow-but-healthy query as failed, and hardens the None-status deref via `_statement_state`. `get_statement` returns the same `StatementResponse` type as `execute_statement` (with `.manifest`/`.result` populated on completion), so the subsequent chunk-follow is valid. Covered by `test_none_status_raises_clean_error_not_attributeerror`. Resolving.

mwojtyczka (commented):
Re-review of the current head (bdc8f4b1) after the fix commits. All three findings from the prior review are verified fixed and covered by regression tests — I've resolved those threads.

- **IDOR / cross-user result disclosure** — `submit_job_async` now stamps `requesting_user` (from the OBO `X-Forwarded-Email`) as a declared job parameter, and `get_run_status` reads it back and denies the read (`not_found`, no result download) unless caller == submitter, with an explicit empty-either-side guard. Verified against the SDK (`Run.job_parameters` round-trip) and the guard is correctly positioned *before* the result download / RUNNING branch. Tests: `test_denies_run_submitted_by_another_user`, `test_denies_when_caller_has_no_identity`, `test_denies_when_run_has_no_submitter`, `test_allows_run_submitted_by_same_user` (all assert `files.download` is not called on deny). Because submit and get_run_result both traverse the same OBO middleware, the guard is symmetric — no legitimate-caller lockout.
- **Temp-view leak on submit failure** — new `_submit_or_drop_view` wrapper drops the OBO view on failure across all view-backed tools; file-backed `load_checks` correctly skips it. Test: `test_drops_temp_view_if_submit_fails`.
- **`execute_sql` 30s-timeout misreport** — now polls `get_statement` to completion (2s/120s) and hardens the None-status deref; `get_statement` returns the same `StatementResponse` (manifest+result populated on SUCCEEDED), so the chunk-follow stays valid. Test: `test_none_status_raises_clean_error_not_attributeerror`.

Proactive hardening in the same commits also checks out: leading-digit identifier rejection (unquoted-FQN safety), partial-grant reporting (`access_granted_to` null unless every created table was granted), and the removed dead `to_local_fuse_path`. The changed deploy/CI/docs (requesting_user declared at job level, `ci_deploy.sh` threads the runner SP, `mcp.yml` stays `pull_request` + fork-gated, docs anchors resolve) are all clean.

No new findings on this head. Remaining open threads are the two I intentionally left for the authors (the docs-placement suggestion) — the type-gating thread was resolved separately once `mcp-check` landed in CI.

mwojtyczka (approved):
LGTM - pending manual testing, and code coverage

docs/dqx/docs/installation.mdx
ghanse: This doc is getting quite long. Maybe we should separate the different installation docs under a `/installation` folder so that each is more scoped?

@mwojtyczka wdyt?

PR #1259, opened 2026-06-20
Migrate Configs and Rules to Pydantic

### LINKED ISSUE #467: [FEATURE]: Migrate data models to Pydantic

### Problem statement

Currently, we define DQX rules using data classes, which requires extensive and repetitive validation logic. This approach is both error-prone and difficult to maintain.

### Proposed Solution

We propose migrating to Pydantic to handle validation automatically. Pydantic would also simplify tasks such as serializing and deserializing rules to and from YAML, improving overall developer productivity and code maintainability.

### Additional Context

Runtime 15.4 uses Pydantic v1.10.6
Runtime 16.4 and newer use Pydantic v2.8.2

Since Pydantic v1 and v2 are not backward-compatible, we’ll need a strategy to ensure compatibility with older runtimes while transitioning to Pydantic v2 for new development in DQX. Installing v2 in 15.4 may introduce some side effects.

fedeflowers:
## Changes

Migrate DQX data models from dataclasses to Pydantic v2 for automatic
validation and simpler YAML serialization/deserialization.

Migrated to `pydantic.BaseModel`:
- `checks_validator.py` — `ChecksValidationStatus` (errors held in a real
  aliased field so it stays constructor-settable)
- `config.py` — 7 storage-config classes; `__post_init__` →
  `@model_validator(mode="after")` with `type(self) is ...` guards for the
  multiple-inheritance composite; `cached_property` → `property`
- `rule.py` — `DQRule`, `DQRowRule`, `DQDatasetRule`, `DQForEachColRule` →
  frozen Pydantic models (`ConfigDict(frozen=True, arbitrary_types_allowed=True)`
  for Spark `Callable`/`Column` fields); class constants marked `ClassVar`
- `pyproject.toml` — add `pydantic>=2.8.2`

Intentionally left as dataclasses: `WorkspaceConfig`, `RunConfig`, and the
nested config tree, because databricks-labs-blueprint's `Installation.load()`
relies on `dataclasses.is_dataclass()` / `__dataclass_fields__`, which Pydantic
models don't satisfy.

### Linked issues

Resolves #467

### Tests

- [x] manually tested
- [ ] added unit tests
- [ ] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests


src/databricks/labs/dqx/rule.py
```diff
@@ -173,18 +172,44 @@ class DQRule(abc.ABC, DQRuleTypeMixin, SingleColumnMixin, MultipleColumnsMixin):
         rules generated from a ``DQForEachColRule``.
     """
 
+    model_config = ConfigDict(frozen=True, arbitrary_types_allowed=True)
```
mwojtyczka, line 227: `DQRule` is now a frozen Pydantic model, but `engine.py:424` still mutates a rule via `dataclasses.replace` (`from dataclasses import replace` at engine.py:7):

```python
# engine.py:424 — _filter_for_original_columns_preselection
return replace(check, check_func_kwargs=rule_kwargs)
```

`dataclasses.replace()` raises `TypeError: replace() should be called on dataclass instances` for Pydantic models, so this crashes whenever a check goes through the original-columns preselection path (the `columns`-injection branch). Switch to `check.model_copy(update={"check_func_kwargs": rule_kwargs})`. This is a production path, so it needs an integration run to confirm — `make integration` / the dataset-check tests will exercise it.

src/databricks/labs/dqx/checks_validator.py
```diff
@@ -2,21 +2,21 @@
...
 logger = logging.getLogger(__name__)
 
 
-@dataclass(frozen=True)
-class ChecksValidationStatus:
+class ChecksValidationStatus(BaseModel):
     """Class to represent the validation status."""
 
-    _errors: list[str] = field(default_factory=list)
+    _errors: list[str] = PrivateAttr(default_factory=list)
```
mwojtyczka: Using `PrivateAttr` means `_errors` is no longer settable via the constructor — Pydantic v2 does not accept private-attr kwargs in `__init__`. That breaks the existing test `tests/unit/test_app_backend.py:1873`:

```python
validation = ChecksValidationStatus(_errors=["Unknown function: bad_func"])
```

The `_errors=` kwarg is dropped, so `validation.has_errors` is `False` and the test's intent is lost. This also contradicts the PR description ("errors held in a real aliased field so it stays constructor-settable") — the field is private, not aliased. Either make `errors` a real aliased field (`Field(alias="errors")` + `model_config = ConfigDict(populate_by_name=True)`) as the description claims, or update the call site to `ChecksValidationStatus()` followed by `add_errors([...])`.

src/databricks/labs/dqx/config.py
```diff
@@ -418,19 +419,26 @@ class LakebaseChecksStorageConfig(BaseChecksStorageConfig):
             When None (default), loads the latest batch.
     """
 
-    location: str
     instance_name: str | None = None
     client_id: str | None = None
     port: str = "5432"
     run_config_name: str = "default"
     mode: str = "append"
```
mwojtyczka, line 459: `LakebaseChecksStorageConfig` is now a Pydantic `BaseModel`, but `tests/integration/test_save_and_load_checks_from_lakebase_table.py:511-512` still constructs updated configs with `dataclasses.replace`:

```python
dq_engine.save_checks(checks=TEST_CHECKS[:1], config=dataclasses.replace(config, mode="append"))
dq_engine.save_checks(checks=TEST_CHECKS[1:], config=dataclasses.replace(config, mode="overwrite"))
```

`dataclasses.replace()` raises `TypeError` on Pydantic models — use `config.model_copy(update={"mode": "append"})` instead.

mwojtyczka (changes requested):
Generally looking good, left some comments. 

We will need follow up PRs for pydantic to cover serialization and validation. That would bring a real value. It's fine to just have base-class swap in this PR. That's a reasonable incremental step; routing validation/serialization through Pydantic is a separate, higher-risk refactor that shouldn't bloat this one.

fedeflowers:
Recap of changes:

**1. `engine.py` — `dataclasses.replace` in `_filter_for_original_columns_preselection`**

Added a small `DQRule.replace()` helper and call `check.replace(check_func_kwargs=rule_kwargs)` instead of `model_copy(update=...)`. The reason for not using `model_copy` directly: it (a) does not re-run validators — DQRule's `mode="after"` validator validates attributes and initialises the name — and (b) shallow-copies the instance dict, so it carries over already-cached `functools.cached_property` values (`rule_fingerprint`, `columns_as_string_expr`). `replace()` rebuilds through the constructor, so validation re-runs and derived state is recomputed from the updated fields. Quick demonstration of the difference:

    r = DQRowRule(name="x", check_func=is_not_null, column="id")
    _ = r.rule_fingerprint  # populate the cached_property
    r.model_copy(update={"column": "y"}).rule_fingerprint == r.rule_fingerprint  # True  (stale)
    r.replace(column="y").rule_fingerprint            == r.rule_fingerprint       # False (recomputed)

**2. `checks_validator.py` — `_errors` not constructor-settable**

Good catch — `PrivateAttr` (and the interim aliased-field variant) were both more trouble than they were worth. Made `errors` a plain public field instead: `errors: list[str] = []`, which is constructor-settable as `ChecksValidationStatus(errors=[...])`; `has_errors` stays a `@property`. Updated the one call site (`test_app_backend.py:1873`) from `_errors=[...]` to `errors=[...]`, so the test once again actually exercises the error path. Also added `model_config = ConfigDict(extra="forbid")` so a stray/unknown kwarg now raises `ValidationError` loudly instead of being silently dropped. This also removed the `Field(...)` that was tripping pylint's `no-member` on `self.errors.append`, so `make lint` is green again.

**3. lakebase test — `dataclasses.replace` on `LakebaseChecksStorageConfig`**

Rather than `model_copy(update={"mode": ...})`, I added a validating `BaseChecksStorageConfig.replace()` helper (mirroring `DQRule.replace()`) and switched the test to `config.replace(mode=...)`, and dropped the `import dataclasses`. The reason for not using `model_copy`: it skips all validators, so an invalid override (e.g. a bad `mode` or a non-3-part `location`) would be accepted silently and only blow up later inside the save path. `replace()` rebuilds through the constructor so the storage-config validators re-run — verified:

    c = LakebaseChecksStorageConfig(location="a.b.c", instance_name="inst")
    c.model_copy(update={"mode": "BOGUS"}).mode  # 'BOGUS'           (accepted silently)
    c.replace(mode="BOGUS")                       # raises InvalidConfigError

**Also did a completeness sweep / added coverage:**
- Grepped the repo for other `dataclasses.replace`/`asdict` on migrated models — the remaining occurrences all target classes still defined as `@dataclass` (`WorkspaceConfig`, workflow `Task`, `DQProfile`, anomaly `ExplanationContext`), so they're unaffected.
- Added unit tests for `DQRule.replace()` (override + field preservation, cached-fingerprint recomputation, validator re-run) and `BaseChecksStorageConfig.replace()` (override + field preservation, validator re-run).
- `make lint` green (10.00/10); affected unit tests pass (validator + storage configs + rule + app dry-run). Ran the affected integration paths: `has_valid_schema` (the `check.replace` preselection path) — 21 passed, 1 skipped; rule-fingerprint apply — 4 passed.

grusin-db:
What is the planned approach for serialisation? that's the part where actually pydantic support matterns the most. 

In dqx we have few serializers, we have yaml mode, where yamls need to be serialized to normal validators, also we have delta serializaer for storage, and probablt few more I am not aware/slipped my mind. This serializers needs considered before making changes in the models, so that we dont deadlock ourselves with **potentially** bad decisions.

grusin-db:
Another topic, since we have now DQRules as pdyantic, have you tried serializing them into `json_schema`? there is `.model_json_schema()`, this should be included in testing, to ensure it passes. 

It might - and probably will trip - on some internal non serializable types, so this needs to be verified to make sure it works e2e

grusin-db:
In case of bad user input, all the errors will now be totally different, if pydantic model is throwing it will thrown its own `ValidationError`, which in turn will trip all validation error tests we have, hence that needs to be covered.

I don't see any tests changes regarding that. I.e. putting "blabla" into "int" field, would for sure throw pydantic validation error about "blabla" being not serializable to an int

src/databricks/labs/dqx/rule.py
mwojtyczka: Verified on `7a8a5465` — `engine.py:423` now calls `check.replace(check_func_kwargs=...)`, and `DQRule.replace()` rebuilds through the constructor so validators re-run (better than `model_copy`, which would skip them). Resolving.

src/databricks/labs/dqx/checks_validator.py
mwojtyczka: Verified — `errors` is now a real field (`errors: list[str] = []`, `extra="forbid"`) and `test_app_backend.py:1873` constructs it with `errors=[...]`. Constructor-settable as intended. Resolving.

src/databricks/labs/dqx/config.py
mwojtyczka: Verified — the lakebase test now uses `config.replace(mode=...)`, and `BaseChecksStorageConfig.replace()` rebuilds through the constructor so the model validators re-run. Resolving.

mwojtyczka:
> What is the planned approach for serialisation? that's the part where actually pydantic support matterns the most.
> 
> In dqx we have few serializers, we have yaml mode, where yamls need to be serialized to normal validators, also we have delta serializaer for storage, and probablt few more I am not aware/slipped my mind. This serializers needs considered before making changes in the models, so that we dont deadlock ourselves with **potentially** bad decisions.

we will do this as a follow up PRs: serialization and validation

mwojtyczka:
> Another topic, since we have now DQRules as pdyantic, have you tried serializing them into `json_schema`? there is `.model_json_schema()`, this should be included in testing, to ensure it passes.
> 
> It might - and probably will trip - on some internal non serializable types, so this needs to be verified to make sure it works e2e

You are right:
DQRule.model_json_schema()       → PydanticInvalidForJsonSchema: Cannot generate a JsonSchema for core_schema.CallableSchema
DQRowRule.model_json_schema()    → same
FileChecksStorageConfig          → OK
ChecksValidationStatus           → OK

The Callable/Column fields need a custom json_schema / WithJsonSchema annotation — and a test

mwojtyczka:
> In case of bad user input, all the errors will now be totally different, if pydantic model is throwing it will thrown its own `ValidationError`, which in turn will trip all validation error tests we have, hence that needs to be covered.
> 
> I don't see any tests changes regarding that. I.e. putting "blabla" into "int" field, would for sure throw pydantic validation error about "blabla" being not serializable to an int

Yes, FileChecksStorageConfig raises `pydantic_core.ValidationError`. So programmatic construction with wrong-typed fields now surfaces Pydantic's error instead of DQX's own. The tests pass because the dict-level `ChecksValidator` still catches metadata errors before construction — but the model-construction path has no test coverage for the new error shape. I think DQX should catch/wrap `ValidationError` into its own error types for a consistent API rather than leak `pydantic.ValidationError`.

mwojtyczka (changes requested):
@fedeflowers I left comments as a follow up requests from @grusin-db , at least 2 of them are valid. Using pydantic for serializaiton can be done as a follow up PR. I kicked tests suite, it seems some tests are failing. Please check.

fedeflowers:
## Summary

Fix pydantic validation for rule/config models, wrap raw pydantic errors as DQX errors, and add regression tests.

### Changes by file

- **`pyproject.toml`** — Added `tests/unit/test_engine.py` to the pylint `protected-access` exception list, since the new engine tests call the private `_preselect_original_columns` method directly.

- **`src/databricks/labs/dqx/anomaly/core.py`** — Removed the unused `DEFAULT_MAX_ROWS` constant (dead-code cleanup, no behavioral impact).

- **`src/databricks/labs/dqx/checks_validator.py`** — Expanded the `ChecksValidationStatus` docstring to document its in-place mutable-accumulator behavior and shallow-copy sharing risk, and fixed a copy-paste docstring on `add_errors`.

- **`src/databricks/labs/dqx/config.py`** — `BaseChecksStorageConfig.__init__` now catches pydantic `ValidationError` and re-raises it as DQX's `InvalidConfigError`. Added comments to the per-class `validate_location` validators explaining why the exact-class guard is needed (diamond inheritance in `InstallationChecksStorageConfig`).

- **`src/databricks/labs/dqx/installer/logs.py`** — Removed the unused `LogRecord` dataclass (dead-code cleanup).

- **`src/databricks/labs/dqx/rule.py`** — Migrated `DQRule`/`DQForEachColRule` Column fields to the new `SparkColumn` type and annotated `check_func` with `WithJsonSchema`, fixing JSON-schema generation and Spark Connect column acceptance. Replaced the imperative `_initialize_column(s)_if_missing` setters with a `mode='before'` validator that promotes `column`/`columns` from `check_func_kwargs`, and wrapped construction to raise `InvalidCheckError` instead of raw pydantic errors.

- **`src/databricks/labs/dqx/utils.py`** — Introduced the `SparkColumn` annotated type with a `PlainValidator` that accepts both classic `pyspark.sql.Column` and the Spark Connect column variant (read at call time so tests can monkeypatch). Reorganized imports to support it.

- **`tests/integration/test_save_and_load_checks_from_table.py`** — Dropped the `@dataclass` decorator from the test `ChecksDummyStorageConfig`, since decorating a pydantic subclass generates a conflicting `__init__`. Documented why in the docstring.

- **`tests/unit/test_build_rules.py`** — Added regression tests covering Spark Connect column acceptance/rejection, `model_json_schema()` generation for all three rule types, and `ValidationError → InvalidCheckError` wrapping on bad construction input.

- **`tests/unit/test_config.py`** — Added a test asserting that a wrong-typed field on a `BaseChecksStorageConfig` subclass raises `InvalidConfigError` rather than a raw pydantic error.

- **`tests/unit/test_engine.py`** — Added three tests for `_preselect_original_columns`: no-op for unregistered functions, injects DataFrame columns via `DQRule.replace()` when registered, and does not overwrite columns already present in `check_func_kwargs`.

mwojtyczka (approved):
LGTM - can merge once all tests are green

pyproject.toml
```diff
@@ -40,6 +40,7 @@ dependencies = [
     "databricks-sdk~=0.73",
     "sqlalchemy>=2.0,<3.0",
     "PyYAML~=6.0.3",
+    "pydantic>=2.8.2",
```
mwojtyczka: **Cap the upper bound on `pydantic`.** This is the only direct dependency without a ceiling — `databricks-sdk~=0.73`, `sqlalchemy>=2.0,<3.0`, `PyYAML~=6.0.3`, and `databricks-labs-blueprint>=0.9.1,<=0.12` all bound the major. `>=2.8.2` permits a future **pydantic 3.0**, which (given the 1→2 break) could break exactly the APIs this PR relies on — `model_validator`, `ConfigDict(frozen=...)`, `WithJsonSchema`, the `ValidationError` wrapping. Since DQX is a distributed library, the floor-only spec propagates that risk to every consumer's resolution.

```suggestion
    "pydantic>=2.8.2,<3",
```
Then re-run `make lock-dependencies` so `uv.lock`'s `specifier = ">=2.8.2"` regenerates to `">=2.8.2,<3"` (no manual lock edit needed).

FWIW the `uv.lock` change itself is correct/minimal: pydantic (2.12.5) was already resolved transitively via mlflow/dspy/presidio, so promoting it to a direct dep only adds the two edges — no new stanza expected. And `.build-constraints.txt` is correctly untouched (it only pins the PEP 517 build backend).

uv.lock
```diff
@@ -1340,6 +1341,7 @@ requires-dist = [
     { name = "mlflow", extras = ["databricks"], marker = "extra == 'anomaly'", specifier = ">=2.19.1,<3.0" },
     { name = "numpy", marker = "extra == 'pii'", specifier = ">=1.20,<2.0" },
     { name = "presidio-analyzer", marker = "extra == 'pii'", specifier = "~=2.2.359" },
+    { name = "pydantic", specifier = ">=2.8.2" },
```
mwojtyczka: Mirror the `pyproject.toml` cap here too. The resolved version (pydantic 2.12.5) already satisfies `<3`, so this matches what `make lock-dependencies` regenerates — no re-resolution needed. Best to regenerate via `make lock-dependencies` after applying the `pyproject.toml` suggestion rather than hand-editing, but the change is just:

```suggestion
    { name = "pydantic", specifier = ">=2.8.2,<3" },
```

PR #1262, opened 2026-06-22
Feat/app UI i18n insights contracts

berrybluecode:
## Changes

This PR adds substantially richer
authoring UI, and overhauls the deployment story. At a high level:

**Data-contract & AI rule generation**
- Import ODCS v3.x data contracts and generate DQX rules (predefined,
  property-based, and natural-language) via a new `contract_rules_service` and
  `/v1/contract` route, including a dedicated "from contract" UI.
- Natural-language (`type: text`) expectations are generated through the app's
  ChatDatabricks Foundation Model path (no dspy/Spark needed in-container).
- Reference-table checks (`foreign_key`, `has_valid_schema`) are now authored in
  the single-table editor with a table picker; schema-validation gets its own
  editor and mutual-exclusion handling for `expected_schema` vs `ref_table`.

**UI features**
- Internationalization (English, Spanish, Italian, Portuguese-BR) with a
  language selector and locale-parity test.
- Insights dashboard is now hosted persistently so the embedded Lakeview iframe
  no longer reloads on every navigation; expandable labels UI; component
  reorganisation (`apx/` → `layout/`); regenerated API client (orval).

**Deployment & docs**
- Updated DEPLOYMENT.md / DEVELOPMENT.md (CLI v1.4.0+ gate, one-button
  `postgres_roles`, bind workflow) and removed duplicated content.

### Linked issues

Resolves #..

### Tests

Added backend unit suites for the Lakebase executor, Postgres/Delta migration
runners, contract rule generation, discovery DDL, dry-run dispatch, app
scheduler lease, quarantine filtering, lint policy, and i18n locale parity.
`make app-test` (658 passing) and `make test` (1160 passing) are green, along
with `make fmt` and `make app-check`.

- [x] manually tested
- [x] added unit tests
- [ ] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

### Documentation and Demos

- [ ] added/updated demos
- [x] added/updated docs


app/src/databricks_labs_dqx_app/backend/routes/v1/contract.py
```diff
@@ -0,0 +1,96 @@
...
+    service: Annotated[ContractRulesService, Depends(get_contract_rules_service)],
+) -> GenerateRulesFromContractOut:
+    """Parse an ODCS v3.x contract and return generated DQX rules grouped per schema."""
+    if body.default_criticality not in ("error", "warn"):
+        raise HTTPException(
+            status_code=400,
+            detail="default_criticality must be 'error' or 'warn'",
+        )
+    try:
+        result = service.generate(
```
mwojtyczka, line 52: **Contract-generated rules are returned without validation, unlike the `/generate` path.** The AI-assisted generation route validates before returning (`routes/v1/generate.py:34` → `DQEngine.validate_checks(checks)`), but here `service.generate(...)` → `_to_out(result)` returns predefined/property-based **and `text_llm` (LLM-produced)** rules with no `DQEngine.validate_checks`, no `is_sql_query_safe()` on generated SQL, and no registry check on generated function names.

Per AGENTS.md (LLM section): *"LLM-generated output — function names, SQL fragments, rule arguments — must be validated before execution: function names must resolve through `CHECK_FUNC_REGISTRY`; any generated SQL must pass `is_sql_query_safe()`."* A hallucinated/malformed rule flows to the UI and only fails later at execution. Please route contract output through the same `DQEngine.validate_checks` gate as `/generate`, and validate any LLM-generated SQL.

Two smaller related points in `contract_rules_service.py`: the `ChatDatabricks` calls have no token/budget cap (AGENTS.md asks for limits on LLM calls), and the catch-all handler surfaces the raw exception string back to the caller (`warnings`) — avoid relaying raw LLM/SDK error detail.

app/src/databricks_labs_dqx_app/ui/components/RunReviewStatusPanel.tsx
```diff
@@ -0,0 +1,301 @@
...
+    );
+  };
+
+  const handleClear = () => {
+    clearMutation.mutate(
+      { runId },
+      {
+        onSuccess: () => {
+          refresh();
+          toast.success("Reverted to default");
```
mwojtyczka: **Hardcoded user-facing strings bypass i18n.** These toasts (`"Reverted to default"`, `"Failed to revert review status"`) are English literals rather than `t(...)` lookups, so they won't translate under es/it/pt-BR. Same pattern in `routes/_sidebar/config.tsx` (e.g. lines 688, 893, 915, 917, 1243: `"Retention settings saved."`, `"Dashboard settings saved."`, `"Cleared dashboard override."`, `"Failed to clear dashboard override."`, `"Review statuses saved."`). The UI CLAUDE.md requires user-facing display text to go through the i18n `t()` function.

mwojtyczka:
### Additional findings — pre-existing (not introduced by this PR)

While reviewing the backend changes I hit three issues in code this PR touches/neighbours. `git blame` puts all three at or before this branch's merge-base, so they're **pre-existing, not regressions from this PR** — flagging here for visibility; fine to fix in a follow-up rather than blocking this PR. (They're posted at PR level because they sit on unchanged lines, which GitHub won't let me anchor an inline comment to.)

**1. Scheduler re-fires a schedule every tick if `_trigger_run` raises — `scheduler_service.py:294-302`.**
`next_run_at` is only advanced to a future time at the `_upsert_tracker` call on line 301, which runs *after* `_trigger_run` returns. Per-table failures are caught inside `_trigger_run` and returned as `errors`, but `_resolve_scope` / `_load_custom_metrics` / DB access sit *outside* that inner try. If any raises, it propagates to the `except` at line 302 (log-only); since line 275 already wrote `next_run_at = now - 1s`, the schedule is "due" again on the next tick → a deterministic failure becomes a tight retry loop that keeps submitting jobs. Relevant now because this PR adds new code paths inside `_trigger_run`. Fix: advance `next_run_at` (or persist a `failed` status with backoff) in a `finally`/`except`.

**2. Monthly `day_of_month` silently capped at 28 — `scheduler_service.py:1089`.**
`candidate = after.replace(day=min(dom, 28), ...)` — a schedule configured for the 29th/30th/31st always fires on the 28th while the UI shows the configured day. Clamp to the actual last day of the target month instead, e.g. `calendar.monthrange(year, month)[1]`.

**3. Non-`TERMINATED` terminal states written verbatim as run status — `dryrun.py:533-539`.**
When `status.state` is `INTERNAL_ERROR` or `SKIPPED`, `update_run_status(..., status=status.state, ...)` writes those raw strings into the run-history `status` column, outside the normal `SUCCESS`/`FAILED`/`CANCELED`/`RUNNING` vocabulary (the synthesized path at 505-513 correctly maps terminal → SUCCESS/FAILED). The row does update (not stuck RUNNING), but downstream UI/queries may not recognize these values. Consider mapping `INTERNAL_ERROR`/`SKIPPED` → `FAILED`.

app/scripts/_align_wheel_version.py
```diff
@@ -0,0 +1,169 @@
+"""Re-align a wheel's METADATA / RECORD / dist-info dir to its filename.
+
+Background
+----------
+``apx build`` post-processes the freshly-built application wheel by
```
mwojtyczka: remove apx references

app/scripts/_align_wheel_version.py
```diff
@@ -0,0 +1,169 @@
...
+``*.dist-info`` directory name.
+
+The result is a wheel whose filename advertises a new version each
+build but whose package metadata still reports the unchanged
+git-derived version.  ``pip`` consults the metadata, sees the version
+is already installed, and silently skips reinstall — so a Databricks
+App container's persistent venv keeps running stale code on every
+redeploy.
+
+This script repairs the wheel produced by ``apx build`` by:
```
mwojtyczka: Remove references to apx, just describe what we want to achieve

mwojtyczka (changes requested):
Mostly looks good, left some comments. There are some pre-existing issues as well.

app/src/databricks_labs_dqx_app/ui/components/RunReviewStatusPanel.tsx
```diff
@@ -0,0 +1,304 @@
...
+  const setMutation = useSetRunReviewStatus();
+  const clearMutation = useClearRunReviewStatus();
+
+  const refresh = () => {
+    queryClient.invalidateQueries({ queryKey: getRunReviewStatusQueryKey(runId) });
+    queryClient.invalidateQueries({ queryKey: getRunReviewStatusHistoryQueryKey(runId) });
+    // Bust the listing cache too so the History page row reflects the
+    // change immediately when the user collapses the row.
+    queryClient.invalidateQueries({ queryKey: ["listValidationRuns"] });
+    queryClient.invalidateQueries({ queryKey: ["/listValidationRuns"] });
```
mwojtyczka: **These invalidations never match the runs-listing cache, so History shows stale review status.** The listing query key is `getListValidationRunsQueryKey()` → `["/api/v1/dryrun/runs"]` (api-custom.ts:194-195), but here we invalidate `["listValidationRuns"]` / `["/listValidationRuns"]`. TanStack Query matches query keys by array prefix, so neither matches — the listing cache is **not** busted, contradicting the "bust the listing cache too" intent. After a user changes a run's review status, the Runs History row badge/filter keeps showing the old value until `staleTime` expiry or a manual refetch. Use `getListValidationRunsQueryKey()` here instead.

app/src/databricks_labs_dqx_app/ui/routes/_sidebar/runs-history.tsx
```diff
@@ -105,6 +114,150 @@ function StatusBadge({ status }: { status: string | null }) {
...
+              {selected.size}
+            </Badge>
+          )}
+          <ChevronDown className="h-3 w-3 opacity-60" />
+        </Button>
+      </PopoverTrigger>
+      <PopoverContent align="start" className="w-72 p-1">
+        {available.length === 0 && (
+          <div className="px-2 py-1.5 text-[11px] text-muted-foreground">
+            No statuses configured yet. Define them in Configuration.
```
mwojtyczka: **Hardcoded English display text bypasses i18n.** This string (and the nearby `"Clear selection"` at ~252, `title="Filter by review status"` at ~177, `title="Default — not yet reviewed"` at ~141, `"Review"`, `"auto"`) is an English literal rather than a `t(...)` lookup, in a file that already calls `t()` ~55 times. ui/CLAUDE.md requires user-facing display text (including titles/placeholders) to go through `t()`; es/it/pt-BR users will see these in English with no locale keys defined. (The toast strings flagged in the prior review are correctly fixed — this is the new review-status filter UI.)

app/src/databricks_labs_dqx_app/backend/services/contract_rules_service.py
```diff
@@ -0,0 +1,530 @@
...
+                "is not configured on this deployment."
+            ]
+
+        expectations = self._extract_text_expectations(contract_text)
+        if not expectations:
+            return [], []
+
+        rules: list[dict[str, Any]] = []
+        warnings: list[str] = []
+        for exp in expectations:
```
mwojtyczka, line 241: **Unbounded LLM fan-out — cost/latency amplification.** This loop issues one `ChatDatabricks` call per `type: text` expectation with no cap on the number of expectations, and `GenerateRulesFromContractIn.contract_text` (models.py:62) has no `max_length`. A RULE_AUTHOR can POST a large ODCS contract with thousands of `text` quality entries and trigger thousands of LLM invocations. Per-call `max_tokens` is capped, but the call **count** is not — AGENTS.md asks for budget limits on LLM calls (OWASP LLM04). It's gated to author+ so it's amplification, not privilege escalation, but please bound `len(expectations)` and add a `max_length` to `contract_text`.

app/src/databricks_labs_dqx_app/backend/pg_executor.py
```diff
@@ -533,17 +539,21 @@ def upsert_with_audit(
...
             qcol = self.q(col)
             if col == increment_on_update:
-                # Bare column reference resolves to the existing row.
-                update_pairs.append(f"{qcol} = {qcol} + 1")
+                # Table-qualified reference resolves unambiguously to
+                # the existing row — even when ``EXCLUDED`` also exposes
+                # the same column name (which is always the case for
+                # ``increment_on_update`` since it has to be in
+                # ``value_cols``).
+                update_pairs.append(f"{qcol} = {table}.{qcol} + 1")
```
mwojtyczka: **Please verify this schema-qualified self-reference against live Lakebase.** With `table = self._sql.fqn("dq_schedule_configs")` this renders as `"version" = "dq"."dq_schedule_configs"."version" + 1`. The Postgres docs describe the existing-row reference in `ON CONFLICT DO UPDATE SET` as "the table's **name** (or an alias)" — it's not obvious that a *schema-qualified* reference is accepted on the RHS, and since this is resolved at parse/analyze time, if Lakebase rejects it the error (`invalid reference to FROM-clause entry for table "dq"`) would fire on the **first** `save` of any schedule config, not just on conflict. The regression test (test_pg_executor.py:546) only asserts the rendered **string shape** and never executes against Postgres, so it wouldn't catch this. Could you confirm a real schedule save+update succeeds against a live Lakebase instance? If it errors, qualify with the bare table name (`dq_schedule_configs."version" + 1`) instead of the schema-qualified FQN.

app/src/databricks_labs_dqx_app/backend/services/role_service.py
```diff
@@ -133,16 +178,118 @@ def create_mapping(self, role: str, group_name: str, user_email: str) -> RoleMap
...
+        WARNING with the stack trace so they're still investigable
+        post-hoc.
+        """
+        try:
+            escaped_role = escape_sql_string(role)
+            escaped_group = escape_sql_string(group_name)
+            escaped_action = escape_sql_string(action)
+            user_sql = f"'{escape_sql_string(user_email)}'" if user_email else "NULL"
+            sql = (
+                f"INSERT INTO {self._history_table} "
```
mwojtyczka, line 279: **SQL injection via `group_name` (and the same pattern in `create_mapping`/`list_history`).** `escape_sql_string()` only doubles single quotes (`value.replace("'", "''")`) and, per its own docstring, deliberately does **not** backslash-escape. On the Delta/Spark executor path the default `spark.sql.parser.escapedStringLiterals=false` means `\` is an escape char, so a free-form `group_name` like `g\' OR 1=1 --` terminates the string literal early and breaks out into the `INSERT`/`DELETE`/`SELECT` against `dq_role_mappings(_history)`. `role` is effectively enum-constrained but `group_name` is arbitrary. Admin-gated, so impact is bounded — but it's the exact class AGENTS.md says to prevent, and the root cause is the shared `escape_sql_string`/`_render_value`. Suggest validating the group identifier (or parameterizing) before interpolation.

app/src/databricks_labs_dqx_app/backend/routes/v1/config.py
```diff
@@ -537,3 +539,293 @@ def save_custom_metrics(
...
+        host = f"https://{host}"
+    return host.rstrip("/")
+
+
+@router.get(
+    "/embedded-dashboard",
+    response_model=EmbeddedDashboardOut,
+    operation_id="getEmbeddedDashboard",
+)
+def get_embedded_dashboard(
```
mwojtyczka, line 623: **`GET /embedded-dashboard` has no `require_role`, but the PUT/DELETE do (lines 647/686).** It returns the dashboard id + workspace host to any authenticated user including `VIEWER`. With `embed_credentials: true` set on the dashboard (databricks.yml:200), the embedded Lakeview iframe renders with the *publisher's* credentials — so the "the underlying dashboard enforces UC permissions" assumption doesn't hold, and a VIEWER can see data they lack UC grants for. Please gate the GET.

app/scripts/post_deploy_grants.sh
```diff
@@ -296,6 +308,33 @@ echo "==> Granting end-user permissions for tmp view creation..."
...
+# For SP-based deployers (CI/CD) ``current-user me`` returns the SP and
+# ``userName`` is the application ID — which is what UC expects in a GRANT.
+# For human deployers it's the email. Either way the GRANT line is identical.
+echo ""
+echo "==> Granting SELECT to deployer (for Insights dashboard queries)..."
+DEPLOYER_JSON=$($CLI current-user me -o json 2>/dev/null || echo "")
+DEPLOYER=$(echo "$DEPLOYER_JSON" | jq -r '.userName // .applicationId // empty' 2>/dev/null || echo "")
+if [[ -n "$DEPLOYER" ]]; then
+  echo "   Deployer: $DEPLOYER"
+  run_sql "GRANT USE SCHEMA ON SCHEMA \`$CATALOG\`.\`$SCHEMA\` TO \`$DEPLOYER\`"
```
mwojtyczka, line 353: **`$DEPLOYER` is interpolated into `GRANT … TO \`$DEPLOYER\`` without `validate_uc_identifier`.** `run_sql`'s own comment states the invariant that every interpolated identifier was validated by `validate_uc_identifier`/`validate_uuid` — `$DEPLOYER` (from `current-user me` JSON) bypasses that. A principal name containing a backtick escapes the quoted identifier in a statement run with deploy privileges. Route it through the same validator used for `$CATALOG`/`$SCHEMA`.

app/src/databricks_labs_dqx_app/backend/services/contract_rules_service.py
```diff
@@ -0,0 +1,551 @@
...
+        top_filter = rule.get("filter")
+        if isinstance(top_filter, str):
+            sql_fragments.append(top_filter)
+        arguments = check.get("arguments")
+        if isinstance(arguments, dict):
+            for key in cls._SQL_BEARING_KEYS:
+                value = arguments.get(key)
+                if isinstance(value, str):
+                    sql_fragments.append(value)
+        return all(is_sql_query_safe(sql) for sql in sql_fragments if sql.strip())
```
mwojtyczka, line 346: **`_is_llm_rule_safe` gates SQL-safety on a fixed key allowlist (`_SQL_BEARING_KEYS`).** Only `expression/query/row_filter/filter` args are run through `is_sql_query_safe()`. A prompt-injected ODCS `type: text` expectation can steer the LLM to emit a registry-resolved check whose SQL lives in any other argument (e.g. `column`, or a future SQL-bearing key), which then passes the gate and flows to save/execute. Consider validating by which args actually reach Spark SQL rather than by name.

app/src/databricks_labs_dqx_app/backend/routes/v1/dryrun.py
```diff
@@ -315,14 +474,19 @@ def get_dry_run_status(
...
+    requesting-user attribution is missing (older run, SDK shape drift),
+    only admins/approvers may proceed.
     """
     try:
         if job_run_id_param is not None:
             requesting_user = obo_ws.current_user.me().user_name or "unknown"
             run_owner = job_svc.get_run_creator(job_run_id_param)
-            if run_owner and run_owner != requesting_user:
-                raise HTTPException(status_code=403, detail="You can only check status of your own runs")
+            if run_owner != requesting_user:
```
mwojtyczka: **Ownership check fails closed with no escape hatch, contradicting the new docstring.** Changed from `if run_owner and run_owner != requesting_user` to `if run_owner != requesting_user`, but — unlike `cancel_dry_run` — this handler has no `can_check_others`/admin bypass, yet the docstring claims "only admins/approvers may proceed." When `get_run_creator` returns `None` (older run / SDK drift), `None != user` is True, so the legitimate owner is 403'd and the status-poll-driven view cleanup never runs. Add the admin bypass the docstring promises, or treat `None` owner as allowed-for-owner.

app/src/databricks_labs_dqx_app/backend/services/contract_rules_service.py
```diff
@@ -0,0 +1,551 @@
...
+                    )
+                    continue
+                user_metadata = {
+                    "contract_id": metadata.contract_id or "unknown",
+                    "contract_version": metadata.version or "unknown",
+                    "odcs_version": metadata.odcs_api_version or "unknown",
+                    "schema": exp.schema_name,
+                    "rule_type": "text_llm",
+                    "text_expectation": exp.description,
+                    **(rule.get("user_metadata") or {}),
```
mwojtyczka: **LLM-supplied `user_metadata` overwrites trusted contract lineage.** The dict literal sets the trusted keys (`contract_id`, `schema`, `rule_type: text_llm`) first, then spreads `**(rule.get("user_metadata") or {})` **last**, so an LLM rule carrying its own `schema`/`rule_type` wins. That corrupts `_bucket_rules_by_schema` (wrong/nonexistent bucket) and lineage, and an overwritten `schema` can route the rule to an unintended target table on save. Spread the LLM metadata first, then the trusted keys.

app/src/databricks_labs_dqx_app/backend/services/review_status_service.py
```diff
@@ -0,0 +1,366 @@
...
+            self._table,
+            key_cols={"run_id": run_id},
+            value_cols={
+                "status": cleaned_status,
+                "updated_by": user_email,
+                "updated_at": RawSql("current_timestamp()"),
+            },
+        )
+
+        self._append_history(
```
mwojtyczka, line 268: **`set_status`/`clear_status` are non-atomic: primary upsert/DELETE and the audit `_append_history` INSERT are separate auto-committed statements with no guard.** Unlike `RoleService._record_history` (best-effort try/except), here a failing history insert propagates after the status change has already committed → the endpoint returns 500 (caller retries) while the live status reflects the change and the audit row — the table's whole purpose — is missing. Wrap in a transaction, or make the history write best-effort like RoleService.

app/src/databricks_labs_dqx_app/backend/services/app_settings_service.py
```diff
@@ -199,3 +199,219 @@ def _get_int_setting(self, key: str) -> int | None:
...
+        Runs detail dropdown is never empty after a fresh deploy.
+        Returned entries are always normalised — ``value`` trimmed,
+        ``description`` defaulted to empty string, ``color`` defaulted
+        to ``"gray"``, ``is_default`` coerced to bool — so call sites
+        can index by field without defensive lookups.
+        """
+        raw = self.get_setting(self._RUN_REVIEW_STATUSES_KEY)
+        if not raw:
+            logger.info("Seeding default run_review_statuses on first read")
+            self._persist_run_review_statuses(self._RUN_REVIEW_STATUSES_SEED, user_email=None)
```
mwojtyczka: **A read performs a write: `get_run_review_statuses()` seeds + persists via `_persist_run_review_statuses` on first read.** It's invoked on the Runs listing path (a GET), so a fresh deploy turns an ordinary read into a `dq_app_settings` write as the service principal, and two concurrent first-reads both attempt the seed (relies on `save_setting` being a true idempotent upsert). It also surprises read-only health probes. Consider seeding at migration/startup time instead of lazily on read.

mwojtyczka (approved):
LGTM

PR #1275, opened 2026-06-25
Guard LLM-generated sql_query rules

### LINKED ISSUE #1123: [FEATURE]: Guard results of AI-assisted rules generation

### Problem statement

Output is not capped, and the generated `sql_query` rule is not validated against malicious SQL. 

### Proposed Solution

Add `max_tokens` to cap input length.
is_sql_query_safe() on any sql_query in the LLM output before returning

### Additional Context

_No response_

aarushisingh04:
## Changes

adds SQL safety validation to both LLM-assisted rule generation paths so that any `sql_query` rule containing unsafe SQL (DML/DDL) is dropped before being returned to the caller.

- added `_filter_unsafe_sql_rules()` in `llm/llm_core.py` : parses the generated JSON rules array and drops any rule whose `check.function` is `sql_query` and whose `query` argument fails `is_sql_query_safe()`, logging a warning with the sanitized query string (CWE-117)
- hooked it into `DspyRuleGeneration.forward()` and `DspyRuleUsingDataStats.forward()` via an `else` clause on the existing JSON-validation block; `DspyRuleGenerationWithSchemaInference` inherits the fix automatically since it delegates to `DspyRuleGeneration`
> note: `max_tokens` was already present in `LLMModelConfig` and forwarded to `dspy.LM()`, it caps completion length, not prompt input, which is sufficient for the issue's intent of bounding unbounded LLM output.
- fixed training example for ai-assisted rules generation

### Linked issues

resolves #1123

### Tests

- [ ] manually tested
- [x] added unit tests
- [ ] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

19 unit tests in `tests/unit/test_llm_core.py` covering: safe SELECT queries pass through, unsafe DROP/DELETE/INSERT/TRUNCATE rules are dropped, non-`sql_query` rules are unaffected, mixed safe+unsafe batches, null/missing `query` argument, non-array JSON input, malformed rule entries, invalid JSON passthrough, and warning log emission.

### Documentation and Demos

- [ ] added/updated demos
- [ ] added/updated docs

### DIFF
--- a/src/databricks/labs/dqx/llm/llm_core.py
+++ b/src/databricks/labs/dqx/llm/llm_core.py
@@ -7,10 +7,60 @@ import dspy  # type: ignore
 
 from databricks.labs.dqx.config import LLMModelConfig
 from databricks.labs.dqx.llm.llm_utils import create_optimizer_training_set, create_optimizer_training_set_with_stats
+from databricks.labs.dqx.utils import is_sql_query_safe
 from databricks.labs.dqx.llm.optimizers import BootstrapFewShotOptimizer
 from databricks.labs.dqx.llm.validators import RuleValidator
 
 logger = logging.getLogger(__name__)
+
+
+def _filter_unsafe_sql_rules(rules_json: str) -> str:
+    """Filter out any sql_query rules whose query argument fails the SQL safety check.
+
+    Parses the JSON rules array and drops every rule where *check.function* is
+    *sql_query* and the *query* argument does not pass *is_sql_query_safe*. All
+    other rules are returned unchanged. If the input cannot be parsed as a JSON
+    array the original string is returned so the caller's JSON-validation step
+    handles it.
+
+    Args:
+        rules_json: JSON string containing the generated rules array.
+
+    Returns:
+        JSON string with unsafe sql_query rules removed, or the original string
+        when it cannot be parsed.
+    """
+    try:
+        rules = json.loads(rules_json)
+    except (json.JSONDecodeError, ValueError):
+        return rules_json
+
+    if not isinstance(rules, list):
+        return "[]"
+
+    safe_rules = []
+    for rule in rules:
+        if not isinstance(rule, dict):
+            continue
+
+        check = rule.get("check", {})
+        if not isinstance(check, dict):
+            continue
+
+        if check.get("function") == "sql_query":
+            arguments = check.get("arguments", {})
+            if not isinstance(arguments, dict):
+                continue
+
+            query = arguments.get("query") or ""
+            if not is_sql_query_safe(query):
+                # Strip control characters before logging to prevent log injection (CWE-117)
+                safe_repr = repr(query.replace("\n", " ").replace("\r", " "))
+                logger.warning(f"LLM-generated sql_query rule dropped: unsafe SQL query {safe_repr}")
+                continue
+        safe_rules.append(rule)
+
+    return json.dumps(safe_rules)
 
 
 class LLMModelConfigurator:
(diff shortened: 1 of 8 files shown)


src/databricks/labs/dqx/llm/llm_core.py
```diff
@@ -7,12 +7,62 @@
...
+        check = rule.get("check", {})
+        if not isinstance(check, dict):
+            continue
+
+        if check.get("function") == "sql_query":
+            arguments = check.get("arguments", {})
+            if not isinstance(arguments, dict):
+                continue
+
+            query = arguments.get("query") or ""
```
mwojtyczka: **A non-string `query` will crash rule generation instead of being filtered.** `arguments.get("query") or ""` keeps a non-string truthy value as-is, and `is_sql_query_safe()` immediately does `re.sub(r"\s+", " ", query)` → `TypeError: expected string or bytes-like object` for an `int`/`list`/`dict`. This function doesn't catch it, and the caller's `try/except` only wraps `json.loads` (the `_filter_unsafe_sql_rules` call is in the `else` branch, outside the try) — so malformed LLM output like `{"query": 123}` raises an unhandled `TypeError` straight out of `forward()` and aborts generation. That's exactly the bad-LLM-output case this guard exists for, and it's the one input shape not already handled (non-dict rule/check/arguments and null/missing query all are). Suggest: `q = arguments.get("query"); if not isinstance(q, str): q = ""` (or treat non-str as unsafe and drop), plus a `query: 123` regression test. (Reproduced locally.)

src/databricks/labs/dqx/llm/llm_core.py
```diff
@@ -147,13 +197,15 @@ def forward(
...
-        # Validate JSON output
+        # Validate JSON output and filter unsafe sql_query rules
         if result.quality_rules:
             try:
                 json.loads(result.quality_rules)
             except json.JSONDecodeError as e:
                 logger.warning(f"Generated invalid JSON: {e}. Returning empty rules.")
                 result.quality_rules = "[]"
+            else:
+                result.quality_rules = _filter_unsafe_sql_rules(result.quality_rules)
```
mwojtyczka: Minor: the `try/except/else` control flow is correct (the `else` runs only when `json.loads` succeeds, which is the intent). But `result.quality_rules` is now parsed twice — once here just to validate, then again inside `_filter_unsafe_sql_rules`. You could parse once and pass the parsed `list[dict]` into the filter (having it return the filtered list, re-serialized by the caller), avoiding the redundant `json.loads`. Not a correctness issue — same remark applies to the identical block in `DspyRuleUsingDataStats.forward`

tests/unit/test_llm_core.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+def test_filter_unsafe_sql_rules_null_query_kept():
+    # null query defaults to "" (safe); engine will reject it later via MissingParameterError
+    rules = [{"check": {"function": "sql_query", "arguments": {"query": None}}}]
+    result = json.loads(_filter_unsafe_sql_rules(json.dumps(rules)))
+    assert len(result) == 1
+
+
+def test_filter_unsafe_sql_rules_missing_query_key_kept():
```
mwojtyczka: Good coverage overall (happy paths, drop/keep, malformed rule/check, null & missing `query`, invalid-JSON passthrough, plus integration for two of the generators). A few gaps worth closing, and notably they overlap the edge cases most likely to break:

1. **Non-string `query` (e.g. `{"query": 123}`)** — not tested, and it's exactly the input that currently raises an unhandled `TypeError` (see my comment on `llm_core.py:55`). Add this test together with the type guard so it's a regression test, asserting the rule is dropped (or kept as `""`) rather than crashing.
2. **`arguments` not a dict** (e.g. `"arguments": "DROP TABLE x"`) — the `if not isinstance(arguments, dict): continue` branch has no test.
3. **`DspyRuleGenerationWithSchemaInference.forward`** — the third public generation path has no test. It inherits the guard via delegation, but a test would catch a future regression in that path.
4. **Outer `if result.quality_rules` falsy case** (empty string / `None`) isn't exercised at the integration level — the "not json" test covers a non-empty invalid value, not an empty/None one.

mwojtyczka (changes requested):
Thank you for the PR. Generally looking good. Left some comments to address.

aarushisingh04:
@mwojtyczka 

appreciate the thorough review! addressed all three points in the latest push:

- added `isinstance(query, str)` guard, non-string query is now dropped with a warning instead of crashing
- refactored `_filter_unsafe_sql_rules` to accept a pre-parsed list, eliminating the double-parse
- added missing tests: non-string query, non-dict arguments, `DspyRuleGenerationWithSchemaInference` path, and empty/None `quality_rules` at the integration level

mwojtyczka (approved):
LGTM

PR #1278, opened 2026-06-27
Added databricks runtime requirements verification

### LINKED ISSUE #1225: [FEATURE]: Verify Databricks runtime environment for the checks with version requirements

### Problem statement

Some functions used the library have DBR minimal version requirements in order to be executed properly. Such as geo verifications that needs 17+ DBR for these purpose. Please, find `skip_if_runtime_not_geo_compatible` test utility function as an example. 

Additionally, the following feature proposal is based on pull request review feedback: https://github.com/databrickslabs/dqx/pull/1164#discussion_r3342788606

### Proposed Solution

Extend `def register_rule(rule_type: str) -> Callable:` core method to accept optional metadata, including DBR versions requirements.

```python
from pydantic import BaseModel

class RuleMeta(BaseModel):
   "Rule optional metadata, such as Databricks runtime environment requirements"
   dbr_min_version: int | None = None 
...

def register_rule(rule_type: str, meta: RuleMeta) -> Callable
```

So later this can be validated using https://docs.databricks.com/aws/en/sql/language-manual/functions/current_version in `DQEngineCore` at `apply_checks` method.

### Additional Context

After implementation cover geo check with additional meta.

IvannKurchenko:
## Changes
Verifies minimal Databricks environment version before checks execution. The implementation differs from the suggestion in the original ticket to preserve backward compatibility. New annotation stores additional property in the function. Additionally covered geo-checks that has DBR requirements in the description.

### Linked issues
Resolves #1225 

### Tests

- [ ] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

### Documentation and Demos

- [ ] added/updated demos
- [ ] added/updated docs

### DIFF
--- a/src/databricks/labs/dqx/engine.py
+++ b/src/databricks/labs/dqx/engine.py
@@ -11,6 +11,7 @@ from typing import Any
 from uuid import uuid4
 
 import pyspark.sql.functions as F
+from pyspark.errors import AnalysisException
 from pyspark.sql import DataFrame, Observation, SparkSession
 from pyspark.sql.streaming import StreamingQuery
 
@@ -40,6 +41,7 @@ from databricks.labs.dqx.manager import DQRuleManager
 from databricks.labs.dqx.reporting_columns import ColumnArguments, DefaultColumnNames, merge_info_columns
 from databricks.labs.dqx.rule import (
     Criticality,
+    CHECK_FUNC_DBR_VERSION_REQUIREMENT_ATTRIBUTE,
     DQRule,
     CHECK_FUNC_REGISTRY_ORIGINAL_COLUMNS_PRESELECTION,
 )
@@ -143,6 +145,8 @@ class DQEngineCore(DQEngineCoreBase):
             raise InvalidCheckError(
                 "All elements in the 'checks' list must be instances of DQRule. Use 'apply_checks_by_metadata' to pass checks as list of dicts instead."
             )
+
+        self._validate_dbr_version_requirements(checks)
 
         warning_checks = self._get_check_columns(checks, Criticality.WARN.value)
         error_checks = self._get_check_columns(checks, Criticality.ERROR.value)
@@ -390,6 +394,49 @@ class DQEngineCore(DQEngineCoreBase):
     def _all_are_dq_rules(checks: list[DQRule]) -> bool:
         """Check if all elements in the checks list are instances of DQRule."""
         return all(isinstance(check, DQRule) for check in checks)
+
+    def _validate_dbr_version_requirements(self, checks: list[DQRule]) -> None:
+        """Raise InvalidParameterError if the current DBR major version is below the highest version required by any check.
+
+        The requirement is declared by decorating a check function with *requires_dbr_version(major_version)*.
+        The current DBR version is resolved via *current_version().dbr_version* Spark SQL function.
+        Validation checks that the current major version is >= the required major version (not an exact match).
+
+        Args:
+            checks: List of DQRule instances to validate.
+
+        Raises:
+            InvalidCheckError: If the current DBR major version is below the maximum required version,
+                or if the DBR version cannot be resolved (non-Databricks environment or null result).
+        """
+        versioned = {
+            c.check_func.__name__: getattr(c.check_func, CHECK_FUNC_DBR_VERSION_REQUIREMENT_ATTRIBUTE)
+            for c in checks
+            if getattr(c.check_func, CHECK_FUNC_DBR_VERSION_REQUIREMENT_ATTRIBUTE, None) is not None
+        }
+        if not versioned:
+            return
+        required = max(versioned.values())
+        try:
+            rows = self.spark.sql("select current_version().dbr_version as dbr_version").collect()
+            dbr_version_str = rows[0]["dbr_version"] if rows else None
+        except AnalysisException as e:
+            raise InvalidCheckError(
+                "Check functions with a DBR version requirement can only run on Databricks Runtime. "
+                f"Failed to resolve the current DBR version: {e}"
+            ) from e
+        if dbr_version_str is None:
+            raise InvalidCheckError(
+                "Check functions with a DBR version requirement can only run on Databricks Runtime. "
+                "current_version().dbr_version returned null."
+            )
+        major = int(dbr_version_str.split(".")[0])
+        if major < required:
+            check_names = ", ".join(sorted(versioned))
+            raise InvalidCheckError(
+                f"Check functions [{check_names}] require Databricks Runtime >= {required}, "
+                f"but the current version is {dbr_version_str} (major: {major})."
+            )
 
     def _preselect_original_columns(self, df: DataFrame, check: DQRule) -> DQRule:
         """
--- a/src/databricks/labs/dqx/engine.py
+++ b/src/databricks/labs/dqx/engine.py
@@ -2,6 +2,7 @@ import copy
 import inspect
 import logging
 import os
+import re
 from concurrent import futures
 from collections.abc import Callable
 from dataclasses import replace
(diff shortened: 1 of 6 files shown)


src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -347,6 +355,7 @@ def is_geometrycollection(column: str | Column) -> Column:
     )
 
 
+@requires_dbr_version(17)
```
mwojtyczka: **Four geo checks that require DBR 17 are missing this annotation, so the new guard silently skips them.** I audited every `@register_rule` geo function; these did not get `@requires_dbr_version(17)` despite needing it:
- **`is_geometrycollection`** (~line 332) — its own docstring says *"requires Databricks serverless compute or runtime 17.1 or above"* and it uses `try_to_geometry`/`st_geometrytype`, same as the decorated `is_geometry`.
- **`is_geo_contains`** (~1091), **`is_geo_covers`** (~1133), **`is_geo_intersects`** (~1195) — all delegate to the same `_has_topological_relationship_approximate` helper as the decorated `is_geo_touches`/`is_geo_within`.

(`is_latitude`/`is_longitude` are correctly left undecorated — plain numeric range checks, no spatial SQL.) Suggest adding the decorator to those four, plus a test that asserts every DBR-17 geo function carries the attribute so the set can't drift again.

Separately (design call): the requirement is major-only, but these geo docstrings say **17.1** — a cluster on **17.0** passes `major < 17` yet may still fail at runtime. Consider minor-version granularity, or note the 17.0↔17.1 gap explicitly.

src/databricks/labs/dqx/engine.py
```diff
@@ -391,6 +395,49 @@ def _all_are_dq_rules(checks: list[DQRule]) -> bool:
         """Check if all elements in the checks list are instances of DQRule."""
         return all(isinstance(check, DQRule) for check in checks)
 
+    def _validate_dbr_version_requirements(self, checks: list[DQRule]) -> None:
```
mwojtyczka, line 399: **Nit:** the summary line says *"Raise InvalidParameterError…"* but the method raises `InvalidCheckError` (the `Raises:` section is correct) — align the first line.

We can also unit test this.

src/databricks/labs/dqx/engine.py
```diff
@@ -391,6 +395,49 @@ def _all_are_dq_rules(checks: list[DQRule]) -> bool:
...
+            raise InvalidCheckError(
+                "Check functions with a DBR version requirement can only run on Databricks Runtime. "
+                f"Failed to resolve the current DBR version: {e}"
+            ) from e
+        if dbr_version_str is None:
+            raise InvalidCheckError(
+                "Check functions with a DBR version requirement can only run on Databricks Runtime. "
+                "current_version().dbr_version returned null."
+            )
+        major = int(dbr_version_str.split(".")[0])
```
mwojtyczka: **Unguarded version parse.** `int(dbr_version_str.split(".")[0])` sits only behind `except AnalysisException`. If `current_version().dbr_version` ever returns a format whose major part isn't a plain integer, this raises a raw `ValueError` instead of the intended friendly `InvalidCheckError`. Wrap the parse (or validate the format) and surface `InvalidCheckError` with the offending value.

src/databricks/labs/dqx/engine.py
```diff
@@ -40,6 +41,7 @@
 from databricks.labs.dqx.reporting_columns import ColumnArguments, DefaultColumnNames, merge_info_columns
 from databricks.labs.dqx.rule import (
     Criticality,
+    CHECK_FUNC_DBR_VERSION_REQUIREMENT_ATTRIBUTE,
```
mwojtyczka: maybe rename to: 
`CHECK_FUNC_MIN_DBR_VERSION_ATTRIBUTE`

mwojtyczka (changes requested):
Thanks for the PR, generally looking good, added some comments

src/databricks/labs/dqx/geo/check_funcs.py
IvannKurchenko: for `is_geo_contains`, `is_geo_covers`, `is_geointersects` documentations adjusted because DBR 17.1 is not required for `precise=False` case, because H3 functions are used then.

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -1106,6 +1131,7 @@ def is_geo_contains(
     )
 
 
+@requires_dbr_version("17.1")
```
mwojtyczka, line 1134: **Unconditional `@requires_dbr_version("17.1")` over-blocks the `precise=False` path.** `is_geo_covers` (and `is_geo_intersects` below) take `precise: bool`, and the docstring now says 17.1 is required *"if `precise` is True"* — `precise=False` runs the H3-approximate path (`_has_topological_relationship_approximate`) that works on lower DBR. But this decorator is function-level: the engine reads the requirement off the function and can't see the `precise` argument, so it blocks the whole run below 17.1 regardless.

*Failure:* `DQRowRule(check_func=is_geo_covers, ..., precise=False, resolution=8)` on DBR 16.x → `InvalidCheckError`, blocking a check that runs fine there and contradicting the docstring just clarified. A function-level decorator can't express a per-argument condition — these two functions likely need no static gate (or per-call enforcement that inspects `precise`). (`is_geo_contains` is correctly unconditional, so its decorator is fine.)

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -1168,6 +1194,7 @@ def is_geo_covers(
     return _has_topological_relationship_approximate(column, reference_geometry, resolution, "COVERS")
 
 
+@requires_dbr_version("17.1")
```
mwojtyczka, line 1197: Same issue as `is_geo_covers` above: this unconditional `@requires_dbr_version("17.1")` blocks the `precise=False` H3 path on DBR < 17.1, contradicting this function's own "…17.1 or above if `precise` is True" docstring.

src/databricks/labs/dqx/engine.py
```diff
@@ -391,6 +395,59 @@ def _all_are_dq_rules(checks: list[DQRule]) -> bool:
...
+        try:
+            rows = self.spark.sql("select current_version().dbr_version as dbr_version").collect()
+            dbr_version_str = rows[0]["dbr_version"] if rows else None
+        except AnalysisException as e:
+            raise InvalidCheckError(
+                "Check functions with a DBR version requirement can only run on Databricks Runtime. "
+                f"Failed to resolve the current DBR version: {e}"
+            ) from e
+
+        if dbr_version_str is None:
```
mwojtyczka: **This blocks Databricks serverless, which the geo checks document as supported.** Every annotated geo docstring says *"Databricks serverless compute or runtime 17.1 or above"*, but on serverless `current_version().dbr_version` is null/empty, so this branch raises `InvalidCheckError`.

*Failure:* any geo check on serverless → blocked, even though serverless supports the spatial functions — a regression for an explicitly-supported environment. Worth confirming what `current_version().dbr_version` returns on serverless; the null/empty case probably needs to *pass* the gate rather than hard-fail. (Relatedly, the parse just below raises on suffixed strings like `'15.4 LTS'` or an empty token — same hard-fail effect on a capable runtime.)

We can have serverless and non-serverless version of the int tests. There are examples of these.

tests/integration/test_engine.py
```diff
@@ -0,0 +1,33 @@
...
+from databricks.labs.dqx.rule import DQRowRule, register_rule, requires_dbr_version
+
+
+@requires_dbr_version("1.0")
+@register_rule("row")
+def _check_requires_dbr_v1(column: str) -> Column:
+    return make_condition(F.col(column).isNull(), f"\'{column}\' is null", f"{column}_is_null")
+
+
+@requires_dbr_version("999.0")
```
mwojtyczka: **The minor-version boundary — the core of this feature — isn't tested.** These integration tests use `1.0` (always passes) and `999.0` (always fails), and the unit tests mock `spark.sql` to cover only the error branches (sql failure / null / unparseable). Nothing exercises a real `(major, minor)` boundary, e.g. required `17.1` with current `17.0` (must fail) vs `17.1`/`17.2` (must pass).

A regression that flips `<`→`<=`, or compares major-only (which the docstring's "major version (not an exact match)" wording would invite), would pass every existing test. Per AGENTS.md "Cover all changes with tests," add a boundary case.

mwojtyczka (changes requested):
Did second pass. Previous comments resolved. Left 2 additional comments.

mwojtyczka (approved):
LGTM

IvannKurchenko:
@mwojtyczka Thank you for adjustments. I'm sorry, missed your comments.

Issue #1214, opened 2026-06-01
[FEATURE]: Add error_if & warn_if settings


feliblo:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Problem statement

In dbt theres an error_if , warn_if setting: https://docs.getdbt.com/reference/resource-configs/severity?version=2.0&name=Fusion

Lets say I only want a warning if 25% or more of my column is null, but an error if 50% is null. This will now require 2 tests, which is.

1. Too much config
2. Too much compute

**Note** this has nothing to do with the `rows`. It should be a setting for each rule. Especially important for custom sql expressions.

### Proposed Solution

Add `error_if` & `warn_if` configurations for each rule. Should be either the amount of rows or the actual value.

### Additional Context
E.g this

```yaml
- name: fct_cdm_customer_order_line_mk_productid_pct_below_zero_error
  description: "error if more than 50% of mk_productid values are < 0 (null business key in source)"
  criticality: error
  check:
    function: sql_expression
    arguments:
      expression: COUNT(CASE WHEN mk_productid <= 0 THEN 1 END) OVER () * 100.0 / COUNT(*) OVER () <= 50
      msg: "More than 50% of mk_productid values are < 0"

- name: fct_cdm_customer_order_line_mk_productid_pct_below_zero_warn
  description: "warn if more than 25% of mk_productid values are < 0 (null business key in source)"
  criticality: warn
  check:
    function: sql_expression
    arguments:
      expression: COUNT(CASE WHEN mk_productid <= 0 THEN 1 END) OVER () * 100.0 / COUNT(*) OVER () <= 25
      msg: "More than 25% of mk_productid values are < 0"
```

Could be condensed into

```yaml
- name: fct_cdm_customer_order_line_mk_productid_pct_below_zero
  description: "warn/error if too many mk_productid values are < 0 (null business key in source)"
  criticality: error
  warn_if: "<=25"
  error_if: "<=50"
  check:
    function: sql_expression
    arguments:
      expression: COUNT(CASE WHEN mk_productid <= 0 THEN 1 END) OVER () * 100.0 / COUNT(*) OVER ()
      msg: "Too many mk_productid values are < 0"
```

_No response_

mwojtyczka:
Thanks for reporting this. This is a similar request to https://github.com/databrickslabs/dqx/issues/610

[...]

PR #1289, opened 2026-06-30
feat(actions): Pluggable DQ Actions & Alerting

### LINKED ISSUE #204: [FEATURE]: Deploy data quality alerts

### Problem statement

The dashboard can currently be used to track quality over time. However, there is no automatic alert triggering (e.g. if the number of errors reach a certain threshold).

### Proposed Solution

As part of the installation in the workspace using Databricks CLI deploy Databricks SQL alert for every run_config. The alert should be triggered when the number of rows on error in the (output or quarantined dataset) reaches the threshold.

The alert should be deactivated by default as it's usage should be optional

Add `errors_threshold` parameter to each run_config.

### Additional Context

_No response_

### LINKED ISSUE #610: [FEATURE]: Conditionally failing checks

### Problem statement

Currently the flagging/quarantining is done for any row that fail a check. However, there are 2 special cases which are not natively-possible today in DQX:
1. only flag/quarantine the records but only if certain amount of records fail for that check
2. flag/quarantine all rows sharing the same column value if any row fails the check for that column

### Proposed Solution

1. Flag/quarantine the records but only if certain amount of records fail for that check

Introduce `threshold_rate` and `threshold_count` to control when a particular check should fail (must reach certain rate or count for that check). If none of the thresholds are breached, no records are failed for that check.

```
criticality: warn
check:
  function: is_in_list
  threshold_rate: 0.1  # max fraction of rows allowed to fail; 10% of rows can fail; default 0
  threshold_count: 12 # max number of rows allowed to fail; fail if more than 12 records fail; default 0
  arguments:
    column: col1
    allowed:
    - 1
    - 2
```

For example, if at least 12 readings of that sensor fails, then all readings for that sensor should be failed.

2. Flag/quarantine all rows sharing the same column value if any row fails the check for that column

Introduce `fail_group_on_any` parameter.
```
criticality: warn
check:
  function: is_in_list
  fail_group_on_any: true
  arguments:
    column: col1
    allowed:
    - 1
    - 2
```

For example, if 1 reading of a sensor fails, then all readings for that sensor should be failed.

### Additional Context

Since rows are evaluated independent, this will have to be implemented as post-processing task. Therefore, it might be good to abstract a `DQAction` class that implements this behavior, saving metrics, and alerts.

mwojtyczka:
## Summary

Adds an extensible **actions** subsystem to DQX. An *action* runs when data checked by DQX violates an (optional) condition evaluated against the summary metrics produced by `DQMetricsObserver`. The concrete actions are:

- **`DQAlert`** — sends a notification to one or more destinations: **Slack**, **Microsoft Teams**, **generic HTTPS webhook**, **Log** (driver logger, no external system) or an in-process **callback**.
- **`FailPipeline`** — raises `PipelineFailedError` to stop the current pipeline run.
- **`NoOpAction`** — fires with no side effect (records the event only). Serves as an observe-only / dry-run mode and as a side-effect-free building block for testing action semantics.

Actions can be defined programmatically (DQX classes) or declaratively as **metadata** (YAML/JSON dicts), and are passed to `DQEngine(ws, observer=..., actions=[...])` — they fire automatically on the save-to-table methods (**batch and streaming**), or explicitly via `engine.evaluate_actions(...)`. Action definitions can be stored/loaded from UC or Lakebase tables (or local YAML/JSON files) via `DQActionManager`, and action events can be persisted to a UC/Lakebase events table so frequency/status-change suppression survives engine restarts. In the installed DQX Workflows and the parallel multi-run-config runner, each `RunConfig` can point at its own `actions_location` (definitions) and `action_events_location` (history) so actions are auto-loaded and applied per run config.

## What's included

- **New package `databricks.labs.dqx.actions`** — `DQAction`, `Action` ABC, `DQAlert`, `FailPipeline`, `NoOpAction`; `AlertDestination` hierarchy (Slack / Teams / webhook / **log** / callback); a **safe AST condition evaluator**; standard message builder; webhook delivery with retry/backoff and an **SSRF guard**; `SecretResolver` (`DQSecret`); `ActionStateStore` + UC/Lakebase event stores; `ActionEvaluator` orchestrator; `ActionSerializer` + UC/Lakebase definition storage; `DQActionManager`. Destinations and actions are Pydantic discriminated unions, so adding a new type is a small, isolated change (OCP).
- **Engine integration** — `actions=` on `DQEngine`/`DQEngineCore` (accepts `DQAction` instances or raw metadata dicts), batch + streaming firing, `evaluate_actions(...)`, and optional `action_events_config` for persistent state/history.
- Engine-level user_metadata (ExtraParams.user_metadata) is included in every alert.
- **Metadata (YAML/JSON) API** — actions can be authored as dicts/YAML with the same wire format as checks; `DQActionManager.load_actions_from_local_file` / `save_actions_in_local_file` round-trip them to disk.
- **Workflow & run-config integration** — `DQEngine.apply_checks_and_save_in_tables` (and the installed quality-checker workflow) auto-load each `RunConfig.actions_location` and fire those actions for that run. Each run config is applied through a dedicated engine (its own actions, a fresh observer, and an optional event store), keeping the parallel runner thread-safe. `action_events_location` persists event history and durable alert suppression across runs.
- **Config** — `DQSecret`, `TableActionsStorageConfig`, `LakebaseActionsStorageConfig`, `ActionEventsConfig`, `RunConfig.actions_location` (definitions) and `RunConfig.action_events_location` (event history).
- **Demo** — `demos/dqx_demo_alerting.py` demonstrating alerting via the log destination (with optional Slack), wired into the e2e demo runner.
- Fixed logging to use the standard blueprint approach, but avoid reconfiguring root logger. This is to fix the current logging bug where logs are not properly printed. Previous fix was wrong: https://github.com/databrickslabs/dqx/pull/1376

### Output from real environment

**Slack**:
Screenshot 2026-08-03 at 15 15 40
Screenshot 2026-07-31 at 13 53 17

**Teams**:
Screenshot 2026-08-08 at 00 16 11

## Known follow-ups (out of scope here)

- **Fully asynchronous off-thread streaming webhook delivery — not needed for practical use.** Streaming action evaluation runs on Spark's `StreamingQueryListener` thread (not the micro-batch data path), so synchronous delivery blocks only that callback, and alerts are rate-limited (`HOURLY`/`DAILY`) and fire only on threshold crossings — the documented small webhook timeout/retry budget keeps the callback fast. Off-thread delivery would only help high-throughput streaming that fires alerts frequently against slow endpoints (where a slow callback could let Spark drop progress events). Documented; deferred unless such a use case appears.
- **`FailPipeline` in streaming is best-effort.** It aborts immediately on the batch path, but in streaming it is raised inside the listener callback and Spark generally isolates/logs listener exceptions rather than stopping the query — so it is not guaranteed to hard-stop a streaming query. Documented; a guaranteed streaming stop (e.g. signalling `query.stop()` from the listener) is a possible follow-up that needs live verification.
- Email notification: https://github.com/databrickslabs/dqx/issues/1400
- Secrets are sourced from the classic secret scope, not using UC secrets. This will be done as a follow up to support UC Secrets in addition. Follow up issue: https://github.com/databrickslabs/dqx/issues/1420
- Teams destinations support standard webhooks integration workflow for teams that require an anonymously-callable Workflows webhook (the default for the "Post to a channel when a webhook request is received" template). A Workflows trigger explicitly configured to require Entra authentication is not supported: DQX sends only the signed URL and does not attach an OAuth bearer token (the delivery layer currently carries HTTP Basic auth only, and minting/refreshing an Entra token is out of scope). The modern Teams Workflows template generates the anonymous SAS (sig=) URL by default. Entra-OAuth on the trigger is an org that deliberately configures it at the Logic App / Easy Auth level — a minority. So this serves a real but very uncommon case; the implemented support for anonymous URL covers the overwhelming majority.

### Linked issues

Resolves #204 #610 (related but not solving it #1214) 

### Tests

- [x] manually tested: webhook: https://play.hook0.com/, real slack and teams, log in notebook
- [x] added unit tests
- [x] added integration tests
- [x] added e2e tests

### Documentation and Demos

- [x] added/updated demos
- [x] added/updated docs


docs/dqx/docs/guide/actions_and_alerts.mdx
```diff
@@ -0,0 +1,1039 @@
...
+sidebar_position: 311
+---
+
+import Admonition from '@theme/Admonition';
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+# Actions and Alerting
+
+DQX **actions** let you react automatically when observed summary metrics cross a threshold. After checks are applied and the Spark observation is materialized, each action evaluates its optional *condition* against the observed metrics and, when the condition is satisfied (or when there is no condition), executes its inner logic.
```
ghanse: Might be helpful to add a small introductory note about when to use actions. Something like:


```suggestion
DQX **actions** let you react automatically when observed summary metrics cross a threshold. Actions can be used to notify stakeholders or fail a pipeline when data quality issues are detected.

After checks are applied and the Spark observation is materialized, each action evaluates its optional *condition* against the observed metrics and, when the condition is satisfied (or when there is no condition), executes its inner logic.
```

docs/dqx/docs/guide/actions_and_alerts.mdx
```diff
@@ -0,0 +1,1039 @@
...
+
+The built-in action types are:
+
+- *DQAlert* — delivers a notification to one or more destinations (Slack, Microsoft Teams, generic HTTPS webhook, or an in-process callback).
```
ghanse: Should we mention logging here as well?

docs/dqx/docs/guide/actions_and_alerts.mdx
```diff
@@ -0,0 +1,1039 @@
...
+    metrics_config=OutputConfig(location="catalog.schema.metrics"),
+)
+```
+
+  </TabItem>
+</Tabs>
+
+<Admonition type="warning" title="Observer required">
+Passing *actions* to *DQEngine* without also passing an *observer* raises `InvalidParameterError` at construction time. Actions evaluate metrics produced by the observer and cannot function without one.
+</Admonition>
```
ghanse, line 150: I think this section is combining 2 concepts that might be clearer if split:

1. Defining an action when initializing `DQEngine`
2. Checking data with end-to-end methods to generate an action

I would suggest adding a small section like `### Configuring actions` before this one:

```
### Configuring actions

To run actions, attach them to a `DQEngine` together with a `DQMetricsObserver`. 
`DQEngine` accepts `DQAction` instances, raw metadata dicts, or a list loaded from 
a file with `DQActionManager` — and mixed lists of instances and dicts are supported.

... CODE BLOCK ...
```

docs/dqx/docs/guide/actions_and_alerts.mdx
```diff
@@ -0,0 +1,1039 @@
...
+synchronously in the save call). In **streaming**, `FailPipeline` is raised inside the listener
+callback, and Spark generally isolates and logs listener exceptions rather than stopping the query —
+so do not rely on `FailPipeline` to hard-stop a streaming query; use a downstream gate if you need a
+guaranteed stop.
+
+In practice, alerts are rate-limited (`HOURLY` / `DAILY`) and typically fire only when a condition
+crosses a threshold, so the callback rarely blocks and the synchronous path is sufficient. Fully
+asynchronous off-thread delivery is a possible future enhancement, mainly for high-throughput
+streaming that fires alerts frequently against slow endpoints.
+</Admonition>
```
ghanse, line 194: Because the condition is optional, I think we should suggest that users specify a condition or frequency for continuous streaming data.

docs/dqx/docs/guide/actions_and_alerts.mdx
```diff
@@ -0,0 +1,1039 @@
...
+            destinations=[SlackDQAlertDestination(name="slack", webhook_url="https://hooks.slack.com/…")]
+        ),
+    )
+])
+
+# Apply checks — returns the DataFrame and an observation handle
+checked_df, observation = engine.apply_checks_by_metadata(input_df, checks)
+
+# Trigger a Spark action to materialize the observation
+checked_df.count()  # or write to a table, etc.
```
ghanse, line 228: It might be useful to show practical examples of which Spark actions to trigger for specific behavior. 

These use cases come to mind:

1. For notifications prefer `write`, then `evaluate_actions` to avoid an unnecessary `count`.
2. To fail the pipeline, use `count`, then `evaluate_actions` to fail before writing bad data.

docs/dqx/docs/guide/actions_and_alerts.mdx
```diff
@@ -0,0 +1,1039 @@
...
+
+Every action is wrapped in a `DQAction` that binds it to an optional [condition](#condition-syntax-and-available-metrics); the action runs when the condition is met (or when no condition is set). DQX ships three built-in action types, each documented in its own section below:
+
+- [DQAlert](#dqalert) — send a notification to one or more destinations.
+- [FailPipeline](#failpipeline) — abort the current run by raising `PipelineFailedError`.
+- [NoOpAction](#noopaction-dry-run--audit) — fire without any side effect (testing / dry-run / audit).
+
+### DQAlert
+
+`DQAlert` delivers a notification to one or more destinations (Slack, Microsoft Teams, generic HTTPS webhook, log, or an in-process callback). A single alert message is dispatched concurrently to every configured destination, and a delivery failure in one destination is isolated — the others still receive the alert, and per-destination errors are captured in `ActionResult.destination_errors`. The available destinations are listed under [Alert types](#alert-types), and [Controlling when alerts are sent](#controlling-when-alerts-are-sent) covers rate-limiting and status-change-only delivery.
```
ghanse: Maybe we should make it explicit that these are the Spark driver logs.

docs/dqx/docs/guide/actions_and_alerts.mdx
```diff
@@ -0,0 +1,1039 @@
...
+      - type: slack
+        name: slack
+        webhook_url: https://hooks.slack.com/…
+  condition: error_row_count > 0
+```
+
+  </TabItem>
+</Tabs>
+
+### FailPipeline
```
ghanse, line 426: Do we need a note or warning about how this will behave for streaming? The listener and query evaluation are independent, so we can't guarantee any immediate stopping.

docs/dqx/docs/guide/actions_and_alerts.mdx
```diff
@@ -0,0 +1,1039 @@
...
+        password:
+          secret: dq-secrets/pd-token
+```
+
+  </TabItem>
+</Tabs>
+
+### Log (driver logger)
+
+`LogDQAlertDestination` writes the alert to the standard Python logger on the Spark driver instead of contacting an external system. Because it performs no network I/O it never triggers SSRF validation and cannot stall a streaming micro-batch, which makes it the ideal destination for local development, demos, and end-to-end tests. Unlike the callback destination it is fully serializable, so it has a `type: log` metadata form and can be persisted. The optional `level` controls the log level (`debug`, `info`, `warning`, `error`, or `critical`; defaults to `warning`).
```
ghanse: I think the part about serializability might be unnecessary here?

```suggestion
`LogDQAlertDestination` writes the alert to the standard Python logger on the Spark driver instead of contacting an external system. Because logging performs no network I/O, it never triggers SSRF validation and cannot stall a streaming micro-batch. This makes it an ideal destination for local development, demos, and end-to-end tests. The optional `level` controls the log level (`debug`, `info`, `warning`, `error`, or `critical`; defaults to `warning`).
```

ghanse (commented):
Leaving some feedback on the docs. Still reviewing the code/tests.

src/databricks/labs/dqx/actions/conditions.py
```diff
@@ -0,0 +1,521 @@
...
+    return cast(ast.Constant, node).value
+
+
+def _eval_name(node: ast.Name, metrics: dict[str, object] | None) -> object:
+    """Evaluate a *ast.Name* node against *metrics*.
+
+    Returns a placeholder when *metrics* is *None* (validate-only mode).
+    """
+    if metrics is None:
+        return 0  # placeholder that satisfies downstream arithmetic
```
ghanse: Will this fail conditions with a divisor (e.g. `error_row_count / input_row_count < 0.1`) in validate mode due to division by zero?

src/databricks/labs/dqx/actions/event_storage.py
```diff
@@ -0,0 +1,438 @@
...
+        result: dict[str, AlertEvent] = {}
+        with engine.connect() as conn:
+            rows = conn.execute(stmt).mappings().all()
+
+        for row in rows:
+            action_name = row["action_name"]
+            if action_name in result:
+                continue  # already have the most recent (ordered by run_time desc)
+            if row["run_time"] is None:
+                continue  # defensive: skip rows with a missing timestamp (e.g. externally-inserted data)
```
ghanse: Consider moving these conditions into the `stmt` to avoid a few issues:

1. Collecting unnecessary rows into the calling Python session
2. Iterating over rows that will be skipped

docs/dqx/docs/guide/actions_and_alerts.mdx
mwojtyczka: true

mwojtyczka:
Thanks for the thorough docs pass, @ghanse — all reasonable. Plan for each:

1. **Intro note** — adopting your suggested two-paragraph framing (adds the "notify stakeholders / fail a pipeline" motivation).
2. **Mention logging in the type list** — yes, adding the **log (driver logger)** destination to the `DQAlert` bullet.
3. **Split out `### Configuring actions`** — agreed, promoting the attach-to-`DQEngine` prose into its own section before "Defining and applying actions."
4. **Streaming condition/frequency** — adding a recommendation that streaming actions set a `condition` and/or `alert_frequency` so continuous data doesn't fire on every micro-batch.
5. **Which Spark action to trigger** — adding your two use-cases: `write` → `evaluate_actions` for notifications (avoids an extra `count`); `count` → `evaluate_actions` to fail before writing bad data.
6. **Spark driver logs** — making it explicit these are the **Spark driver** logs.
7. **FailPipeline in streaming** — adding a note/cross-reference in the FailPipeline section (the streaming admonition already explains the listener/query independence, so I'll link to it rather than duplicate).
8. **Log serializability** — trimming as you suggest; I'll keep one short clause explaining that (unlike the callback destination) log is serializable and therefore has a `type: log` metadata form, since that answers a common "why can I persist this one?" question. Shout if you'd rather drop it entirely.

Will push the doc updates shortly.

src/databricks/labs/dqx/actions/state.py
```diff
@@ -0,0 +1,383 @@
...
+            *True* if the action should execute this run (and the slot is now reserved); else *False*.
+        """
+        with self._lock:
+            if not self._should_fire_locked(dq_action, context, condition_result):
+                return False
+            # Reserve the slot so a concurrent try_fire for the same alert is suppressed before this
+            # thread's execute()/record() completes. Only DQAlert is frequency-gated.
+            if isinstance(dq_action.action, DQAlert):
+                self._last_fired[dq_action.name] = context.run_time
+            return True
```
ghanse, line 266: Should this also set `self._last_status[dq_action.name]` so that status-change rules are not duplicated?

src/databricks/labs/dqx/actions/evaluator.py
```diff
@@ -0,0 +1,223 @@
...
+                    # on, so a later critical action (FailPipeline, another alert) still runs instead of
+                    # the whole loop aborting.
+                    logger.warning(
+                        f"Action '{safe_name}' skipped: condition '{safe_condition}' could not be "
+                        f"evaluated: {_sanitize(str(exc))}"
+                    )
+                    self._state_store.record(
+                        self._build_event(dq_action, context, fired=False, status=ActionStatus.UNHEALTHY)
+                    )
+                    continue
```
ghanse, line 125: Should we consider a separate `ActionStatus` for this type of behavior? The data is not really unhealthy, it is more a failure of the alert/condition configuration.

src/databricks/labs/dqx/actions/destinations/webhook.py
```diff
@@ -0,0 +1,95 @@
...
+        """
+        return {
+            "title": message.title,
+            "summary": message.summary,
+            "condition": message.condition,
+            "table": message.table,
+            "run_id": message.run_id,
+            "run_time": message.run_time.isoformat(),
+            "severity": message.severity,
+            "observed_metrics": dict(message.observed_metrics),
```
ghanse, line 116: Do we need to normalize the `message.observed_metrics` values? Custom metrics might have some unexpected types.

src/databricks/labs/dqx/actions/destinations/slack.py
```diff
@@ -0,0 +1,92 @@
...
+            Block Kit block array.
+        """
+        condition_text = message.condition if message.condition is not None else "unconditional"
+        table_text = message.table if message.table is not None else "unspecified"
+        metrics_text = ", ".join(f"{k}: {v}" for k, v in message.observed_metrics.items())
+
+        blocks: list[dict] = [
+            {
+                "type": "header",
+                "text": {"type": "plain_text", "text": message.title, "emoji": True},
```
ghanse, line 61: I think we need some methods to ensure these payload arguments comply. These could error at instantiation or warn and trim text.

For example, [Section](https://docs.slack.dev/reference/block-kit/blocks/section-block/#fields) is capped at 3000 characters, and [Header](https://docs.slack.dev/reference/block-kit/blocks/header-block#fields) is capped at 150 characters.

docs/dqx/docs/guide/actions_and_alerts.mdx
mwojtyczka: added

docs/dqx/docs/guide/actions_and_alerts.mdx
mwojtyczka: added

docs/dqx/docs/guide/actions_and_alerts.mdx
mwojtyczka: added

docs/dqx/docs/guide/actions_and_alerts.mdx
mwojtyczka: added

docs/dqx/docs/guide/actions_and_alerts.mdx
mwojtyczka: fixed

docs/dqx/docs/guide/actions_and_alerts.mdx
mwojtyczka: added

mwojtyczka:
> Thanks for the thorough docs pass, @ghanse — all reasonable. Plan for each:
> 
> 1. **Intro note** — adopting your suggested two-paragraph framing (adds the "notify stakeholders / fail a pipeline" motivation).
> 2. **Mention logging in the type list** — yes, adding the **log (driver logger)** destination to the `DQAlert` bullet.
> 3. **Split out `### Configuring actions`** — agreed, promoting the attach-to-`DQEngine` prose into its own section before "Defining and applying actions."
> 4. **Streaming condition/frequency** — adding a recommendation that streaming actions set a `condition` and/or `alert_frequency` so continuous data doesn't fire on every micro-batch.
> 5. **Which Spark action to trigger** — adding your two use-cases: `write` → `evaluate_actions` for notifications (avoids an extra `count`); `count` → `evaluate_actions` to fail before writing bad data.
> 6. **Spark driver logs** — making it explicit these are the **Spark driver** logs.
> 7. **FailPipeline in streaming** — adding a note/cross-reference in the FailPipeline section (the streaming admonition already explains the listener/query independence, so I'll link to it rather than duplicate).
> 8. **Log serializability** — trimming as you suggest; I'll keep one short clause explaining that (unlike the callback destination) log is serializable and therefore has a `type: log` metadata form, since that answers a common "why can I persist this one?" question. Shout if you'd rather drop it entirely.
> 
> Will push the doc updates shortly.

All updates pushed

src/databricks/labs/dqx/actions/destinations/webhook.py
```diff
@@ -0,0 +1,95 @@
...
+
+        Args:
+            services: Injected services containing the secret resolver.
+
+        Returns:
+            A *WebhookAuth* instance when both credentials are present, or
+            *None* when either is absent.
+        """
+        if self.username is None or self.password is None:
+            return None
```
ghanse, line 92: This will silently fallback to no authentication either `username` or `password` are provided individually. We should probably catch this.

src/databricks/labs/dqx/actions/destinations/slack.py
```diff
@@ -0,0 +1,92 @@
...
+        the app-based one is the supported path. The URL carries its own token, so no additional
+        authentication is sent.
+
+    Class attributes:
+        allowed_host_suffixes: Restricts delivery to *hooks.slack.com*.
+
+    Attributes:
+        type: Discriminator literal, always *"slack"*.
+        name: Logical name for this destination instance.
+        webhook_url: The Slack incoming webhook URL (plain string or *DQSecret*).
```
ghanse, line 33: Slack webhook URLs are contain a token ([docs](https://docs.slack.dev/messaging/sending-messages-using-incoming-webhooks/#create_a_webhook)). We should probably require `DQSecret` here?

src/databricks/labs/dqx/actions/destinations/teams.py
```diff
@@ -0,0 +1,83 @@
...
+        currently supported.
+
+    Class attributes:
+        allowed_host_suffixes: Restricts delivery to the Workflows webhook hosts
+            *logic.azure.com* and *environment.api.powerplatform.com*.
+
+    Attributes:
+        type: Discriminator literal, always *"teams"*.
+        name: Logical name for this destination instance.
+        webhook_url: The Teams Workflows webhook URL (plain string or *DQSecret*).
```
ghanse, line 52: I believe the Teams webhook URL also contains sensitive info.

src/databricks/labs/dqx/actions/destinations/teams.py
```diff
@@ -0,0 +1,83 @@
...
+    """
+
+    type: Literal["teams"] = "teams"
+    allowed_host_suffixes: ClassVar[list[str] | None] = ["logic.azure.com", "environment.api.powerplatform.com"]
```
ghanse, line 63: `"environment.api.powerplatform.com"` is noted in the docstring as requiring an Entra bearer token which  we do not send. Should this be allowed?

src/databricks/labs/dqx/config.py
```diff
@@ -29,9 +29,62 @@
...
+        Returns:
+            A new *DQSecret* instance.
+
+        Raises:
+            InvalidParameterError: If *ref* does not contain a ``/``, or if
+                either the scope or key part is empty.
+        """
+        if "/" not in ref:
+            raise InvalidParameterError(f"Secret reference must be in the form 'scope/key', got {ref!r}.")
+        scope, _, key = ref.partition("/")
```
ghanse:
```suggestion
        raw_scope, _, raw_key = ref.partition("/")
        scope = raw_scope.strip()
        key = raw_key.strip()
```

src/databricks/labs/dqx/engine.py
```diff
@@ -790,6 +844,178 @@ def apply_checks_by_metadata_and_split(
...
+            checks_location: Path/URI of the checks file, or *None*.
+            rule_set_fingerprint: Fingerprint of the applied rule set, or *None*.
+        """
+        # Actions-only path: when actions are configured but metrics_config is absent, the Spark
+        # Observation has not yet been triggered (no output/quarantine write was done, and
+        # _populate_batch_observation only fires count() when metrics_config is set).  The two
+        # count() triggers are therefore mutually exclusive — this branch handles the complementary
+        # actions-only path to ensure the observation is populated before evaluate_actions reads it.
+        if self._actions and batch_observation is not None and metrics_config is None and metrics_only_df is not None:
+            metrics_only_df.count()
```
ghanse, line 1012: Can we ever have this condition? I think `metrics_only_df` is created only when no output or quarantine table are requested. In this case we should always have a `metrics_config`?

src/databricks/labs/dqx/engine.py
```diff
@@ -1671,6 +1959,143 @@ def _apply_checks_for_run_config(self, run_config: RunConfig) -> None:
...
+                instance_name=run_config.lakebase_instance_name,
+                client_id=run_config.lakebase_client_id,
+                port=run_config.lakebase_port or "5432",
+                run_config_name=run_config.name,
+            )
+        if not is_table_location(location):
+            raise InvalidConfigError(
+                "action_events_location must be a Unity Catalog table (catalog.schema.table) or a Lakebase table; "
+                "file locations are not supported for action events."
+            )
```
ghanse: Should these be re-ordered to check if the location is a table first?

```suggestion
        if not is_table_location(location):
            raise InvalidConfigError(
                "action_events_location must be a Unity Catalog table (catalog.schema.table) or a Lakebase table; "
                "file locations are not supported for action events."
            )
        if run_config.lakebase_instance_name:
            return LakebaseActionsStorageConfig(
                location=location,
                instance_name=run_config.lakebase_instance_name,
                client_id=run_config.lakebase_client_id,
                port=run_config.lakebase_port or "5432",
                run_config_name=run_config.name,
            )
```

src/databricks/labs/dqx/actions/event_storage.py
```diff
@@ -0,0 +1,438 @@
...
+        values are coerced to *str* before serialization into JSONB.
+
+        Args:
+            events: One or more *AlertEvent* records to persist.
+        """
+        if not events:
+            return
+
+        engine = self._get_engine()
+        self._bootstrap(engine)
```
ghanse: Do we need to call this each time the user calls `append`? Maybe run it once and cache that the storage has been bootstrapped?

src/databricks/labs/dqx/actions/event_storage.py
```diff
@@ -0,0 +1,438 @@
...
+        last_fired: dict[str, datetime] = {}
+        with engine.connect() as conn:
+            rows = conn.execute(stmt).mappings().all()
+        for row in rows:
+            action_name = row["action_name"]
+            if row["run_time"] is None:
+                continue  # defensive: skip rows with a missing timestamp (e.g. externally-inserted data)
+            if action_name not in last_fired:  # rows are run_time-descending, so first seen is latest
+                last_fired[action_name] = to_utc(row["run_time"])
+        return last_fired
```
ghanse, line 404: Similar pattern here. Can we offload some of the filtering to the database instead of bringing back all the rows and filtering in the Python session?

src/databricks/labs/dqx/actions/definition_storage.py
```diff
@@ -0,0 +1,481 @@
...
+
+        df = self._spark.read.table(config.location).filter(F.col("run_config_name") == config.run_config_name)
+
+        result: list[DQAction] = []
+        for row in df.collect():
+            try:
+                action_dict = json.loads(row["action_json"])
+                result.append(ActionSerializer.from_dict(action_dict))
+            except Exception as exc:  # broad catch: isolate deserialization errors
+                logger.warning(f"Failed to deserialize action from table '{safe_location}': {_sanitize(str(exc))}")
```
ghanse: I can see how we wouldn't want to disrupt the ongoing pipeline if we can't deserialize the action.

On the other hand, there's a risk that if an action can't be loaded, we won't fail and the pipeline won't behave as intended. For example, a pipeline previously configured to fail might start processing bad data.

ghanse (commented):
Implementation looks very good overall. Left some comments.

src/databricks/labs/dqx/actions/definition_storage.py
mwojtyczka: I would lean toward fail-loud (raise) since silently dropping a FailPipeline action is dangerous.

src/databricks/labs/dqx/actions/conditions.py
mwojtyczka: fixed

src/databricks/labs/dqx/actions/event_storage.py
mwojtyczka: fixed

src/databricks/labs/dqx/actions/state.py
mwojtyczka: fixed

src/databricks/labs/dqx/actions/evaluator.py
mwojtyczka: Yes added

src/databricks/labs/dqx/actions/destinations/webhook.py
mwojtyczka: added

src/databricks/labs/dqx/actions/destinations/slack.py
mwojtyczka: fixed

src/databricks/labs/dqx/actions/destinations/webhook.py
mwojtyczka: fixed

src/databricks/labs/dqx/actions/destinations/slack.py
mwojtyczka: added as documentation info, for testing it's still good to be able to define it as plain text

src/databricks/labs/dqx/actions/destinations/teams.py
mwojtyczka: fixed

src/databricks/labs/dqx/actions/destinations/teams.py
mwojtyczka: fixed

src/databricks/labs/dqx/config.py
mwojtyczka: fixed

src/databricks/labs/dqx/engine.py
mwojtyczka: fixed

src/databricks/labs/dqx/engine.py

mwojtyczka:
@ghanse — thanks for the detailed code/test review. Addressed the feedback in `f75169ec`. Summary of what changed:

**Correctness / safety**
- **Webhook Basic-auth** now fails fast at construction if exactly one of `username`/`password` is set (previously it silently sent an unauthenticated request).
- **Slack/Teams webhook URLs** that embed a token now warn when supplied as a plaintext string, nudging toward `DQSecret` (plaintext still allowed for local dev, per the thread).
- **Teams** `environment.api.powerplatform.com` removed from the allowlist — it requires an Entra bearer token we don't send, so it could never work; only `logic.azure.com` remains.
- **Action deserialization** now fails loud (raises `InvalidActionError`) instead of silently dropping an undeserializable stored action — a silently-dropped `FailPipeline` could let bad data through. Applied to both the UC-table and Lakebase loaders.
- **Condition evaluation errors** at runtime (e.g. unknown metric, divide-by-zero on live values) are now isolated to the offending action and recorded as a new `ActionStatus.CONFIG_ERROR` (not `UNHEALTHY`, so they don't seed status-change suppression); other actions still run.
- **Divisor conditions** like `error_row_count / input_row_count < 0.1` no longer fail validation (validate-mode now uses a non-zero placeholder).
- **`try_fire`** stamps `_last_status` alongside `_last_fired` so a concurrent `STATUS_CHANGE` evaluation can't double-fire.
- **`DQSecret` scope/key** are now stripped of surrounding whitespace.

**Design**
- **Event-store location validation** moved onto the config (`ActionEventsConfig` / Lakebase configs validate at construction) instead of a guard in the engine, so the table-vs-file rule holds regardless of call path. The table/file predicate (`is_table_location`) is now a single source of truth in `config.py`, re-exported by `io.py`/`checks_storage.py` (previously duplicated).

**Efficiency**
- Fixed the Lakebase event-store row filtering and per-append bootstrap caching

**Tests**: added coverage for every behavior change above (auth-pair rejection, plaintext-URL warning, fail-loud deserialization on both backends, CONFIG_ERROR path, divisor-condition validation, `try_fire` status-change suppression, DQSecret whitespace, and file/extension/2-part location validation).

src/databricks/labs/dqx/actions/event_storage.py
mwojtyczka: fixed, cache

src/databricks/labs/dqx/actions/event_storage.py
mwojtyczka: fixed

berrybluecode:
## Follow-up: custom alert message content + rule `user_metadata` (e.g. owner)

### Question
Can callers customise the alert message and include a rule’s `user_metadata` (e.g. rules owner) in the Slack / Teams / webhook payload?

### Current behaviour
**Not supported as-is via config/templates.**

- Alert text is fixed by `StandardMessageBuilder` (`title` / `summary` from action name, table, condition, metrics only) — see `actions/message.py`.
- `DQAlert` has no message-template fields (`name`, `destinations`, `alert_frequency`, `notify_on`, `severity` only) — see `actions/alert.py`.
- Rule-level `user_metadata` is written into row-level `_errors` / `_warnings` result structs, **not** into summary metrics or the alert message path.
- `ActionContext.user_metadata` is **engine-level** `ExtraParams.user_metadata` only, and even that is not passed into `StandardMessageBuilder` today (Slack/Teams/webhook/log ignore it).
- `check_metrics` exposes per-check `check_name` / `error_count` / `warning_count`, but **not** rule fingerprints or `user_metadata`.

### Workarounds today
1. **Custom action** (`@register_action`) — build your own notification payload in `execute`.
2. **`DQCallbackAlertDestination`** — read `message` + `context` and post a custom payload (not persistable / no YAML form).
3. Neither automatically surfaces **per-rule** owners without also joining failed checks back to the check definitions (and duplicate check names make name-only joins ambiguous).

### Suggested product follow-up
1. Pass engine-level `context.user_metadata` into `StandardMessageBuilder` / `AlertMessage.fields` (quick win).
2. Optional `DQAlert` config: e.g. `title_template` / `summary_template` / `include_user_metadata` (or `user_metadata_keys`).
3. For **per-rule owners**: enrich `check_metrics` (or alert context) with failing-rule `rule_fingerprint` + `user_metadata`, then allow templates such as `{owners}` — fingerprints matter because check names can collide.

mwojtyczka:
> ## Follow-up: custom alert message content + rule `user_metadata` (e.g. owner)
> ### Question
> Can callers customise the alert message and include a rule’s `user_metadata` (e.g. rules owner) in the Slack / Teams / webhook payload?
> 
> ### Current behaviour
> **Not supported as-is via config/templates.**
> 
> * Alert text is fixed by `StandardMessageBuilder` (`title` / `summary` from action name, table, condition, metrics only) — see `actions/message.py`.
> * `DQAlert` has no message-template fields (`name`, `destinations`, `alert_frequency`, `notify_on`, `severity` only) — see `actions/alert.py`.
> * Rule-level `user_metadata` is written into row-level `_errors` / `_warnings` result structs, **not** into summary metrics or the alert message path.
> * `ActionContext.user_metadata` is **engine-level** `ExtraParams.user_metadata` only, and even that is not passed into `StandardMessageBuilder` today (Slack/Teams/webhook/log ignore it).
> * `check_metrics` exposes per-check `check_name` / `error_count` / `warning_count`, but **not** rule fingerprints or `user_metadata`.
> 
> ### Workarounds today
> 1. **Custom action** (`@register_action`) — build your own notification payload in `execute`.
> 2. **`DQCallbackAlertDestination`** — read `message` + `context` and post a custom payload (not persistable / no YAML form).
> 3. Neither automatically surfaces **per-rule** owners without also joining failed checks back to the check definitions (and duplicate check names make name-only joins ambiguous).
> 
> ### Suggested product follow-up
> 1. Pass engine-level `context.user_metadata` into `StandardMessageBuilder` / `AlertMessage.fields` (quick win).
> 2. Optional `DQAlert` config: e.g. `title_template` / `summary_template` / `include_user_metadata` (or `user_metadata_keys`).
> 3. For **per-rule owners**: enrich `check_metrics` (or alert context) with failing-rule `rule_fingerprint` + `user_metadata`, then allow templates such as `{owners}` — fingerprints matter because check names can collide.

1. I added engine level user metadata as suggested. 
2. Regarding customizable alert message template: The general platform direciton is to not provide this kind of flexiblity. Can you please provide more info why is this needed? this would require a template for each destination. I would rather extend the current standard message than provide customization for this. Please let us know what is missing from the standard message and we can add it. 
 To send a fully custom payload, use a custom `@register_action` action or a `DQCallbackAlertDestination`, which receive the `AlertMessage` and run-time context and can build any payload.
3. The user metadata and fingerprint have to be added to the summary metrics first (out of scope for this PR). I logged issue for this here: https://github.com/databrickslabs/dqx/issues/1409. Once this is fixed, the user metadata field and fingerprint will appear automatically as part of the metrics field.

berrybluecode:
LGTM

berrybluecode (approved):

PR #1300, opened 2026-07-03
Simplify dqx-studio deployment

laurencewells:
## Changes

This pull request removes legacy deployment steps and updates documentation and configuration to reflect a streamlined, natively declarative Databricks bundle deploy process. The changes eliminate the need for manual post-deploy grant scripts and one-time resource binding, clarify the deployment workflow, and update references to the latest resource types and Python version.

**Deployment workflow simplification and documentation updates:**

* Removed the `app-bind` and `app-grant-permissions` Makefile targets and all related documentation, as Unity Catalog grants are now handled natively by the Databricks bundle during deploy. There is no longer a post-deploy grants script or a need to bind pre-existing resources; only a one-time manual `USE CATALOG` grant is required per catalog. [[1]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L209-L286) [[2]](diffhunk://#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9L38-R38) [[3]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L331-R275)
* Updated documentation in `CLAUDE.md` and `CUSTOMER_QA.md` to reflect the new resource types (`postgres_projects` instead of `database_instances`) and clarified the Lakebase deployment and privilege model. [[1]](diffhunk://#diff-cb82e5c2bd201ba350a51a9ba43c9ccd760d7de7d64f37e39728880e0207fabeL106-R112) [[2]](diffhunk://#diff-cb82e5c2bd201ba350a51a9ba43c9ccd760d7de7d64f37e39728880e0207fabeL88-R88) [[3]](diffhunk://#diff-85c88a1c61aa4c2fe444cbb78d2e66c8ef9877a25633bc1d5f66a4e52a7aa090L44-R44)
* Updated the backend stack to Python 3.12+ in documentation.

**Build and dependency updates:**

* Simplified the `app-build` Makefile target to only run the frontend and backend build steps, removing wheel builds from this step.
* Cleaned up `app/.build-constraints.txt` by removing dependencies related to `uv-dynamic-versioning` and updating `trove-classifiers` to the latest version. [[1]](diffhunk://#diff-31e72afadd39cbdf95fa0b79b1cdec02b5388c9b05decb48ac24ac089a63f2c1L1-R7) [[2]](diffhunk://#diff-31e72afadd39cbdf95fa0b79b1cdec02b5388c9b05decb48ac24ac089a63f2c1L118-L128)

**Developer experience and configuration:**

* Added a `.gitignore` rule for untracked deploy target files to avoid accidental commits of real deployment configurations.

These changes make the deployment process more robust, less error-prone, and easier for new developers to understand.


app/src/databricks_labs_dqx_app/backend/pg_executor.py
```diff
@@ -94,14 +93,18 @@ def token(self, value: str) -> None:
...
+def _generate_token(ws: WorkspaceClient, endpoint: str) -> str:
+    """Generate a fresh Lakebase OAuth token (1-hour TTL) for *endpoint*.
+
+    Endpoint-scoped credential generation (the Lakebase *projects* model):
+    *endpoint* is a full resource path, e.g.
+    ``projects/dqx-studio-db/branches/dev/endpoints/primary``. This is the
+    same value passed to :meth:`WorkspaceClient.postgres.get_endpoint`, so
+    host resolution and credential issuance stay in lockstep.
+    """
+    cred = ws.postgres.generate_database_credential(endpoint=endpoint)
```
mwojtyczka, line 105: **Blocking: `ws.postgres.generate_database_credential(...)` does not exist in the pinned SDK.**

In databricks-sdk 0.120.0 (what `app/uv.lock` resolves), `generate_database_credential` lives on `ws.database`, not `ws.postgres` — `PostgresAPI` has no such method. Its real signature also takes `instance_names` / `request_id` / `claims`, with no `endpoint` kwarg.

*Failure:* `AttributeError: 'PostgresAPI' object has no attribute 'generate_database_credential'` on the first token generation (called from `PgExecutor.__init__` and the refresh loop) — every Lakebase deployment fails to start. The unit tests pass only because they use `MagicMock`, which fabricates the attribute.

Worth confirming which SDK version this projects-model credential API actually ships in — `~=0.120` may be too old (or the wrong service).

app/src/databricks_labs_dqx_app/backend/pg_executor.py
```diff
@@ -762,18 +777,22 @@ def build_pg_executor(
...
+    user locally) before opening the pool. The token-refresh tuning
     kwargs are passed straight through to :class:`PgExecutor` — see
     its :meth:`__init__` and :meth:`_token_refresh_loop` for the
     full back-off / escalation contract.
     """
-    instance = ws.database.get_database_instance(name=instance_name)
-    host = instance.read_write_dns
+    endpoint_obj = ws.postgres.get_endpoint(name=endpoint)
+    status = endpoint_obj.status
+    host = status.hosts.host if status and status.hosts else None
```
mwojtyczka, line 791: **Blocking: wrong attribute path `status.hosts.host`.**

The SDK's `EndpointStatus` exposes the hostname directly as `.host`; there is no `.hosts` sub-object. The guard `status.hosts.host if status and status.hosts` does not save you — evaluating `status.hosts` raises before the boolean can short-circuit.

*Failure:* `AttributeError: 'EndpointStatus' object has no attribute 'hosts'` in `build_pg_executor`. Correct path is `status.host`. Masked by `MagicMock` in the tests.

Consider adding a test that asserts the real SDK attribute paths exist (e.g. `getattr` against the SDK model types) so a mock can't hide a nonexistent API again.

app/databricks.yml
```diff
@@ -236,8 +346,10 @@ resources:
           spec:
             client: "5"
             dependencies:
+              # Local task-runner wheel (not published) + the library pinned
+              # from the registry — same version as the app, no local dqx build.
               - ./.build/tasks/databricks_labs_dqx_task_runner-*.whl
-              - ./.build/databricks_labs_dqx-*.whl
+              - databricks-labs-dqx==0.15.0
```
mwojtyczka, line 352: **Third DQX version pin, uncovered by the lockstep test.**

`TestDqxVersionLockstep` in `test_lint_policy.py` only compares `app/pyproject.toml` and `app/tasks/pyproject.toml`; this task-runner job dependency (`databricks-labs-dqx==0.15.0`) is not checked.

*Failure:* on the 0.16.0 release, if both pyprojects are bumped but this line is missed, `make app-test` stays green while the serverless task-runner installs stale 0.15.0 — the exact author-time/run-time engine skew the lockstep test claims to prevent, shipped silently. Consider extending the test to cover this pin too.

app/tests/test_lint_policy.py
```diff
@@ -435,3 +446,58 @@ def test_migrations_postgres_does_not_transitively_import_psycopg(self):
...
+                return match.group("version")
+        raise AssertionError(
+            f"{pyproject_path.relative_to(REPO_ROOT.parent)} has no exact "
+            "``databricks-labs-dqx==<version>`` pin in [project].dependencies. "
+            "The app and task-runner must both pin the engine to the SAME "
+            "exact version (see the lockstep comments in both files); a "
+            "range or unpinned dep breaks that guarantee."
+        )
+
+    def test_app_and_task_runner_pin_the_same_dqx_version(self):
```
mwojtyczka, line 492: **No pin is tied to the library's actual version.**

The lockstep test enforces app == task-runner, but neither is checked against `src/databricks/labs/dqx/__about__.py` (0.15.0), and the app's own `version` was switched from dynamic to a static `0.15.0`.

*Failure:* the two pins can sit in lockstep at a version that no longer matches the repo (or a nonexistent one) with no CI signal until a deploy-time pip resolution failure. On each release, five spots now must move together (`__about__.py`, two pyproject pins, `databricks.yml` task-runner dep, and the static app version) with nothing enforcing it. Consider asserting the pins equal `__about__.__version__`.

mwojtyczka (commented):
Automated review (medium effort). The `instance_name` -> `endpoint` rename is clean and the removed grant/bind/wheel-align machinery leaves no dangling references. However, the Lakebase SDK migration has two runtime bugs that the unit tests can't catch because they mock `WorkspaceClient`, plus a couple of version-pin maintenance gaps. Findings inline.

mwojtyczka (changes requested):
Generally looking good. Left a few comments that needs fixing

app/databricks.yml
laurencewells: This is expected as we may trail the core engine

app/tests/test_lint_policy.py
laurencewells: This is expected as we may trail the core engine

app/src/databricks_labs_dqx_app/backend/pg_executor.py
laurencewells: Verified against the pinned SDK (databricks-sdk 0.120.0). This code uses the projects-model ws.postgres API, not the instance-model ws.database API. ws.postgres.generate_database_credential(endpoint=...) exists (service/postgres.py:3986) with endpoint as a real kwarg, and EndpointStatus.hosts.host is the correct path (postgres.py:1563 → EndpointHosts.host at 1345). Both APIs coexist in 0.120.0. Confirmed working against a live Lakebase deployment.

app/src/databricks_labs_dqx_app/backend/pg_executor.py
laurencewells: Verified against the pinned SDK (databricks-sdk 0.120.0). This code uses the projects-model ws.postgres API, not the instance-model ws.database API. ws.postgres.generate_database_credential(endpoint=...) exists (service/postgres.py:3986) with endpoint as a real kwarg, and EndpointStatus.hosts.host is the correct path (postgres.py:1563 → EndpointHosts.host at 1345). Both APIs coexist in 0.120.0. Confirmed working against a live Lakebase deployment.

mwojtyczka (approved):
LGTM, can be merged as soon as all steps are Green

ghanse (approved):
LGTM

PR #1301, opened 2026-07-03
Add support for summary metrics in Lakeflow Declarative Pipelines (LDP) / DLT

### LINKED ISSUE #947: [BUG]: Materialize metrics in data streaming lakeflow pipelines

### Current Behavior

Thank you for DQX development and a comprehensive ambition on Data Quality in Databricks.
I implemented DQX framework successfully in Lakehouse pipelines (e.g., Data quality controls, quarantine work flawlessly), but I don't seem to succeed in materializing DQ results via "observer" and listeners as described in documentation (see https://databrickslabs.github.io/dqx/docs/guide/summary_metrics/#writing-metrics-to-a-table-with-streaming). Screenshot of working pipeline implementation:

Image

Tested both with serverless and defined job cluster.

Relevant pipeline code is attached below for bronze->silver transformation, I hope to have implemented correctly observer/listener metrics pattern:

```python
# Import modules
from datetime import datetime
from pyspark import pipelines as dp
from pyspark.sql.types import StructType, TimestampType
from databricks.labs.dqx.config import InputConfig, OutputConfig
from utilities import conf, utils

# parameters
catalog_name = spark.conf.get(
    "catalog_name", default="XXX"
)
metrics_table = "monitoring.dqx_summary_metrics"

# Define silver tables
def define_dq_view(table_name: str, schema: StructType, data_quality):
    source_table = f"bronze.{table_name}"
    dq_view = f"bronze.{table_name}_dq"
    select_cols = [field.name for field in schema.fields]

    @dp.table(
        name=dq_view,
        temporary=True,
        comment=f"Data Quality (DQX) view for table `{source_table}`"
    )
    def _dq_view():
        df = spark.readStream.table(source_table).select(select_cols)

        # Extract timestamp field names
        timestamp_fields = (field.name
                            for field in schema.fields
                            if isinstance(field.dataType, TimestampType))
        for field_name in timestamp_fields:
            if field_name in df.columns:
                df = df.withColumn(field_name, utils.safe_to_timestamp(field_name))

        df, observer = conf.dq_engine.apply_checks_by_metadata(df, data_quality)

        return df
    
    return _dq_view

def define_silver(table_name: str, table_comment: str, schema: StructType):
    source_table = f"bronze.{table_name}"
    dq_view = f"bronze.{table_name}_dq"

    @dp.table(
        name=f"silver.{table_name}",
        comment=table_comment,
        schema=schema
    )
    def _silver():
        df = (
            spark.readStream.table(dq_view)
        )
        # get rows without errors or warnings, and drop auxiliary columns
        return conf.dq_engine.get_valid(df)

    @dp.table(
        name=f"silver.{table_name}_quarantine",
        comment=f"Data Quality (DQX) quarantine for table `{source_table}`"
    )
    def _silver_quarantine():
        df = (
            spark.readStream.table(dq_view)
        )
        # get only rows with errors or warnings
        return conf.dq_engine.get_invalid(df)

    return _silver, _silver_quarantine

# Create pipeline
for data_entity in conf.get_all_data_entities():
  data_entity_conf = conf.get_data_entity_conf(data_entity)
  name = data_entity_conf["metadata"]["table"]["name"]

  globals()[f"silver_{name}_dq"] = define_dq_view(
    table_name=name,
    schema=data_entity_conf["schemas"]["silver"],
    data_quality = data_entity_conf["data_quality"]
  )
    
  globals()[f"silver_{name}"], globals()[f"silver_{name}_quarantine"] = define_silver(
    table_name=name,
    table_comment=data_entity_conf["metadata"]["table"]["comment"],
    schema=data_entity_conf["schemas"]["silver"],
  )
  
  # Write per-microbatch metrics to a Delta table  
  listener = conf.dq_engine.get_streaming_metrics_listener(
    input_config=InputConfig(location=f"bronze.{name}", is_streaming=True),
    output_config=OutputConfig(location=f"silver.{name}"),
    quarantine_config=OutputConfig(location=f"silver.{name}_quarantine"),
    metrics_config=OutputConfig(location=metrics_table, mode="append"),
  )
  
  # add once per session; safe to call if not already present
  spark.streams.addListener(listener)
```

DQ Engine is initialized as such in a separate config file:
```python
observer = DQMetricsObserver(name="dq_metrics_ingestion")
dq_engine = DQEngine(WorkspaceClient(), observer=observer)
```

### Expected Behavior

Metrics table materialized in configured destination `monitoring.dqx_summary_metrics` at the end of pipeline run.

### Steps To Reproduce

1. Define data streaming pipeline via lakeflow APIs
2. Configure metrics materialization
3. Run pipeline and wait till end of successful execution
4. Check for metrics persistence in table (none found)

### Cloud

Azure

### Operating System

Windows

### Relevant log output

```shell

```

### LINKED ISSUE #1291: [FEATURE]: `DQMetricsObserve` is incompatible with Spark Declarative Pipelines (SDP)

### Problem statement

`DQMetricsObserver` relies on Spark’s `DataFrame.observe()` mechanism, which requires an explicit action e.g.,(`.count (), write`) to populate metrics. In Spark Declarative Pipelines (SDP formerly DLT), the user's `@dlt.table` function returns a DataFrame - DLT itself triggers the write action after the function returns. This means `observation.get` is inaccessible within the pipeline function and will stall if called there. 

### Proposed Solution

A supported pattern for collecting DQX metrics within an SDP pipeline or documented guidance for deriving metrics from the quality columns (`_errors`, `_warnings`) as a downstream `@dlt.table`

Essentially, I want to be able to do this example from https://databrickslabs.github.io/dqx/docs/guide/summary_metrics/#accessing-metrics-when-applying-checks in Spark Declarative Pipeline. 

```python
from databricks.labs.dqx.metrics_observer import DQMetricsObserver

# Create observer
observer = DQMetricsObserver(name="dq_metrics")

# Create the engine with the optional observer
engine = DQEngine(WorkspaceClient(), observer=observer)

# Apply checks and get metrics
checked_df, observation = engine.apply_checks_by_metadata(df, checks)

# Apply checks, split and get metrics
#valid_df, quarantine_df, observation = engine.apply_checks_by_metadata_and_split(df, checks)

# Trigger an action to populate metrics (e.g., count, save to a table).
# Without triggering an action, metrics will not be populated, and accessing them will result in a stall.
row_count = checked_df.count()

# Access metrics
metrics = observation.get
print(f"Input row count: {metrics['input_row_count']}")
print(f"Error row count: {metrics['error_row_count']}")
print(f"Warning row count: {metrics['warning_row_count']}")
print(f"Valid row count: {metrics['valid_row_count']}")
print(f"Check metrics: {metrics['check_metrics']}")  # per-check error/warning counts as JSON
```

### Additional Context

_No response_

mwojtyczka:
## Changes

Adds aggregation-based data quality summary metrics that work inside Spark Declarative Pipelines (SDP / Lakeflow / DLT), where the existing `observe()`-based observer and the streaming metrics listener cannot be used because the pipeline runtime — not the caller — owns the write action.

- **New `DQEngine.compute_summary_metrics(...)`** — computes the same metrics as the observer path (row counts, per-check breakdown, and observer custom metrics) but as a lazy aggregation over the result columns, returning a DataFrame with the `OBSERVATION_TABLE_SCHEMA`. This can back a materialized view over the checked table inside a pipeline. It uses the engine's `DQMetricsObserver` (including its `custom_metrics`) and raises `InvalidParameterError` when no observer is configured.
- **`DQEngineCore._observe_metrics` now detects a pipeline context** (via `is_dlt_pipeline`) and skips wiring `observe()` there, returning the DataFrame unchanged. This keeps `return engine.apply_checks*(df, checks)` usable directly as a `@dlt.table` definition even with an observer configured; metrics are computed downstream by `compute_summary_metrics`.
- **`DQMetricsObserver.build_metrics_df_from_aggregation`** — reshapes the one-row wide aggregation into the long `OBSERVATION_TABLE_SCHEMA` without triggering an action; robust to dotted custom-metric aliases.
- Docs updated (`summary_metrics.mdx`, `quality_checks_apply.mdx`, `demos.mdx`) and DLT demos added.
- Updated demos to show how to use summary metrics for annotation and quarantine scenarios.
- **Lakeflow materialized-view (default) demos** — `dqx_dlt_demo.py` (report-as-columns) and `dqx_dlt_demo_quarantine.py` (quarantine), applying checks in `@dlt.table` datasets with summary metrics computed as a materialized view. These remain the simple, recommended default.
- **Lakeflow foreachBatch demos (new)** — `dqx_dlt_demo_foreach_batch.py` (report-as-columns) and `dqx_dlt_demo_foreach_batch_quarantine.py` (quarantine), applying checks inside a Lakeflow [foreachBatch sink](https://docs.databricks.com/aws/en/ldp/for-each-batch) (`@dp.foreach_batch_sink`) with arbitrary per-batch writes and per-batch summary metrics. Use this pattern when the write needs custom logic (MERGE/upsert, multiple destinations) a declarative dataset cannot express.
- **When to use which** — added a "materialized view vs foreachBatch" comparison (pros/cons, write model, metrics behavior) in `quality_checks_apply.mdx` under a new "Option 3: foreachBatch sink" section.
- All four Lakeflow demos are wired into the `test_run_dqx_dlt_demo` e2e parametrization.
- AGENTS.md: strengthened the testing guidance on not accessing private/protected members from tests

Quarantine scenario using MV:
Screenshot 2026-07-03 at 18 35 02

Annotation scenario using MV:
Screenshot 2026-07-03 at 18 35 04

Quarantine scenario using foreach batch sink:
Screenshot 2026-07-28 at 16 42 32

Annotation scenario using foreach batch sink:
Screenshot 2026-07-28 at 16 42 34

### Linked issues

Resolves #1291
Resolves #947

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests
- [x] added end-to-end tests
- [ ] added performance tests

### Documentation and Demos

- [x] added/updated demos
- [x] added/updated docs


demos/dqx_dlt_demo.py
```diff
@@ -109,26 +108,31 @@ def bronze():
...
-# get only rows with errors or warnings
+# Summary Metrics: materialized view computed by aggregation over the silver table. 
+# One row per metric (input / error / warning / valid row counts and per-check breakdown).
 @dlt.table
-def quarantine():
-  df = dlt.read_stream("bronze_dq_check")
-  return dq_engine.get_invalid(df)
\ No newline at end of file
+def dq_summary_metrics():
+  df = dlt.read("silver")
```
ghanse: This defines an MV on the entire silver table with all of its historical data and does not scope the metrics to the latest run or micro-batch. Should we show a `foreachBatch` sink that reads data with the streaming APIs instead? This seems more scalable.

demos/dqx_dlt_demo_quarantine.py
```diff
@@ -0,0 +1,154 @@
...
+  df = dlt.read_stream("bronze_dq_check")
+  return dq_engine.get_invalid(df)
+
+# COMMAND ----------
+
+# Summary Metrics: materialized view computed by aggregation over the checked table. 
+# One row per metric (input / error / warning / valid row counts and per-check breakdown).
+@dlt.table
+def dq_summary_metrics():
+  df = dlt.read("bronze_dq_check")
```
ghanse: Same applies here.

docs/dqx/docs/guide/quality_checks_apply.mdx
```diff
@@ -1019,6 +1019,53 @@ To enable summary metrics programmatically, create and pass a `DQMetricsObserver
   </TabItem>
 </Tabs>
 
+### Enabling summary metrics in Lakeflow Pipelines
+
+Inside a Lakeflow Pipeline (Spark Declarative Pipeline / DLT), do the following:
+
+1. Persist the checked data as a table (so the result columns are available downstream).
+2. Add a materialized view that calls `compute_summary_metrics` on that table, passing the same `checks` you applied.
```
ghanse: An MV will recompute metrics over the entire checked table. There are 2 possible challenges with this:
1. It might not be scalable for large tables where data is checked incrementally
2. MV will not have per-run or per-micro-batch quality results, the metrics will be recomputed every time

Should we instruct users to use a foreachBatch sink in these scenarios?

docs/dqx/docs/guide/summary_metrics.mdx
```diff
@@ -306,6 +309,85 @@ Metrics-only streaming writes are not supported because DQX needs an output or q
...
+observer = DQMetricsObserver()  # pass check names to get_metrics(check_names) to also include the per-check breakdown
+
+@dlt.table
+def dq_summary_metrics():
+    df = dlt.read_stream("silver").withWatermark("event_time", "10 minutes")
+    metric_exprs = [F.expr(m) for m in observer.get_metrics()]
+    return df.groupBy(F.window("event_time", "1 hour")).agg(*metric_exprs)
+```
+
+This needs a timestamp column to window on (an event or ingestion time) and a watermark; each window's metrics are emitted only once the watermark passes it, and very late rows are dropped. The output carries a `window` column plus the metric columns (it is not the long `OBSERVATION_TABLE_SCHEMA` shape) — so it is **not** schema-compatible with the companion-job output below, and the two cannot be appended into the same metrics table. Pick one history strategy per table.
```
ghanse: ForeachBatch sink may be useful to avoid recomputing the results for large tables.

ghanse (commented):
Overall LGTM, left a few comments regarding documentation and usage patterns.

demos/dqx_dlt_demo.py
mwojtyczka: You are right, although `foreachBatch` is more complex and is for streaming pipelines only. I guess we should document all 3 approaches:

- MV — simple, fully managed, broad runtime support, gives current-state snapshot (scalable when incrementalizable). Default for a current-quality view and the simplest starting point.
- foreach_batch sink — recommended for large/streaming/incremental tables and when you want per-run/per-batch history; needs a streaming source, a recent runtime, and an idempotent write.
- windowed streaming table — per-time-window history.

As for the demo, I'm not sure, maybe we can have a separate one with foreach_batch? the MV one will definitely be simpler.

docs/dqx/docs/guide/quality_checks_apply.mdx
mwojtyczka: Will add additional example showcasing the usage of foreachBatch sink

docs/dqx/docs/guide/summary_metrics.mdx
mwojtyczka: Will add additional example showcasing the usage of foreachBatch sink

demos/dqx_dlt_demo.py
mwojtyczka: Will add additional example showcasing the usage of for_each_batch sink

mwojtyczka:
@ghanse added foreach batch sink in addition. Seems like overall the best solution.

demos/dqx_dlt_demo.py
mwojtyczka: added

demos/dqx_dlt_demo_quarantine.py
mwojtyczka: added

docs/dqx/docs/guide/quality_checks_apply.mdx
mwojtyczka: added

demos/dqx_dlt_demo.py
```diff
@@ -109,26 +113,37 @@ def bronze():
...
 
-# get only rows with errors or warnings
+# Summary Metrics: materialized view computed by aggregation over the silver table.
+# One row per metric (input / error / warning / valid row counts and per-check breakdown).
+# Note: this MV is a cumulative snapshot over the whole table (input_row_count is the running
+# total, not a per-run count). It refreshes incrementally only when the query is deterministic —
+# set static run_time_overwrite / run_id_overwrite in ExtraParams for that.
+# To keep a per-batch history of metrics instead of a cumulative snapshot — and to avoid re-aggregating
+# the whole table on each update (potentially more performant on large or growing tables) — compute the
+# metrics inside a foreachBatch sink over each micro-batch instead; see `dqx_dlt_demo_foreach_batch.py`.
```
ghanse: Maybe lift this into a markdown cell?

demos/dqx_ldp_demo.py
ghanse, line 1: I am thinking it would be nice to have a demo with SDP where we set the `run_time_overwrite` and `run_id_overwrite` for checking MVs similar to [this pattern](https://databrickslabs.github.io/dqx/docs/guide/quality_checks_apply/#applying-checks-in-lakeflow-pipelines) in our docs.

src/databricks/labs/dqx/engine.py
```diff
@@ -631,6 +634,19 @@ def _observe_metrics(
...
+        # triggers the write, so an attached observe() never has an accessible result (observation.get
+        # stalls, the streaming listener receives no events). Skip wiring observe() there: apply_checks*
+        # returns the DataFrame unchanged (no tuple, no wasted/inaccessible observation), and metrics are
+        # instead computed by DQEngine.compute_summary_metrics over the checked table. The engine's
+        # observer (incl. its custom_metrics) is still used by that method.
+        if is_dlt_pipeline(self.spark):
+            logger.info(
+                "Spark Declarative Pipeline detected: observe()-based summary metrics are disabled. "
+                "Compute metrics with DQEngine.compute_summary_metrics(...) in a materialized view instead."
+            )
```
ghanse: `warning` may be a more appropriate level for this. We are skipping the intended behavior when the user instantiates the engine with a metrics observer.

src/databricks/labs/dqx/engine.py
```diff
@@ -1517,6 +1533,116 @@ def save_checks(
...
+                "Configure a DQMetricsObserver on the engine, e.g. DQEngine(workspace_client, observer=DQMetricsObserver(...))."
+            )
+
+        check_names: list[str] | None = None
+        rule_set_fingerprint: str | None = None
+        if checks:
+            rules = deserialize_checks(checks, custom_check_functions)
+            # Duplicate check names are preserved so check_metrics reports each occurrence separately.
+            check_names = [rule.name for rule in rules]
+            rule_set_fingerprint = compute_rule_set_fingerprint(rules)
```
ghanse, line 1673: Will we not get a `rule_set_fingerprint` if the user passes the checks location?

src/databricks/labs/dqx/engine.py
```diff
@@ -1517,6 +1533,116 @@ def save_checks(
...
+
+        aggregated_df = checked_df.selectExpr(*observer.get_metrics(check_names))
```
ghanse, line 1675: `selectExpr` will fail if the user passes an input `checked_df` without the DQX metadata columns. Should we handle this error gracefully?

src/databricks/labs/dqx/engine.py
```diff
@@ -1517,6 +1533,116 @@ def save_checks(
...
+            quarantine_config: Optional quarantine configuration recorded for traceability.
+            checks_location: Optional checks location recorded for traceability.
+            rule_set_fingerprint: Optional SHA-256 fingerprint of the rule set used for this run.
+
+        Returns:
+            A *DQMetricsObservation* populated from the engine's run id, run time, result column names, and metadata.
+        """
+        return DQMetricsObservation(
+            run_id=self._engine.run_id,
+            run_name=self._engine.observer.name if self._engine.observer else DQMetricsObserver().name,
```
ghanse: I think `self._engine.observer` can only be `None` if this method is called from `save_results_in_table`. In other call sites, we validate that an observer is present.

Not sure we should instantiate a new observer for the `run_name` here. I think it may be misleading. Maybe default to an empty string or `None`?

src/databricks/labs/dqx/metrics_observer.py
```diff
@@ -211,3 +211,77 @@ def build_metrics_df(spark: SparkSession, observation: DQMetricsObservation) ->
...
+        metric_pairs = F.array(
+            *[
+                F.struct(
+                    F.lit(metric_names[i]).alias("metric_name"),
+                    renamed[f"metric_{i}"].cast("string").alias("metric_value"),
+                )
+                for i in range(len(metric_names))
+            ]
+        )
+        long_df = renamed.select(F.explode(metric_pairs).alias("metric"))
```
ghanse, line 263: I think we should validate that `aggregated_df` contains only 1 row. It's documented in the docstring but not explicitly checked.

ghanse (changes requested):
Left a few comments.

src/databricks/labs/dqx/engine.py
mwojtyczka: done

src/databricks/labs/dqx/engine.py
mwojtyczka: done

src/databricks/labs/dqx/engine.py
mwojtyczka: fixed

src/databricks/labs/dqx/engine.py

demos/dqx_ldp_demo.py
mwojtyczka: fixed

demos/dqx_dlt_demo.py
mwojtyczka: fixed

src/databricks/labs/dqx/metrics_observer.py
mwojtyczka: I decided **not** to add a runtime row-count check, and instead made the guarantee explicit in the docstring — reasoning:

- The single-row property is **guaranteed by construction**: the only caller (`DQEngine.compute_summary_metrics`) feeds this an aggregation from `observer.get_metrics(...)` selected with **no `groupBy`**, which always yields exactly one row. There's no code path that can hand it a multi-row frame.
- A literal `aggregated_df.count() != 1` check would **trigger a Spark action**, which defeats this method's whole purpose: it's intentionally lazy so its result can back a materialized view in a Spark Declarative Pipeline, where the pipeline runtime — not us — owns the write. Forcing a `count()` here would break that (and double-scan on the eager path).

So I strengthened the docstring to state the invariant and why it isn't enforced at runtime, rather than adding a check that would regress the lazy behavior:

```
The single-row property is guaranteed by construction — the only caller
(DQEngine.compute_summary_metrics) feeds get_metrics aggregates selected without groupBy,
which always yield exactly one row — and is intentionally not enforced with a runtime row-count
check: counting the rows would trigger a Spark action and defeat this method's whole purpose of
staying lazy so it can back a Spark Declarative Pipeline materialized view.
```

ghanse: Makes sense

ghanse (approved):
LGTM

PR #1302, opened 2026-07-03
Feature/pydantic serialization

### LINKED ISSUE #1285: [FEATURE]: Use Pydantic for validation and serialization

### Problem statement

Rules validation and serialization is done with bespoke methods

### Proposed Solution

Use pydantic for validation and serialization/deserialization of rules. 

### Additional Context

This is a follow up of https://github.com/databrickslabs/dqx/pull/1272

fedeflowers:
## Changes

Follow-up to #1259, completing the Pydantic migration for the declarative checks path.
Also addresses the review comment on #1259 about covering Pydantic `ValidationError`
behaviour with tests: DQX callers never see raw `pydantic_core.ValidationError` —
every entry point raises DQX error types with the pre-migration message format.

**Validation & deserialization through Pydantic:**
- New `CheckSpec` / `CheckBlock` Pydantic schemas in `checks_validator.py` own the
  structural validation of check metadata dicts (required `check` key, criticality enum,
  `for_each_column` non-empty list, argument dict shape). Signature-based argument
  validation (function exists, required params, types) is unchanged and runs after.
- Pydantic errors are translated back into the pre-existing human-readable messages
  (`_translate_pydantic_error`), so users and existing tests see identical error strings.
  Unmatched errors name the offending field path (e.g. `user_metadata -> confidence`)
  instead of a generic message. Translation matches on stable Pydantic error *types*,
  not version-dependent message text.
- `ChecksDeserializer.deserialize` reads typed attributes off the validated `CheckSpec`
  instead of raw-dict `.get()` calls, so schema, defaults, and validation live in one place.
  Note: it re-parses via `model_validate` after `validate_checks` — structural parsing is
  cheap, and having the validator return parsed specs would be a larger API change.
- New `project_to_check_schema` strips storage-only columns (`run_config_name`,
  `created_at`, `rule_fingerprint`, `rule_set_fingerprint`) when loading checks from
  Lakebase, so stored checks round-trip cleanly into `apply_checks_by_metadata` and match
  the Delta path shape. The allow-list is derived from `CheckSpec.model_fields`.

**`extra` policy on the migrated models (deliberate split):**
- `extra="forbid"` on `DQRule`/`DQForEachColRule`/`BaseChecksStorageConfig`: a misspelled
  kwarg raises `InvalidCheckError`/`InvalidConfigError` at construction. This *restores
  released behaviour* — v0.15.0 (last release) predates the Pydantic migration and its
  dataclasses raised `TypeError` on unknown kwargs; the lenient `ignore` default was never
  shipped.
- `extra="ignore"` on `CheckSpec`/`CheckBlock`: check *metadata* legitimately carries
  extra keys (storage columns, user files), and rejecting them would break the
  load -> apply round-trip.

**Cleanup:**
- Removed dead `checks_formats.py` (+ its tests): `FILE_SERIALIZERS`/`FILE_DESERIALIZERS`
  had no consumers anywhere in `src/` (already unused on `main`); `SerializerFactory` is
  the single source of truth for file formats.

### Linked issues

Resolves #1285

### Tests

- [x] manually tested
- [ ] added unit tests
- [ ] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

New `tests/unit/test_pydantic_serialization.py` covers:
- **Fingerprint byte-parity corpus**: SHA-256 values captured *before* the migration,
  asserted byte-identical after — stored fingerprints in existing tables stay valid.
- Every translated Pydantic error path (missing `check`, bad criticality value *and* type,
  non-list/empty `for_each_column`, wrong-typed fields reported with their location).
- YAML/JSON round-trip incl. the `__decimal__` wire format; deserializer dispatch to
  `DQRowRule`/`DQDatasetRule`/`for_each_column`; storage-column tolerance + projection.
- Regression tests for `extra="forbid"` (unknown kwargs rejected, wrapped in DQX errors).


src/databricks/labs/dqx/checks_validator.py
```diff
@@ -5,14 +5,75 @@
...
+    definitions and would fail the load -> apply round-trip for stored checks.
+    """
+
+    model_config = ConfigDict(extra="ignore")
+
+    check: CheckBlock
+    criticality: str = Criticality.ERROR.value
+    name: str | None = None
+    filter: str | None = None
+    user_metadata: dict[str, str] | None = None
```
mwojtyczka: **Narrowed validation on a previously-unchecked field.** `user_metadata: dict[str, str] | None` now rejects non-string metadata values. The pre-migration validator never inspected `user_metadata`, and Pydantic v2 won't coerce `int`/`float`/`bool`/nested values to `str` — so a check carrying e.g. `user_metadata: {"confidence": 0.95}` that loaded fine before now raises `InvalidCheckError`. The built-in datacontract generators use string values, but user-authored or other programmatic metadata may not. Consider widening (e.g. `dict[str, Any]`) or coercing to preserve backward compatibility. **(most important finding — a behavior regression.)**

src/databricks/labs/dqx/checks_serializer.py
```diff
@@ -253,21 +273,23 @@ def deserialize(self, checks: list[dict]) -> list[DQRule]:
         for check_def in checks:
             logger.debug(f"Processing check definition: {check_def}")
 
-            check = check_def.get("check", {})
-            name = check_def.get("name") or ""
-            func_name = check.get("function")
+            # Parse through the already-validated Pydantic schema to avoid re-reading raw dicts.
+            # model_validate is safe here because ChecksValidator already confirmed the structure.
+            spec = CheckSpec.model_validate(check_def)
```
mwojtyczka: **Double validation.** By the time `deserialize` runs, `ChecksValidator.validate_checks` has already built a `CheckSpec` per check via `_pydantic_validate_structure`; here each check is parsed through `CheckSpec.model_validate` again. That's two parse passes over the same data (scaling with rule-set size) and two validation entry points that can drift apart. Consider having validation return the parsed specs so `deserialize` reuses them instead of re-parsing.

src/databricks/labs/dqx/checks_validator.py
```diff
@@ -86,60 +160,149 @@ def validate_checks(
...
+            # Cannot resolve the function; return all structural errors without signature validation.
+            return sibling_errors + blocking_errors
+
+        # The 'check' block is well-formed; run signature validation and merge sibling-field errors.
+        return sibling_errors + ChecksValidator._validate_check_block(
+            check, custom_check_functions, validate_custom_check_functions
+        )
+
+    @staticmethod
+    def _pydantic_validate_structure(check: dict) -> tuple[list[str], list[str]]:
```
mwojtyczka: **The validated `CheckSpec` is discarded, then the raw dict is re-read.** `_pydantic_validate_structure` builds a `CheckSpec` only to shape-check it; `_validate_check_block` and downstream code then re-reach into the untyped dict (`check['check']`, `check_block.get(...)`). Field access stays stringly-typed, so the migration doesn't actually give you a single typed representation. Consider threading the parsed `CheckSpec` through instead of re-walking the dict.

src/databricks/labs/dqx/checks_validator.py
```diff
@@ -86,60 +160,149 @@ def validate_checks(
...
+            return blocking_errors, sibling_errors
 
-        if "criticality" in check and check["criticality"] not in [c.value for c in Criticality]:
-            errors.append(
-                f"Invalid 'criticality' value: '{check['criticality']}'. "
-                f"Expected '{Criticality.WARN.value}' or '{Criticality.ERROR.value}'. "
-                f"Check details: {check}"
-            )
+    @staticmethod
+    def _translate_pydantic_error(error: ErrorDetails, check: dict) -> str | None:
```
mwojtyczka: **This translation shim couples to Pydantic's internal error `type`/`loc` codes** (`dict_type`, `list_type`, `value_error`, plus stripping the `'Value error, '` prefix) purely to reproduce the legacy message text. A Pydantic upgrade that renames an error type will silently change which branch fires and degrade messages — caught only by string-match tests. Worth pinning the Pydantic version and/or adding a test that asserts on the error `type` codes this relies on.

mwojtyczka (commented):
## Automated review — Pydantic serialization migration

Ran a multi-angle review pass over this PR. Overall the migration is careful and well-tested — `CheckSpec`/`CheckBlock`, `extra="forbid"`, `project_to_check_schema`, and the error-message parity/blocking split are handled deliberately, and no crash/inversion bug surfaced. The notes below are ranked; inline comments are attached to the relevant lines.

**Ordered summary:**
1. **`user_metadata` type narrowing** (`checks_validator.py`) — the one likely **regression**: a previously-unvalidated field is now strictly `dict[str, str]`, so checks with non-string metadata values that loaded before now raise `InvalidCheckError`. Worth resolving before merge.
2. **Double validation** (`checks_serializer.py`) — every check is parsed through `CheckSpec` twice (validate → deserialize).
3. **Validated `CheckSpec` discarded** (`checks_validator.py`) — Pydantic is used for shape-checking only; downstream code re-reads the untyped dict, so the migration doesn't yet become the single typed boundary.
4. **Error-translation shim coupled to Pydantic internals** (`checks_validator.py`) — reproduces legacy message text via internal error `type` codes; fragile across Pydantic upgrades.
5. **`extra="forbid"` behavior change** (`config.py`) — intended, but rejects extra kwargs some callers may pass; confirm construction sites.
6. **`project_to_check_schema` duplicates `model_dump` and is applied only on the Lakebase path** (`checks_serializer.py` / `checks_storage.py`) — the Delta loader hand-picks keys, so the two loaders can diverge.
7. **Docstring double-backticks** (`checks_validator.py`) — AGENTS.md says use italics, not backticks (renders wrong in API docs).

Nice work on the test coverage. #1 is the main one to decide on; the rest are cleanup/altitude to make Pydantic the actual serialization boundary rather than a shape-check in front of the existing dict-walking code.

src/databricks/labs/dqx/config.py
```diff
@@ -335,6 +335,8 @@ class BaseChecksStorageConfig(BaseModel, ABC):
         location: The file path or table name where checks are stored.
     """
 
+    model_config = ConfigDict(extra="forbid")
```
mwojtyczka, line 338: **`extra="forbid"` is a broad behavior change.** Any caller constructing a `*ChecksStorageConfig` from a loosely-typed dict (`**run_config`, installation-loaded data, forward-compat/vestigial keys) will now raise `InvalidConfigError` instead of silently ignoring unknown keys. Reasonable hardening, but worth grepping the construction sites (CLI/installer/loaders) before merge to confirm none pass extra keys.

src/databricks/labs/dqx/checks_serializer.py
```diff
@@ -22,6 +22,26 @@
 logger = logging.getLogger(__name__)
 
 
+def project_to_check_schema(check: dict) -> dict:
```
mwojtyczka, line 25: **`project_to_check_schema` re-implements key-filtering Pydantic already does**, and it's only applied on the Lakebase load path (see `checks_storage.py:783`) — the Delta `from_dataframe` path hand-cherry-picks keys instead. Two consequences: (1) `CheckSpec` uses `extra='ignore'`, so `CheckSpec.model_validate(check).model_dump()` yields the projected dict natively; (2) a new `CheckSpec` field is auto-retained by Lakebase but silently dropped by the Delta loader, so the two loaders can disagree on the loaded shape. Consider one shared projection used by both paths.

src/databricks/labs/dqx/checks_storage.py
```diff
@@ -779,7 +780,7 @@ def _load_checks_from_lakebase(self, config: LakebaseChecksStorageConfig, engine
                     f"for run_config_name='{config.run_config_name}'. "
                     f"Make sure the profiler has run successfully and saved checks to this location."
                 )
-            checks_dict = [dict(check) for check in checks]
+            checks_dict = [project_to_check_schema(dict(check)) for check in checks]
```
mwojtyczka, line 785: Only the Lakebase load path projects rows through `project_to_check_schema`; the Delta `from_dataframe` path hand-builds the check dict field-by-field. Adding a logical field to `CheckSpec` would be retained here but silently dropped by the Delta loader — the two loaders should share one projection so they agree on the loaded shape. (See the note on the `project_to_check_schema` definition.)

src/databricks/labs/dqx/checks_validator.py
```diff
@@ -86,60 +160,149 @@ def validate_checks(
...
+    @staticmethod
+    def _pydantic_validate_structure(check: dict) -> tuple[list[str], list[str]]:
+        """Run *CheckSpec.model_validate* on a single check dict and partition the errors.
+
+        Translates Pydantic *ValidationError* into human-readable strings that match the
+        existing error format expected by callers and tests.  The original *check* dict is
+        included in error messages for context, as the hand-rolled validator did before.
+
+        Errors are split by whether they block signature validation.  Errors on the 'check' block
+        (``loc`` starting with ``check``) are *blocking*: without a well-formed check block there is
```
mwojtyczka: **Docstrings use double-backticks around object names** (``loc``, ``check`` here; also `value_error`/`list_type`/`ValueError` around lines 274–278 and the `[]`/`model_copy()` docstrings near line 81). AGENTS.md → Docstrings: *"No backticks around object names — use italics instead (e.g., *arg1*). Backticks cause rendering issues in API docs."* Use `*italics*` for consistency with the rest of this PR.

mwojtyczka (changes requested):
Thanks for the PR. Run automated review, issues seems plausible. 

I was also hoping we could replace the ChecksValidator with Pydantic native validation.

fedeflowers:
Thanks for the review @mwojtyczka. All findings are addressed in e2f02d7. Recap below, plus a direct answer on the ChecksValidator question.

**Ranked findings:**

| # | Finding | Resolution |
|---|---|---|
| 1 | `user_metadata` narrowed to `dict[str, str]` (regression) | Widened back to `dict[str, Any]` on `CheckSpec`, `DQRule`, and `DQForEachColRule`, so non-string values (e.g. `{"confidence": 0.95}`) load exactly as pre-migration. Regression test added. |
| 2 | Double validation (validate then deserialize re-parses) | `_validate_and_parse` now returns the parsed specs; `ChecksDeserializer.deserialize` reuses them. Each check is parsed through `CheckSpec.model_validate` exactly once. |
| 3 | Validated `CheckSpec` discarded, raw dict re-read | `deserialize` now reads typed fields off the spec (`spec.check.function`, `spec.criticality`, ...), with no more re-walking of the untyped dict. |
| 4 | Error-translation shim coupled to Pydantic error `type` codes | Added a test pinning the exact error-type codes the translator branches on (`missing`, `dict_type`, `list_type`, `value_error`), so a Pydantic rename fails loudly instead of silently degrading messages. `pydantic<3` already caps the major version. |
| 5 | `extra="forbid"` on storage configs | Audited construction sites: all pass explicit kwargs, none spread loosely-typed dicts. No code change; behavior locked by `test_storage_config_rejects_unknown_kwarg`. |
| 6 | `project_to_check_schema` only on Lakebase path | Delta `from_dataframe` now routes through the same projection; the allow-list is derived from `CheckSpec.model_fields`, so both loaders agree on the loaded shape automatically. Guard test added. |
| 7 | Docstring double-backticks | Replaced with italics per AGENTS.md. |

**On replacing ChecksValidator with Pydantic native validation:**

The validation logic is now fully Pydantic-native. All rules live on `CheckSpec` (field types + validators for shape, a `model_validator(mode="after")` for criticality/function resolution/argument checks), and `CheckSpec.model_validate(check, context=...)` fully validates a check on its own. No hand-rolled dict-walking validation remains.

What's left of `ChecksValidator` is ~75 lines of pure error-string translation, mapping Pydantic's `ValidationError` back to the exact legacy DQX messages (`'check' field is missing: {...}`, etc.) that existing callers and tests assert on. Message compatibility is its only remaining reason to exist.

Two ways to finish the job:

1. **Adopt Pydantic's native error surface**: let `ValidationError` text through, delete the translation layer and pinning test. Cleanest end state, but a breaking change for anyone matching on the current message strings.
2. **Keep the strings, kill the class**: fold the loop + translation into `CheckSpec` (e.g. a `validate_many` classmethod returning status + parsed specs), so `ChecksValidator` disappears as a concept while messages stay byte-identical.

**I lean Option 2** to avoid breaking downstream matchers on error text.

fedeflowers:
## Follow-up (out of scope for this PR): table backends silently drop/coerce check metadata, flagged by AI

Flagging two pre-existing gaps between `CheckSpec` and the table storage backends. Both exist on `main` unchanged by this PR (the serialization migration only touched the load path), so I'd suggest tracking them as a separate follow-up rather than expanding scope here.

### 1. Non-string `user_metadata` values are coerced to strings on the Delta path

`CheckSpec.user_metadata` is `dict[str, Any]` and intentionally permissive (e.g. `{"confidence": 0.95}`), but `CHECKS_TABLE_SCHEMA` types the column as `MAP<STRING, STRING>` and the row builder writes the raw dict into it — PySpark's StringType converter coerces any value via `str()` (see `pyspark/sql/conversion.py`, "Coercion to StringType is allowed"), so `0.95` is silently stored as `"0.95"` and `True` as `"true"`. Lakebase (`JSONB`) and File/Volume (YAML/JSON) preserve the original types; only Delta loses them.

Note that `arguments` in the same schema already avoids this via `json.dumps` on write + `safe_json_load` on read — `user_metadata` just never got the same treatment.

### 2. `message_expr` is not persisted by either table backend

`message_expr` is a first-class `CheckSpec` / `DQRule` field (added in #1092) and is in the load-path allow-list (derived from `CheckSpec.model_fields`), but:

- **Delta**: no column in `CHECKS_TABLE_SCHEMA`, and the row builder never emits it.
- **Lakebase**: omitted from both the table column definitions and the normalized row dict.

So a save→load round-trip through either table backend silently drops it. It survives only via File/Volume/YAML/JSON storage. Since the loaders would return it if it were stored, the drop happens purely on the write side.

### Why a separate PR

Any fix changes the at-rest format:

- `test_save_checks_to_table_with_unresolved_for_each_column` pins raw `user_metadata` as plain strings, so JSON-encoding on write breaks that pinned contract.
- Both fixes need a back-compat plan for existing tables (legacy rows without the new encoding/column).

### Suggested direction

- **`user_metadata`**: `json.dumps` on Delta write + `safe_json_load` on read (mirror `arguments`), with the read path tolerant of legacy plain-string rows.
- **`message_expr`**: add a nullable `STRING` / `Text` column to both table schemas; legacy rows read back as `None`.
- Update the pinned contract test to the new encoded shape.

Happy to open the follow-up issue/PR if that's the preferred path.

mwojtyczka (commented):
## Re-review of `e2f02d7` — thanks for the thorough turnaround, @fedeflowers 🙏

Re-ran the review against `e2f02d7`. **All 7 original findings are verified fixed in code** — `user_metadata` → `dict[str, Any]`; single-pass validate→deserialize reusing the parsed specs; typed `spec.*` reads; Delta + Lakebase both routing through `project_to_check_schema`; docstring italics; plus the error-type pinning test and the `extra="forbid"` construction-site audit + test. I've **resolved those 8 review threads**.

**On retiring `ChecksValidator` (your question):** agree with your **Option 2** — fold the loop + translation into `CheckSpec` (e.g. a `validate_and_parse`/`validate_many` classmethod) so the class disappears as a concept while error strings stay byte-identical. That also removes the cross-module `_validate_and_parse` access flagged below, so it's worth doing here rather than deferring. Option 1 (native Pydantic error text) is cleaner long-term but breaks downstream matchers, so not now.

**On the out-of-scope follow-up (Delta metadata coercion + `message_expr`):** agreed, and the AI flag is correct — my re-review independently hit the same two gaps (non-string `user_metadata` coerced via `str()` into the Delta `MAP<STRING,STRING>` column; `message_expr` in the `CheckSpec` allow-list but persisted by neither table backend). Both are pre-existing on `main` and change the at-rest format + need a back-compat plan, so they belong in a **separate follow-up**, not here. Please go ahead and open it — your suggested direction (`json.dumps`/`safe_json_load` for `user_metadata` mirroring `arguments`; a nullable `message_expr` column on both table schemas + tolerant reads; update the pinned contract test) looks right. Please open new feature improvement issues for these.

**New findings from the re-review** (inline; all introduced by the rework — none blocking, but 1–2 are worth addressing here):
1. **Diagnostics parity gap** — a bad `criticality` is no longer reported alongside a malformed `check` block (after-validator skipped on field errors); contradicts the parity goal, test gap.
2. **Cross-module protected access** (`checks_serializer.py:273`) — `_validate_and_parse` reached from another module (Critical Rule 6); folds away under Option 2.
3. **Re-implements `pydantic.TypeAdapter`** — the `_check_*` type-walker duplicates what `TypeAdapter` provides.
4. **`model_validate` without `context` footgun** — silent defaults run strict function validation; tolerance only works via `ChecksValidator`.
5. **Ineffective `lru_cache`** (minor) — per-call nested cache reuses nothing across checks.

Great migration overall — the single-`CheckSpec`-boundary direction is the right one.

src/databricks/labs/dqx/checks_validator.py
```diff
@@ -5,201 +5,177 @@
...
+    filter: str | None = None
+    # Values are intentionally *Any*, not *str*: the pre-migration validator never inspected
+    # user_metadata, so checks carrying non-string values (e.g. {"confidence": 0.95}) loaded fine.
+    # Narrowing to dict[str, str] would reject them (Pydantic v2 won't coerce int/float/bool to str)
+    # and regress released behaviour. Keep it permissive to preserve the load -> apply round-trip.
+    user_metadata: dict[str, Any] | None = None
+    message_expr: str | None = None
+
+    @model_validator(mode="after")
+    def _validate_semantics(self, info: ValidationInfo) -> "CheckSpec":
```
mwojtyczka, line 118: **New (re-review): diagnostics parity gap.** `criticality` is validated here in `_validate_semantics` (a `model_validator(mode="after")`), and Pydantic skips after-validators when *any* field-level validation fails — so a bad `criticality` combined with a malformed `check` block no longer surfaces together; only the check-block error is returned. The old hand-rolled validator returned both.

Repro: `{"criticality": "bogus", "check": {"arguments": {"column": "a"}}}` (bad criticality + missing `function`) previously yielded **both** the invalid-criticality and missing-function errors; now only the missing-function one. This contradicts the stated parity goal ("reports as many problems as the hand-rolled validator did"), and the regression test only exercises a well-formed check block so it misses the case. Consider validating `criticality` at the field level (e.g. an enum/`Literal` or `field_validator`) so it's collected alongside field errors — or add a test for the combined case and document the reduced-in-one-pass behavior.

src/databricks/labs/dqx/checks_serializer.py
```diff
@@ -245,29 +270,30 @@ def deserialize(self, checks: list[dict]) -> list[DQRule]:
         Raises:
             InvalidCheckError: If any dictionary is invalid or unsupported.
         """
-        status = ChecksValidator.validate_checks(checks, self.custom_checks)
+        status, specs = ChecksValidator._validate_and_parse(checks, self.custom_checks)
```
mwojtyczka: **New (re-review): cross-module protected-member access.** `deserialize` reaches into `ChecksValidator._validate_and_parse` (leading underscore) from another module using private method. This is what AGENTS.md **Critical Rule 6** calls out: *"never … let tests or callers reach private/protected members (`_name`) … make it public or extract a shared helper."* Reusing the parsed specs is now a deliberate cross-module contract (it's how you avoid double-validation), so promote it to a public `validate_and_parse` (or a shared module-level helper). This also disappears naturally under your **Option 2** for retiring `ChecksValidator`. (pylint `protected-access`/W0212 should be flagging this — worth confirming `make lint` is green.)

src/databricks/labs/dqx/checks_validator.py
```diff
@@ -5,201 +5,177 @@
...
+                errors.extend(self._validate_func_args(effective_arguments, func, check, func_parameters))
             return errors
-        return ChecksValidator._validate_func_args(effective_arguments, func, check, func_parameters)
+        return self._validate_func_args(effective_arguments, func, check, func_parameters)
 
     @staticmethod
     def _validate_func_args(arguments: dict, func: Callable, check: dict, func_parameters: Any) -> list[str]:
-        """
-        Validates the arguments passed to a function against its signature.
+        """Validate argument names and types against the function signature.
```
mwojtyczka, line 228: **New (re-review): re-implements `pydantic.TypeAdapter`.** The `_check_type`/`_check_union_type`/`_check_list_type`/… family hand-rolls runtime type validation against arbitrary annotations (`Union`, `list`, `dict`, `tuple`, `str | Column`, …) — exactly what `pydantic.TypeAdapter(annotation).validate_python(value)` provides, in a PR whose whole premise is moving validation onto Pydantic. Bespoke type-walking is easy to get subtly wrong on nested generics and duplicates a library you now depend on. Consider a `TypeAdapter` per parameter annotation to stay consistent with how `CheckSpec`/`DQRule` validate.

src/databricks/labs/dqx/checks_validator.py
```diff
@@ -5,201 +5,177 @@
...
+        Raises:
+            ValueError: If the criticality value is invalid, the function is unresolved, or the
+                supplied arguments do not satisfy the function signature.
         """
-        Validates a check block within a configuration.
+        context = info.context or {}
+        raw_check = context.get("raw_check")
+        check_repr = raw_check if raw_check is not None else self.model_dump(exclude_none=True)
+        custom_check_functions = context.get("custom_check_functions")
+        validate_custom_check_functions = context.get("validate_custom_check_functions", True)
```
mwojtyczka, line 145: **New (re-review): `CheckSpec.model_validate(check)` without `context` is a footgun.** `_validate_semantics` reads `info.context or {}` with silent defaults, so a direct call without the context dict that `ChecksValidator._validate_and_parse` supplies runs full function/argument validation with `validate_custom_check_functions=True` — reporting every custom/unregistered function as "not defined" even on paths meant to tolerate them (LLM/profiler). The tolerance flag only works when routed through `ChecksValidator`. Relatedly, when `raw_check` is absent the error context falls back to `model_dump(exclude_none=True)`, so messages reflect the projected model (extras/None dropped) rather than the user's input. Consider requiring context (or a classmethod that binds it) so the model can't be mis-called.

src/databricks/labs/dqx/checks_validator.py
```diff
@@ -5,201 +5,177 @@
...
-
-        return ChecksValidator._validate_check_function_arguments(arguments, func, for_each_column, check)
-
-    @staticmethod
-    def _validate_check_function_arguments(
-        arguments: dict, func: Callable, for_each_column: list, check: dict
-    ) -> list[str]:
-        """
-        Validates the provided arguments for a given function and updates the errors list if any validation fails.
+    def _validate_check_function_arguments(self, func: Callable, check: dict) -> list[str]:
```
mwojtyczka, line 201: **New (re-review, minor): this `lru_cache` caches almost nothing.** `@ft.lru_cache` decorates `cached_signature`, but it's defined *inside* this method, so a fresh empty cache is created on every call — `inspect.signature` is re-computed once per check rather than once per distinct function across the whole validation pass. Move the cache to a module-level function keyed on the resolved `func` (or just drop it) so it actually reuses across checks.

src/databricks/labs/dqx/checks_validator.py
```diff
@@ -228,24 +248,32 @@ def _validate_func_args(arguments: dict, func: Callable, check: dict, func_param
                 expected_type = func_parameters[arg].annotation
                 if get_origin(expected_type) is list:
                     expected_type_args = get_args(expected_type)
-                    errors.extend(ChecksValidator._validate_func_list_args(arg, func, check, expected_type_args, value))
-                elif not ChecksValidator._check_type(value, expected_type):
-                    expected_type_name = getattr(expected_type, '__name__', str(expected_type))
+                    errors.extend(CheckSpec._validate_func_list_args(arg, func, check, expected_type_args, value))
```
mwojtyczka, line 257: **New (round 3): strict `TypeAdapter` tightens `bool`→`int`, and the list path is inconsistent with it.**

The scalar branch validates via `_matches_type` → `TypeAdapter.validate_python(value, strict=True)`, which **rejects `bool` for an `int`/`float`/`str` parameter** — whereas the old hand-rolled checker used `isinstance`, and `isinstance(True, int)` is `True`. So a stored check with a YAML `true`/`false` literal reaching an int/float/str(-union) parameter that validated before now fails, breaking the load→apply round-trip this PR aims to preserve (verified: old `isinstance(True, int)` → True vs new strict → False).

Separately, the **list branch** (`get_origin(...) is list` → `_validate_func_list_args`, ~line 344) still uses non-strict `isinstance(item, ...)`, so `bool` is *accepted* for `list[int]` but *rejected* for scalar `int` — the same logical type validated by two different rules. `TypeAdapter` handles `list[int]` natively, so routing lists through `_matches_type` too would remove both the divergence and the bespoke list validator. Worth deciding whether strict is intended (and adding a bool carve-out + test), but scalar and list should agree.

src/databricks/labs/dqx/checks_validator.py
```diff
@@ -261,67 +289,39 @@ def _missing_required_arguments(
...
-    @staticmethod
-    def _check_tuple_type(args, value):
-        if not isinstance(value, tuple):
+        Returns:
+            True if *value* is valid for *expected_type* (or the type cannot be schematised).
+        """
+        if expected_type is inspect.Parameter.empty:
+            return True  # no type hint, assume valid
+        try:
+            adapter = _type_adapter(expected_type)
```
mwojtyczka, line 324: **New (round 3): an unhashable annotation crashes instead of the documented lenient fallback.** The docstring says annotations pydantic can't schematise are “treated as satisfied,” and `except (PydanticSchemaGenerationError, SchemaError)` guards `_type_adapter(...)`. But `_type_adapter` is `@ft.lru_cache`-decorated, so an **unhashable** `expected_type` (e.g. a custom check function whose parameter annotation is a non-hashable typing/metadata object) raises `TypeError` from `lru_cache` *before* `TypeAdapter` is built — and `TypeError` isn't caught, so it propagates out of `_matches_type` → `validate_and_parse` as a hard crash. Add `TypeError` to the caught set (or guard hashability) to preserve the lenient behavior.

src/databricks/labs/dqx/checks_validator.py
```diff
@@ -2,204 +2,224 @@
...
 from databricks.labs.dqx.checks_resolver import resolve_check_function
 from databricks.labs.dqx.rule import Criticality
 
 logger = logging.getLogger(__name__)
 
 
-class ChecksValidationStatus(BaseModel):
-    """Class to represent the validation status.
+@ft.lru_cache(maxsize=None)
+def _cached_signature(func: Callable) -> inspect.Signature:
```
mwojtyczka, line 32: **New (round 3): unbounded module-level caches retain user callables for the process lifetime.** `_cached_signature` (and `_type_adapter`, line 36) use `@ft.lru_cache(maxsize=None)`. `_cached_signature` is called with user-supplied `custom_check_functions`, so every distinct callable (a fresh closure per notebook/session, possibly capturing large state) is pinned forever — monotonic memory growth in a long-lived driver, and broader than the docstring’s “once per validation pass” (it’s process-wide). This is the flip side of hoisting the cache to module scope (which was the right call for reuse): give both a bounded `maxsize`, or only cache functions resolved from the built-in registry.

src/databricks/labs/dqx/checks_validator.py
```diff
@@ -2,204 +2,224 @@
...
-    def _validate_check_block(
-        check: dict, custom_check_functions: dict[str, Callable] | None, validate_custom_check_functions: bool
-    ) -> list[str]:
+        Returns:
+            A single-element list with the error message, or an empty list when valid.
         """
-        Validates a check block within a configuration.
+        criticality = check.get("criticality", Criticality.ERROR.value)
+        valid = {c.value for c in Criticality}
+        if not isinstance(criticality, str) or criticality in valid:
```
mwojtyczka, line 172: **New (round 3, minor): non-string `criticality` loses the descriptive message.** For `criticality: 123`, `criticality_errors` returns `[]` (the `isinstance(criticality, str)` guard), so the user gets the generic field-type message (“Input should be a valid string”) instead of the pre-migration “Invalid criticality value: 123. Expected warn or error.” The docstring notes this is intentional (avoid double-reporting), which is reasonable — flagging only in case a downstream test/caller matches the old message text for non-string criticality.

mwojtyczka (commented):
## Round-3 re-review of `3b8cab1` — thanks @fedeflowers 🙏

All **5 findings from the previous round are verified fixed** in `3b8cab1`: `validate_and_parse` is public (no more cross-module protected access); `_cached_signature` is a module-level cache keyed on `func`; the hand-rolled type-walker is replaced by `pydantic.TypeAdapter`; the criticality diagnostics gap is reproduced on the field-failure path via `criticality_errors`; and a blessed `validate_check` entry point documents the context requirement.

The `TypeAdapter` switch introduced a few new edges (inline), the top one worth a decision before merge:

1. **`bool`→`int` strict tightening + list/scalar inconsistency** — strict `TypeAdapter` rejects a `bool` for an `int`/`float`/`str` param that the old `isinstance` check accepted, while the list branch still uses `isinstance` (so `bool` passes for `list[int]`). A stored check with a `true`/`false` literal on an int-ish param can now fail the load→apply round-trip, and the scalar/list paths disagree. (Verified empirically.)
2. **Unhashable annotation → uncaught `TypeError`** from the cached `_type_adapter`, crashing instead of the documented lenient fallback.
3. **Unbounded module-level caches** (`maxsize=None`) retaining user callables for the process lifetime.
4. **Minor:** non-string `criticality` loses its descriptive parity message (documented tradeoff).

Also a related *loosening*, not flagged inline: `int` is now accepted for a `float`-only param where the old checker rejected it — arguably an improvement, but a behavior change if any test pins the old rejection.

Overall very close — #1 is the one I'd resolve (pick strict-vs-lenient deliberately and make scalar + list consistent); the rest are small.

fedeflowers:
fixed the behavior of re-review 3

src/databricks/labs/dqx/checks_validator.py
mwojtyczka: Verified fixed on the current head — the caches are now bounded via `maxsize=_CACHE_MAXSIZE` instead of `maxsize=None`. Resolving.

src/databricks/labs/dqx/checks_validator.py
mwojtyczka: Verified fixed on the current head — `except (SchemaError, TypeError)` now catches the unhashable-annotation `TypeError` from `lru_cache`, preserving the lenient fallback. Resolving.

src/databricks/labs/dqx/checks_validator.py
mwojtyczka: Addressed in the current revision — the divergence is gone. `_validate_func_list_args` no longer uses `isinstance`; each list element is now checked with the same `CheckSpec._matches_type(item, item_type)` used for scalar args, so `int` and `list[int]` agree: a stringy "5" is rejected for both, and a bool is accepted for both. Resolving.

src/databricks/labs/dqx/checks_validator.py
mwojtyczka: Addressed in the current revision. `criticality_errors` now returns the descriptive `Invalid 'criticality' value: '...'. Expected 'warn' or 'error'.` message for any invalid *string* value (the realistic typo case, e.g. "critical"). Only a non-*string* type (int/bool/list) is deliberately left to pydantic's generic `str` type error, as documented in the method's docstring — the input is still rejected either way. Resolving.

mwojtyczka (approved):
LGTM - kicked tests

mwojtyczka:
@fedeflowers PR merged, created follow up issue: https://github.com/databrickslabs/dqx/issues/1336

PR #1303, opened 2026-07-04
Guard check filter and row_filter against unsafe SQL

### LINKED ISSUE #1089: [FEATURE]: Guard check filter

### Problem statement

`filter` (and `row_filter`) from rules are passed directly as sql expresssion without any safety check

### Proposed Solution

For check.filter and row_filter — these are filter predicates, not full SQL queries. SELECT is never a valid keyword in a predicate expression, so it can be safely blocked. Add "select" to the forbidden_statements list in `is_sql_query_safe` and call it in DQRuleManager.filter_condition and at each row_filter call site.

For sql_expression — this is explicitly a "write SQL here" parameter where subqueries (EXISTS (SELECT ...)) may be legitimate. The right mitigation here is a clear docstring security warning rather than a keyword block.

Documentation — add a section to the user guide noting which parameters accept arbitrary SQL and that callers are responsible for ensuring values come from trusted sources when using them in automated or multi-tenant pipelines.

### Additional Context

_No response_

SaptarshiAcharyya99:
## Changes
A check's `filter` / `row_filter` was compiled directly via `F.expr(...)` with no validation. This PR routes all filter compile sites through a shared `safe_filter_expr` helper that rejects destructive SQL keywords, and treats a check with an unsafe filter as **skipped** rather than failing the run.

**Scope / threat model (important).** Filters compile to a boolean **expression** (`F.expr(...)` / `df.filter("<str>")`), which Spark parses with the *expression* grammar, not the *statement* grammar. A destructive statement like `DROP TABLE ...` therefore cannot execute in a filter context — it fails to parse (`PARSE_SYNTAX_ERROR`). So this guard on the filter paths is **defense-in-depth plus better diagnostics**, not a fix for an exploitable hole. The genuinely statement-executing paths (`sql_query` → `spark.sql`, and the LLM query path) were already guarded by `is_sql_query_safe` before this PR and are unchanged here.

What the PR does:
- New shared helper `safe_filter_expr` validates a filter with `is_sql_query_safe` and compiles it. Used for both the check-level `filter` and check-function `row_filter` params. A tree-wide sweep routed 12 sites through it: `manager.py`, 9 `row_filter` sites in `check_funcs.py`, and one each in `anomaly/scoring_utils.py` and `geo/check_funcs.py` — including the previous `has_no_outliers` path that applied the raw `row_filter` string with no validation.
- **Unsafe filter ⇒ skip, not abort.** `DQRuleManager` detects an unsafe filter (via `has_unsafe_filter`, non-raising) and skips just that check — reported with `skipped = true` and a message identifying the unsafe filter — while the rest of the rule set still runs. This matches how unresolvable columns/filters are already handled and avoids one bad filter aborting the whole `apply_checks` run. The `sql_query` check's `spark.sql` path is intentionally **not** downgraded to a skip — it still hard-rejects.
- `SELECT` / subqueries are **allowed** in filters (a working, powerful feature — inline referential filters, e.g. `customer_id IN (SELECT ... FROM ref)`), authored by trusted operators. Only destructive keywords are blocked.
- `is_sql_query_safe` was hardened to avoid false positives: quoted string literals and backtick-quoted identifiers are stripped before scanning (so `status = 'drop'`, `col = 'use case'`, `` `drop` `` are safe, including backslash-escaped quotes), and inline comments are handled so a keyword hidden in (`/* delete */`) or split by (`dr/**/op`) a comment is still caught.
- `sql_expression` is intentionally **not** keyword-blocked (subqueries can be legitimate); it gets a docstring security note flagging it as trusted-input-only.

### Linked issues
Resolves #1089

### Tests
- [x] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

Unit tests cover `is_sql_query_safe` (destructive-keyword rejection, keyword-as-data / quoted-identifier / backslash-escape false-positive cases, comment handling, `SELECT` allowed) and `safe_filter_expr` (raises on destructive, allows `SELECT`). `test_manager.py` asserts an unsafe filter causes the check to be **skipped** (row, dataset, and `row_filter`-in-kwargs cases; plus `suppress_skipped`). Integration test asserts `has_no_outliers` rejects a destructive `row_filter` at the check-function level (regression test for the previous MAD bypass).

### Documentation and Demos
- [ ] added/updated demos
- [x] added/updated docs

Updated the `filter` note in `quality_checks_definition.mdx` (bullet + table) to state that subqueries are allowed and that an unsafe filter is skipped (`skipped = true`), and added a trusted-input security note to the `sql_expression` docstring.


mwojtyczka:
Suggestion: I'd **keep `SELECT` allowed in filters** rather than blocking it, and scope this PR to the destructive keywords.

**Why keep `SELECT`:**
- **Filters are authored by trusted operators** — the same people who define and run the DQX checks. The only real risk from a subquery is a confused-deputy *read* (a filter reading data its author shouldn't, via the engine's identity). When the filter author and the engine operator are the same trusted principal — which is the case for check definitions — that risk doesn't materialize.
- **A `SELECT` subquery in a filter is genuinely more powerful and works today.** It lets you express referential / dynamic filters inline (scope a check to rows that exist in a reference table, an allowlist sourced from a table, etc.). Blocking it removes real, working capability.

**Why destructive keywords should still be blocked (and this stays a simple change):**
- A filter compiles to a boolean *expression* (`F.expr(...)` / `df.filter("<str>")`), so `DELETE`/`DROP`/`INSERT`/… can't parse as statements and can't execute anyway. Keeping `is_sql_query_safe`'s existing block (`delete, insert, update, drop, truncate, alter, create, replace, grant, revoke, merge, use, refresh, analyze, optimize, zorder`) is the right, low-cost guarantee. So the change is just: **block the destructive set, allow `SELECT`.**

**Tests should cover the destructive keywords:**
- Parametrize `is_sql_query_safe` over the full forbidden set, asserting each is rejected.
- Add a filter-path test that a `drop`/`delete` filter is rejected (via `safe_filter_expr` and via `apply_checks_by_metadata` for a representative check).
- Add a test asserting a `SELECT`/subquery filter is **allowed** and runs end-to-end.

**Example of a `SELECT` filter that should keep working** — only check rows whose `customer_id` is in a reference/allowlist table:

```yaml
- criticality: error
  filter: "customer_id IN (SELECT customer_id FROM main.ref.active_customers)"
  check:
    function: is_not_null
    arguments:
      column: email
```

equivalently in code:

```python
DQRowRule(
    check_func=check_funcs.is_not_null,
    column="email",
    criticality="error",
    filter="customer_id IN (SELECT customer_id FROM main.ref.active_customers)",
)
```

Could you also add an example like this — a filter that references another table — to the PR (and ideally the docs), so the intended "allow `SELECT`" behavior is demonstrated end-to-end?

mwojtyczka (changes requested):
Thanks for the PR! Let's keep SELECT allowed in filters rather than blocking it, and scope this PR to the destructive keywords only. Explanation in the previous comment.

mwojtyczka (commented):
Inline notes to accompany the summary comment above: keep `SELECT` in filters (trusted-operator authored, and subqueries are a working, powerful feature), while keeping the destructive-keyword block and testing it.

src/databricks/labs/dqx/utils.py
```diff
@@ -260,9 +237,32 @@ def is_sql_query_safe(query: str) -> bool:
...
+    Args:
+        filter_expr: The filter predicate as a string, or None.
+
+    Returns:
+        The compiled filter column, or a literal true column when no filter is given.
+
+    Raises:
+        UnsafeSqlQueryError: If the filter contains a forbidden statement such as SELECT.
+    """
+    if filter_expr and not is_sql_query_safe(filter_expr, forbid_select=True):
```
mwojtyczka: I'd **not forbid `SELECT` for filters** here. A subquery in a filter is a working, powerful feature (inline referential/dynamic filters), and check filters are authored by trusted operators — the same principals who run the engine — so the confused-deputy *read* risk doesn't apply. Suggest dropping `forbid_select=True` and keeping only the destructive-keyword block. If you want a safety valve for deployments that load untrusted filters, expose `forbid_select` as an opt-in (default off) rather than always-on.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1246,7 +1264,7 @@ def apply(df: DataFrame) -> DataFrame:
                 f"Column '{col_expr_str}' must be of numeric type to perform outlier detection using MAD method, "
                 f"but got type '{column_type.simpleString()}' instead."
             )
-        filter_condition = F.expr(row_filter) if row_filter else F.lit(True)
+        filter_condition = safe_filter_expr(row_filter)
         median, mad = _calculate_median_absolute_deviation(df, col_expr_str, row_filter)
```
mwojtyczka: This passes the **raw** `row_filter` string into `_calculate_median_absolute_deviation`, which applies it via `df.filter(<str>)` — bypassing `safe_filter_expr` entirely, so even the destructive-keyword guard doesn't cover this path (`has_no_outliers`). Route it through `safe_filter_expr` (or pass the already-built `filter_condition`) so validation is consistent across all checks.

tests/unit/test_utils.py
```diff
@@ -258,6 +259,32 @@ def test_mixed_content():
     assert not is_sql_query_safe("WITH cte AS (UPDATE users SET x=1) SELECT * FROM cte")
 
 
+def test_select_blocked_when_forbid_select():
```
mwojtyczka: Two coverage asks: (1) parametrize `is_sql_query_safe` over the full destructive set (`delete, insert, update, drop, truncate, alter, create, replace, grant, revoke, merge, use, refresh, analyze, optimize, zorder`), asserting each is rejected — not just `select`. (2) Since the recommendation is to keep `SELECT` allowed, add a test that a `SELECT`/subquery filter is **allowed** and runs end-to-end (e.g. `customer_id IN (SELECT customer_id FROM ...)`), rather than asserting it's blocked.

SaptarshiAcharyya99:
Thanks for the detailed review — this makes sense and I agree with the direction. Happy to keep
`SELECT` allowed in filters and scope the guard to the destructive keywords.

Plan:

1. **Allow `SELECT` in filters.** Drop `forbid_select=True` in `safe_filter_expr` so filters use
   the existing destructive-keyword block and subqueries work. I'll keep `forbid_select` on
   `is_sql_query_safe` as an off-by-default opt-in, per your suggestion, for deployments that load
   untrusted filters.

2. **Fix the `has_no_outliers` bypass** — good catch. `_calculate_median_absolute_deviation`
   applies the raw `row_filter` string directly. I'll pass the already-built, validated
   `filter_condition` into it instead, so that path is guarded like the rest.

3. **Tests.** `is_sql_query_safe` already parametrizes the full destructive set in
   `test_query_with_forbidden_commands`; I'll add filter-path coverage on top — `safe_filter_expr`
   and `apply_checks_by_metadata` reject a `drop`/`delete` filter — plus tests that a
   `SELECT`/subquery filter is **allowed** (through `safe_filter_expr` and end-to-end), and flip
   the existing SELECT assertions from blocked to allowed.

4. **Docs.** Update the `filter` note in `quality_checks_definition.mdx` and the `sql_expression`
   docstring to reflect: destructive keywords blocked, `SELECT`/subqueries allowed. I'll add your
   `customer_id IN (SELECT customer_id FROM …)` example so the intended behavior is demonstrated.

Pushing the updates shortly.

docs/dqx/docs/guide/quality_checks_definition.mdx
```diff
@@ -342,7 +342,7 @@ Checks defined using declarative syntax must contain the following fields:
...
+- (optional) `filter`: spark expression to filter the rows for which the check is applied (e.g. `"business_unit = 'Finance'"`). The check function will run only on the rows matching the filter condition. The condition can reference any column of the validated dataset, not only the one where you apply the check function. When using dataset-level checks, the filter condition is pushed down as `row_filter` to the check function and applied before aggregation, ensuring that the check operates only on the relevant subset of rows rather than on the aggregated results. For security, `filter` and `row_filter` are restricted to simple predicates: they cannot contain a `SELECT` statement, and an unsafe filter raises an error. If you need arbitrary SQL (including subqueries), use the `sql_expression` check instead, and only accept check definitions from trusted sources in automated or multi-tenant pipelines.
```
ghanse: Instead of saying _"For security, `filter` and `row_filter` are restricted to simple predicates: they cannot contain a `SELECT` statement, and an unsafe filter raises an error."_, I think we should just tell the user what is not allowed with a few examples:

> _"For security, `filter` and `row_filter` cannot contain unsafe DML keywords like (e.g. `UPDATE`, `DELETE`). An unsafe filter raises an error during execution._

docs/dqx/docs/guide/quality_checks_definition.mdx
```diff
@@ -699,7 +699,7 @@ The table used to store checks has the following structure. You can use `for_eac
...
+| `filter`               | `string`              | (Optional) Spark SQL expression to filter rows to which the check is applied, e.g. `"business_unit = 'Finance'"`. The check function will run only on the rows matching the filter condition. The condition can reference any column of the validated dataset, not only the one where you apply the check function. When using dataset-level checks, the filter condition is pushed down as `row_filter` to the check function and applied before aggregation, ensuring that the check operates only on the relevant subset of rows rather than on the aggregated results. For security, `filter` and `row_filter` are restricted to simple predicates: they cannot contain a `SELECT` statement, and an unsafe filter raises an error. If you need arbitrary SQL (including subqueries), use the `sql_expression` check instead, and only accept check definitions from trusted sources in automated or multi-tenant pipelines.|
```
ghanse: Same comment applies here.

src/databricks/labs/dqx/check_funcs.py
ghanse: Agree with passing `filter_condition`

src/databricks/labs/dqx/utils.py
ghanse: +1

ghanse (changes requested):
Looking good. I agree with the previous feedback. Left a few additional comments on the documentation.

SaptarshiAcharyya99:
Thanks @mwojtyczka , @ghanse — addressed in the latest commit:

- **Allow `SELECT` in filters.** `safe_filter_expr` now uses the default (destructive-keyword)
  block; `SELECT`/subqueries pass through. `forbid_select` stays on `is_sql_query_safe` as an
  off-by-default opt-in.
- **`has_no_outliers` bypass** fixed — `_calculate_median_absolute_deviation` now receives the
  already-validated `filter_condition` instead of the raw string.
- **Tests:** flipped the `SELECT` assertions to "allowed"; added a destructive-filter rejection
  test and an allowed-`SELECT` test using the `customer_id IN (SELECT ... FROM ref)` example
  (end-to-end at the check level; full cross-table execution would be an integration test).
- **Docs:** rewrote the `filter` note (bullet + table) to state plainly that subqueries are
  allowed with the reference-table example, and that destructive statements are rejected — and
  removed the now-inaccurate `SELECT` line from the `sql_expression` docstring.

src/databricks/labs/dqx/utils.py
```diff
@@ -237,7 +237,7 @@ def normalize_col_str(col_str: str) -> str:
     return re.sub(COLUMN_NORMALIZE_EXPRESSION, "_", col_str[:max_chars].lower()).rstrip("_")
 
 
-def is_sql_query_safe(query: str) -> bool:
+def is_sql_query_safe(query: str, forbid_select: bool = False) -> bool:
```
mwojtyczka: this is not needed anymore since we are not blocking select

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1956,7 +1961,7 @@ def apply(df: DataFrame) -> DataFrame:
                 f"but got type '{time_col_type.simpleString()}' instead."
             )
 
-        filter_col = F.expr(row_filter) if row_filter else F.lit(True)
+        filter_col = filter_col = safe_filter_expr(row_filter)
```
```suggestion
        filter_col = safe_filter_expr(row_filter)
```

mwojtyczka:
> Thanks @mwojtyczka , @ghanse — addressed in the latest commit:
> 
> * **Allow `SELECT` in filters.** `safe_filter_expr` now uses the default (destructive-keyword)
>   block; `SELECT`/subqueries pass through. `forbid_select` stays on `is_sql_query_safe` as an
>   off-by-default opt-in.
> * **`has_no_outliers` bypass** fixed — `_calculate_median_absolute_deviation` now receives the
>   already-validated `filter_condition` instead of the raw string.
> * **Tests:** flipped the `SELECT` assertions to "allowed"; added a destructive-filter rejection
>   test and an allowed-`SELECT` test using the `customer_id IN (SELECT ... FROM ref)` example
>   (end-to-end at the check level; full cross-table execution would be an integration test).
> * **Docs:** rewrote the `filter` note (bullet + table) to state plainly that subqueries are
>   allowed with the reference-table example, and that destructive statements are rejected — and
>   removed the now-inaccurate `SELECT` line from the `sql_expression` docstring.

Thanks for the fixes. They are all resolved. There are a few more issues that I posted that need fixing before merging

src/databricks/labs/dqx/utils.py
```diff
@@ -260,9 +260,33 @@ def is_sql_query_safe(query: str) -> bool:
         "optimize",
         "zorder",
     ]
+    if forbid_select:
+        forbidden_statements = forbidden_statements + ["select"]
     return not any(re.search(rf"\b{kw}\b", normalized_query) for kw in forbidden_statements)
```
mwojtyczka: Pre-existing issue: **False positives: the blocklist matches keywords inside string literals / values.** `normalized_query` only lowercases and collapses whitespace, and `re.search(rf"\b{kw}\b", ...)` matches a forbidden keyword *anywhere*, including as data. So legitimate filter predicates now raise `UnsafeSqlQueryError`:

- `status = 'drop'`
- `action IN ('DELETE','UPDATE')`
- `event_type = 'merge'`
- `note LIKE '%replace%'`
- `col = 'use case'` (the bare word `use`)

All of these compiled fine via `F.expr` before this PR. Please add a keyword-inside-string-literal test and avoid rejecting predicates where the keyword isn't actually a statement (inspect outside quoted literals, or take a different approach for filter predicates).

src/databricks/labs/dqx/utils.py
```diff
@@ -237,7 +237,7 @@ def normalize_col_str(col_str: str) -> str:
     return re.sub(COLUMN_NORMALIZE_EXPRESSION, "_", col_str[:max_chars].lower()).rstrip("_")
 
 
-def is_sql_query_safe(query: str) -> bool:
+def is_sql_query_safe(query: str, forbid_select: bool = False) -> bool:
     # Normalize the query by removing extra whitespace and converting to lowercase
```
mwojtyczka: Pre-existing issue: **The blocklist isn't robust — no comment/literal handling.** Normalization only collapses whitespace and lowercases; it doesn't strip SQL comments, so `is_sql_query_safe("dr/**/op table users")` returns `True`. That's moot for a `filter` predicate (a statement can't execute in an `F.expr` expression context anyway), but `is_sql_query_safe` is the **shared** helper the `sql_query` check uses to gate `spark.sql` — a real statement context where comment/case/unicode/whitespace bypasses matter. If the guard is meant to protect that path, it needs comment stripping (ideally parse-based validation), not a substring blocklist.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1246,8 +1251,8 @@ def apply(df: DataFrame) -> DataFrame:
                 f"Column '{col_expr_str}' must be of numeric type to perform outlier detection using MAD method, "
                 f"but got type '{column_type.simpleString()}' instead."
             )
-        filter_condition = F.expr(row_filter) if row_filter else F.lit(True)
-        median, mad = _calculate_median_absolute_deviation(df, col_expr_str, row_filter)
+        filter_condition = safe_filter_expr(row_filter)
```
mwojtyczka: Minor: `safe_filter_expr(row_filter)` returns `F.lit(True)` when `row_filter` is `None`, so `filter_condition` is **never** `None` here. `_calculate_median_absolute_deviation` then always runs `df.filter(F.lit(True))` even with no filter (previously it skipped filtering), and the helper's new `Column | None` hint is effectively dead (no caller passes `None`). Consider passing `None` when `row_filter` is falsy (keeping the helper's `is not None` guard) so the no-filter path stays a true no-op and the type hint is honest.

src/databricks/labs/dqx/utils.py
mwojtyczka: Minor (efficiency/consistency): this rebuilds `forbidden_statements` and compiles ~17 `rf"\b{kw}\b"` patterns on **every call** via `re.search`. Elsewhere this module hoists patterns to module-level `re.compile` constants (e.g. `COLUMN_NORMALIZE_EXPRESSION`). Since `is_sql_query_safe` is also on the LLM / rule-validation path, consider one precompiled alternation (`re.compile(r"\b(?:delete|drop|insert|...)\b")`) at module scope.

mwojtyczka (commented):
Follow-up inline notes from re-reviewing `3941ec1` — the three earlier review threads are all addressed (thanks). A few concrete items below.

src/databricks/labs/dqx/utils.py
```diff
@@ -260,9 +260,33 @@ def is_sql_query_safe(query: str) -> bool:
         "optimize",
         "zorder",
     ]
+    if forbid_select:
+        forbidden_statements = forbidden_statements + ["select"]
```
mwojtyczka: not needed anymore

mwojtyczka (approved):
LGTM

mwojtyczka:
### Scope check: what does the filter guard actually protect against?

I dug into whether a destructive keyword can even execute through the filter path, and I think it changes how we should frame (and scope) this PR.

**Finding:** `filter` / `row_filter` / `sql_expression` compile via `F.expr(...)` / `df.filter("<str>")`, which parse with Spark's **expression** grammar, not the statement grammar. A destructive *statement* is not a valid expression, so it fails at analysis time. Confirmed in a real runtime:

```
F.expr("id = 1 OR DROP TABLE users")          # lazy — just returns Column<'...'>, no parse yet
df.filter("id = 1 OR DROP TABLE users").collect()
# [PARSE_SYNTAX_ERROR] Syntax error at or near 'users': extra input 'users'. SQLSTATE: 42601
```

Nothing is dropped — `DROP`/`TABLE` are read as an identifier + alias, then parsing chokes on the trailing token. Same for `DELETE FROM ...`, and `id = 1; DROP TABLE users` doesn't stack a statement either. The only genuinely powerful expression is a `SELECT` subquery (`id IN (SELECT ...)`) — which this PR correctly **allows**.

So on the filter paths, the destructive-keyword blocklist is guarding something Spark's grammar already makes impossible. Its observable effect is turning a `ParseException` (which `_is_invalid_column` already catches → check skipped) into an eager `UnsafeSqlQueryError`. **The real statement-execution surface is `spark.sql(query)` (the `sql_query` check) and `run_config_name` interpolation** — and those pre-date this PR.

**That said — I don't think the guard is worthless.** There's real *diagnostic / DX* value:
- A specific `UnsafeSqlQueryError: Unsafe filter expression: '...'` is far more actionable than `[PARSE_SYNTAX_ERROR] ... extra input 'users'`, which points at the wrong token and gives no hint the problem is a statement-shaped filter.
- A typed exception is a stable signal for callers/tests/telemetry, vs. a `ParseException` that conflates "you wrote a destructive statement" with "you typo'd a column name."

**Suggestions before merge:**

1. **Reword the security claim.** The description says a destructive filter "would run a subquery against another table with the permissions of the process." That's not true for the filter path (the grammar blocks it) — it's only true for the `spark.sql` path. Please scope the wording so this isn't approved as a security fix it isn't, and frame the filter guard as defense-in-depth + better diagnostics.

2. **Deliver the message via skip, not a batch-aborting raise.** Today `manager.filter_condition` raises out of the whole `apply_checks` loop (there's no try/except around `manager.process()` in `engine.py`), so one bad filter aborts *all* checks and produces zero results. Every other invalid filter is handled by skipping just that check with a clear message. Routing the "unsafe filter" message through that same skip path keeps the good diagnostic **and** the per-check blast radius — best of both.

Net: keep the explicit guard for the DX win, but (a) don't over-claim security, and (b) surface it through the existing skip mechanism rather than a hard raise. Non-blocking — the code is correct; this is about scope/framing and one behavior change.

ghanse (approved):
LGTM pending the cosmetic change to the PR description

mwojtyczka:
### Follow-up: does this PR actually touch the paths that need guarding?

I mapped every `is_sql_query_safe` consumer against what this PR changes. The answer sharpens the scope concern above: **the PR's only *new* guarding is on the filter paths (non-exploitable), and it leaves the real statement-execution paths exactly as they were on `main`.**

| Call site | Path type | Existed on `main`? | What this PR did |
|---|---|---|---|
| `check_funcs.py` `_validate_sql_query_params` | **`spark.sql` statement** (the `sql_query` check) | ✅ yes | **untouched** |
| `check_funcs.py:~1607` | statement-execution helper | ✅ yes | **untouched** |
| `llm/llm_core.py` | LLM-generated query validation | ✅ yes | **untouched** |
| `safe_filter_expr` (`utils.py`) | **filter expression** via `F.expr` | ❌ new | **added by this PR** |

So the paths where a `DROP`/`DELETE` genuinely *does* parse and execute (`spark.sql`, LLM query) were **already guarded before this PR** — it neither strengthens nor extends them. The new `safe_filter_expr` wiring targets the `F.expr` filter paths, which the expression grammar already makes non-executable (the `PARSE_SYNTAX_ERROR` evidence above).

There is one indirect effect: the PR **rewrites the body of `is_sql_query_safe`** (literal/comment stripping, `forbid_select` removal). Because the statement paths *share* that function, they inherit the new behavior — including the double-quoted-literal stripping, which slightly widens the false-negative surface on the one path that actually executes statements.

**Which raises the honest question: do we need this change at all?** Options:

1. **Drop the filter-guarding entirely.** The grammar already prevents statement execution on those paths; the pre-existing skip behavior already handled unparseable filters. Net risk removed by the PR ≈ 0.
2. **Keep only the diagnostic value** — a specific "unsafe filter" message beats a raw `PARSE_SYNTAX_ERROR` — but deliver it through the existing *skip* path (not a batch-aborting raise), and be careful the shared-validator rewrite doesn't regress the `spark.sql` path.
3. **Re-aim at the real surface.** If we want added security value, invest in the `spark.sql` / `run_config_name` statement paths (ideally parser-based validation), which is where statements actually run.

I'd lean toward (2) if we value the DX, or (1) if we don't — but I don't think the PR as scoped closes a real hole. Flagging so we decide deliberately rather than merging it as a security fix.

ghanse:
I can see value in detecting and surfacing this type of error and skipping checks that attempt to inject any malicious DML.

There are a few instances of `spark.sql(...)` that option 3 may address. I think it makes sense to open this as a follow-up issue.

mwojtyczka:
> I can see value in detecting and surfacing this type of error and skipping checks that attempt to inject any malicious DML.
> 
> There are a few instances of `spark.sql(...)` that option 3 may address. I think it makes sense to open this as a follow-up issue.

makes sense, we should skip such malicious checks rather than halting the whole process, treat them as invalid

mwojtyczka:
> ### Follow-up: does this PR actually touch the paths that need guarding?
> I mapped every `is_sql_query_safe` consumer against what this PR changes. The answer sharpens the scope concern above: **the PR's only _new_ guarding is on the filter paths (non-exploitable), and it leaves the real statement-execution paths exactly as they were on `main`.**
> 
> Call site	Path type	Existed on `main`?	What this PR did
> `check_funcs.py` `_validate_sql_query_params`	**`spark.sql` statement** (the `sql_query` check)	✅ yes	**untouched**
> `check_funcs.py:~1607`	statement-execution helper	✅ yes	**untouched**
> `llm/llm_core.py`	LLM-generated query validation	✅ yes	**untouched**
> `safe_filter_expr` (`utils.py`)	**filter expression** via `F.expr`	❌ new	**added by this PR**
> So the paths where a `DROP`/`DELETE` genuinely _does_ parse and execute (`spark.sql`, LLM query) were **already guarded before this PR** — it neither strengthens nor extends them. The new `safe_filter_expr` wiring targets the `F.expr` filter paths, which the expression grammar already makes non-executable (the `PARSE_SYNTAX_ERROR` evidence above).
> 
> There is one indirect effect: the PR **rewrites the body of `is_sql_query_safe`** (literal/comment stripping, `forbid_select` removal). Because the statement paths _share_ that function, they inherit the new behavior — including the double-quoted-literal stripping, which slightly widens the false-negative surface on the one path that actually executes statements.
> 
> **Which raises the honest question: do we need this change at all?** Options:
> 
> 1. **Drop the filter-guarding entirely.** The grammar already prevents statement execution on those paths; the pre-existing skip behavior already handled unparseable filters. Net risk removed by the PR ≈ 0.
> 2. **Keep only the diagnostic value** — a specific "unsafe filter" message beats a raw `PARSE_SYNTAX_ERROR` — but deliver it through the existing _skip_ path (not a batch-aborting raise), and be careful the shared-validator rewrite doesn't regress the `spark.sql` path.
> 3. **Re-aim at the real surface.** If we want added security value, invest in the `spark.sql` / `run_config_name` statement paths (ideally parser-based validation), which is where statements actually run.
> 
> I'd lean toward (2) if we value the DX, or (1) if we don't — but I don't think the PR as scoped closes a real hole. Flagging so we decide deliberately rather than merging it as a security fix.

refactored to use option 2, skip the filter

mwojtyczka:
### Follow-up fixes (commit `f1b5e3d4`)

Addressed the findings from the latest code-review pass:

- **Positional `row_filter` could bypass the skip gate (correctness).** `_effective_filter` now resolves the filter from the check function's *bound arguments* (via `inspect.signature.bind_partial`), so a `row_filter` supplied positionally through `check_func_args` — not just `check.filter` or the `row_filter` kwarg — is detected and skipped. Previously such a filter slipped past the manager and hard-raised in the executor, reintroducing the batch-abort the skip change was meant to remove.
- **No engine-level skip-not-abort coverage (tests).** Added an integration test that drives the real `DQEngine.apply_checks` with an unsafe-filter check alongside a normal check, and asserts the unsafe one is reported `skipped = true` while the normal check still evaluates. Also added a manager unit test for the positional-`row_filter` case.
- **Log injection on the skip message (CWE-117).** The filter string is now routed through a new `sanitize_for_logging` helper (escapes CR/LF) before being interpolated into the warning log and the result message. Applied to both the unsafe- and invalid-filter branches.
- **Cleanup.** `_effective_filter` is now a `cached_property` (was recomputed ~4× per check); and `has_no_aggr_outliers` / `_is_aggr_compare` no longer compute a discarded `safe_filter_expr(None)` → `F.lit(True)` when there is no filter (inlined into the existing `if row_filter` guard, matching the `has_no_outliers` fix).

`make fmt` is clean (Black, Ruff, mypy, Pylint 10.00/10). Unit suites pass; the new engine-level and `has_no_outliers` filter tests are integration tests (need a workspace) and will run under `make integration`.

PR #1309, opened 2026-07-07
Feature/validate upstream tables

### LINKED ISSUE #1046: [FEATURE]: Validation of upstream tables

### Problem statement

In ingestion scenarios, it is often critical to validate that the number of records matches the upstream source

### Proposed Solution

New Dataset-level check function to verify metrics (e.g., row count or custom SQL logic) against the upstream table(s).

### Additional Context

_No response_

STEFANOVIVAS:
## Changes

* Added new check function: `aggr_matches_dataset`— a new dataset-level check that compares an aggregate metric (row count by default, or any curated/built-in aggregate) computed on the checked DataFrame against the same aggregate computed on a reference (upstream) table or DataFrame, with optional abs_tolerance/rel_tolerance and independent row_filter/ref_row_filter on each side.

### Linked issues

Resolves #1046 

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [x] added performance tests

### Documentation and Demos

- [ ] added/updated demos
- [x] added/updated docs


src/databricks/labs/dqx/check_funcs.py
```diff
@@ -2061,6 +2061,159 @@ def apply(df: DataFrame) -> DataFrame:
     return condition, apply
 
 
+@register_rule("dataset")
+def validate_upstream_table(
```
ghanse: I think a naming convention like "matches_upstream_table" might be clearer. For example `aggr_matches_upstream_table` for this check.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -2061,6 +2061,159 @@ def apply(df: DataFrame) -> DataFrame:
...
+            F.lit(f"{aggr_display_name} value "),
+            F.col(metric_col).cast("string"),
+            F.lit(f" in column '{aggr_col_str}' does not match {aggr_display_name} value "),
+            F.col(ref_metric_col).cast("string"),
+            F.lit(f" in column '{ref_col_str}' of upstream {ref_label}"),
+        ),
+        alias=normalize_col_str(f"{aggr_col_str_norm}_{aggr_type.lower()}_not_equal_to_upstream".lstrip("_")),
+    )
+
+    return condition, apply
```
ghanse: It feels like we should use `_is_aggr_compare(...)` here to reduce duplication. We may need to update that method to conditionally create the `message` value. What do you think?

ghanse (dismissed):
Headed in the right direction. Left some comments. Do you also want to add a similar check for a custom SQL query?

src/databricks/labs/dqx/check_funcs.py
STEFANOVIVAS: Hmm..that makes sense!

STEFANOVIVAS: Hi, @ghanse ! The updated _is_aggr_compare function ended up with too many parameters and is activating some pylint warnings. "max-args=12 (14 args), too-many-locals (28 vs 24), too-many-statements (54 vs 50), and McCabe complexity (12, extension defaults to 10)."

I added 3 new parameters to the older version of the function: 

`null_safe`: If True, forces a tolerance-based comparison (even when both tolerances are 0)  and treats two NULL metrics as matching while a one-sided NULL is treated as a mismatch. Only meaningful for *compare_op* in (operator.eq, operator.ne). Used when *limit* is a  dynamic Column (e.g., an aggregate computed from another DataFrame) that may be NULL.

`message_builder`: Optional callable that receives the metric Column expression and returns  the full violation message Column. Overrides the default generic message, letting callers  (e.g., upstream table comparisons) surface additional context such as a reference table name.
        
`alias`: Optional name for the check result column. Overrides the default name derived from *column*, *aggr_type*, *group_by* and *compare_op_name*.

Should I create a new class to accommodate some of the attributes or do you see another better solution?

docs/dqx/docs/reference/quality_checks.mdx
```diff
@@ -1828,6 +1828,7 @@ You can also define your own custom dataset-level checks (see [Creating custom c
...
...umn name or a column expression), pass `'*'` for `count(*)` over all rows; `ref_table`: (optional) fully qualified reference table name; `ref_df_name`: (optional) name of the reference DataFrame (dictionary of DataFrames can be passed when applying checks); either `ref_df_name` or `ref_table` must be provided but never both; `ref_column`: (optional) column to aggregate in the reference DataFrame or Table, defaults to `column` when not provided; `aggr_type`: aggregation function (default: "count"), supports 20 curated functions (count, sum, avg, stddev, percentile, etc.) plus any Databricks built-in aggregate; `aggr_params`: (optional) dict of parameters for aggregates requiring them; `row_filter`: (optional) SQL expression to filter rows in the checked DataFrame before aggregation; `ref_row_filter`: (optional) SQL expression to filter rows in the reference DataFrame or Table before aggregation; `abs_tolerance`: (optional) absolute tolerance for equality comparison of numeric aggregations (formula: `abs(a - b) <= tolerance`); `rel_tolerance`: (optional) relative tolerance for equality comparison of numeric aggregations (formula: `abs(a - b) <= rel_tolerance * max(abs(a), abs(b))`) |
```
mwojtyczka: maybe make the name more generic: `validate_upstream_dataset` since it can be a dataframe or a table

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: Maybe keep it even more generic: `aggr_matches_upstream_dataset` since it can be a table or a dataframe

mwojtyczka (commented):
Review of `validate_upstream_table`. Useful check and largely correct; main themes are: it **reimplements existing aggregate machinery** (`_is_aggr_compare` + `_get_ref_df`) rather than composing it, and a few edges around filter validation, the empty-upstream message, and the exact-equality default. Inline below.

**Cleared:** `rel_tolerance` has no div-by-zero (guarded by `max()`); the `apply` closure captures `spark`/`ref_dfs` correctly (standard dataset-check pattern); `.limit(1)` matches `_is_aggr_compare`; behavioral coverage lives in the integration tests (dataset checks can't be unit-tested without Spark).

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: **Altitude / DRY.** This largely re-implements machinery that already exists: tolerance validation + the non-curated-aggregate warning + aggregate-expression building + the null-safe tolerance match duplicate `_is_aggr_compare`, and the `ref_df_name`/`ref_table` resolution duplicates `_get_ref_df`. The message strings have already drifted slightly from `_is_aggr_compare`. Consider composing instead: resolve the reference via `_get_ref_df`, compute its scalar aggregate, and delegate the comparison to `_is_aggr_compare` (or a thin shared helper). Also collapses the extra Spark jobs noted below.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -2061,6 +2061,159 @@ def apply(df: DataFrame) -> DataFrame:
...
+            ref_dfs: Dictionary of reference DataFrames (by name), used to resolve *ref_df_name*.
+
+        Returns:
+            The DataFrame with additional condition and metric columns for upstream comparison.
+        """
+        ref_df = _get_ref_df(ref_df_name, ref_table, ref_dfs, spark)
+
+        # Filter rows (rather than nulling out excluded values via F.when) so that
+        # column="*" (count(*) over all rows) works correctly with row_filter/ref_row_filter.
+        filtered_df = df.filter(row_filter) if row_filter else df
```
mwojtyczka: `row_filter` / `ref_row_filter` are passed to `df.filter(<raw str>)` (here and line 2174) without `is_sql_query_safe()` validation. This is the same gap PR #1303 is closing project-wide (other aggregate checks are raw today too), so it's consistent — but a new user-facing filter param should validate. Ideally route both through the `safe_filter_expr` helper #1303 introduces once it lands. (Note: a `; DROP TABLE` string can't execute here — `df.filter` parses an expression, not statements; the real risk is a subquery reading other data with the engine's identity.)

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -2061,6 +2061,159 @@ def apply(df: DataFrame) -> DataFrame:
...
+        return result_df.withColumn(condition_col, ~match)
+
+    condition = make_condition(
+        condition=F.col(condition_col),
+        message=F.concat_ws(
+            "",
+            F.lit(f"{aggr_display_name} value "),
+            F.col(metric_col).cast("string"),
+            F.lit(f" in column '{aggr_col_str}' does not match {aggr_display_name} value "),
+            F.col(ref_metric_col).cast("string"),
```
mwojtyczka: When the reference aggregate is NULL (empty upstream on `sum`/`avg`), `concat_ws("", …)` silently drops this cast and leaves a **blank value** — the message reads `…does not match Sum value  in column 'b'…` (double space). The integration test (test file ~line 481) asserts this broken output, enshrining it. Suggest `F.coalesce(F.col(ref_metric_col).cast("string"), F.lit("NULL"))` (same for the checked side) and updating the test to the readable message.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -2061,6 +2061,159 @@ def apply(df: DataFrame) -> DataFrame:
...
+            - if *abs_tolerance* or *rel_tolerance* is negative.
+    """
+    if ref_df_name is not None and ref_table is not None:
+        raise InvalidParameterError(
+            "Both 'ref_df_name' and 'ref_table' were provided. Please provide only one to avoid ambiguity."
+        )
+    if not ref_df_name and not ref_table:
+        raise MissingParameterError("Either 'ref_df_name' or 'ref_table' is required but neither was provided.")
+
+    abs_tolerance = 0.0 if abs_tolerance is None else abs_tolerance
```
mwojtyczka: With neither tolerance provided (both → `0.0`), the check requires **exact** equality. Because it compares two separately-computed aggregates, `sum`/`avg` over floating-point columns can differ across runs/clusters (non-associative summation) even for identical data → spurious violations. Worth documenting in the docstring that a non-zero tolerance is recommended for non-integer aggregates (and/or defaulting to a small `rel_tolerance`).

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -2061,6 +2061,159 @@ def apply(df: DataFrame) -> DataFrame:
...
+              columns.
+
+    Raises:
+        MissingParameterError:
+            - if neither *ref_df_name* nor *ref_table* is provided.
+        InvalidParameterError:
+            - if both *ref_df_name* and *ref_table* are provided.
+            - if *abs_tolerance* or *rel_tolerance* is negative.
+    """
+    if ref_df_name is not None and ref_table is not None:
```
mwojtyczka: Mutual-exclusion guard uses `is not None`, but the presence check just below uses truthiness — inconsistent. `ref_df_name="my_ref", ref_table=""` (empty string, e.g. a config-template default) hits this branch and raises "both provided" instead of routing to `ref_df_name`. Use `if ref_df_name and ref_table:`.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -2061,6 +2061,159 @@ def apply(df: DataFrame) -> DataFrame:
...
+        aggr_expr = _build_aggregate_expression(aggr_type, aggr_col_expr, aggr_params)
+
+        ref_filtered_df = ref_df.filter(ref_row_filter) if ref_row_filter else ref_df
+        ref_aggr_expr = _build_aggregate_expression(aggr_type, ref_col_expr, aggr_params)
+
+        metric_df = filtered_df.select(aggr_expr.alias(metric_col)).limit(1)
+        ref_metric_df = ref_filtered_df.select(ref_aggr_expr.alias(ref_metric_col)).limit(1)
+
+        if aggr_type not in CURATED_AGGR_FUNCTIONS:
+            _validate_aggregate_return_type(metric_df, aggr_type, metric_col)
```
mwojtyczka: Minor / pre-existing: `_validate_aggregate_return_type` rejects Array/Map/Struct but not `StringType`/`BooleanType`. A non-curated aggregate like `first` on a string column passes this check, then `_match_values_with_tolerance`'s `F.abs(str - str)` throws a raw Spark `AnalysisException` at execution rather than a clean `InvalidParameterError`. Shared with the existing `is_aggr_*` checks — noting for awareness. Also, this is called twice (once per side) — folding into the composed approach above avoids the duplicate validation pass.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -2061,6 +2061,159 @@ def apply(df: DataFrame) -> DataFrame:
...
+    ref_column = column if ref_column is None else ref_column
+
+    aggr_col_str_norm, aggr_col_str, aggr_col_expr = get_normalized_column_and_expr(column)
+    _, ref_col_str, ref_col_expr = get_normalized_column_and_expr(ref_column)
+    aggr_display_name = _get_aggregate_display_name(aggr_type, aggr_params)
+    ref_label = f"table '{ref_table}'" if ref_table else f"DataFrame '{ref_df_name}'"
+
+    unique_str = uuid.uuid4().hex  # make sure any column added to the dataframe is unique
+    condition_col = f"__condition_{aggr_col_str_norm}_{aggr_type}_upstream_{unique_str}"
+    metric_col = f"__metric_{aggr_col_str_norm}_{aggr_type}_{unique_str}"
```
```suggestion
    metric_col = f"__metric_{aggr_col_str_norm}_{aggr_type}_upstream_{unique_str}"
```

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -2061,6 +2061,159 @@ def apply(df: DataFrame) -> DataFrame:
...
+    ref_metric_col = f"__ref_metric_{aggr_col_str_norm}_{aggr_type}_{unique_str}"
```
```suggestion
    ref_metric_col = f"__ref_metric_{aggr_col_str_norm}_{aggr_type}_upstream_{unique_str}"
```

mwojtyczka (changes requested):
Going in the right direction, left some comments

src/databricks/labs/dqx/check_funcs.py
STEFANOVIVAS: In fact there isn’t even an obligation to be an upstream table, right? It’s just as just a validation between tables/dataframes.

src/databricks/labs/dqx/check_funcs.py
ghanse: This is what I'm thinking.

**Reducing arguments:**
We can group the following into an internal config class like `AggrComparisonConfig`:

- `aggr_type`
- `aggr_params`
- `compare_op`
- `compare_op_label`
- `compare_op_name`
- `abs_tolerance`
- `rel_tolerance`

**Reducing statements and complexity:**
We can move some validations to internal functions. For example, this can be moved to an `_validate_aggr_functions(...)` method:

```
    is_curated = aggr_type in CURATED_AGGR_FUNCTIONS
    if not is_curated:
        warnings.warn(
            f"Using non-curated aggregate function '{aggr_type}'. "
            f"Curated functions: {', '.join(sorted(CURATED_AGGR_FUNCTIONS))}. "
            f"Non-curated aggregates must return a single numeric value per group.",
            UserWarning,
            stacklevel=3,
        )
```

This can be moved to an `_validate_aggr_tolerances(...)` method:

```
    abs_tolerance = 0.0 if abs_tolerance is None else abs_tolerance
    rel_tolerance = 0.0 if rel_tolerance is None else rel_tolerance
    if abs_tolerance < 0 or rel_tolerance < 0:
        raise InvalidParameterError("Absolute and/or relative tolerances if provided must be non-negative")
```

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: @STEFANOVIVAS  Correct, can you pls rename it to `aggr_matches_dataset`? this will also require update to docs and docstrings. We can have upstream table validation as one of the examples where this could be used.

mwojtyczka (approved):
LGTM. Just one request to rename the function to `aggr_matches_dataset` to make it more generic and update docs and docs strings accordingly

STEFANOVIVAS:
> LGTM. Just one request to rename the function to `aggr_matches_dataset` to make it more generic and update docs and docs strings accordingly

Hi, @mwojtyczka! One thing to point out: Should we implement a group by clause in this check? _is_aggr_compare function already let us apply it in the input Dataframe. Could make sense because in many situations we need to make a per-group reconciliation (date, region,source_system,etc).

mwojtyczka:
> > LGTM. Just one request to rename the function to `aggr_matches_dataset` to make it more generic and update docs and docs strings accordingly
> 
> Hi, @mwojtyczka! One thing to point out: Should we implement a group by clause in this check? _is_aggr_compare function already let us apply it in the input Dataframe. Could make sense because in many situations we need to make a per-group reconciliation (date, region,source_system,etc).

yes makes sense

STEFANOVIVAS:
Hi @mattfaltyn! I don't know if I do understand your bug report correctly but i think that this pull request already fix it. I created a helper function called _build_aggr_condition_result which accepts a param called `null_safe_limit_compare`.

`null_safe_limit_compare`: Only applies when *compare_op* is *operator.eq* or *operator.ne*. When
            True, a NULL metric or NULL limit is never left to standard SQL NULL propagation: both NULL
            is treated as a match (no violation), a one-sided NULL is treated as a mismatch (violation).
            Use this when the limit can legitimately be NULL (e.g., an aggregate computed over an
            external/reference dataset that may be empty), so a NULL limit isn't silently treated as
            "no violation".

mattfaltyn:
@STEFANOVIVAS Thanks! I tested this against PR head `96ae5f4c`. `null_safe_limit_compare` handles NULL aggregate values or limits, but #1354 concerns NULL join-key components preventing computed results from being reattached. The `sql_query`, grouped aggregate, and outlier joins still use ordinary equality, and the original reproduction still fails. I therefore don’t think #1309 resolves #1354; those result-reattachment joins still need null-safe equality and focused regression coverage.

mwojtyczka:
@STEFANOVIVAS pls also add the check to `tests/resources/all_dataset_checks.yaml`, and adjust the relevant tests. We also need a performance test so that we can validate regressions in the future

mwojtyczka:
## Code review — `aggr_matches_dataset`

Overall this is a high-quality, well-tested addition (23 integration tests + unit exception cases, careful null-safety handling, behavior-preserving refactor of `_is_aggr_compare`). No blocking bugs found in the happy paths. Notes below are minor correctness edge-cases and test-coverage gaps.

> Note: the non-null-safe grouped-join issue I initially flagged for `count_distinct` + `group_by` is already being fixed in #1356 (it replaces the `df.join(agg_df, on=join_cols, how="left")` in `_is_aggr_compare` with a null-safe helper). Once #1356 lands, `aggr_matches_dataset`'s grouped `count_distinct` path inherits the fix — so it's excluded here.

### 1. Not wired into the "all dataset checks" coverage (test-coverage)
`tests/resources/all_dataset_checks.yaml` (+ the all-checks integration tests) — `aggr_matches_dataset` isn't included, unlike sibling dataset checks (`compare_datasets`, `foreign_key`, …). The declarative metadata / YAML round-trip for this check is therefore unexercised.

### 2. No `aggr_matches_dataset` test for `count_distinct` **with** `group_by` (test-coverage)
`tests/integration/test_dataset_checks.py` — the check is tested with `count_distinct` (no group_by) and with `group_by` (count/sum), but not both together. #1356 adds a grouped-`count_distinct` regression for the base `is_aggr_*` path; a small `aggr_matches_dataset`-specific case (ideally with a NULL group key, to lock in the #1356 fix for this check) would close the loop.

mwojtyczka (changes requested):
Code looks good, left comments regading tests

mwojtyczka:
> ## Code review — `aggr_matches_dataset`
> Overall this is a high-quality, well-tested addition (23 integration tests + unit exception cases, careful null-safety handling, behavior-preserving refactor of `_is_aggr_compare`). No blocking bugs found in the happy paths. Notes below are minor correctness edge-cases and test-coverage gaps.
> 
> > Note: the non-null-safe grouped-join issue I initially flagged for `count_distinct` + `group_by` is already being fixed in #1356 (it replaces the `df.join(agg_df, on=join_cols, how="left")` in `_is_aggr_compare` with a null-safe helper). Once #1356 lands, `aggr_matches_dataset`'s grouped `count_distinct` path inherits the fix — so it's excluded here.
> 
> ### 1. Not wired into the "all dataset checks" coverage (test-coverage)
> `tests/resources/all_dataset_checks.yaml` (+ the all-checks integration tests) — `aggr_matches_dataset` isn't included, unlike sibling dataset checks (`compare_datasets`, `foreign_key`, …). The declarative metadata / YAML round-trip for this check is therefore unexercised.
> 
> ### 2. No `aggr_matches_dataset` test for `count_distinct` **with** `group_by` (test-coverage)
> `tests/integration/test_dataset_checks.py` — the check is tested with `count_distinct` (no group_by) and with `group_by` (count/sum), but not both together. #1356 adds a grouped-`count_distinct` regression for the base `is_aggr_*` path; a small `aggr_matches_dataset`-specific case (ideally with a NULL group key, to lock in the #1356 fix for this check) would close the loop.

I fixed these

mwojtyczka (approved):
LGTM

mwojtyczka (approved):
LGTM

mwojtyczka:
Heads-up: I pushed a small CI addition to this branch (`96ef0896`) — an `mcp-tests` job in `.github/workflows/nightly.yml` so the MCP integration suite runs on the nightly schedule.

It's unrelated to the `aggr_matches_dataset` change here — it's a gap we missed when the MCP server landed: `mcp.yml` only triggers on PRs touching `mcp-server/`, so nightly never exercised the suite. Bundling the fix here for expediency.

PR #1317, opened 2026-07-10
FEATURE: Added profile build for "has_no_outliers" check

### LINKED ISSUE #977: [FEATURE]: Add has_no_outliers function to automated generated rules with profiler

### Problem statement

The newly created has_no_outlier check function is only available for manual user input.

### Proposed Solution

Add has_no_outliers function to automated generated rules with profiler/generator.

### Additional Context

_No response_

IvannKurchenko:
## Changes
Summary of changes:
- Extracted MAD related calculations to common package;
- Refactored profiler internal - extracted configs and default to constants;
- Implemented `has_no_outliers` profile build to generate corresponding check;
- Corresponding documentation updated;

### Linked issues

Resolves https://github.com/databrickslabs/dqx/issues/977

### Tests

- [ ] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

### Documentation and Demos

- [ ] added/updated demos
- [x] added/updated docs


src/databricks/labs/dqx/profiler/profile_builder.py
```diff
@@ -646,3 +676,58 @@ def _round_decimal(value: decimal.Decimal, rounding_direction: str) -> decimal.D
...
+    """
+    if not _is_numeric(column_type):
+        return None
+
+    total_count = profiler_metrics.get("count", 0)
+    if total_count == 0:
+        return None
+
+    filter_condition = profiler_options.get(PROFILE_OPTION_FILTER, None)
+    bounds = calculate_median_absolute_deviation_bounds(df, column_name, filter_condition)
```
mwojtyczka: **🔴 Blocking — crashes when a profiler `filter` is set.** The profiler passes a single-column `df.select(field).dropna()` to builders, and the dataset filter is already applied upstream in `_sample`. Re-passing `filter_condition` here makes `calculate_median_absolute_deviation_bounds` run `df.filter(<filter>)` on that single-column frame, so any filter referencing another column (e.g. `filter="status = 'active'"` while profiling `value`) raises `AnalysisException: Column 'status' not found`. It's also a redundant second application. Suggest not re-applying the filter here — the frame is already filtered.

src/databricks/labs/dqx/profiler/profile_builder.py
```diff
@@ -9,11 +9,28 @@
...
+    PROFILE_OPTION_ROUND,
+    PROFILE_OPTION_TRIM_STRINGS,
+)
 
 # Type alias for annotations; use TEXT_TYPES for isinstance() checks.
 TextType = T.CharType | T.StringType | T.VarcharType
 TEXT_TYPES: tuple[type[TextType], ...] = (T.CharType, T.StringType, T.VarcharType)
 
+NumericType = T.IntegerType | T.LongType | T.FloatType | T.DoubleType
+NUMERICAL_TYPES: tuple[type[NumericType], ...] = (T.IntegerType, T.LongType, T.FloatType, T.DoubleType)
```
mwojtyczka: `NUMERICAL_TYPES` (Integer/Long/Float/Double) is narrower than what the `has_no_outliers` **check** accepts (`T.NumericType`), so `ByteType`/`ShortType`/`DecimalType` columns are silently skipped for profiling even though the check supports them. Also, the local `NumericType = T.IntegerType | …` alias shadows Spark's `T.NumericType` (confusing — `_supports_min_max` already uses `isinstance(x, T.NumericType)`). Prefer `isinstance(column_type, T.NumericType)`.

src/databricks/labs/dqx/profiler/profile_builder.py
```diff
@@ -646,3 +676,58 @@ def _round_decimal(value: decimal.Decimal, rounding_direction: str) -> decimal.D
...
+    if total_count == 0:
+        return None
+
+    filter_condition = profiler_options.get(PROFILE_OPTION_FILTER, None)
+    bounds = calculate_median_absolute_deviation_bounds(df, column_name, filter_condition)
+    if bounds is None:
+        return None
+
+    lower_bound, upper_bound = bounds
+    outliers_count = df.filter((F.col(column_name) < lower_bound) | (F.col(column_name) > upper_bound)).count()
```
mwojtyczka: `outliers_count` is counted on the `dropna()`'d single-column frame, but `total_count` is `profiler_metrics["count"]`. If that count includes nulls, the ratio is under-reported (nulls sit in the denominator but never the numerator), so `has_no_outliers` can be emitted/suppressed incorrectly on nullable columns. Please confirm `metrics["count"]` is the non-null count so numerator and denominator are consistent.

src/databricks/labs/dqx/profiler/profile_options.py
```diff
@@ -0,0 +1,75 @@
...
+PROFILE_OPTION_FILTER = "filter"
+"""SQL expression used to pre-filter the DataFrame before profiling."""
+
+PROFILE_OPTION_LLM_PRIMARY_KEY_DETECTION = "llm_primary_key_detection"
+"""Whether to invoke the LLM engine to detect primary-key columns and emit an *is_unique* rule."""
+
+PROFILE_OPTION_OUTLIERS_RATIO = "outliers_ratio"
+"""Outlier ratio threshold; emit a *has_no_outliers* profile only when the outlier fraction is below this value."""
+
+DEFAULT_PROFILE_OPTIONS: dict[str, Any] = {
```
mwojtyczka: `dict[str, Any]` — AGENTS.md: *"Avoid `Any` … New code outside [anomaly] must not introduce it."* These options are `bool | int | float | None`; typing as `dict[str, bool | int | float | None]` keeps the safety.

src/databricks/labs/dqx/calculations_utils.py
```diff
@@ -0,0 +1,60 @@
+"""Common calculations shared between different layers of the framework."""
+
+from pyspark.sql import DataFrame
+from pyspark.sql import functions as F
+
+
+def calculate_median_absolute_deviation_bounds(
```
mwojtyczka: Altitude: shared PySpark-stats helpers already have homes (`profiling_utils.py`, `profiler/common.py`). A new top-level `calculations_utils.py` fragments the utility surface. Please place these MAD helpers in one of the existing modules rather than a new root-level one.

mwojtyczka (changes requested):
Re-review of the `has_no_outliers` profile builder. The MAD extraction and option-constant refactor are clean, but there are **two blocking issues** plus some cleanup. Most points are inline; two that can't be anchored to this diff:

### 🔴 Blocking — `has_no_outliers` profiles never become checks
`DQGenerator._checks_mapping` (`src/databricks/labs/dqx/profiler/generator.py`) maps only `is_not_null, is_in, min_max, is_not_null_or_empty, is_not_empty, is_unique`, and `generate_dq_rules()` skips any profile whose name isn't a key. This PR builds a `DQProfile(name="has_no_outliers")` but never adds a mapping entry / `dq_generate_has_no_outliers`, and doesn't touch `generator.py` — so the profile is silently dropped and **no check is generated end-to-end**, which is the PR's stated goal.

### Test coverage
Unit tests cover only the non-numeric early-exit; the positive / threshold-negative paths and the MAD math are integration-only, and **nothing tests the profile → `generate_dq_rules` → check path**. That's exactly why the mapping gap above and the filter crash (inline) both slip through — the tests call the builder directly with a single-column frame and no cross-column filter. A generate-path test and a cross-column-filter test would catch both blocking issues.

See inline comments for the filter crash (blocking), numeric-type narrowing, ratio denominator, the `Any` convention, and module placement.

src/databricks/labs/dqx/profiler/profile_options.py
ghanse, line 1: Stylistically I think we should move the comments above the options and maybe just use single-line comments instead of block comments. This is more aligned with the rest of the repo.

src/databricks/labs/dqx/profiler/profile_builder.py
ghanse: Yes, we should use PySpark's `NumericType` instead of creating our own. The reason we created `TextType` was that PySpark does not have a built-in base class for text types.

src/databricks/labs/dqx/calculations_utils.py
ghanse: I would add these to `profiling_utils.py`

src/databricks/labs/dqx/calculations_utils.py
```diff
@@ -0,0 +1,60 @@
...
+    Args:
+        df: PySpark DataFrame
+        column: Name of the numeric column to calculate MAD for
+        filter_condition: SQL filter expression to apply before calculation (optional)
+
+    Returns:
+        A (median, mad) tuple. Both values are None when the filtered DataFrame is empty.
+    """
+    if filter_condition is not None:
+        df = df.filter(filter_condition)
```
ghanse: Filtering is handled in `profiler.py`. I think it's best to leave that responsibility to the `profile(...)` method and assume the input to these functions is a pre-filtered DataFrame.

tests/integration/test_profile_builder.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def test_make_has_no_outliers_profile_outliers_below_threshold(spark):
+    """
+    Test expects profile to be created for the following case: 11 values with one extreme value.
+    1 outlier out of 11 ≈ 9 %. The threshold configuration is 10 %, hence profile is expected to return.
+    MAD bounds: median=6; MAD=3; lower=-4.5; upper=16.5; Hence only 1000 is an outlier.
+    """
+    data = [(1,), (2,), (3,), (4,), (5,), (6,), (7,), (8,), (9,), (10,), (1000,)]
+    df = spark.createDataFrame(data, "col: int")
```
ghanse: It would be good to test the other numeric types. Maybe you can `pytest.mark_parametrize` and pass the column type as a string to parameterize the DataFrame schema?

tests/integration/test_profile_builder.py
```diff
@@ -0,0 +1,54 @@
...
+def test_make_has_no_outliers_profile_outliers_above_threshold(spark):
+    """
+    Test expects no profile for the following case: 7 values with three extreme values.
+    3 outliers out of 7 ≈ 43 %. The threshold configuration is 10 %, hence no profile is expected to return.
+    MAD bounds: median=4; MAD=3; lower=-6.5; upper=14.5; Hence 100, 200, 300 are outliers.
+    """
+    # [1..4] + three extreme values → 3 outliers out of 7 ≈ 43 %, threshold 10 % → None
+    # MAD bounds: median=4, MAD=3 → lower=-6.5, upper=14.5 → 100, 200, 300 are outliers
+    data = [(1,), (2,), (3,), (4,), (100,), (200,), (300,)]
+    df = spark.createDataFrame(data, "col: int")
```
ghanse: Same parameterization can be applied here.

tests/integration/test_profile_builder.py
ghanse, line 1: It would also be good to add an integration test for a non-empty DataFrame with all null values. We should get `None` as a result of profiling.

ghanse (changes requested):
This will be a nice addition. Agree with the other feedback. Left a few small comments.

src/databricks/labs/dqx/calculations_utils.py
IvannKurchenko: Thank you so much for the feedback and review. Small note: this logic is also used for `has_no_outliers` check as well, where filter can be applied.

ghanse: Yes, makes sense. We can leave it for the check functions.

IvannKurchenko:
@mwojtyczka  @mwojtyczka Thank you so much for your review. May I ask for second round?

src/databricks/labs/dqx/profiler/profile_builder.py
```diff
@@ -646,3 +661,57 @@ def _round_decimal(value: decimal.Decimal, rounding_direction: str) -> decimal.D
...
+
+    outliers_ratio = float(outliers_count) / total_non_null_count
+    outliers_ratio_threshold = profiler_options.get(PROFILE_OPTION_OUTLIERS_RATIO, 0.01)
+
+    if outliers_ratio < outliers_ratio_threshold:
+        return DQProfile(
+            name="has_no_outliers",
+            description=f"Column {column_name} has {outliers_ratio * 100:.1f}% of outliers (allowed: {outliers_ratio_threshold * 100:.1f}%). Lower boundary - {lower_bound}, upper boundary - {upper_bound}.",
+            column=column_name,
+            filter=None,
```
mwojtyczka: **The generated `has_no_outliers` rule drops the profiler `filter`.** Every other builder emits `filter=profiler_options.get(PROFILE_OPTION_FILTER, None)` (e.g. lines 117, 214, 276, 462), but this hardcodes `filter=None`. With `options={'filter': "region='EU'"}`, the bounds/ratio are profiled on the EU subset and the rule is emitted, but at apply time the check has no `row_filter` and re-derives MAD over **all** rows — a different population than was profiled, and inconsistent with the `is_not_null`/`min_max` rules generated in the same run (see `test_profile_with_dataset_filter`). Propagate the filter like the sibling builders.

src/databricks/labs/dqx/profiler/profiler_runner.py
```diff
@@ -65,18 +76,18 @@ def run(
...
+            PROFILE_OPTION_LIMIT: run_config.profiler_config.limit,
+            PROFILE_OPTION_FILTER: run_config.profiler_config.filter,
+            PROFILE_OPTION_LLM_PRIMARY_KEY_DETECTION: run_config.profiler_config.llm_primary_key_detection,
         }
         if run_config.profiler_config.max_null_ratio is not None:
-            options["max_null_ratio"] = run_config.profiler_config.max_null_ratio
+            options[PROFILE_OPTION_MAX_NULL_RATIO] = run_config.profiler_config.max_null_ratio
         if run_config.profiler_config.max_empty_ratio is not None:
-            options["max_empty_ratio"] = run_config.profiler_config.max_empty_ratio
+            options[PROFILE_OPTION_MAX_EMPTY_RATIO] = run_config.profiler_config.max_empty_ratio
```
mwojtyczka, line 92: **`outliers_ratio` isn't plumbed through `ProfilerConfig` / `ProfilerRunner`.** Its sibling ratios `max_null_ratio` / `max_empty_ratio` are wired here (and are fields on `ProfilerConfig` in `config.py`), but there's no `outliers_ratio` field or plumbing. So the documented `outliers_ratio` knob only takes effect via a direct `DQProfiler.profile(options=...)` call — through the installed workflow/CLI it silently stays at the `0.01` default. Add an `outliers_ratio` field to `ProfilerConfig` and thread it here (and in `run_for_patterns`) alongside the other two.

src/databricks/labs/dqx/profiler/profile_builder.py
```diff
@@ -646,3 +661,57 @@ def _round_decimal(value: decimal.Decimal, rounding_direction: str) -> decimal.D
     if rounding_direction == "up":
         return value.to_integral_value(rounding=decimal.ROUND_CEILING)
     return value
+
+
+@register_profile_builder("has_no_outliers")
+def make_has_no_outliers_profile(
```
mwojtyczka, line 768: Outlier profiling runs for **every numeric column** with no enable/disable toggle (unlike `remove_outliers`), and issues 3 extra Spark actions per column — two `percentile_approx` collects in `calculate_median_absolute_deviation_bounds` plus the `filter().count()` below — whose computed bounds are then **discarded** (the `has_no_outliers` check recomputes them at apply time). On wide tables this roughly triples per-numeric-column profiling cost with no opt-out. Consider a profiler option to gate it, and/or folding the outlier count into a single aggregation.

src/databricks/labs/dqx/profiler/profile_builder.py
```diff
@@ -646,3 +661,57 @@ def _round_decimal(value: decimal.Decimal, rounding_direction: str) -> decimal.D
...
+    if bounds is None:
+        return None
+
+    lower_bound, upper_bound = bounds
+    outliers_count = df.filter((F.col(column_name) < lower_bound) | (F.col(column_name) > upper_bound)).count()
+
+    outliers_ratio = float(outliers_count) / total_non_null_count
+    outliers_ratio_threshold = profiler_options.get(PROFILE_OPTION_OUTLIERS_RATIO, 0.01)
+
+    if outliers_ratio < outliers_ratio_threshold:
```
mwojtyczka: Degenerate-distribution edge: when all non-null values are identical, MAD=0 so the bounds collapse to `median±0`; `outliers_count` is 0, the ratio passes, and a `has_no_outliers` rule is emitted. At apply time the check recomputes MAD=0 and then flags **any** value ≠ median as an outlier — an overly strict rule for a constant column, unlikely to be the intent. Consider skipping emission (or widening) when MAD == 0 / the distribution is degenerate.

src/databricks/labs/dqx/profiler/profile_builder.py
```diff
@@ -646,3 +661,57 @@ def _round_decimal(value: decimal.Decimal, rounding_direction: str) -> decimal.D
...
+    total_non_null_count = profiler_metrics.get("count_non_null", 0)
+    if total_non_null_count == 0:
+        return None
+
+    bounds = calculate_median_absolute_deviation_bounds(df, column_name)
+    if bounds is None:
+        return None
+
+    lower_bound, upper_bound = bounds
+    outliers_count = df.filter((F.col(column_name) < lower_bound) | (F.col(column_name) > upper_bound)).count()
```
mwojtyczka: Edge (Decimal columns): outliers are counted here with raw Python-float bounds (`F.col(...) < lower_bound`), while the `has_no_outliers` check builds its comparison via `get_limit_expr(lower_bound)`. For `Decimal` columns (now accepted since the guard is `NumericType`), a float-vs-Decimal comparison can coerce/round differently than the typed check expression, so the builder's emit/skip decision may not match what the generated check actually flags near the boundary. Consider building the bounds comparison the same way the check does.

src/databricks/labs/dqx/profiler/generator.py
```diff
@@ -443,11 +443,33 @@ def dq_generate_is_unique(column: str, criticality: str = "error", **params: dic
             "user_metadata": user_metadata,
         }
 
+    @staticmethod
+    def dq_generate_has_no_outliers(column: str, criticality: str = "error", **params: dict):
```
mwojtyczka, line 447: Docstring uses reST fields (`:param:` / `:return:`), but the project standard (AGENTS.md — Google Style docstrings) and every sibling generator (e.g. `dq_generate_is_unique`) use `Args:` / `Returns:`. Please convert for consistent API-doc rendering.

src/databricks/labs/dqx/profiler/profile_builder.py
```diff
@@ -646,3 +661,57 @@ def _round_decimal(value: decimal.Decimal, rounding_direction: str) -> decimal.D
...
+    column_name: str,
+    column_type: T.DataType,
+    profiler_metrics: dict[str, Any],
+    profiler_options: dict[str, Any],
+) -> DQProfile | None:
+    """
+    Creates a *has_no_outliers* profile using the same MAD method as the *has_no_outliers* check rule.
+
+    A profile is returned when all the following conditions are met:
+    - The column type is numeric (integer, long, float, or double).
```
mwojtyczka: Minor: the docstring says the column type is "integer, long, float, or double", but the guard is `isinstance(column_type, T.NumericType)`, which also accepts Byte/Short/Decimal (correctly covered by the parametrized tests). Update the wording so it doesn't re-introduce the narrow-list description.

mwojtyczka (commented):
Re-review of `794cf15`. **All prior findings (mine + @ghanse's) are addressed** — the profiler-`filter` crash is fixed, the builder uses `NumericType`, the outlier ratio uses `count_non_null` consistently, `dict[str, Any]` is gone, `calculations_utils.py` was folded into `profiling_utils.py` (MAD shared by the check and the builder), and the requested tests (numeric-type parametrization, all-nulls→None) were added. Thanks for the thorough turnaround.

New findings below. **#1 (filter dropped) and #2 (`outliers_ratio` not plumbed through `ProfilerConfig`/`ProfilerRunner`)** are the two I'd fix before merge — they make the feature behave inconsistently with the rest of the profiler and silently ignore a documented config. The rest are a perf/toggle decision, two edge cases (zero-MAD, Decimal bounds), and doc/style cleanup.

Minor (not inline): the `0.01` default is duplicated between `DEFAULT_PROFILE_OPTIONS` and the inline `.get(..., 0.01)` fallback; `dq_generate_has_no_outliers` has an unused `params` (`params = params or {}` is dead); and `column_name` is interpolated into the logged `DQProfile.description` without newline sanitization (CWE-117, low — schema-derived).

tests/integration/test_profiler.py
```diff
@@ -339,6 +353,13 @@ def test_profiler_non_default_profile_options(spark, ws):
             parameters={"min": 1, "max": 3},
             filter="t1 > 0",
         ),
+        DQProfile(
+            name='has_no_outliers',
+            column='t1',
+            description='Column t1 has 0.0% of outliers (allowed: 1.0%). Lower boundary - -1.5, upper boundary - 5.5.',
+            parameters=None,
+            filter=None,
```
ghanse: This expectation is now stale and this test will fail. Following the earlier review, the builder was updated to propagate the dataset filter (`profile_builder.py`: `filter=profiler_options.get(PROFILE_OPTION_FILTER, None)`), and this run passes `"filter": "t1 > 0"`. Every sibling profile in this same `expected_profiles` list carries `filter="t1 > 0"`, so `has_no_outliers` will too — which makes `assert profiles == expected_profiles` fail here.

```suggestion
            filter="t1 > 0",
```

The unit test `test_has_no_outliers_filter_propagated` already asserts the propagated value; these integration tests need a live workspace, which is likely why the stale expectation slipped through.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

tests/integration/test_profiler.py
```diff
@@ -1714,6 +1749,13 @@ def test_profile_with_dataset_filter(spark, ws):
             filter="machine_id IN ('MCH-002', 'MCH-003') AND maintenance_type = 'preventive'",
             description="Real min/max values were used",
         ),
+        DQProfile(
+            name='has_no_outliers',
+            column='cost',
+            description='Column cost has 0.0% of outliers (allowed: 1.0%). Lower boundary - -25.0, upper boundary - 325.0.',
+            parameters=None,
+            filter=None,
```
ghanse: Same stale-filter problem as in `test_profiler_non_default_profile_options`. The builder now propagates the profiler filter, and every sibling profile in this list carries the `machine_id IN (...)` filter, so the generated `has_no_outliers` profile will too — this `filter=None` makes `assert profiles == expected_profiles` fail.

```suggestion
            filter="machine_id IN ('MCH-002', 'MCH-003') AND maintenance_type = 'preventive'",
```

This is also the exact consistency point raised in the previous round: at apply time the generated check must re-derive MAD over the same filtered population it was profiled on, so the emitted profile has to carry the filter.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

src/databricks/labs/dqx/profiler/profile_builder.py
```diff
@@ -646,3 +736,75 @@ def _round_decimal(value: decimal.Decimal, rounding_direction: str) -> decimal.D
...
+    below_lower_bound_expr = F.col(column_name) < get_limit_expr(lower_bound)
+    above_upper_bound_expr = F.col(column_name) > get_limit_expr(upper_bound)
+    outside_bounds_expr = below_lower_bound_expr | above_upper_bound_expr
+    outliers_count = df.filter(outside_bounds_expr).count()
+
+    if lower_bound == upper_bound:
+        logger.info(
+            f"MAD bounds are equal for column '{column_name}'. All values are equal in the distribution. Skipping profile generation."
+        )
+        return None
```
ghanse, line 824: `outliers_count` triggers a Spark `count()` action on line 790, but when the distribution is degenerate (`lower_bound == upper_bound`, i.e. MAD == 0) that work is discarded and the function returns `None` two lines later. Move the degenerate-bounds early-return above the count so the extra Spark job is skipped for constant columns — exactly the columns most likely to hit this path:

```suggestion
    lower_bound, upper_bound = bounds
    if lower_bound == upper_bound:
        logger.info(
            f"MAD bounds are equal for column '{column_name}'. All values are equal in the distribution. Skipping profile generation."
        )
        return None

    below_lower_bound_expr = F.col(column_name) < get_limit_expr(lower_bound)
    above_upper_bound_expr = F.col(column_name) > get_limit_expr(upper_bound)
    outside_bounds_expr = below_lower_bound_expr | above_upper_bound_expr
    outliers_count = df.filter(outside_bounds_expr).count()
```

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

src/databricks/labs/dqx/profiler/profile_builder.py
```diff
@@ -646,3 +736,75 @@ def _round_decimal(value: decimal.Decimal, rounding_direction: str) -> decimal.D
...
+
+    outliers_ratio = float(outliers_count) / total_non_null_count
+    outliers_ratio_threshold = profiler_options[PROFILE_OPTION_OUTLIERS_RATIO]
```
ghanse: Nit (optional): this direct subscript `KeyError`s if the builder is invoked without merged defaults, whereas every sibling builder reads options via `.get(key, default)`. Through `DQProfiler.profile` the key is always present (merged from `DEFAULT_PROFILE_OPTIONS`), so it's safe today — but `make_has_no_outliers_profile` is a module-level registered builder, and a direct caller (or a future unit test) that omits `outliers_ratio` would crash instead of falling back.

The previous round flagged duplicating the `0.01` literal, so rather than re-introducing `\.get(..., 0.01)`, you can default to the canonical value to stay both robust and DRY:

```suggestion
    outliers_ratio_threshold = profiler_options.get(
        PROFILE_OPTION_OUTLIERS_RATIO, DEFAULT_PROFILE_OPTIONS[PROFILE_OPTION_OUTLIERS_RATIO]
    )
```

(requires adding `DEFAULT_PROFILE_OPTIONS` to the existing `profile_options` import.)

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

tests/integration/test_profiler.py
IvannKurchenko: Thank you

mwojtyczka:
## Code review — profile build for `has_no_outliers`

Solid, well-tested feature. The MAD extraction into `profiling_utils` (shared by the check and the profiler) is a clean DRY win, the option-constants extraction into `profile_options.py` improves maintainability, and edge handling is careful (empty df, constant column `lower==upper`, log-injection guard on the column name). Test coverage is strong — unit tests cover non-numeric, zero-count, no-outliers, exceeds-threshold, none-bounds, allow/deny columns, both-lists error, and filter propagation, plus integration + generator tests. No import cycle introduced (`profiling_utils` doesn't import `check_funcs`).

A few smaller notes:

### Sampling caveat (inference, not a bug — worth a doc note)
The outlier ratio is measured on the **sampled** DataFrame (default `sample_fraction=0.3`, `limit=1000`), but the emitted `has_no_outliers` check runs against the **full** dataset at apply time. A column that looks clean in-sample may be noisier at full scale (and vice-versa), so the emit/skip decision is sample-based. Worth a one-line note in the profiler docs so users understand the threshold is evaluated on the sample.

### Minor: unused `params` in `dq_generate_has_no_outliers`
`generator.py` — the method accepts `**params`, then does `params = params or {}` and never uses it. Understandable for signature-compatibility with the other `dq_generate_*` methods (and the docstring says so), but the reassignment is dead code and could be dropped.

### Nit: profiled bounds aren't carried into the generated check
The generated rule is `{"function": "has_no_outliers", "arguments": {"column": ...}}` with no bounds — so `outliers_ratio` only gates whether the rule is *emitted*; the check recomputes its own `median ± 3.5×MAD` at apply time. This appears intentional (the check is self-contained), but since the profiler already computes `lower_bound`/`upper_bound`, it's worth confirming there's no expectation that the profiled bounds be baked into the rule.

Nice, thorough PR overall.

mwojtyczka (approved):
LGTM

mwojtyczka:
@IvannKurchenko — I pushed a commit (`0e8bf389`) to this branch addressing the review points from my earlier comment. Heads-up since it's a direct push to your PR; 

What it changes:

- **Emission gate `outliers_ratio`**: switched the strict `<` to inclusive `<=`, to match the sibling `max_null_ratio` / `max_empty_ratio` gates (a column whose outlier fraction exactly equals the threshold now emits a rule). Docstring + docs updated to "at or below".
- **allow/deny mutual-exclusion**: moved the `has_no_outliers_allow_columns` / `has_no_outliers_deny_columns` "only one of them" validation to a single up-front `validate_profile_options()` call in `DQProfiler.profile` (before any sampling/profiling work), instead of raising per-column deep in the run. The per-column check stays as defense-in-depth.
- **Degenerate-distribution guard**: `make_has_no_outliers_profile` now skips when the MAD band width is non-positive or near-zero *relative to the column scale*, not only when the bounds are exactly equal — so a tiny-but-nonzero MAD doesn't emit a rule that flags nearly every row at apply time.
- **Docstring**: fixed the stale `_build_profiles_for_column` note (order is now `null_or_empty → is_in → min_max → has_no_outliers`; `min_max` is no longer last).
- **Docs**: added a note that ratio thresholds are evaluated on the **sampled** data (via the existing `<Admonition>` component).
- **Tests**: added unit tests for at-threshold emission, the near-degenerate MAD skip, and the up-front validation raise/pass paths.

Two things I intentionally did **not** change:
- The `params = params or {}` line in `dq_generate_has_no_outliers` — it's the shared house pattern for consuming the `**params` dispatch arg (all sibling `dq_generate_*` methods do the same, and removing it trips pylint's unused-argument check).
- Baking the profiled `lower/upper` bounds into the generated rule — the `has_no_outliers` check is intentionally self-contained (recomputes MAD at apply time) and doesn't take bound parameters, so that would be a separate API change.

src/databricks/labs/dqx/profiler/profile_options.py
```diff
@@ -0,0 +1,85 @@
...
+    PROFILE_OPTION_MAX_EMPTY_RATIO: 0.01,
+    PROFILE_OPTION_SAMPLE_FRACTION: 0.3,
+    PROFILE_OPTION_SAMPLE_SEED: None,
+    PROFILE_OPTION_SAMPLE_BY_COLUMN: None,
+    PROFILE_OPTION_SAMPLE_BY_VALUES_LIMIT: 1000,
+    PROFILE_OPTION_LIMIT: 1000,
+    PROFILE_OPTION_FILTER: None,
+    PROFILE_OPTION_LLM_PRIMARY_KEY_DETECTION: True,
+    PROFILE_OPTION_OUTLIERS_RATIO: 0.01,
+    PROFILE_OPTION_HAS_NO_OUTLIERS: True,
```
ghanse: We are enabling this option by default. There may be a significant performance impact for customers who use the profiler today. IMO we should either default to `False` or communicate this change clearly in our release notes.

ghanse (approved):
Implementation LGTM.

Profiling for outliers will be enabled by default due to `PROFILE_OPTION_HAS_NO_OUTLIERS: True`. Should we disable by default to preserve the existing behavior and performance?

src/databricks/labs/dqx/profiler/profile_options.py
IvannKurchenko: Thanks for raising the concern. Disabled

mwojtyczka: makes sense, let's disable

IvannKurchenko:
@mwojtyczka Im sorry for failed integration tests, I'll work on fix

mwojtyczka:
> @mwojtyczka Im sorry for failed integration tests, I'll work on fix

no worries, let me know if you need help

mwojtyczka:
> > @mwojtyczka Im sorry for failed integration tests, I'll work on fix
> 
> no worries, let me know if you need help

I think you just missed enabling it in the test, just updated

PR #1318, opened 2026-07-10
Fix rule source mismatch

### LINKED ISSUE #1315: [BUG] YAML import fails to save rules: UI sends source "imported" but DB constraint only allows "import"

### Current Behavior

Saving rules through **Rules → Import → From DQX YAML** in DQX Studio always fails with a CHECK constraint violation. The import page sends `source: "imported"`, but the `dq_quality_rules` table constraint only accepts `('ui', 'sql', 'profiler', 'import', 'ai')`.

From the user's perspective the failure is silent: the UI shows a generic "failed to save rules" toast, nothing appears under Drafts & Review, and the real error is only visible in the app logs.

Root cause: `app/src/databricks_labs_dqx_app/ui/routes/_sidebar/rules.import.tsx` sends `source: "imported"` in both `handleSaveAsDrafts` and `handleSubmitForReview` (lines ~398 and ~412). The contract import tab (`rules.from-contract.tsx:324`) correctly sends `"import"`, which is why only the YAML tab is affected. Both the Delta migration and the Postgres/Lakebase DDL enforce the same allowed values, so both OLTP backends should be affected.

### Expected Behavior

Rules pasted in the YAML import tab are saved as draft rules (visible under Drafts & Review) and can be submitted for approval — same as rules created via the single-table editor or the contract import tab.

A one-word fix resolves it: `source: "imported"` → `source: "import"` (2 occurrences in `rules.import.tsx`). Verified locally: after the change, YAML import saves and submits normally.

Suggestion: typing `SaveRulesIn.source` as `Literal["ui", "sql", "profiler", "import", "ai"]` in `backend/models.py` would surface this class of mismatch as a 422 at the API boundary instead of a storage-layer 500.

### Steps To Reproduce

1. Deploy DQX Studio as a Databricks App with the Delta OLTP fallback (`DQX_LAKEBASE_INSTANCE_NAME="-"`)
2. Open **Rules → Import → From DQX YAML**
3. Paste any valid check YAML (e.g. a simple `is_not_null` check), click Validate (passes)
4. Select a target table
5. Click **Save as drafts** (or **Submit for review**)
6. Observe the error toast; check Drafts & Review — nothing was saved. App logs show the constraint violation.

### Cloud

Azure

### Operating System

Windows

```shell
ERROR | rules.save_rules | Failed to save rules for <catalog>.<schema>.<table>: SQL execution failed:
[DELTA_VIOLATE_CONSTRAINT_WITH_VALUES] CHECK constraint chk_dq_quality_rules_source
(source IN ('ui', 'sql', 'profiler', 'import', 'ai')) violated by row with values:
 - source : imported.
SQL: INSERT INTO <catalog>.dqx_studio.dq_quality_rules (rule_id, table_fqn, `check`, version, status, source, created_by, created_at, updated_by, updated_at)
VALUES ('<rule_id>', '<catalog>.<schema>.<table>', parse_json('{"name": "...", "criticality": "warn", "check": {...}}'), 1, 'draft', 'imported', '<user>', now(), '<user>', now())

Traceback (most recent call last):
  File ".../backend/routes/v1/rules.py", line 187, in save_rules
    entries = svc.save(body.table_fqn, body.checks, user_email, source=body.source)
  File ".../backend/services/rules_catalog_service.py", line 177, in save
    self._sql.execute(sql)
  File ".../backend/sql_executor.py", line 399, in execute
    raise RuntimeError(f"SQL execution failed: {msg}\nSQL: {sql}")
RuntimeError: SQL execution failed: [DELTA_VIOLATE_CONSTRAINT_WITH_VALUES] ...
```

ghanse:
## Changes

This PR fixes a mismatch in the rule source value for rules imported through the DQX App.

### Linked issues

Resolves #1315 

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

### Documentation and Demos

- [ ] added/updated demos
- [ ] added/updated docs

### DIFF
--- a/app/src/databricks_labs_dqx_app/backend/models.py
+++ b/app/src/databricks_labs_dqx_app/backend/models.py
@@ -1,3 +1,4 @@
+from enum import Enum
 from typing import Any
 
 from databricks.labs.dqx.config import RunConfig, WorkspaceConfig
@@ -147,17 +148,46 @@ class RuleCatalogEntryOut(BaseModel):
     updated_at: str | None = None
 
 
+class RuleSource(Enum):
+    """Source (e.g. 'ui', 'profiler') where the rule was created."""
+
+    ui = "ui"
+    sql = "sql"
+    profiler = "profiler"
+    user_import = "import"
+    ai = "ai"
+
+    @classmethod
+    def sql_in_list(cls) -> str:
+        """Renders the members as a SQL-safe list for 'IN' expressions."""
+        return ", ".join(f"'{member.value}'" for member in cls)
+
+
+class RuleStatus(Enum):
+    """Lifecycle status of a rule in the catalog."""
+
+    draft = "draft"
+    pending_approval = "pending_approval"
+    approved = "approved"
+    rejected = "rejected"
+
+    @classmethod
+    def sql_in_list(cls) -> str:
+        """Renders the members as a SQL-safe list for 'IN' expressions."""
+        return ", ".join(f"'{member.value}'" for member in cls)
+
+
 class SaveRulesIn(BaseModel):
     table_fqn: str = Field(description="Fully qualified table name (catalog.schema.table)")
     checks: list[dict[str, Any]] = Field(description="List of check metadata dictionaries")
-    source: str = Field(default="ui", description="Origin of the rules: ui, imported, or ai")
+    source: RuleSource = Field(default=RuleSource.ui, description="Origin of the rules: ui, imported, or ai")
     rule_id: str | None = Field(default=None, description="If set, update existing rule instead of creating")
 
 
 class BatchSaveRulesIn(BaseModel):
     table_fqns: list[str] = Field(description="Fully qualified table names to apply the checks to")
     checks: list[dict[str, Any]] = Field(description="List of check metadata dictionaries")
-    source: str = Field(default="ui", description="Origin of the rules: ui, imported, or ai")
+    source: RuleSource = Field(default=RuleSource.ui, description="Origin of the rules: ui, imported, or ai")
 
 
 class BatchSaveRulesOut(BaseModel):
@@ -208,7 +238,7 @@ class FilterTablesByColumnsOut(BaseModel):
 
 
 class SetStatusIn(BaseModel):
-    status: str = Field(description="New status: draft | pending_approval | approved | rejected")
+    status: RuleStatus = Field(description="New status: draft | pending_approval | approved | rejected")
     expected_version: int | None = Field(
         default=None,
         description="If provided, the update is rejected when the current version does not match (optimistic concurrency).",
--- a/app/src/databricks_labs_dqx_app/backend/models.py
+++ b/app/src/databricks_labs_dqx_app/backend/models.py
@@ -134,20 +134,6 @@ class GenerateRulesFromContractOut(BaseModel):
     )
 
 
-class RuleCatalogEntryOut(BaseModel):
-    table_fqn: str
-    display_name: str = ""
-    checks: list[dict[str, Any]]
-    version: int
-    status: str
-    source: str = "ui"
-    rule_id: str | None = None
-    created_by: str | None = None
-    created_at: str | None = None
-    updated_by: str | None = None
-    updated_at: str | None = None
-
-
 class RuleSource(Enum):
     """Source (e.g. 'ui', 'profiler') where the rule was created."""
 
@@ -161,6 +147,20 @@ class RuleSource(Enum):
     def sql_in_list(cls) -> str:
         """Renders the members as a SQL-safe list for 'IN' expressions."""
         return ", ".join(f"'{member.value}'" for member in cls)
+
+
+class RuleCatalogEntryOut(BaseModel):
+    table_fqn: str
+    display_name: str = ""
+    checks: list[dict[str, Any]]
+    version: int
+    status: str
+    source: RuleSource = RuleSource.ui
+    rule_id: str | None = None
+    created_by: str | None = None
+    created_at: str | None = None
+    updated_by: str | None = None
+    updated_at: str | None = None
 
 
 class RuleStatus(Enum):
@@ -180,14 +180,18 @@ class RuleStatus(Enum):
 class SaveRulesIn(BaseModel):
     table_fqn: str = Field(description="Fully qualified table name (catalog.schema.table)")
     checks: list[dict[str, Any]] = Field(description="List of check metadata dictionaries")
-    source: RuleSource = Field(default=RuleSource.ui, description="Origin of the rules: ui, imported, or ai")
+    source: RuleSource = Field(
+        default=RuleSource.ui, description="Origin of the rules: ui, sql, profiler, import, or ai"
+    )
     rule_id: str | None = Field(default=None, description="If set, update existing rule instead of creating")
 
 
 class BatchSaveRulesIn(BaseModel):
     table_fqns: list[str] = Field(description="Fully qualified table names to apply the checks to")
     checks: list[dict[str, Any]] = Field(description="List of check metadata dictionaries")
-    source: RuleSource = Field(default=RuleSource.ui, description="Origin of the rules: ui, imported, or ai")
+    source: RuleSource = Field(
+        default=RuleSource.ui, description="Origin of the rules: ui, sql, profiler, import, or ai"
+    )
 
 
 class BatchSaveRulesOut(BaseModel):
--- a/app/src/databricks_labs_dqx_app/backend/services/rules_catalog_service.py
+++ b/app/src/databricks_labs_dqx_app/backend/services/rules_catalog_service.py
@@ -6,7 +6,7 @@ from datetime import datetime, timezone
 from typing import Any
 from uuid import uuid4
 
-from databricks_labs_dqx_app.backend.models import RuleStatus
+from databricks_labs_dqx_app.backend.models import RuleSource, RuleStatus
 from databricks_labs_dqx_app.backend.sql_executor import OltpExecutorProtocol
 from databricks_labs_dqx_app.backend.sql_utils import escape_sql_string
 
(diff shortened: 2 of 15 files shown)


app/src/databricks_labs_dqx_app/ui/routes/_sidebar/rules.active.tsx
```diff
@@ -632,14 +632,14 @@ function ByTableView({ groups, expandedTables, onToggle, onNavigate, canDelete,
...
-                                rule.source === "imported"
+                                rule.source === "import"
                                   ? "border-blue-500 text-blue-600"
                                   : rule.source === "ai"
                                     ? "border-purple-500 text-purple-600"
                                     : "border-emerald-500 text-emerald-600"
                               }`}
                             >
-                              {rule.source === "imported" ? t("rulesActive.imported") : rule.source === "ai" ? t("rulesActive.ai") : t("rulesActive.ui")}
+                              {rule.source === "import" ? t("rulesActive.imported") : rule.source === "ai" ? t("rulesActive.ai") : t("rulesActive.ui")}
```
mwojtyczka: **Missed call site — the fix is incomplete.** You corrected the badge here (`rule.source === "import"`), but the **source-filter dropdown at line 403** still compares `src === "imported"`:

```tsx
{src === "imported" ? t("rulesActive.imported") : src === "ai" ? t("rulesActive.ai") : t("rulesActive.ui")}
```

Since imported rules now carry `source: "import"`, that branch is never true, so imported rules are labeled **"UI"** in the filter selector — the same mislabeling #1315 is about, still present. Please update line 403 to `=== "import"` too (and grep for any other `"imported"` literals).

app/src/databricks_labs_dqx_app/backend/models.py
```diff
@@ -147,17 +148,46 @@ class RuleCatalogEntryOut(BaseModel):
     updated_at: str | None = None
 
 
+class RuleSource(Enum):
```
mwojtyczka, line 137: **The read/output path isn't typed as `RuleSource`, which is why the missed `"imported"` literal above couldn't be caught.** The input models were migrated (`SaveRulesIn.source`/`BatchSaveRulesIn.source: RuleSource`), but the **output** model is still a plain string — `RuleCatalogEntryOut.source: str = "ui"` (line 143), mirrored as `source?: string` in the generated `api.ts`. So the value the UI badge/filter compares is never narrowed to the union: TypeScript can't flag `src === "imported"`, and a stray/legacy source flows through unvalidated (falling through to the "ui" label). Typing the read model as `RuleSource` (and regenerating `api.ts`) would make this class of drift a compile error rather than a silent mislabel.

app/src/databricks_labs_dqx_app/backend/services/rules_catalog_service.py
```diff
@@ -55,7 +56,7 @@ class RulesCatalogService:
     so existing callers don't have to change.
     """
 
-    VALID_STATUSES = {"draft", "pending_approval", "approved", "rejected"}
+    VALID_STATUSES = {member.value for member in RuleStatus}
 
     VALID_TRANSITIONS: dict[str, set[str]] = {
```
mwojtyczka, line 62: You validate status transitions here (`VALID_TRANSITIONS`/`VALID_STATUSES`), but there's no equivalent guard for `source`: the `save`/history layer types it as a plain `str`, and only the request models enforce `RuleSource`. A non-route caller (or the batch partial-failure path) passing an unknown source string reaches the `INSERT` and trips the DB `CHECK` constraint, surfacing as a **500** rather than a clean validation error. Consider a `VALID_SOURCES` guard mirroring `VALID_STATUSES`, or typing the service param as `RuleSource`.

src/databricks/labs/dqx/engine.py
```diff
@@ -460,12 +463,11 @@ def _validate_dbr_version_requirements(self, checks: list[DQRule]) -> None:
...
             return
 
-        # Resolve the leading "major.minor" prefix, tolerating suffixes such as "18.2.x-photon-scala2.13"
-        # (serverless) or "15.4 LTS".
-        match = re.match(r"\s*(\d+)\.(\d+)", dbr_version_str)
+        match = re.match(r"\s*(\d+)(?:\.(\d+))?", dbr_version_str)
         if match is None:
             raise InvalidCheckError(f"Cannot parse Databricks Runtime version: '{dbr_version_str}'.")
-        current = (int(match.group(1)), int(match.group(2)))
+        minor = int(match.group(2)) if match.group(2) is not None else sys.maxsize
```
mwojtyczka, line 469: **Fail-open on same major, and this change looks out of scope for a "rule source" PR.** With `minor = ... else sys.maxsize`, a bare-major runtime like `"17.x"` becomes `(17, sys.maxsize)`, so it satisfies **any** same-major requirement — e.g. a serverless runtime reporting `"17.x"` (actually 17.0) passes `requires_dbr_version("17.5")`, and a check needing a 17.5-only feature then fails at execution instead of being blocked up front. If that leniency is intended, worth a comment/test pinning it; either way this DBR-version-parsing change (and its test) is unrelated to the rule-source fix — consider splitting it out or calling it out in the PR description.

app/src/databricks_labs_dqx_app/backend/models.py
```diff
@@ -147,17 +148,46 @@ class RuleCatalogEntryOut(BaseModel):
...
+    def sql_in_list(cls) -> str:
+        """Renders the members as a SQL-safe list for 'IN' expressions."""
+        return ", ".join(f"'{member.value}'" for member in cls)
+
+
 class SaveRulesIn(BaseModel):
     table_fqn: str = Field(description="Fully qualified table name (catalog.schema.table)")
     checks: list[dict[str, Any]] = Field(description="List of check metadata dictionaries")
-    source: str = Field(default="ui", description="Origin of the rules: ui, imported, or ai")
+    source: RuleSource = Field(default=RuleSource.ui, description="Origin of the rules: ui, imported, or ai")
```
mwojtyczka: Minor: the field description still says `"... ui, imported, or ai"` while the value is now `"import"` (same on `BatchSaveRulesIn` at line 190). It's the exact string that caused the bug — worth updating to avoid future confusion.

tests/unit/test_engine.py
```diff
@@ -527,3 +529,9 @@ def test_apply_checks_parses_dbr_version_with_suffix(dbr_version, required_check
         match=rf"require Databricks Runtime >= .*but the current version is {re.escape(dbr_version)}",
     ):
         engine.apply_checks(df, [DQRowRule(check_func=required_check, column="a")])
+
+
+@pytest.mark.parametrize("dbr_version", ["17.x-photon-scala2.13", "18.x-photon-scala2.13", "17.x"])
+def test_validate_dbr_version_bare_major_serverless_satisfies_same_major(dbr_version):
+    engine = _engine_returning_dbr_version(dbr_version)
+    engine._validate_dbr_version_requirements([DQRowRule(check_func=_check_requires_dbr_17_1, column="a")])
```
mwojtyczka: This new test calls the protected `engine._validate_dbr_version_requirements([...])` directly. AGENTS.md testing standards / Critical Rule 6: *"Do not call or access private/protected members … from tests — exercise the public API instead."* The sibling tests drive the same version-parsing logic through the public `engine.apply_checks(df, [...])`, so the public seam is available here too.

mwojtyczka (commented):
## Review — Fix rule source mismatch (resolves #1315)

The fix (align the UI/models from `"imported"` to `"import"`, matching the DB `CHECK` constraint that always expected `'import'`, and migrate the input models to a `RuleSource` enum) is the right direction, and no DB backfill is needed — old `"imported"` inserts would have been rejected by the constraint, so there are no stranded rows. But the alignment is **incomplete**, and an unrelated `engine.py` change is bundled in.

**Findings (inline), most-severe first:**
1. **Missed call site (`rules.active.tsx:403`)** — the source-**filter dropdown** still compares `"imported"`, so imported rules are mislabeled "UI" in the filter. The PR doesn't fully resolve #1315 without this.
2. **Read path not typed as `RuleSource`** — `RuleCatalogEntryOut.source` is still `str` (backend + generated `api.ts`), which is exactly why #1 wasn't caught at compile time; typing it is the durable fix that prevents recurrence.
3. **Service layer accepts unvalidated `source: str`** — non-route callers can hit a raw `CHECK`-constraint 500 instead of a clean validation error.
4. **`engine.py` bare-major fail-open + out-of-scope** — `"17.x"` satisfies even `"17.5"`; also unrelated to the rule-source fix.
5. Minor: `RuleSource` field descriptions still say "imported".
6. New unit test reaches into a protected method.

Nice catch on the root mismatch — #1 and #2 are the two I'd land before merge (fully close the loop and type the read path so it can't drift again).

mwojtyczka (changes requested):

src/databricks/labs/dqx/engine.py
ghanse: A _"17.X"_ DBR version on serverless typically matches the highest available minor version within a major version family.

This change is to resolve issues with broken DBR version checks during testing.

tests/unit/test_engine.py
ghanse: Removed this test. This should be covered by integration tests.

app/src/databricks_labs_dqx_app/backend/models.py
ghanse: Updated the descriptions.

app/src/databricks_labs_dqx_app/backend/services/rules_catalog_service.py
ghanse: Added a set of `VALID_SOURCES` and a validation block.

app/src/databricks_labs_dqx_app/backend/models.py
ghanse: Updated the `RuleCatalogEntryOut` model to use `RuleSource`.

app/src/databricks_labs_dqx_app/ui/routes/_sidebar/rules.active.tsx
ghanse: Fixed

mwojtyczka (approved):
LGTM

PR #1319, opened 2026-07-11
Persist check metadata across table storage backends (message_expr + typed user_metadata)

### LINKED ISSUE #1308: `[FEATURE]: Faithfully persist check metadata across table storage backends (user_metadata types + message_expr)`

### Problem statement

The table storage backends do not faithfully persist all `CheckSpec` / `DQRule` fields, so a check can lose information on a save→load round-trip through a table — but not through File/Volume storage. Two concrete gaps, sharing the same root cause (the table at-rest schema doesn't fully represent a check):

**1. Non-string `user_metadata` values are coerced to strings on the Delta path.**
`user_metadata` is `dict[str, Any]`, so a check can carry `{"confidence": 0.95}` or `{"enabled": true}`. `CHECKS_TABLE_SCHEMA` types the column as `MAP<STRING, STRING>` and writes the raw dict, so PySpark's StringType converter coerces every value via `str()` — `0.95` → `"0.95"`, `true` → `"true"`. Lakebase (`JSONB`) and File/Volume (YAML/JSON) preserve the types; only Delta loses them. Note that `arguments` in the same schema already avoids this via `json.dumps`-on-write + `safe_json_load`-on-read.

**2. `message_expr` is not persisted by either table backend.**
`message_expr` (custom check message, added in #1092 / #958) is a first-class field and is in the load-path allow-list, but there is no column for it in `CHECKS_TABLE_SCHEMA` (Delta) and it is omitted from the Lakebase column definitions and row dict. So it round-trips only via File/Volume; through any table backend it is silently dropped on write.

### Proposed Solution

Make the table backends faithfully persist all logical check fields, with a back-compat path for existing tables:

- **`user_metadata`**: `json.dumps` on Delta write + `safe_json_load` on read (mirroring `arguments`), so value types survive. Read path stays tolerant of legacy plain-string rows.
- **`message_expr`**: add a nullable `STRING` (Delta) / `Text` (Lakebase) column to both table schemas and emit it in both row builders; legacy rows without the column read back as `None`.
- Update the pinned storage contract test `test_save_checks_to_table_with_unresolved_for_each_column`, which currently asserts raw plain-string `user_metadata`, to the new encoded shape.

### Additional Context

- Flagged during review of PR #1302 (tracking issue #1285) and independently confirmed by @mwojtyczka in his re-review.
- Precedent: #591 ("save_checks doesn't store row filter") — same class of bug, accepted and fixed.
- Both fixes change the at-rest format, which is why they are tracked here as a separate follow-up rather than expanded into #1302.

fedeflowers:
## Changes

Faithfully persist all logical check fields through the **table** storage backends (Delta / Lakebase).
Previously a check could lose information on a save→load round-trip through a table — but not through
File/Volume — because the at-rest table schema didn't fully represent a check. Two gaps, one root cause:

1. **Non-string `user_metadata` values were coerced to strings on the Delta path.** `user_metadata` is
   `dict[str, Any]`, so a check can carry `{"confidence": 0.95}` or `{"enabled": true}`.
   `CHECKS_TABLE_SCHEMA` types the column `MAP<STRING, STRING>` and wrote the raw dict, so PySpark's
   `StringType` converter coerced every value via `str()` (`0.95` → `"0.95"`, `true` → `"true"`).
   Lakebase (`JSONB`) and File/Volume (YAML/JSON) preserved the types; only Delta lost them. `arguments`
   in the same schema already avoided this via `json.dumps`-on-write + `safe_json_load`-on-read.
2. **`message_expr` was not persisted by either table backend.** It is a first-class check field (#1092 /
   #958) and is in the load-path allow-list, but had no column in `CHECKS_TABLE_SCHEMA` (Delta) and was
   omitted from the Lakebase columns and row dict — so it round-tripped only via File/Volume and was
   silently dropped through any table backend.

**Fix** (with a back-compat path for existing tables):

- **`user_metadata`** — `json.dumps` on Delta write + `safe_json_load` on read (mirroring `arguments`), so
  value types survive. Read stays tolerant of legacy plain-string rows (`safe_json_load` returns the
  original string on `JSONDecodeError`). Lakebase's `JSONB` path is unchanged (already type-preserving).
- **`message_expr`** — added a nullable `STRING` (Delta) / `Text` (Lakebase) column to both schemas and
  emit it in both row builders. Legacy rows/tables without the column read back as `None` (Delta read uses
  `getattr(row, "message_expr", None)`).
- **Back-compat migration for pre-existing tables** — saving to a table created before this change adds the
  new column in place rather than failing: Delta generalizes the existing `_ensure_versioning_columns` ALTER
  path (now `_ensure_columns`) to also add `message_expr`; Lakebase runs an idempotent
  `ADD COLUMN IF NOT EXISTS message_expr TEXT` on save and detects the column's presence on load so a
  `SELECT` never references a missing column.
- **Legacy NULL tolerance** — `safe_json_load` now returns `None` for a `None` input, so a legacy Delta
  table with a SQL NULL inside a `MAP<STRING, STRING>` value decodes safely instead of raising `TypeError`.
- Updated the pinned contract test `test_save_checks_to_table_with_unresolved_for_each_column` to the new
  JSON-encoded `user_metadata` shape.

Fingerprints are unaffected — they are computed from the semantic check dict **before** row encoding
(`compute_rule_fingerprint` in `rule.py` is untouched by this PR).

### Compatibility note — Delta `user_metadata` at-rest format

On Delta, `user_metadata` values are now stored **JSON-encoded** in the `MAP<STRING, STRING>` column
(e.g. `check_type -> "completeness"` with literal quotes), mirroring how `arguments` is already stored.
Any external consumer that reads the checks table directly (dashboards, ad-hoc SQL) will see the quoted
form, and a predicate like `user_metadata['check_owner'] = 'someone@email.com'` must account for the
JSON-quoted value. Legacy rows written by the old code (raw strings) still load: `safe_json_load` returns
the original string on a parse failure, though a legacy value that happens to look like JSON (`"5"`,
`"true"`) will decode to its typed form (`5`, `True`) — the same behavior `arguments` already has. Lakebase
is unaffected (native `JSONB`), as are File/Volume backends.

### Known limitation — metadata-only edits are a no-op on re-save

`compute_rule_fingerprint` hashes only `name / criticality / function / arguments / filter /
for_each_column` — **not** `message_expr` or `user_metadata`. Because `save()` short-circuits when a rule
set with a matching `rule_set_fingerprint` already exists (regardless of `mode`, including `overwrite`),
**editing *only* `message_expr` (or *only* `user_metadata`) on an existing rule set and re-saving is a
no-op** — the previously stored value persists.

This is **pre-existing fingerprint/idempotency behavior**, not introduced here: `user_metadata` already
behaved this way before this PR, and the fingerprint logic is unchanged. This PR simply makes
`message_expr` newly persistable, so it inherits the same limitation. It is documented in the Delta and
Lakebase `save()` docstrings. Changing the overwrite-skip so metadata-only edits persist would alter
existing semantics for `user_metadata` too and is out of scope here — tracked as a follow-up.

| Symbol | Change |
|---|---|
| `CHECKS_TABLE_SCHEMA` | Insert nullable `message_expr STRING` between `user_metadata` and the versioning columns. |
| `DataFrameConverter.to_dataframe` | `json.dumps` each `user_metadata` value (`None`→`None`); append `message_expr` to the row at the matching position. |
| `DataFrameConverter._row_to_check_dict` *(new)* | Extracted per-row read; `safe_json_load` over `user_metadata` values; `getattr(row, "message_expr", None)` for legacy-table safety. |
| `TableChecksStorageHandler._ensure_columns` *(was `_ensure_versioning_columns`)* | Also ALTERs `message_expr` into a pre-existing Delta table before write. |
| `LakebaseChecksStorageHandler.get_table_definition` | Add nullable `Column("message_expr", Text)`, gated by `include_message_expr` for legacy-table loads. |
| `LakebaseChecksStorageHandler._normalize_checks` | Emit `"message_expr"`; `user_metadata` left native (JSONB). |
| `LakebaseChecksStorageHandler._ensure_message_expr_column` / `_column_exists` *(new)* | Idempotent `ADD COLUMN IF NOT EXISTS` on save; column-presence detection on load. |
| `safe_json_load` *(utils.py)* | Returns `None` for a `None` input so legacy NULL map values decode safely. |

Blast radius is import-only — the two edited classes live in the `checks_storage` core module; downstream
modules merely import it and no caller signatures changed:

```mermaid
graph LR
  checks_storage["checks_storage · 2 changed<br/><i>modified</i>"]
  core["core (engine, cli)<br/><i>impacted</i>"]
  contexts["contexts.workflow_context<br/><i>impacted</i>"]
  profiler["profiler.profiler_runner<br/><i>impacted</i>"]
  core -->|imports ×2| checks_storage
  contexts -->|imports| checks_storage
  profiler -->|imports| checks_storage
  style checks_storage stroke:#2563eb,stroke-width:2px
  style core stroke:#d97706,stroke-width:2px
  style contexts stroke:#d97706,stroke-width:2px
  style profiler stroke:#d97706,stroke-width:2px
```

⚠️ Breaking change: user_metadata at-rest encoding (Delta table backend)

To preserve non-string `user_metadata` types through the Delta MAP<STRING, STRING> column, values are now JSON-encoded on write and decoded on read. Save→load via DQX is transparent (you get the original typed value back), but the stored representation changes:
- Direct table consumers (SQL/dashboards reading user_metadata without DQX) now get JSON-encoded values — e.g. user_metadata['check_type'] returns "completeness" (quoted), not completeness. Decode with `from_json`.
- Existing tables aren't migrated, so old and new rows may hold different encodings. Legacy string values that look like JSON atoms ("true", "1", "null") will now read back as True / 1 / None; re-save affected rule sets after upgrading to normalize.
- Not affected: File/Volume (YAML/JSON) and Lakebase (JSONB) backends — native types, unchanged on disk.

### Linked issues

Resolves #1308

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

Unit (`tests/unit/`, no Spark/DB): Lakebase `get_table_definition` exposes a nullable `message_expr`
column (and omits it when gated off); `_normalize_checks` emits `message_expr` (present + `None`) and passes
non-string `user_metadata` through unchanged; `_row_to_check_dict` handles legacy rows (no `message_expr`
field) and NULL map values; `safe_json_load(None)` returns `None`. Full unit suite: **1282 passed** (only
3 pre-existing unrelated failures in `test_app_backend`/`test_cli_extras`); `make lint` clean
(black + ruff + mypy + pylint 10.00/10).

Integration (`tests/integration/`, live Delta table, `TZ=UTC` + warehouse): the fix-relevant suites are
green — **35 passed, 0 skipped** across `test_save_and_load_checks_from_table.py`, `test_build_rules.py`,
and `test_apply_checks_with_rule_fingerprint.py`. This includes the decisive round-trips
(`..._preserves_user_metadata_value_types`, `..._preserves_message_expr`), the updated pinned test, the
string round-trip regression, and the back-compat migration tests
(`test_save_to_legacy_delta_table_without_message_expr_adds_column_and_round_trips`,
`test_save_to_legacy_delta_table_adds_versioning_columns`). The Lakebase migration is covered by unit tests
(`get_table_definition` gating, `_column_exists`).

### Documentation and Demos

- [ ] added/updated demos
- [ ] added/updated docs


src/databricks/labs/dqx/checks_storage.py
```diff
@@ -244,14 +262,22 @@ def to_dataframe(
 
             name, rule_fingerprint = DataFrameConverter._resolve_name_and_fingerprint(original_check, check)
 
+            # Values are already normalized by ChecksNormalizer.normalize; json.dumps for MAP<STRING, STRING>,
+            # mirroring the arguments encoding so non-string user_metadata types survive the round-trip.
+            user_metadata = check.get("user_metadata")
+            json_user_metadata = (
```
mwojtyczka, line 273: Encoding all `user_metadata` values with `json.dumps` into `MAP<STRING,STRING>` changes the **at-rest** representation: the string `"completeness"` is now stored as `'"completeness"'` (embedded quotes). It faithfully preserves types on load (good), but two consequences worth a conscious call + an upgrade note: (1) anything reading the raw table directly (e.g. `SELECT user_metadata['check_type'] …` in SQL/dashboards) now gets JSON-escaped values; (2) existing tables aren't migrated, so rows written before vs after this change carry different encodings for the same key.

mwojtyczka (commented):
Follow-up review of the table-storage metadata persistence. The round-trip works and is well-tested for **new** data; inline notes below. The main point is the first two comments: `user_metadata` values are now JSON-encoded into the `MAP<STRING,STRING>` column to preserve types — a reasonable fix, but it changes the **at-rest** format (direct-SQL/dashboard consumers see quoted strings) and existing tables aren't migrated, so it deserves a BREAKING CHANGE note. Lower-priority (not inlined): the Lakebase load path does a second schema introspection (`_column_exists` after `_rule_set_columns_exists`) that could be unified, and migrated-legacy tables get `message_expr` in a different physical column position than fresh tables (harmless for DQX's name-based reads).

src/databricks/labs/dqx/checks_storage.py
```diff
@@ -158,29 +163,42 @@ def from_dataframe(
...
+                "function": row.check["function"],
+                "arguments": ({k: safe_json_load(v) for k, v in arguments.items()} if arguments is not None else {}),
+            },
+        }
+        if "for_each_column" in row.check and row.check["for_each_column"]:
+            check_dict["check"]["for_each_column"] = row.check["for_each_column"]
+        if row.filter is not None:
+            check_dict["filter"] = row.filter
+        if row.user_metadata is not None:
+            check_dict["user_metadata"] = {k: safe_json_load(v) for k, v in row.user_metadata.items()}
```
mwojtyczka, line 202: On read, `safe_json_load` is applied to every value, including legacy rows. For a table written by the **old** code (values stored as plain strings), a value the user intended as the string `"true"` / `"1"` / `"null"` now decodes to `True` / `1` / `None` — a silent type change for pre-existing data, with no way to distinguish an original string from an original scalar. Narrow (only JSON-atom-looking strings), but worth an upgrade note; pairs with the write-side encoding change.

src/databricks/labs/dqx/checks_storage.py
```diff
@@ -662,6 +706,7 @@ def _save_checks_to_lakebase(self, checks: list[dict], config: LakebaseChecksSto
                 f"Successfully created or verified table '{config.database_name}.{config.schema_name}.{config.table_name}'."
             )
             LakebaseChecksStorageHandler._ensure_versioning_columns(engine, config)
+            LakebaseChecksStorageHandler._ensure_message_expr_column(engine, config)
```
mwojtyczka, line 714: `_ensure_message_expr_column` runs on every save and issues `ALTER TABLE … ADD COLUMN IF NOT EXISTS message_expr` unconditionally. It's idempotent at the DB, but it's a DDL round-trip on every save (including fresh tables that `create_all` just created with the column). It also duplicates `_ensure_versioning_columns` — consider folding both into one method parameterized by a `column→type` mapping (mirroring `_ADDITIONAL_DELTA_COLUMN_TYPES` on the Delta side) and/or guarding with `_column_exists`.

src/databricks/labs/dqx/checks_storage.py
```diff
@@ -158,29 +163,42 @@ def from_dataframe(
...
-            if row.user_metadata is not None:
-                check_dict["user_metadata"] = row.user_metadata
-            # Denormalize special markers back to objects
-            checks.append(ChecksNormalizer.denormalize_value(check_dict))
-        return checks
+        # Denormalize special markers back to objects
+        return [ChecksNormalizer.denormalize_value(DataFrameConverter._row_to_check_dict(row)) for row in check_rows]
+
+    @staticmethod
+    def _row_to_check_dict(row: Row) -> dict:
```
mwojtyczka: Nit (convention): return type is a bare `dict`. Per AGENTS.md, use parameterized generics — `dict[str, object]` — so mypy can check the reconstructed check dict at the call sites.

src/databricks/labs/dqx/utils.py
```diff
@@ -240,14 +240,19 @@ def is_sql_query_safe(query: str) -> bool:
     return not any(re.search(rf"\b{kw}\b", normalized_query) for kw in forbidden_statements)
 
 
-def safe_json_load(value: str):
+def safe_json_load(value: str | None):
```
mwojtyczka: Nit (convention): the PR edits this signature, but there's still no return-type annotation. AGENTS.md: *"Every parameter and return value must be annotated."* `safe_json_load` returns the decoded value or the passthrough — `-> object` (or a precise union) would satisfy it.

src/databricks/labs/dqx/checks_storage.py
mwojtyczka:
```suggestion
    def _row_to_check_dict(row: Row) -> dict[str, object]:
```

src/databricks/labs/dqx/utils.py
mwojtyczka:
```suggestion
def safe_json_load(value: str | None) -> object:
```
Adds the missing return annotation — returns the decoded JSON value, the original string on parse failure, or `None`.

src/databricks/labs/dqx/checks_storage.py
mwojtyczka:
```suggestion
    def _row_to_check_dict(row: Row) -> dict[str, object]:
```
Parameterize the return generic per the type-hint convention.

mwojtyczka (commented):
Suggested fixes for the two signature nits (one-click applicable).

src/databricks/labs/dqx/utils.py
```diff
@@ -263,14 +240,19 @@ def is_sql_query_safe(query: str) -> bool:
...
```
mwojtyczka:
```suggestion
def safe_json_load(value: str | None) -> object:
```

mwojtyczka (approved):
LGTM - just kicked the full CI, will merge if all green

mwojtyczka:
@fedeflowers seems like build and some tests are failing

fedeflowers:
I think the issue was the signature dict[str, str]
The dbx runtime parsing error shouldn't be related to this pr

mwojtyczka:
@fedeflowers there are some merge conflicts, can you pls resolve them?

PR #1321, opened 2026-07-12
Add registry rules import with batch API and migration backfill.

berrybluecode:
Move YAML and contract import into the Rules Registry, add a batch-import endpoint to avoid slow per-rule saves, and backfill monitored-tables schema in Lakebase/Delta migrations.

### Linked issues

Resolves #..

### Tests

- [x] manually tested
- [x] added unit tests
- [ ] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

### Documentation and Demos

- [ ] added/updated demos
- [ ] added/updated docs


app/src/databricks_labs_dqx_app/backend/migrations/__init__.py
```diff
@@ -741,6 +741,42 @@ class Migration:
...
+    "  schedule_cron STRING,"
+    "  schedule_tz STRING,"
+    "  last_profiled_at TIMESTAMP,"
+    "  created_by STRING,"
+    "  created_at TIMESTAMP,"
+    "  updated_by STRING,"
+    "  updated_at TIMESTAMP,"
+    "  CONSTRAINT pk_dq_monitored_tables PRIMARY KEY (binding_id) RELY"
+    ") CLUSTER BY (table_fqn, status);"
+    f"CREATE TABLE IF NOT EXISTS {_PLACEHOLDER}.dq_applied_rules ("
```
mwojtyczka, line 764: it would be better to version rules and refer to them from the registry rather than log every rule set applied

app/src/databricks_labs_dqx_app/backend/migrations/__init__.py
```diff
@@ -765,7 +801,8 @@ class Migration:
 # enabled (the Postgres mirror is v5 in ``backend.migrations.postgres``), so
 # this only runs against Delta when Lakebase is disabled.
 _V9_MONITORED_TABLES_STATUS_CONVERGE = (
-    f"ALTER TABLE {_PLACEHOLDER}.dq_monitored_tables "
+    _V9_MONITORED_TABLES_BACKFILL
```
mwojtyczka, line 804: **Migration mechanism.** The backfill is *prepended to the existing v9 migration* rather than added as a new version. Migrations only run when `version > recorded`, and versions already go to 16 on this branch — so this backfill runs **only on DBs still below v9** and never on any DB already at/above it. If the schema-missing DBs you're targeting are already past v9 it silently never reaches them, and editing an applied migration won't re-run on dev/staging that already ran the old v9. Consider adding a **new** migration version instead. (Same for the PG v5 change below.)

mwojtyczka (commented):
Review of the registry-rules import + batch API + migration backfill. Overall solid; the main things to look at are the **migration mechanism** (prepending to an already-applied version), the **batch endpoint's robustness** (no size bound + error leak), and a few frontend type/i18n loose ends — inline comments below.

**Cleared:** YAML uses `safe_load`; the batch endpoint does cut browser↔server round-trips (its stated goal, though it still does N sequential server↔DB writes); PG DDL is transactional.

app/src/databricks_labs_dqx_app/backend/migrations/postgres.py
```diff
@@ -94,6 +94,52 @@ class PgMigration:
...
 _S = "{schema}"
 
+# Backfill for Lakebase DBs where v1 was recorded before Phase 3A
+# piggybacked ``dq_monitored_tables`` / ``dq_applied_rules`` onto the v1
+# baseline. Those DBs skip v1 (already in ``dq_migrations``), so the v5
+# converge step must create the tables before it can ALTER them.
+# ``IF NOT EXISTS`` makes this a no-op on fresh installs (v1 already
+# created them). DDL mirrors the current v1 baseline shape (without the
+# ``version`` column — v6 adds that).
+_PG_MONITORED_TABLES_BACKFILL = (
```
mwojtyczka, line 104: Same concern as the Delta side: this is folded into the existing PG v5 migration, so it only runs on DBs below v5 and never re-runs where v5 was already applied. A new migration version would also reach already-migrated DBs.

app/src/databricks_labs_dqx_app/backend/models.py
```diff
@@ -459,6 +459,33 @@ class CreateRegistryRuleOut(BaseModel):
     )
 
 
+class BatchImportRegistryRulesIn(BaseModel):
+    """Bulk-create registry drafts from imported check dicts (YAML, data contract, …)."""
+
+    rules: list[CreateRegistryRuleIn] = Field(min_length=1)
```
mwojtyczka, line 465: No upper bound on the batch (`min_length=1` only). The endpoint runs a synchronous per-rule loop of DB writes, so a very large `rules` payload blocks the uvicorn worker and holds DB connections (DoS). Add a `max_length`, and consider a bulk insert / threadpool.

app/src/databricks_labs_dqx_app/backend/routes/v1/registry_rules.py
```diff
@@ -166,6 +169,68 @@ def create_registry_rule(
...
+                        "Batch import created rule %s but submit failed: %s",
+                        rule.rule_id,
+                        submit_err,
+                    )
+                    submit_failed += 1
+        except UnsafeSqlQueryError as e:
+            failed.append(BatchImportRegistryRulesFailure(index=index, error=str(e)))
+        except Exception as e:
+            logger.error("Batch import failed for rule index %s: %s", index, e, exc_info=True)
+            failed.append(BatchImportRegistryRulesFailure(index=index, error=str(e)))
```
mwojtyczka, line 223: CWE-209: `str(e)` from the generic `except Exception` is returned verbatim in the response `failed[]`. DB/driver exceptions (SQLAlchemy/psycopg) can leak table names, SQL, or connection detail to the caller. You already log it server-side with `exc_info=True` — return a generic message here instead. (The `UnsafeSqlQueryError` branch above is intentionally user-facing, so that one's fine.)

app/src/databricks_labs_dqx_app/ui/lib/api-custom.ts
```diff
@@ -53,6 +53,35 @@ export const useBatchSaveRules = <
...
+  rules: CreateRegistryRuleIn[];
+  also_submit?: boolean;
+}
+
+export interface BatchImportRegistryRulesFailure {
+  index: number;
+  error: string;
+}
+
+export interface BatchImportRegistryRulesOut {
```
mwojtyczka, line 70: `BatchImportRegistryRulesOut` / `batchImportRegistryRules` are declared here **and** auto-generated in `api.ts` (orval) with the same names but divergent types — required fields here vs optional in `api.ts`. Importing from `@/lib/api` type-checks but `data.failed.map()` throws at runtime, and `make app-regen-api` perpetuates the conflict. Consolidate on one definition.

app/src/databricks_labs_dqx_app/ui/lib/i18n/locales/en.json
```diff
@@ -606,13 +602,16 @@
...
     "validationErrors": "Validation errors",
-    "targetTable": "Target table",
+    "registryHint": "Rules are saved to the Rules Registry as drafts. Assign them to monitored tables after approval.",
     "validate": "Validate",
     "dryRun": "Dry run",
     "cancelDryRun": "Cancel dry run",
     "saveAsDrafts": "Save as drafts",
     "submitForReview": "Submit for review",
     "yamlMustBeList": "YAML must be a list of check definitions, for example:\n- criticality: error\n  check:\n    function: is_not_null\n    arguments:\n      column: id",
+    "yamlCommentsOnly": "Only comments were found — paste at least one rule starting with a dash (-).",
```
mwojtyczka, line 612: These new `rulesImport.*` keys (`yamlCommentsOnly`, `yamlEmptyList`, `yamlInvalidEntry`, `importing_one/_other`, `importingAndSubmitting_one/_other`) are only in `en.json` — missing from `es.json`, `it.json`, `pt-BR.json`. `ui/CLAUDE.md`: *"Add every new key to all four locales."* Non-English users get English fallback for the most visible import-flow strings.

app/src/databricks_labs_dqx_app/ui/types/routeTree.gen.ts
```diff
@@ -200,6 +201,12 @@ const SidebarRegistryRulesNewRoute = SidebarRegistryRulesNewRouteImport.update({
   path: '/registry-rules/new',
   getParentRoute: () => SidebarRouteRoute,
 } as any)
+const SidebarRegistryRulesImportRoute =
+  SidebarRegistryRulesImportRouteImport.update({
+    id: '/registry-rules/import',
```
mwojtyczka, line 206: `ui/CLAUDE.md` lists `types/routeTree.gen.ts` under *Auto-Generated Files — Do Not Edit*. This looks hand-edited — regenerate it (`make app-build` / dev server) so a future regen doesn't wipe/diverge the manual entries (and to confirm the new route actually registers at runtime).

app/src/databricks_labs_dqx_app/ui/lib/import-registry-rules.ts
```diff
@@ -0,0 +1,259 @@
...
+      error: null,
+      hint: messages.invalidEntry.replace("{{index}}", String(invalidIndex + 1)),
+    };
+  }
+
+  const normalized = rules.map((raw) => normalizeImportedCheck(raw));
+  return { checks: normalized, error: null, hint: null };
+}
+
+function parseChecksForImport(
```
mwojtyczka, line 175: `normalizeImportedCheck` ends up applied twice to YAML-imported checks — once in `parseImportYamlText` before they're stored in state, then again here. Idempotent today, but any future non-idempotent change would corrupt YAML imports only (contract imports go through a single pass). Normalize in exactly one place.

app/src/databricks_labs_dqx_app/ui/routes/_sidebar/rules.import.tsx
```diff
@@ -1,734 +1,28 @@
...
 
-function _coerceTab(value: unknown): ImportTab | undefined {
-  return value === "yaml" || value === "contract" ? value : undefined;
-}
-
+// Legacy URL — import now lives under Rules Registry. Keep this route as a
+// permanent redirect so bookmarks and old sidebar links keep working.
 export const Route = createFileRoute("/_sidebar/rules/import")({
-  component: ImportRulesPage,
+  component: RulesImportRedirect,
```
mwojtyczka, line 12: The old `/rules/import` page had a `canCreateRules` guard; this redirect drops it and sends everyone to `/registry-rules/import`. Safe today (the new page has its own guard, so unauthorized users double-redirect), but it's a latent bypass if that guard is ever relaxed, plus a redirect flicker for unauthorized users.

mwojtyczka (changes requested):

berrybluecode:
closed the PR as new code updated by Oli in branch dqx-dqlake-integration, will make the PR to the new branch

PR #1347, opened 2026-07-19
Add valid string case check (#1335)

### LINKED ISSUE #1335: [FEATURE]: New check to validate string case

### Problem statement

Some use cases require checking string casing (e.g. upper, lower, title, sentence) to ensure values are consistent. DQX lacks a built-in check for string casing.

### Proposed Solution

Add a `has_valid_string_case` check that accepts the `column` to check and a `case` argument which can be one of:

- `"upper"`
- `"lower"`
- `"title"`
- `"sentence"`

### Additional Context

_No response_

SyedIshmumAhnaf:
## Changes

- Add a `has_valid_string_case` row-level check for validating consistent string casing.
- Support `upper`, `lower`, `title`, and `sentence` case modes.
- Cast input values to strings before normalization and comparison so the check can handle non-string columns consistently.
- Validate the `case` argument and reject unsupported values.
- Register the check for declarative rule construction.
- Add unit and integration test coverage for the new behavior.
- Document the check and update the DQX check-definition agent skill.
- Updated tests to avoid flakiness

### Linked issues

Resolves #1335 

### Tests

- [ ] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

### Documentation and Demos

- [ ] added/updated demos
- [x] added/updated docs


docs/dqx/docs/reference/quality_checks.mdx
```diff
@@ -60,6 +60,7 @@ You can also define your own custom checks in Python (see [Creating custom check
...
+| `has_valid_string_case`            | Checks whether the values match the requested string case. `upper` and `lower` use the corresponding Spark case functions; `title` uses Spark `initcap`; `sentence` splits on literal `.` characters, cases each segment independently, and preserves delimiters and surrounding whitespace. | `column`: column to check (can be a string column name or a column expression); `case`: one of `upper`, `lower`, `title`, or `sentence` |
```
ghanse: I think it's better to say what each case is rather than how it is implemented. For example:

```
...`upper` and `lower` check that all characters are uppercase or lowercase, respectively; `title` checks that the 1st character of each word is uppercase; `sentence` checks that the 1st character of each sentence is uppercase
```

skills/dqx-define-checks/SKILL.md
```diff
@@ -86,7 +86,7 @@ Full reference: <https://databrickslabs.github.io/dqx/docs/reference/quality_che
 
 ## Fallback: custom SQL
 
-**Search `check_funcs` first** — the built-ins cover null/empty, range, set membership, regex, referential, aggregate, uniqueness, schema, freshness, comparison, and outlier cases with typed error messages and tested edge handling. Drop down to SQL only when no built-in fits.
+**Search `check_funcs` first** - the built-ins cover null/empty, string case, range, set membership, regex, referential, aggregate, uniqueness, schema, freshness, comparison, and outlier cases with typed error messages and tested edge handling. Use `has_valid_string_case` for `upper`, `lower`, `title`, or `sentence` string-case validation. Drop down to SQL only when no built-in fits.
```
ghanse: Let's remove the specific sentence about case validation. Details about most individual checks are intentionally left out of the skill file to prevent it from getting too large.

tests/integration/test_row_checks.py
```diff
@@ -39,13 +39,71 @@
...
+            [
+                violation("Upper", "upper", "upper"),
+                violation("Lower", "lower", "lower"),
+                violation("Spark SQL", "title", "title"),
+                violation("first SEGMENT. second SEGMENT", "sentence", "sentence"),
+                violation("Nested", "nested['value']", "lower"),
+                None,
+            ],
+            [None, None, None, None, None, None],
+            [None, None, None, violation(" first SEGMENT. second SEGMENT ", "sentence", "sentence"), None, None],
```
ghanse: I think the `title` case should have a violation here?

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -249,6 +250,61 @@ def is_null_or_empty(column: str | Column, trim_strings: bool | None = False) ->
...
+        normalized_case = F.initcap(col_expr_cast)
+    else:
+        segments = F.split(col_expr_cast, r"\.")
+        normalized_case = F.array_join(
+            F.transform(
+                segments,
+                lambda segment: F.concat(
+                    F.regexp_extract(segment, r"^(\s*)", 1),
+                    F.upper(F.substring(F.regexp_replace(segment, r"^\s*", ""), 1, 1)),
+                    F.lower(F.substring(F.regexp_replace(segment, r"^\s*", ""), 2, 2147483647)),
```
ghanse: Maybe best to use an internal constant for `2147483647`

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -249,6 +250,61 @@ def is_null_or_empty(column: str | Column, trim_strings: bool | None = False) ->
...
+
+    col_str_norm, col_expr_str, col_expr = get_normalized_column_and_expr(column)
+    col_expr_cast = col_expr.cast("string")
+
+    if case == "upper":
+        normalized_case = F.upper(col_expr_cast)
+    elif case == "lower":
+        normalized_case = F.lower(col_expr_cast)
+    elif case == "title":
+        normalized_case = F.initcap(col_expr_cast)
```
ghanse: This will flag strings with uppercased abbreviations like `"Notes From IEEE Meeting"`.

Maybe we should use a pattern similar to what's been used for sentence case?

tests/unit/test_build_rules.py
```diff
@@ -55,6 +56,35 @@
...
+
+    assert serialize_checks([rule]) == expected_metadata
+
+    deserialized_rules = deserialize_checks(expected_metadata)
+    assert len(deserialized_rules) == 1
+    deserialized_rule = deserialized_rules[0]
+    assert deserialized_rule.name == "a_has_invalid_lower_string_case"
+    assert deserialized_rule.check_func is has_valid_string_case
+    assert deserialized_rule.column == "a"
+    assert deserialized_rule.check_func_kwargs == {"case": "lower"}
```
ghanse: I don't think we need a specific test for this check. We can add it to the rulesets in the existing tests.

ghanse (changes requested):
Overall looks good. Left a few small comments. Can you add some performance tests for this check?

Here's an example:
https://github.com/databrickslabs/dqx/blob/9c8fcafae8be88dabe0153a76eaa67cdc216f7ce/tests/perf/test_apply_checks.py#L2144-L2167

SyedIshmumAhnaf:
Thanks for the review and the helpful comments. I’ll work through the requested changes, add the performance test coverage, and update the PR accordingly.

SyedIshmumAhnaf:
Hi @ghanse, I’ve incorporated the feedback from the review into PR #1347.

The updates include the requested documentation changes, revised title-case behavior for uppercase abbreviations, updated integration expectations, use of an internal substring-length constant, consolidation of the rule-construction coverage, and performance tests for all four supported case modes.

The PR is ready for another review. Thanks!

ghanse (approved):
LGTM

@SyedIshmumAhnaf FYI, I modified the `sentence` case to require uppercase for the 1st character of the sentence only. The remaining characters can have any case. I think this better matches the intent.

SyedIshmumAhnaf:
Thanks @ghanse for the update and for making that adjustment. The revised sentence behavior makes sense to me.

docs/dqx/docs/reference/quality_checks.mdx
```diff
@@ -1288,6 +1296,12 @@ checks = [
       column="col1"
     ),
 
+    # has_valid_string_case check
+    DQRowRule(
+      criticality="error",
+      check_func=check_funcs.has_valid_string_case,
+      column="col1",
+      check_func_kwargs={"case": "lower"}
```
mwojtyczka, line 1304: **Syntax error in the documented example.** This `DQRowRule(` is never closed — line 1304 ends with `check_func_kwargs={"case": "lower"}` and the next line jumps straight to `# is_valid_national_id check` / `DQRowRule(`. Copy-pasting this block raises `SyntaxError` (unclosed paren). Add the closing `),`:

```python
    # has_valid_string_case check
    DQRowRule(
      criticality="error",
      check_func=check_funcs.has_valid_string_case,
      column="col1",
      check_func_kwargs={"case": "lower"}
    ),
```

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -265,6 +267,73 @@ def is_null_or_empty(column: str | Column, trim_strings: bool | None = False) ->
...
+        normalized_case = F.array_join(
+            F.transform(
+                words,
+                lambda word: F.concat(
+                    F.upper(F.substring(word, 1, 1)),
+                    F.substring(word, 2, _MAX_SUBSTRING_LENGTH),
+                ),
+            ),
+            " ",
+        )
```
mwojtyczka, line 314: **`title` / `sentence` only uppercase the first character and leave the remainder unchanged.** For `title`, `normalized_case` of `'HELLO WORLD'` is `'HELLO WORLD'` (each word's first char is already uppercase, the rest is untouched), so it equals the input and the check reports **no violation** — a fully-uppercase string passes the title-case check. Same for `'HELLO WORLD.'` under `sentence`.

If proper Title/Sentence case is intended, the tail needs to be lowercased too — e.g. `F.initcap` gives true title-casing, or lowercase the remainder before concatenating. The current integration tests don't cover an all-caps input for `title`/`sentence`, which is why this slips through — please add one.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -265,6 +267,73 @@ def is_null_or_empty(column: str | Column, trim_strings: bool | None = False) ->
...
+    else:
+        segments = F.split(col_expr_cast, r"\.")
+        normalized_case = F.array_join(
+            F.transform(
+                segments,
+                lambda segment: F.concat(
+                    F.regexp_extract(segment, r"^(\s*)", 1),
+                    F.upper(F.substring(F.regexp_replace(segment, r"^\s*", ""), 1, 1)),
+                    F.substring(F.regexp_replace(segment, r"^\s*", ""), 2, _MAX_SUBSTRING_LENGTH),
```
mwojtyczka: The stripped segment is recomputed three times per row: `regexp_extract(segment, '^(\\s*)', 1)` for the leading whitespace, then `regexp_replace(segment, '^\\s*', '')` **twice**. Compute the stripped value once and reuse it (the leading-whitespace prefix can also be derived once) to avoid three regex passes over the same string on every row.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -265,6 +267,73 @@ def is_null_or_empty(column: str | Column, trim_strings: bool | None = False) ->
...
+                ),
+            ),
+            ".",
+        )
+
+    condition = F.when(col_expr.isNotNull(), col_expr_cast != normalized_case).otherwise(F.lit(None))
```
mwojtyczka: The `F.when(col_expr.isNotNull(), ...).otherwise(F.lit(None))` wrapper is redundant: `col_expr_cast != normalized_case` already evaluates to NULL when the input is NULL (SQL null propagation), and `make_condition` treats NULL as a pass. This can be simplified to `condition = col_expr_cast != normalized_case`.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -31,6 +31,8 @@
 _IPV4_CIDR_SUFFIX = r"(3[0-2]|[12]?\d)"
 IPV4_MAX_OCTET_COUNT = 4
 IPV4_BIT_LENGTH = 32
```
mwojtyczka, line 33: Using `2_147_483_647` (Int32.MAX) as a substring length to mean "rest of the string" is a magic constant tied to the 32-bit limit. Spark supports the 2-argument `substring(str, pos)` form which returns to end-of-string, e.g. `F.expr("substring(word, 2)")` — clearer intent, no sentinel.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -265,6 +267,73 @@ def is_null_or_empty(column: str | Column, trim_strings: bool | None = False) ->
...
+                ),
+            ),
+            ".",
+        )
+
+    condition = F.when(col_expr.isNotNull(), col_expr_cast != normalized_case).otherwise(F.lit(None))
+    message = F.concat_ws(
+        "",
+        F.lit("Value '"),
+        col_expr.cast("string"),
```
mwojtyczka: `col_expr.cast("string")` is rebuilt here even though `col_expr_cast` (line 295) already holds the same string-cast. Reuse `col_expr_cast` for consistency and one fewer duplicate node in the plan.

mwojtyczka (commented):
## Code review — `has_valid_string_case`

Thanks for the thorough PR (docs, unit, integration, and perf coverage all included). Reviewed at high effort. Two items are worth resolving before merge (a broken docs example and a casing-semantics question); the rest are cleanup/efficiency suggestions.

### Correctness
1. **Docs Python example has a syntax error** — the `has_valid_string_case` `DQRowRule` in the Python checks example is missing its closing `),`, so it runs straight into the next `# is_valid_national_id` rule. Copy-pasting the example fails with a `SyntaxError`. (inline)

### Efficiency / simplification
3. **Sentence branch runs 3 regex passes per segment** — `regexp_replace(segment, '^\\s*', '')` is computed twice plus a `regexp_extract`; the stripped value can be computed once and reused. (inline)
4. **Redundant `isNotNull` guard** — `col_expr_cast != normalized_case` already returns NULL for NULL input (SQL null propagation), and `make_condition` maps NULL to a pass, so the `F.when(...).otherwise(F.lit(None))` wrapper never changes the result. (inline)
5. **Magic `2_147_483_647` sentinel** — using Int32.MAX as a substring length to mean "to end of string" is obscure; Spark supports `substring(x, 2)` (2-arg) directly. (inline)
6. **Message re-casts the column** — `col_expr.cast("string")` is rebuilt in the message even though `col_expr_cast` already holds it. (inline)

mwojtyczka (changes requested):
Generally looking good, left some comments that need to be addressed before we merge

SyedIshmumAhnaf:
Hi @mwojtyczka, thanks for the review and the comments. I'll look into them so that the PR can be merged.

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: However, sentences like the following should be valid: "IEEE Meeting Notes", "I live in Miami, FL." So I suppose, let's skip this.

mwojtyczka (commented):
Pushed fixes for the outstanding review comments:

- **Docs syntax error** — added the missing `),` so the `has_valid_string_case` `DQRowRule(` example in `quality_checks.mdx` closes and copy-pastes cleanly.
- **Magic `_MAX_SUBSTRING_LENGTH`** — dropped the `2_147_483_647` sentinel; "rest of string" is now `substring(x, 2, length(x))`, and the constant is removed.
- **Triple regex in the sentence branch** — the stripped segment is computed once per element and the leading-whitespace prefix is derived from it, instead of three regex passes per row.
- **Redundant NULL wrapper** — simplified to `condition = col_expr_cast != normalized_case` (SQL null propagation already yields NULL for NULL input, which `make_condition` passes).
- **Duplicate cast in the message** — reuses `col_expr_cast`.

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: Fixed in 56137845: the stripped segment is computed once per element (stripped = regexp_replace(...)) and the leading-whitespace prefix is derived from it, so there is a single regex pass instead of three. Resolving.

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: Fixed in 56137845: simplified to condition = col_expr_cast != normalized_case; SQL null propagation yields NULL for NULL input and make_condition treats NULL as a pass. Resolving.

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: Fixed in 56137845: the message now reuses col_expr_cast instead of re-casting. Resolving.

mwojtyczka (approved):
LGTM

mwojtyczka (commented):
Follow-up fixes pushed (commit `2fb17110`) for a deeper pass on `has_valid_string_case`:

- **Whitespace/delimiter scope** — documented that `title` splits on the ASCII space only and `sentence` on the literal period only (other whitespace — tabs, newlines, non-breaking spaces — is not a word/segment boundary), and that only the first character of each word/segment is checked. Not switched to `\s+` splitting on purpose: that would regress the `split`/`array_join` round-trip (non-space whitespace can't be faithfully rejoined, producing false violations on already-correct values).
- **All-uppercase quirk now tested** — added `test_has_valid_string_case_all_uppercase_passes_title_and_sentence`, pinning that `HELLO WORLD` passes `title` and `HELLO WORLD. GOODBYE WORLD` passes `sentence` (the intended first-character-only semantics), so a future switch to `initcap`-style casing can't silently change behavior.
- **Sentence prefix handling** — reconstruct the leading-whitespace prefix directly via `regexp_extract(segment, '^(\s*)', 1)` instead of `length(segment) - length(stripped)`; removes the coupling to the strip semantics and the duplicate `length(stripped)` computation.
- `F.initcap` reuse was intentionally **not** adopted — the first-char-only semantics are deliberate (now documented and pinned by the test above).

Verified: `ruff` + `mypy` clean; 13 sentence edge cases (leading/trailing/interior whitespace, empty, all-whitespace, multi-segment, trailing period, and violation cases) confirmed behavior-identical to before the refactor against a workspace; both string-case integration tests pass.

PR #1356, opened 2026-07-21
Fix null-safe result joins

### LINKED ISSUE #1354: Null-unsafe result joins silently erase violations across custom SQL and grouped checks

- [x] I searched the existing issues and pull requests.

Related items are not duplicates:

- #1296 concerns DQX Studio generating a `sql_query` payload without a condition column.
- #1309 documents an analogous null-group limitation for a new aggregate check, but does not fix the existing joins described here.

### Current Behavior

Several dataset-level checks compute the correct result for a nullable key, then join that result back to the input with ordinary equality. Spark does not match `NULL = NULL`, so the computed condition is lost and the affected input row silently passes.

The broadest affected path is row-level `sql_query` validation:

```python
user_query_df_unique = user_query_df.groupBy(*merge_columns).agg(
    F.max(F.col(unique_condition_column)).alias(unique_condition_column)
)
joined_df = df.join(user_query_df_unique, on=merge_columns, how="left")
```

The same result-reattachment pattern exists in grouped, window-incompatible aggregate comparisons such as `count_distinct`, and in grouped `has_no_aggr_outliers` checks.

### Expected Behavior

When the SQL query or grouped aggregate produces a violation for a key, that violation should be attached to the matching input row even when one or more key components are null.

### Steps To Reproduce

Run against current `main` (`9c8fcafae8be88dabe0153a76eaa67cdc216f7ce`) with PySpark 4.0.0:

```python
from databricks.labs.dqx.check_funcs import sql_query

spark = SparkSession.builder.master("local[2]").getOrCreate()


def evaluate(key):
    df = spark.createDataFrame([(key, 100)], "row_id string, amount int")
    condition, apply = sql_query(
        """
        SELECT row_id, amount > 0 AS condition
        FROM {{ input_view }}
        """,
        merge_columns=["row_id"],
        msg="positive amount",
    )
    return [
        (row["row_id"], row["amount"], row["violation"])
        for row in apply(df, spark, {})
        .select("row_id", "amount", condition.alias("violation"))
        .collect()
    ]


print("null_key=", evaluate(None))
print("non_null_control=", evaluate("row-1"))
```

Actual output, reproduced twice:

```text
null_key= [(None, 100, None)]
non_null_control= [('row-1', 100, 'positive amount')]
```

The SQL query returns `TRUE` for both rows. The non-null control is flagged, while the unique null-key row receives no violation.

### Impact

This creates silent false negatives in DQX's custom SQL escape hatch. Nullable order IDs, business keys, countries, segments, or composite-key components can suppress arbitrary custom SQL violations without an exception or a `skipped=True` result.

The same root cause can suppress:

- grouped `count_distinct` and other window-incompatible aggregate comparisons;
- grouped rolling aggregate-outlier violations.

### Affected Code

- `src/databricks/labs/dqx/check_funcs.py`: `sql_query`, result join near line 1626.
- `src/databricks/labs/dqx/check_funcs.py`: `has_no_aggr_outliers`, joins near lines 1994 and 2040.
- `src/databricks/labs/dqx/check_funcs.py`: `_is_aggr_compare`, grouped result join near line 3377.

Existing `sql_query` integration coverage includes null merge-key rows, but their query conditions are false or null. It does not cover a true violation on a null key.

### Suggested Fix

Use aliased DataFrames and null-safe equality (`Column.eqNullSafe`) for every merge/group key when reattaching computed results. A small shared helper could handle composite keys and select only the original left-side columns plus the computed result columns.

Suggested regression coverage:

1. `sql_query` with a unique null merge key returning `TRUE`, plus a non-null control.
2. A composite merge key with one null component.
3. Grouped `count_distinct` with a violating null group.
4. Grouped `has_no_aggr_outliers` with a null segment whose current bucket is a violation.

### Cloud

Not applicable to this reproduction. The defect was reproduced locally with stock PySpark 4.0.0 and is expected to be cloud-independent.

### Operating System

macOS (arm64)

### Environment

- DQX revision: `main@9c8fcafae8be88dabe0153a76eaa67cdc216f7ce`
- Latest affected release checked: `v0.15.0`
- PySpark: `4.0.0`
- Java: OpenJDK `17.0.19`

mattfaltyn:
## Changes

- Reattach custom SQL and grouped-check results with null-safe key matching.
- Reuse the existing null-safe row matcher through one focused left-join helper while preserving the original row and column shape.
- Cover single and composite nullable SQL keys, nullable `count_distinct` groups, and nullable grouped aggregate-outlier violations.

Spark equality joins do not match `NULL` keys, so these paths could compute a real violation and then silently discard it while joining the result back to the input. The shared helper aliases both sides, uses null-safe equality for every key component, and selects only the original left-side columns plus the computed result columns.

### Linked issues

Fixes #1354

### Tests

- [x] manually tested — five focused regressions passed with stock PySpark 4.0.0
- [ ] added unit tests — the affected behavior requires Spark
- [x] added integration tests
- [ ] added end-to-end tests — no workflow or external-resource behavior changed
- [ ] added performance tests — the change only replaces four existing result joins

Validation performed:

- `make fmt`
- `make lint`
- Full unit suite: 1,446 passed
- Focused Spark regressions: single nullable SQL key, two composite-key null positions, grouped `count_distinct`, and grouped aggregate outliers

### Documentation and Demos

- [ ] added/updated demos — not applicable
- [ ] added/updated docs — no public API, configuration, or documented behavior changed
- [ ] added/updated agent skills — no public API changed


src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1991,7 +1986,7 @@ def apply(df: DataFrame) -> DataFrame:
         # Step 5: join current bucket + stats (left-join so current bucket always survives)
         if group_by:
             join_keys = [c if isinstance(c, str) else get_column_name_or_alias(c) for c in group_by]
-            joined = current.join(stats, on=join_keys, how="left")
+            joined = _join_results_on_null_safe_columns(current, stats, join_keys, ["__dq_mu", "__dq_sigma", "__dq_n"])
```
mwojtyczka: **Pre-existing (not introduced here), but this PR now routes it through the shared helper:** `__dq_mu` / `__dq_sigma` / `__dq_n` (and `__dq_metric`, `__dq_grain`, `__dq_current`, `__dq_rn` above) are **fixed** internal column names, unlike the sibling checks whose internal columns are uuid-suffixed (`condition_col = f"__..._{unique_str}"`). A user DataFrame column named e.g. `__dq_metric` would shadow/collide in the `groupBy().agg()` pipeline. The `__dq_` prefix makes it unlikely, but for guaranteed safety these should carry a per-invocation uuid suffix like the other checks. (Not reachable via the helper's `df.*` select here — the left side of this particular join is the internal `current`, not user data.)

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -2922,6 +2917,20 @@ def _match_rows(
...
+) -> DataFrame:
+    """Left-join computed result columns while matching null join keys."""
+    joined = _match_rows(
+        df.alias("df"),
+        result_df.alias("ref_df"),
+        join_columns,
+        join_columns,
+        check_missing_records=False,
+    )
+    return joined.select("df.*", *[F.col(f"ref_df.{column}").alias(column) for column in result_columns])
```
mwojtyczka, line 3204: **Column-name collision (latent).** `select("df.*", ref_df.<col> AS <col>)` will emit two columns with the same name if any `result_column` also exists on the left `df` (and isn't a join key). Can't trigger with today's four callers (result cols are uuid'd, or the left side is a controlled intermediate), but the helper has no guard. Consider asserting `result_columns` are disjoint from `df.columns`, or documenting it.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -2922,6 +2917,20 @@ def _match_rows(
     return results
 
 
+def _join_results_on_null_safe_columns(
```
mwojtyczka, line 3171: **Unique-key assumption (latent).** This left-joins `df` to `result_df` on `join_columns`; if `result_df` has >1 row per key it multiplies `df` rows. Safe for all four current callers (each feeds a `groupBy().agg()`), but as a shared helper the invariant should be stated/enforced so a future caller passing a non-deduplicated `result_df` doesn't silently corrupt the row count.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -2922,6 +2917,20 @@ def _match_rows(
...
+    df: DataFrame, result_df: DataFrame, join_columns: list[str], result_columns: list[str]
+) -> DataFrame:
+    """Left-join computed result columns while matching null join keys."""
```
mwojtyczka: The docstring omits the two load-bearing contracts a reuser must satisfy: (1) `result_df` unique per join key, (2) `result_columns` disjoint from `df`'s columns — plus that the helper aliases the sides internally as `df` / `ref_df`. Worth spelling out given the helper is explicitly designed for reuse.

tests/integration/test_dataset_checks.py
```diff
@@ -31,6 +32,48 @@
...
+    test_df = spark.createDataFrame(rows, schema)
+    query_columns = ", ".join(merge_columns)
+    condition, apply_method = sql_query(
+        f"SELECT {query_columns}, amount > 0 AS condition FROM {{{{ input_view }}}}",
+        merge_columns=merge_columns,
+        msg="positive amount",
+    )
+
+    actual = apply_method(test_df, spark, {}).select(*merge_columns, "amount", condition.alias("violation"))
+    violations = {tuple(row[column] for column in merge_columns): row["violation"] for row in actual.collect()}
```
mwojtyczka, line 68: This asserts the per-key violation map but not that the output row count is preserved. Since `violations` is keyed by the merge columns, a regression that duplicated rows on reattachment (fan-out) could still pass. An `assert actual.count() == len(rows)` would close that gap.

mwojtyczka (commented):
### Review — null-safe result joins

Solid, correct fix for the null-key silent-drop bug (#1354). I verified the four rewired joins are row-count-equivalent to the old left-joins for non-null keys (every `result_df` is a `groupBy().agg()` — one row per key) and now additionally match null keys, and ran the new regressions locally (3 passed). No blocking issues.

The notes below are robustness/hardening observations on the new **shared** helper and one pre-existing naming gap it now routes through — none block merge. Inline comments follow.

mwojtyczka (changes requested):
Thank you for the PR! Generally looking good, left some small comments

mattfaltyn:
Thank you @mwojtyczka for your review. All comments have been addressed in commit 2abf5c2b.

mwojtyczka (approved):
LGTM - adjusted docstring

PR #1359, opened 2026-07-22
Stop reconfiguring the root logger when importing DQX

### LINKED ISSUE #1136: [BUG]: Python Root Logger is Reconfigured When Importing DQX

### Current Behavior

When importing anything from the DQX library, the Python root logger is reconfigured entirely, replacing any existing logging handlers with the DQX-configured handler.
Issue lies in the `src/databricks/labs/dqx/__init__.py` calling `install_logger` of the Databricks Labs Blueprint, see also [#54](https://github.com/databrickslabs/dqx/pull/54/changes#r3152633806).

### Expected Behavior

It should be up to the user of the library to configure their application's logging handlers and/or formatting. Having the option to use the Databricks Labs blueprint logger as a recommendation is fine but the user should have a choice and it should not happen hidden inside the package import.

### Steps To Reproduce

1. Execute the following minimal example
  ```python
  import logging
  logging.error("Default root logger handler")
  logging.error("Root logger handler overwritten after import")
  ```
2. Observe that the logging format has changed after the import.

### Cloud

AWS

### Operating System

Linux

```shell
ERROR:root:Default root logger handler
11:57:46 ERROR [root] Root logger handler overwritten after import
```

SreeramaYeshwanthGowd:
## Changes

Importing anything from DQX currently reconfigures the Python root logger, because `src/databricks/labs/dqx/__init__.py` calls blueprint's `install_logger()` at import time, which removes all existing root handlers and installs its own. That overwrites the logging setup of any application using DQX as a library (see #1136).

This changes the package import to leave the root logger alone. Following standard library logging practice, it attaches a `NullHandler` to the `databricks.labs.dqx` logger instead of installing a handler on the root logger. The CLI is unaffected; it configures logging separately via `get_logger` / `install_logger` in `cli.py` and `installer/logs.py`.

### Linked issues

Resolves #1136

### Tests

- [ ] manually tested
- [x] added unit tests
- [ ] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

Unit tests in `tests/unit/test_logging_setup.py` check that importing DQX does not remove the application's root handlers or change the root level, and that a `NullHandler` is attached to the `databricks.labs.dqx` logger.

### Documentation and Demos

- [ ] added/updated demos
- [ ] added/updated docs


mwojtyczka (approved):
LGTM

PR #1368, opened 2026-07-23
Add is_valid_currency_code row-level check

### LINKED ISSUE #1340: [FEATURE]: New check and generation for currency codes for ISO-4217

### Problem statement

https://en.wikipedia.org/wiki/ISO_4217 standard describes currency codes that are widely used in the industry. Although we have `is_in` check, using this for such a long list might be a bit tedious. Also because list is clearly defined this check can be easily automated for generation 

### Proposed Solution

This list can be reused for dedicated check. Additionally profiler and generation can be added using condition: generate check if string column match ratio avoid threshold with default 99%.

### Additional Context

_No response_

SreeramaYeshwanthGowd:
## Changes

Adds a new row-level check `is_valid_currency_code(column, code_format="alphabetic")` that validates values against the ISO 4217 currency codes. ISO 4217 defines a three-letter **alphabetic** code (e.g. `USD`) and a three-digit **numeric** code (e.g. `840`); both are supported via the `code_format` argument (`"alphabetic"` default, or `"numeric"`). Validation is a case-sensitive membership test; null values pass with no violation reported.

The code lists are stored as data files under `src/databricks/labs/dqx/resources/` and loaded via `importlib.resources`, generated from the ISO 4217 standard data in the `pycountry` package (dev-time generation; the regeneration steps are in a comment). The implementation follows the existing pattern-validator checks such as `is_valid_email` and `is_valid_national_id`.

### Linked issues

Resolves #1340

### Tests

- [ ] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [x] added performance tests

- Unit tests in `tests/unit/test_row_checks.py` cover the auto-generated column name (default and numeric) and `code_format` validation (None, non-string, unsupported value).
- Integration tests in `tests/integration/test_row_checks.py` cover both the alphabetic and numeric formats (valid, invalid, case-sensitive, empty and null cases) plus column-expression input.
- `is_valid_currency_code` is also added to `tests/resources/all_row_checks.yaml` and the "all checks" integration tests (`test_apply_checks_all_checks_as_yaml`, `test_apply_checks_all_row_checks_as_yaml_with_streaming`, and `test_apply_checks_all_checks_using_classes`).
- A performance benchmark (`test_benchmark_is_valid_currency_code`) is added under `tests/perf/`.

### Documentation and Demos

- [ ] added/updated demos
- [x] added/updated docs

`docs/dqx/docs/reference/quality_checks.mdx` gets a reference-table entry and YAML plus programmatic (`DQRowRule`) usage examples.


src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1022,6 +1022,203 @@ def is_valid_email(column: str | Column) -> Column:
     return _matches_pattern(column, DQPattern.EMAIL_ADDRESS)
 
 
+# Active ISO 4217 alpha-3 currency codes, derived from the CLDR currency data bundled with Babel:
```
IvannKurchenko: Thank you for the contribution. 

> Active ISO 4217 alpha-3 currency codes,

I'm having hard times finding the notion of `alpha-3` in ISO 4217 standard,  are you sure it is present there? The standard supports alphanumeric and numeric codes, would it be possible to support both of lists?

>  derived from the CLDR currency data bundled with Babel:

It might be a better idea to use official resources for standard values list - https://www.iso.org/iso-4217-currency-codes.html.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1022,6 +1022,203 @@ def is_valid_email(column: str | Column) -> Column:
...
+        "YER",
+        "ZAR",
+        "ZMW",
+        "ZWG",
+    }
+)
+
+
+@register_rule("row")
+def is_valid_currency_code(column: str | Column) -> Column:
```
IvannKurchenko: It might not a bad idea to validate ISO 4217 currency alpha code and numeric code. Either by argument or as separate method.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1022,6 +1022,203 @@ def is_valid_email(column: str | Column) -> Column:
...
+    """Checks whether the values in the input column are valid ISO 4217 currency codes
+    (for example, *USD*, *EUR*, *JPY*).
+
+    Validation is a case-sensitive membership test against the set of active ISO 4217
+    alpha-3 currency codes. Historical codes and non-national special codes (for example
```
IvannKurchenko: I doubt the notion of `alpha-3` is present in ISO-4217. Can you double check the standard or wiki page, please?

SreeramaYeshwanthGowd:
@IvannKurchenko thanks for the review, good points. You're right that ISO 4217 uses alphabetic and numeric codes rather than the alpha-3 term, so I fixed the wording and added a `code_format` argument to validate either the alphabetic or the numeric code. I also switched the embedded list to pycountry, which packages the official ISO 4217 standard data, and rebased on main.

SreeramaYeshwanthGowd:
@IvannKurchenko following your earlier note on using official sources, I moved the currency code lists into resource files loaded with importlib.resources, and referenced the official ISO 4217 source (https://www.iso.org/iso-4217-currency-codes.html) in the docstring and docs. Both the alphabetic and numeric formats have unit and integration tests.

IvannKurchenko (commented):
Thank you for all your efforts. I don't have further feedback from my end. Please, wait for approval from other reviewers.

mwojtyczka (commented):
## Code review — `is_valid_currency_code`

Well-built PR. I verified the two things most likely to break a resource-backed check, and both are correct:

- **Wheel packaging** ✅ — I built the wheel and confirmed both `iso_4217_*.txt` files ship inside it (712 bytes each), following the existing `llm/resources/*.yml` precedent. The `importlib.resources` loading pattern matches `llm_utils.py`.
- **Data integrity** ✅ — both files are 178 codes, all 3 chars, numeric zero-padding preserved (`008`, `036`), and `.split()` handles the trailing newline.

Findings below are edges and polish; the numeric-format ones (#1, #4) and the doc note (#2) are the ones worth resolving before merge.

### Correctness / robustness
1. **`case_sensitive` is a silent no-op for `numeric`** — digits have no case, but the flag is accepted and still runs `F.lower` over every row and all 178 literals. Reject it or document it as ignored for numeric. (inline)
2. **`code_format` matching is case-insensitive but undocumented** — `.lower()` makes `'NUMERIC'`/`'Alphabetic'` valid (a unit test asserts this), but the docstring lists only lowercase literals. (inline)
4. **Integer input column not guarded** — the docstring warns leading zeros must be preserved, but nothing enforces string input; an int column compared against zero-padded string literals fails silently, and no test covers a non-string column. (inline)
6. **`_load_iso_codes` has no empty/missing guard** — runs at import time; a future packaging regression would crash the whole module import or silently yield an empty set flagging every value. A non-empty assertion fails fast. (Files ship correctly today — this is defensive.) (inline)

### Efficiency
5. **Literal list rebuilt per call** — `sorted()` + 178 `F.lit(...)` are recomputed on every invocation even though the sets are module constants; cache the two literal lists at module scope. (inline)

### Tests
3. **No test for `numeric` + `case_sensitive=False`** — that interaction (and the wasted-lower path) is unverified. (inline)

### Nit
The PR body says the codes are "embedded as module-level frozensets" — stale; they're `.txt` resources loaded via `importlib.resources`. Worth a one-line body edit.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1081,6 +1083,95 @@ def is_valid_national_id(column: str | Column, country: str = "US") -> Column:
...
+            f"Unsupported code_format for currency code validation: '{code_format}'. Supported: [{supported}]."
+        )
+    col_str_norm, col_expr_str, col_expr = get_normalized_column_and_expr(column)
+    if case_sensitive:
+        col_expr_compare = col_expr
+        allowed = [F.lit(code) for code in sorted(allowed_codes)]
+    else:
+        col_expr_compare = to_lowercase(col_expr)
+        allowed = [F.lit(code.lower()) for code in sorted(allowed_codes)]
+    condition = F.when(col_expr.isNotNull(), ~col_expr_compare.isin(*allowed)).otherwise(F.lit(None))
```
mwojtyczka: **`case_sensitive` is a silent no-op for `numeric`.** ISO 4217 numeric codes are all digits, so `to_lowercase` / `.lower()` changes nothing — yet with `code_format="numeric", case_sensitive=False` this still wraps the column and all 178 literals in `F.lower`, adding per-row work with no effect and no signal to the user that the flag is meaningless. Either reject the combination (`raise InvalidParameterError` when `code_format == "numeric"` and `not case_sensitive`) or document explicitly that `case_sensitive` is ignored for the numeric format, and skip the folding in that case.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1081,6 +1083,95 @@ def is_valid_national_id(column: str | Column, country: str = "US") -> Column:
...
+
+    Raises:
+        MissingParameterError: if *code_format* is None.
+        InvalidParameterError: if *code_format* is not a string, or is not a supported representation.
+    """
+    if code_format is None:
+        raise MissingParameterError("'code_format' is not provided.")
+    if not isinstance(code_format, str):
+        raise InvalidParameterError(f"'code_format' must be a string, got {type(code_format)} instead.")
+    normalized_format = code_format.lower()
```
mwojtyczka, line 1260: `normalized_format = code_format.lower()` makes matching case-insensitive, so `'NUMERIC'` / `'Alphabetic'` are accepted (your `test_is_valid_currency_code_format_is_case_insensitive` relies on this). The docstring only lists lowercase `alphabetic`/`numeric`, so the accepted input set is wider than documented. Add a sentence noting that `code_format` matching is case-insensitive.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1081,6 +1083,95 @@ def is_valid_national_id(column: str | Column, country: str = "US") -> Column:
...
+    allowed_codes = _ISO_4217_CODES_BY_FORMAT.get(normalized_format)
+    if allowed_codes is None:
+        supported = ", ".join(sorted(_ISO_4217_CODES_BY_FORMAT))
+        raise InvalidParameterError(
+            f"Unsupported code_format for currency code validation: '{code_format}'. Supported: [{supported}]."
+        )
+    col_str_norm, col_expr_str, col_expr = get_normalized_column_and_expr(column)
+    if case_sensitive:
```
mwojtyczka: **Numeric check on a non-string (integer) column fails silently.** The docstring warns that numeric input "must preserve the leading zeros", but nothing enforces string input. If the column is an int type, `isin` compares integer values against zero-padded string literals like `'008'`, so canonical codes (e.g. `8` vs `'008'`) never match and every row is flagged as invalid. There's no guard/warning and no test with an integer input column. Consider casting the column to string for the comparison (`col_expr.cast("string")`) or documenting/validating that numeric input must already be a zero-padded string, and add a test for an int-typed column.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1081,6 +1083,95 @@ def is_valid_national_id(column: str | Column, country: str = "US") -> Column:
...
+        col_expr_compare = col_expr
+        allowed = [F.lit(code) for code in sorted(allowed_codes)]
+    else:
+        col_expr_compare = to_lowercase(col_expr)
```
mwojtyczka, line 1280: `sorted(allowed_codes)` and the 178-element `[F.lit(code) ...]` list are rebuilt on every call. Since the code sets are fixed module constants, the sorted literal lists (per format, and per case) could be precomputed once at module load and reused. With a `DQForEachColRule` spanning many columns this rebuild happens once per column at plan-build time.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1081,6 +1083,95 @@ def is_valid_national_id(column: str | Column, country: str = "US") -> Column:
     return _matches_pattern(column, pattern)
 
 
+def _load_iso_codes(resource_name: str) -> frozenset[str]:
+    """Load a set of standard codes from a newline-delimited data file in the resources package.
+
+    The large standard code lists are stored as data files rather than inline literals to keep them
+    readable and easy to regenerate. See the files under *databricks/labs/dqx/resources* for the
+    values and their authoritative sources.
+    """
```
mwojtyczka: `_load_iso_codes` runs at import time and returns `frozenset(text.split())` with no validation. The `.txt` files ship in the wheel today (I verified the built wheel), but if a future packaging change drops or empties one, this either raises `FileNotFoundError` at import (breaking **all** checks, since the module fails to load) or silently returns an empty set that flags every non-null value as invalid. A minimal guard — assert the loaded set is non-empty and raise a clear error naming the missing resource — would fail fast and localize the problem.

tests/integration/test_row_checks.py
```diff
@@ -2031,6 +2032,122 @@ def test_col_is_valid_national_id_column_expr_and_lowercase_country(spark):
...
+        [violation("USD")],
+        [violation("")],
+        [None],
+    ]
+    expected = spark.createDataFrame(checked_data, checked_schema)
+
+    assertDataFrameEqual(actual, expected)
+
+
+def test_col_is_valid_currency_code_case_insensitive(spark):
```
mwojtyczka, line 2228: `case_sensitive=False` is tested only for the alphabetic format. There's no test for `code_format="numeric"` with `case_sensitive=False`, so the (no-op) interaction and the wasted-lower path are unverified. Add a case so a future change to the case-folding path can't silently break numeric matching.

mwojtyczka (changes requested):
Thanks for the PR. Generally looking good, left some comments that need to be addressed before merging.

SreeramaYeshwanthGowd:
@mwojtyczka Changes are done, please take a look.

• Case sensitive skipped
• Code format note
• Integer column cast
• Precomputed literal lists
• Empty resource guard
• New int test

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: Addressed: docstring now states case_sensitive has no effect for numeric, the numeric branch is checked first (no folding), and test_col_is_valid_currency_code_numeric_case_insensitive_is_noop covers it. Resolving.

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: Addressed: docstring now notes code_format matching is case-insensitive. Resolving.

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: Addressed: numeric branch casts col_expr to string, docstring explains the int leading-zero behavior, and test_col_is_valid_currency_code_numeric_int_column covers it. Resolving.

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: Addressed: literals are precomputed once at module scope now. Resolving.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1081,6 +1088,121 @@ def is_valid_national_id(column: str | Column, country: str = "US") -> Column:
...
+    resource = Path(str(files("databricks.labs.dqx.resources") / resource_name))
```
mwojtyczka: **Resource read is not zip-import safe.** `Path(str(files("databricks.labs.dqx.resources") / resource_name))` assumes the package is on a real filesystem. `files()` returns a `Traversable`; under a zipped wheel / zipimport, `str(traversable)` is not a filesystem path and `Path(...).read_text()` raises `FileNotFoundError`. Because this runs at import (populating the module-level dicts), it would break importing `check_funcs` entirely, not just this check. Use the idiomatic form that works for both on-disk and zipped resources: `(files("databricks.labs.dqx.resources") / resource_name).read_text(encoding="utf-8")` — no `Path`/`str`. Databricks installs are unpacked so it works today; this removes a latent failure with no downside.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1081,6 +1088,121 @@ def is_valid_national_id(column: str | Column, country: str = "US") -> Column:
...
+# the active codes as of July 2026. The code lists are stored as data files under the resources
+# package and loaded via importlib.resources. To regenerate them, iterate pycountry.currencies
+# (which packages the ISO 4217 data) as a convenience, reading each entry's alphabetic code (exposed
+# by pycountry as the alpha_3 attribute) and numeric code, then reconcile against the official ISO
+# list above before committing.
+_ISO_4217_CODES_BY_FORMAT: dict[str, frozenset[str]] = {
+    "alphabetic": _load_iso_codes("iso_4217_alphabetic.txt"),
+    "numeric": _load_iso_codes("iso_4217_numeric.txt"),
+}
+# Precomputed once at module load: the literal lists never change at runtime, so building them per
```
mwojtyczka: **Import-time load couples the whole module to this check's resources.** `_ISO_4217_CODES_BY_FORMAT = {..: _load_iso_codes(..)}` runs at import, so a missing/empty resource raises during `import check_funcs` and makes `DQEngine` and every unrelated check unimportable. The empty-set guard you added is good — consider also deferring the load to first use of `is_valid_currency_code` (lazy + cached) so a resource problem is confined to this check rather than taking down the whole library import.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1081,6 +1088,121 @@ def is_valid_national_id(column: str | Column, country: str = "US") -> Column:
...
+# call (sorted() + one F.lit() per code) would repeat needless work on every check evaluation.
+_ISO_4217_LITERALS_BY_FORMAT: dict[str, list[Column]] = {
+    fmt: [F.lit(code) for code in sorted(codes)] for fmt, codes in _ISO_4217_CODES_BY_FORMAT.items()
+}
+_ISO_4217_LITERALS_BY_FORMAT_LOWER: dict[str, list[Column]] = {
```
mwojtyczka: **Numeric lowercase literals are dead.** `_ISO_4217_LITERALS_BY_FORMAT_LOWER` is built for *both* formats, but the numeric branch (checked first) never uses the lower list and case-folding applies only to alphabetic. So the numeric-lower `F.lit()` list is constructed at import and never referenced. Minor: build the lower list for `alphabetic` only.

tests/perf/conftest.py
```diff
@@ -341,3 +341,21 @@ def generated_national_id_df(spark):
...
     return gen.build()
+
+
+@pytest.fixture
+def generated_currency_code_df(spark):
+    currency_schema_str = "col1_currency_code: string, col2_currency_code: string"
+    schema = _parse_datatype_string(currency_schema_str)
+
+    currency_templates = {
+        "col1_currency_code": r"AAA",
```
mwojtyczka: **Perf fixture generates only invalid codes.** The templates `AAA` / `KKK` are not real ISO 4217 codes, so every generated row misses `isin(*allowed)` and is flagged invalid — the benchmark exercises only the all-invalid path. `isin` can behave differently for all-miss vs mostly-hit inputs. If the intent is a realistic workload, seed with valid codes (USD/EUR/...) or a mix; if all-invalid is deliberate, a short comment noting that would help.

mwojtyczka (commented):
Re-review of the current head. Thanks for the updates — my earlier comments are all addressed and I've resolved those threads. A few **new** items from this pass, inline below. The strongest is the `Path(str(files()))` resource read (one-line change to the idiomatic Traversable API). Minor (no inline): `col_expr.isin(*allowed)` builds a ~178-way OR — `is_in_list` already warns to prefer `foreign_key` for large lists, so a similar note here would help.

mwojtyczka (changes requested):

SreeramaYeshwanthGowd:
@mwojtyczka Changes are done, please take a look.

• Zip import safe read
• Lazy loaded caches
• Dead literals removed
• Perf fixture realistic mix

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1081,6 +1088,136 @@ def is_valid_national_id(column: str | Column, country: str = "US") -> Column:
...
+        raise InvalidParameterError(
+            f"Unsupported code_format for currency code validation: '{code_format}'. Supported: [{supported}]."
+        )
+    col_str_norm, col_expr_str, col_expr = get_normalized_column_and_expr(column)
+    if normalized_format == "numeric":
+        # Numeric codes have no case, so case_sensitive is a no-op. Cast explicitly to string: without
+        # it, comparing a non-string (e.g. int) column against the string literals below makes Spark
+        # coerce the literals to the column's type instead, silently stripping leading zeros (e.g.
+        # "008" -> 8) and letting an unpadded value like 8 match a code that requires "008".
+        col_expr_compare = col_expr.cast("string")
```
mwojtyczka, line 1274: For `code_format="numeric"` on an **integer** column, the value is cast to string without re-adding zero-padding, so a legitimately-typed int code like `8` (Albania `008`) or `36` (Australia `036`) is compared as `'8'`/`'36'` and flagged invalid. This is now documented and tested as intended, but it makes the numeric check effectively unusable on integer-typed columns holding real ISO codes — a natural storage choice. Consider left-padding the cast for integer inputs, e.g. `F.lpad(col_expr.cast("string"), 3, "0")`, so `8` matches `008`. If the current behavior is deliberate, no change needed — just flagging that users will likely hit it as a surprise.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1081,6 +1088,136 @@ def is_valid_national_id(column: str | Column, country: str = "US") -> Column:
...
+# https://www.iso.org/iso-4217-currency-codes.html; the values were verified against it and cover
+# the active codes as of July 2026. The code lists are stored as data files under the resources
+# package and loaded via importlib.resources. To regenerate them, iterate pycountry.currencies
+# (which packages the ISO 4217 data) as a convenience, reading each entry's alphabetic code (exposed
+# by pycountry as the alpha_3 attribute) and numeric code, then reconcile against the official ISO
+# list above before committing.
+#
+# Loaded lazily (on first call, then cached) rather than at module import time, so that a resource
+# packaging problem only breaks this check, not the import of the whole check_funcs module.
+@lru_cache(maxsize=1)
```
mwojtyczka, line 1195: Minor: three separate `lru_cache` helpers (`_iso_4217_codes_by_format`, `_iso_4217_literals_by_format`, `_iso_4217_alphabetic_literals_lower`) encode overlapping state — the lowercase-alphabetic variant duplicates the `sorted() + F.lit()` pipeline of the general builder. For a 178-element constant list this could collapse to a single cached map keyed by `(format, case)`, or just cache the sorted code tuples and build the `F.lit` lists inline. Not blocking.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1081,6 +1088,136 @@ def is_valid_national_id(column: str | Column, country: str = "US") -> Column:
...
+
+    The large standard code lists are stored as data files rather than inline literals to keep them
+    readable and easy to regenerate. See the files under *databricks/labs/dqx/resources* for the
+    values and their authoritative sources.
+    """
+    # Read directly off the Traversable returned by files(): wrapping it in Path(str(...)) assumes a
+    # real filesystem path, which is not guaranteed under a zipped wheel (zipimport).
+    codes = frozenset((files("databricks.labs.dqx.resources") / resource_name).read_text(encoding="utf-8").split())
+    if not codes:
+        raise ComputationError(
```
mwojtyczka: `ComputationError` is used everywhere else in the codebase (`anomaly/core.py`, `anomaly/transformers.py`) for **Spark aggregate/statistic computation** failures. A missing or empty packaged resource file is a packaging/installation fault, not a computation failure — using `ComputationError` here conflates two unrelated failure classes (someone catching `ComputationError` around anomaly stats could unexpectedly swallow this). A packaging/environment-oriented `DQXError` subclass would classify it more accurately. Minor.

mwojtyczka (commented):
## Re-review — `is_valid_currency_code`

Thanks for the thorough follow-up. I verified the round-1 items are all addressed and re-ran the currency tests against a live workspace (**5 passed**):

- numeric `case_sensitive` no-op — documented + tested
- integer-column behavior — documented + tested
- case-insensitive `code_format` — documented
- `_load_iso_codes` empty/missing guard — added
- literals cached + lazy load — done

Only three minor, non-blocking items remain (none are correctness bugs):

1. **numeric check on integer columns** — the documented behavior means int-typed numeric codes lacking zero-padding are always flagged; worth considering `lpad` for int inputs. (inline)
2. **three overlapping `lru_cache` helpers** — could collapse to one. (inline)
3. **`ComputationError` semantics** — a missing packaged resource is a packaging fault, not a compute error. (inline)

mwojtyczka:
I think we also need CI step to ensure the codes are up to date

SreeramaYeshwanthGowd:
@mwojtyczka Changes are done, please take a look.

• New resource error class
• Caches consolidated
• Int column note added
• Pycountry parity test

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: Addressed in e445da0: collapsed to a single parameterized cache _iso_4217_literals(fmt, lower) keyed by (format, case), as suggested. Resolving.

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: Addressed in e445da0: the docstring now documents the integer-column behavior and gives the F.lpad(col_expr.cast('string'), 3, '0') workaround. The behavior is intended (per the original note, documentation was the ask). Resolving.

tests/unit/test_iso_4217_freshness.py
```diff
@@ -0,0 +1,22 @@
...
+
+def test_iso_4217_codes_match_pycountry():
+    """Guards against the packaged ISO 4217 resource files drifting from the upstream standard data.
+
+    Fails when pycountry is upgraded to a release whose currency data no longer matches
+    src/databricks/labs/dqx/resources/iso_4217_{alphabetic,numeric}.txt, so a maintainer knows to
+    regenerate those files (see the regeneration steps in check_funcs.py) rather than silently
+    validating against a stale code list.
+    """
+    expected_alphabetic = frozenset(currency.alpha_3 for currency in pycountry.currencies)
```
mwojtyczka: **Two concerns on the freshness assertion:**

1. **pycountry becomes the source of truth, contradicting the reconcile-against-ISO intent.** This asserts the committed lists *exactly* equal `pycountry.currencies`, but the `check_funcs.py` comment says pycountry is only a convenience and the lists are reconciled against the official ISO 4217 list before committing. `pycountry` is pinned `~=26.2.16`, which permits `26.x` data updates; if a future release adds/withdraws a currency or a funds code (or includes entries the curated list intentionally excludes), this exact-equality test red-builds even when the committed list is *more* correct. Prefer a documented direction/subset assertion, or pin pycountry exactly.

2. **Attribute coupling.** `currency.alpha_3` / `currency.numeric` — if a pycountry release renames these or makes `numeric` optional, the comprehension raises `AttributeError`/`TypeError` at collection time rather than a meaningful assertion failure. (Same freshness-test design as #1369.)

tests/unit/test_iso_4217_freshness.py
```diff
@@ -0,0 +1,22 @@
+from importlib.resources import files
+
+import pycountry
+
+
+def _read_packaged_codes(resource_name: str) -> frozenset[str]:
```
mwojtyczka: `_read_packaged_codes` reimplements `_load_iso_codes`'s reader verbatim (`frozenset((files(...) / name).read_text(encoding="utf-8").split())`). If the production reader ever changes (encoding, split rules, comment/blank-line handling), the test keeps validating the old logic and silently diverges. Import and exercise the real loader — `_iso_4217_codes_by_format()` or `_load_iso_codes` — so it covers what actually runs.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1081,6 +1088,134 @@ def is_valid_national_id(column: str | Column, country: str = "US") -> Column:
...
+        # "008" -> 8) and letting an unpadded value like 8 match a code that requires "008".
+        col_expr_compare = col_expr.cast("string")
+        allowed = _iso_4217_literals(normalized_format, lower=False)
+    elif case_sensitive:
+        col_expr_compare = col_expr
+        allowed = _iso_4217_literals(normalized_format, lower=False)
+    else:
+        col_expr_compare = to_lowercase(col_expr)
+        allowed = _iso_4217_literals(normalized_format, lower=True)
+    condition = F.when(col_expr.isNotNull(), ~col_expr_compare.isin(*allowed)).otherwise(F.lit(None))
```
mwojtyczka: Minor / consistency: the `F.when(col_expr.isNotNull(), ~...isin(...)).otherwise(F.lit(None))` wrapper is redundant — `~col_expr_compare.isin(*allowed)` already yields NULL for NULL input (SQL null propagation), and `make_condition` treats NULL as a pass. `condition = ~col_expr_compare.isin(*allowed)` suffices. The identical simplification was applied to `has_valid_string_case` in #1347, so keeping the wrapper here (and in #1369) is inconsistent across the row checks.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1081,6 +1088,134 @@ def is_valid_national_id(column: str | Column, country: str = "US") -> Column:
     return _matches_pattern(column, pattern)
 
 
+def _load_iso_codes(resource_name: str) -> frozenset[str]:
```
mwojtyczka: **Duplicates the shared helper with the still-open #1369.** This PR and #1369 (`is_valid_country_code`) each add an identical `_load_iso_codes`, `resources/__init__.py`, and the `importlib.resources` import + lazy-cache scaffolding; neither is on `main`. The second to merge will conflict here. The two have also **diverged** — this PR uses a single `_iso_4217_literals(fmt, lower)` and `MissingResourceError`, while #1369 uses two zero-arg helpers and `ComputationError` — so the cleanest path is to extract the shared `_load_iso_codes` (and the freshness-test scaffolding) once and rebase both onto it, which would also resolve the freshness-test findings for both PRs in one place. Coordinate merge order.

mwojtyczka (commented):
Re-review of the current head. The updates are solid and a step ahead of the sibling #1369: zip-safe read, lazy `@lru_cache` loading, a single parameterized literal cache (`_iso_4217_literals(fmt, lower)`), a purpose-built `MissingResourceError`, the `F.lpad` docstring tip for integer columns, a realistic perf fixture, and the isin perf caveat. My two prior open threads (cache consolidation, numeric-int) are addressed and resolved. A few new items on the freshness test and one consistency nit below.

mwojtyczka (approved):
LGTM

mwojtyczka:
**Merge coordination with the sibling PR (#1368 ↔ #1369).** Both this PR and its sibling add an identical public `load_iso_codes` helper, the `databricks/labs/dqx/resources/__init__.py` package marker, the `importlib.resources` import, and the `MissingResourceError` error type — none of which are on `main`. Whichever merges second will hit a conflict on these shared additions.

The two have also diverged slightly (currency uses `_iso_4217_literals(fmt, lower)`; country uses a `(format, case)`-keyed cache), so the conflict will need a manual resolution rather than a clean auto-merge. Suggested options for the maintainer:
- Merge one, then rebase the other and resolve the shared-helper conflict, or
- Extract the shared scaffolding (`load_iso_codes`, `resources/__init__.py`, `MissingResourceError`, freshness-test pattern) into a small precursor PR that both rebase onto.

Flagging so the merge order is a deliberate choice. All code-level review threads here are resolved.

PR #1369, opened 2026-07-23
Add is_valid_country_code row-level check

### LINKED ISSUE #1341: [FEATURE]: New check and generation for currency codes for ISO-3166

### Problem statement

https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes is widely used standard for encoding country codes and names. This can be re-used easily with `is_in`, but supporting long list is tedious.

### Proposed Solution

This list can be reused for dedicated check. Additionally profiler and generation can be added using condition: generate check if string column match ratio avoid threshold with default 99%.

### Additional Context

_No response_

SreeramaYeshwanthGowd:
## Changes

Adds a new row-level check `is_valid_country_code(column, code_format="alpha-2")` that validates values against the ISO 3166-1 country codes. ISO 3166-1 defines three representations, selected with the `code_format` argument: the two-letter **alpha-2** code (default, e.g. `US`), the three-letter **alpha-3** code (e.g. `USA`) and the three-digit **numeric** code (e.g. `840`). Validation is a case-sensitive membership test; null values pass with no violation reported.

The code lists are embedded as module-level frozensets generated from the ISO 3166-1 standard data in the `pycountry` package (dev-time generation; the regeneration steps are in a comment). The implementation follows the existing pattern-validator checks such as `is_valid_email` and `is_valid_national_id`.

### Linked issues

Resolves #1341

### Tests

- [ ] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [x] added performance tests

- Unit tests in `tests/unit/test_row_checks.py` cover the auto-generated column name (all three formats) and `code_format` validation (None, non-string, unsupported value).
- Integration tests in `tests/integration/test_row_checks.py` cover the alpha-2, alpha-3 and numeric formats (valid, invalid, case-sensitive, empty and null cases) plus column-expression input.
- `is_valid_country_code` is also added to `tests/resources/all_row_checks.yaml` and the "all checks" integration tests (`test_apply_checks_all_checks_as_yaml`, `test_apply_checks_all_row_checks_as_yaml_with_streaming`, and `test_apply_checks_all_checks_using_classes`).
- A performance benchmark (`test_benchmark_is_valid_country_code`) is added under `tests/perf/`.

### Documentation and Demos

- [ ] added/updated demos
- [x] added/updated docs

`docs/dqx/docs/reference/quality_checks.mdx` gets a reference-table entry and YAML plus programmatic (`DQRowRule`) usage examples.


src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1022,6 +1022,298 @@ def is_valid_email(column: str | Column) -> Column:
     return _matches_pattern(column, DQPattern.EMAIL_ADDRESS)
 
 
+# ISO 3166-1 alpha-2 country codes (the 249 officially assigned codes), derived from the CLDR
```
IvannKurchenko: Thank you for the contribution. Please, consider supporting alpha-3 codes as well: https://en.wikipedia.org/wiki/ISO_3166-1_alpha-3

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1022,6 +1022,298 @@ def is_valid_email(column: str | Column) -> Column:
...
+        "YT",
+        "ZA",
+        "ZM",
+        "ZW",
+    }
+)
+
+
+@register_rule("row")
+def is_valid_country_code(column: str | Column) -> Column:
```
IvannKurchenko: Might be not a bad idea to parametrise code category: `alpha-2` or `alpha-3`

IvannKurchenko (commented):
Thank you for the contribution. Left some suggestions

SreeramaYeshwanthGowd:
@IvannKurchenko thanks for the review. I added a code_format argument so the check validates the alpha-2 (default), alpha-3 or numeric representation, and switched the embedded list to be generated from pycountry, which packages the ISO 3166-1 standard data.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1081,6 +1081,833 @@ def is_valid_national_id(column: str | Column, country: str = "US") -> Column:
     return _matches_pattern(column, pattern)
 
 
+# ISO 3166-1 country codes, generated from the pycountry package (which bundles the ISO 3166-1
+# standard data). ISO 3166-1 defines a two-letter alpha-2 code (e.g. US), a three-letter alpha-3
+# code (e.g. USA) and a three-digit numeric code (e.g. 840); all three are supported. To
+# regenerate, iterate pycountry.countries and collect the alpha_2, alpha_3 and numeric attributes.
+_ISO_3166_1_ALPHA_2_CODES: frozenset[str] = frozenset(
```
IvannKurchenko: Thank you for extending the list. Because of the size of these lists, this package might become challenging to support. Please, consider using https://docs.python.org/3/library/importlib.resources.html and move these lists as separate files under `resources` folder within same package.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1081,6 +1081,833 @@ def is_valid_national_id(column: str | Column, country: str = "US") -> Column:
     return _matches_pattern(column, pattern)
 
 
+# ISO 3166-1 country codes, generated from the pycountry package (which bundles the ISO 3166-1
+# standard data). ISO 3166-1 defines a two-letter alpha-2 code (e.g. US), a three-letter alpha-3
+# code (e.g. USA) and a three-digit numeric code (e.g. 840); all three are supported. To
+# regenerate, iterate pycountry.countries and collect the alpha_2, alpha_3 and numeric attributes.
```
IvannKurchenko: I'd suggest to prioritise reference to official resources instead of another implementation: https://www.iso.org/iso-3166-country-codes.html

tests/integration/test_apply_checks.py
```diff
@@ -7244,6 +7267,7 @@ def test_apply_checks_all_checks_using_classes(ws, spark):
                 '{"key1": "[1, 2, 3]"}',
                 '{ "a" : 1, "b": 1023455,  "c": null }',
                 "323-45-6789",
+                "DE",
```
IvannKurchenko, line 7318: Please, consider adding integration tests for both alpha-3 and numeric codes.

tests/integration/test_row_checks.py
```diff
@@ -2031,6 +2032,122 @@ def test_col_is_valid_national_id_column_expr_and_lowercase_country(spark):
...
+    schema = "a: string"
+    test_df = spark.createDataFrame(
+        [
+            # Valid alpha-2 codes
+            ["US"],
+            ["GB"],
+            ["DE"],
+            ["FR"],
+            # Invalid
+            ["USA"],  # alpha-3 code, not alpha-2
```
IvannKurchenko, line 2156: Might be a good idea to test also numeric code

tests/integration/test_row_checks.py
```diff
@@ -2031,6 +2032,122 @@ def test_col_is_valid_national_id_column_expr_and_lowercase_country(spark):
...
+
+
+def test_col_is_valid_country_code_alpha_3(spark):
+    schema = "a: string"
+    test_df = spark.createDataFrame(
+        [
+            ["USA"],
+            ["GBR"],
+            ["DEU"],
+            ["US"],  # alpha-2 code, not alpha-3
```
IvannKurchenko, line 2194: Might be a good idea to test also numeric code.

tests/integration/test_row_checks.py
```diff
@@ -2031,6 +2032,122 @@ def test_col_is_valid_national_id_column_expr_and_lowercase_country(spark):
...
+def test_col_is_valid_country_code_numeric(spark):
+    schema = "a: string"
+    test_df = spark.createDataFrame(
+        [
+            ["840"],
+            ["826"],
+            ["276"],
+            ["8"],  # too short, codes are zero-padded to three digits
+            ["0840"],  # extra leading zero
+            ["US"],  # alpha-2 code, not numeric
```
IvannKurchenko, line 2234: Might be a good idea to test also alpha-3 code

docs/dqx/docs/reference/quality_checks.mdx
```diff
@@ -61,6 +61,7 @@ You can also define your own custom checks in Python (see [Creating custom check
...
+| `is_valid_country_code`            | Checks whether the values in the input column are valid ISO 3166-1 country codes (alpha-2, e.g. US, alpha-3, e.g. USA, or numeric, e.g. 840).                                                                                                                                                                                                                                                                                     | `column`: column to check (can be a string column name or a column expression); `code_format`: ISO 3166-1 representation, `alpha-2` (default), `alpha-3`, or `numeric`                                                                                                                                                                                                                                                                                                                                                                               |
```
IvannKurchenko: I'd recommend to mention the source of values, which ideally should be : https://www.iso.org/iso-3166-country-codes.html

IvannKurchenko (commented):
Thank you a lot of your feedback and actions

SreeramaYeshwanthGowd:
@IvannKurchenko thanks, I moved the code lists into resource files under a resources package and load them with importlib.resources. I also referenced the official ISO 3166 source (https://www.iso.org/iso-3166-country-codes.html) in the docstring and the docs. On tests, alpha-3 and numeric already have dedicated integration tests (test_col_is_valid_country_code_alpha_3 and test_col_is_valid_country_code_numeric) plus unit coverage.

IvannKurchenko (commented):
Thank you for addressing all the feedback. Please wait for approval from other reviewers.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1081,6 +1083,92 @@ def is_valid_national_id(column: str | Column, country: str = "US") -> Column:
...
+            f"Unsupported code_format for country code validation: '{code_format}'. Supported: [{supported}]."
+        )
+    col_str_norm, col_expr_str, col_expr = get_normalized_column_and_expr(column)
+    if case_sensitive:
+        col_expr_compare = col_expr
+        allowed = [F.lit(code) for code in sorted(allowed_codes)]
+    else:
+        col_expr_compare = to_lowercase(col_expr)
+        allowed = [F.lit(code.lower()) for code in sorted(allowed_codes)]
+    condition = F.when(col_expr.isNotNull(), ~col_expr_compare.isin(*allowed)).otherwise(F.lit(None))
```
mwojtyczka: **`case_sensitive` is a silent no-op for `numeric`.** ISO 3166-1 numeric codes are all digits, so `to_lowercase` / `.lower()` changes nothing — yet `code_format="numeric", case_sensitive=False` still folds the column and all 249 literals with `F.lower`, adding per-row work with no effect and no signal that the flag is meaningless. Either reject the combination or document that `case_sensitive` is ignored for the numeric format and skip the folding there.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1081,6 +1083,92 @@ def is_valid_national_id(column: str | Column, country: str = "US") -> Column:
...
+    if code_format is None:
+        raise MissingParameterError("'code_format' is not provided.")
+    if not isinstance(code_format, str):
+        raise InvalidParameterError(f"'code_format' must be a string, got {type(code_format)} instead.")
+    normalized_format = code_format.lower()
+    allowed_codes = _ISO_3166_1_CODES_BY_FORMAT.get(normalized_format)
+    if allowed_codes is None:
+        supported = ", ".join(sorted(_ISO_3166_1_CODES_BY_FORMAT))
+        raise InvalidParameterError(
+            f"Unsupported code_format for country code validation: '{code_format}'. Supported: [{supported}]."
```
mwojtyczka: **Numeric check on a non-string (integer) column fails silently.** The docstring warns numeric input "must preserve the leading zeros", but nothing enforces string input. If the column is an int type, `isin` compares integer values against zero-padded string literals like `'004'`, so valid codes (e.g. `4` for Afghanistan, stored as `'004'`) never match and every row is flagged. There's no guard/warning and no test with an integer input column. Consider casting the column to string for the comparison (`col_expr.cast("string")`), or documenting/validating that numeric input must already be a zero-padded string, and add an int-column test.

mwojtyczka (commented):
## Code review — `is_valid_country_code`

Well-built PR, and structurally the sibling of #1368. I verified the risky parts and both are correct:

- **Wheel packaging** ✅ — all three `iso_3166_1_*.txt` files ship in the built wheel (`importlib.resources` pattern matches `llm_utils.py`).
- **Data integrity** ✅ — 249 codes each (matches ISO 3166-1's officially-assigned count), no width anomalies, numeric zero-padding preserved, and **XK (Kosovo) correctly excluded** as the docstring claims.

> Note (not inline): this PR and #1368 both add a byte-for-byte identical `_load_iso_codes` helper and a new `resources/__init__.py`, neither of which is on `main`. Whichever merges second will hit a conflict (duplicate function definition). Worth coordinating so the shared helper lands once and the other PR imports it — flagging here rather than as an inline comment.

Findings below; the numeric-format ones (case-folding no-op, integer input) and the doc note are the ones worth resolving before merge.

### Correctness / robustness
2. **`case_sensitive` is a silent no-op for `numeric`** — digits have no case, but the flag is accepted and still runs `F.lower` over every row and all 249 literals. (inline)
3. **Integer input column fails silently** — numeric compared against zero-padded string literals (`'004'`); no guard, no int-column test. (inline)
4. **`code_format` matching is case-insensitive but undocumented** — `.lower()` accepts `'Alpha-2'` (a test relies on it), docstring lists only lowercase. (inline)
5. **`_load_iso_codes` has no empty/missing guard** — import-time crash or silent empty-set on a packaging regression. (inline)

### Efficiency
6. **Literal list rebuilt per call** — `sorted()` + 249 `F.lit(...)` recomputed on every invocation. (inline)

### Tests
7. **No `numeric`/`alpha-3` + `case_sensitive=False` test** — only alpha-2 exercises the case-insensitive path. (inline)

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1081,6 +1083,92 @@ def is_valid_national_id(column: str | Column, country: str = "US") -> Column:
...
+
+    Raises:
+        MissingParameterError: if *code_format* is None.
+        InvalidParameterError: if *code_format* is not a string, or is not a supported representation.
+    """
+    if code_format is None:
+        raise MissingParameterError("'code_format' is not provided.")
+    if not isinstance(code_format, str):
+        raise InvalidParameterError(f"'code_format' must be a string, got {type(code_format)} instead.")
+    normalized_format = code_format.lower()
```
mwojtyczka: `normalized_format = code_format.lower()` makes matching case-insensitive, so `'Alpha-2'` / `'NUMERIC'` are accepted (your `test_is_valid_country_code_format_is_case_insensitive` relies on this). The docstring lists only lowercase `alpha-2`/`alpha-3`/`numeric`, so the accepted input set is wider than documented — add a sentence noting the matching is case-insensitive.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1081,6 +1083,92 @@ def is_valid_national_id(column: str | Column, country: str = "US") -> Column:
     return _matches_pattern(column, pattern)
 
 
+def _load_iso_codes(resource_name: str) -> frozenset[str]:
```
mwojtyczka: `_load_iso_codes` runs at import time and returns `frozenset(text.split())` with no validation. The `.txt` files ship in the wheel today (I verified the built wheel), but if a future packaging change drops or empties one, this either raises `FileNotFoundError` at import — breaking **all** checks, since the module fails to load — or silently returns an empty set that flags every non-null value as invalid. A minimal guard (assert the loaded set is non-empty, raising a clear error that names the missing resource) would fail fast and localize the problem.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1081,6 +1083,92 @@ def is_valid_national_id(column: str | Column, country: str = "US") -> Column:
...
+    allowed_codes = _ISO_3166_1_CODES_BY_FORMAT.get(normalized_format)
+    if allowed_codes is None:
+        supported = ", ".join(sorted(_ISO_3166_1_CODES_BY_FORMAT))
+        raise InvalidParameterError(
+            f"Unsupported code_format for country code validation: '{code_format}'. Supported: [{supported}]."
+        )
+    col_str_norm, col_expr_str, col_expr = get_normalized_column_and_expr(column)
+    if case_sensitive:
+        col_expr_compare = col_expr
+        allowed = [F.lit(code) for code in sorted(allowed_codes)]
```
mwojtyczka: `sorted(allowed_codes)` and the 249-element `[F.lit(code) ...]` list are rebuilt on every call. Since the code sets are fixed module constants, the sorted literal lists (per format, and per case) could be precomputed once at module load and reused. With a `DQForEachColRule` spanning many columns this rebuild happens once per column at plan-build time.

tests/integration/test_row_checks.py
```diff
@@ -2031,6 +2032,155 @@ def test_col_is_valid_national_id_column_expr_and_lowercase_country(spark):
...
+        [violation("US")],
+        [violation("")],
+        [None],
+    ]
+    expected = spark.createDataFrame(checked_data, checked_schema)
+
+    assertDataFrameEqual(actual, expected)
+
+
+def test_col_is_valid_country_code_case_insensitive(spark):
```
mwojtyczka, line 2262: `case_sensitive=False` is exercised only for the alpha-2 format. There's no test for `code_format="numeric"` (no-op folding) or `"alpha-3"` with `case_sensitive=False`, so those interactions are unverified. Add cases so a future change to the case-folding path can't silently break numeric/alpha-3 matching.

mwojtyczka (changes requested):
Thanks for the PR. Generally looking good, left some comments that need to be addressed before merging.

SreeramaYeshwanthGowd:
@mwojtyczka Changes are done, please take a look.

• Case sensitive skipped
• Integer column cast
• Code format note
• Empty resource guard
• Precomputed literal lists
• Two new tests

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: Addressed: docstring states case_sensitive has no effect for numeric, the numeric branch is checked first (no folding), and test_col_is_valid_country_code_numeric_case_insensitive_is_noop covers it. Resolving.

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: Addressed: numeric branch casts col_expr to string, docstring explains the int leading-zero behavior, and test_col_is_valid_country_code_numeric_int_column covers it. Resolving.

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: Addressed: docstring now notes code_format matching is case-insensitive. Resolving.

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: Partially addressed: _load_iso_codes now raises on an empty set. Two related angles remain open and are posted as new comments (Path(str(files())) is not zip-import safe; import-time load breaks the whole module import). Resolving this original thread.

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: Addressed: literals are precomputed once at module scope now (_ISO_3166_1_LITERALS_BY_FORMAT / _LOWER). Resolving.

tests/integration/test_row_checks.py
mwojtyczka: Addressed: test_col_is_valid_country_code_numeric_case_insensitive_is_noop now covers numeric + case_sensitive=False. Resolving.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1081,6 +1088,119 @@ def is_valid_national_id(column: str | Column, country: str = "US") -> Column:
     return _matches_pattern(column, pattern)
 
 
+def _load_iso_codes(resource_name: str) -> frozenset[str]:
```
mwojtyczka: **Duplicates the shared helper with the still-open #1368.** This PR and #1368 (`is_valid_currency_code`) each independently add the identical `_load_iso_codes`, `resources/__init__.py`, and the `from importlib.resources import files` / `from pathlib import Path` imports (plus the `ComputationError` import edit). Neither is merged. Whichever lands second will hit a git conflict on `check_funcs.py` (redefined function + duplicate imports). Suggest coordinating: land the shared `_load_iso_codes` once (its own small PR or one of these), and rebase the other onto it.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1081,6 +1088,119 @@ def is_valid_national_id(column: str | Column, country: str = "US") -> Column:
...
+    """Load a set of standard codes from a newline-delimited data file in the resources package.
+
+    The large standard code lists are stored as data files rather than inline literals to keep them
+    readable and easy to regenerate. See the files under *databricks/labs/dqx/resources* for the
+    values and their authoritative sources.
+    """
+    resource = Path(str(files("databricks.labs.dqx.resources") / resource_name))
```
mwojtyczka: **Resource read is not zip-import safe.** `Path(str(files("databricks.labs.dqx.resources") / resource_name))` assumes the package is on a real filesystem. `files()` returns a `Traversable`; under a zipped wheel / zipimport, `str(traversable)` is not a filesystem path and `Path(...).read_text()` raises `FileNotFoundError`. Since this runs at import (populating the module-level dicts), it would break importing `check_funcs` entirely, not just this check. Use the idiomatic form that works on-disk and zipped: `(files("databricks.labs.dqx.resources") / resource_name).read_text(encoding="utf-8")` — no `Path`/`str`. Databricks installs are unpacked so it works today; removes a latent failure with no downside.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1081,6 +1088,119 @@ def is_valid_national_id(column: str | Column, country: str = "US") -> Column:
...
+    return codes
+
+
+# ISO 3166-1 country codes. The authoritative source is
+# https://www.iso.org/iso-3166-country-codes.html; the values were verified against it and cover the
+# officially assigned codes as of July 2026. The code lists are stored as data files under the
+# resources package and loaded via importlib.resources. To regenerate them, iterate
+# pycountry.countries (which packages the ISO 3166-1 data) as a convenience, then reconcile against
+# the official ISO list above before committing.
+_ISO_3166_1_CODES_BY_FORMAT: dict[str, frozenset[str]] = {
```
mwojtyczka: **Import-time load couples the whole module to this check's resources.** `_ISO_3166_1_CODES_BY_FORMAT = {..: _load_iso_codes(..)}` runs at import, so a missing/empty resource raises during `import check_funcs` and makes `DQEngine` and every unrelated check unimportable. The empty-set guard you added is good — consider also deferring the load to first use of `is_valid_country_code` (lazy + cached) so a resource problem is confined to this check.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1081,6 +1088,119 @@ def is_valid_national_id(column: str | Column, country: str = "US") -> Column:
...
+    "alpha-2": _load_iso_codes("iso_3166_1_alpha_2.txt"),
+    "alpha-3": _load_iso_codes("iso_3166_1_alpha_3.txt"),
+    "numeric": _load_iso_codes("iso_3166_1_numeric.txt"),
+}
+# Precomputed once at module load: the literal lists never change at runtime, so building them per
+# call (sorted() + one F.lit() per code) would repeat needless work on every check evaluation.
+_ISO_3166_1_LITERALS_BY_FORMAT: dict[str, list[Column]] = {
+    fmt: [F.lit(code) for code in sorted(codes)] for fmt, codes in _ISO_3166_1_CODES_BY_FORMAT.items()
+}
+_ISO_3166_1_LITERALS_BY_FORMAT_LOWER: dict[str, list[Column]] = {
```
mwojtyczka: **Numeric lowercase literals are dead.** `_ISO_3166_1_LITERALS_BY_FORMAT_LOWER` is built for all three formats, but the numeric branch (checked first) never uses the lower list and case-folding applies only to alpha-2/alpha-3. So the numeric-lower `F.lit()` list (~249) is constructed at import and never referenced. Minor: build the lower list for the alpha formats only.

tests/perf/conftest.py
```diff
@@ -341,3 +341,21 @@ def generated_national_id_df(spark):
...
     return gen.build()
+
+
+@pytest.fixture
+def generated_country_code_df(spark):
+    country_schema_str = "col1_country_code: string, col2_country_code: string"
+    schema = _parse_datatype_string(country_schema_str)
+
+    country_templates = {
+        "col1_country_code": r"AA",
```
mwojtyczka: **Perf fixture generates only invalid codes.** `AA` / `KK` are not officially assigned ISO 3166-1 codes, so every generated row misses `isin(*allowed)` and is flagged invalid — the benchmark exercises only the all-invalid path. `isin` can behave differently for all-miss vs mostly-hit inputs. Seed with valid codes (US/GB/DE/...) or a mix for a realistic workload; if all-invalid is deliberate, a short comment noting that would help.

mwojtyczka (commented):
Re-review of the current head. Thanks for the updates — my earlier comments (case-sensitive no-op, numeric int-column cast, case-insensitive `code_format`, empty-resource guard, precomputed literals, numeric case-insensitive test) are all addressed, and I've resolved those threads. Coverage here is good (alpha-2/alpha-3/numeric, case-insensitive for both alpha formats, numeric int-column).

A few **new** items below. The most important is cross-PR: this shares `_load_iso_codes` verbatim with the still-open #1368. Minor (no inline): `col_expr.isin(*allowed)` builds a ~249-way OR — `is_in_list` already advises `foreign_key` for large lists, so a similar note here would help.

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: Just flagging, no action needed now.

mwojtyczka (changes requested):

SreeramaYeshwanthGowd:
@mwojtyczka Changes are done, please take a look.

• Zip import safe read
• Lazy loaded caches
• Dead literals removed
• Perf fixture realistic mix
• Duplicate helper acknowledged

mwojtyczka:
I think we also need CI step to ensure the codes are up to date, maybe as nightly

SreeramaYeshwanthGowd:
@mwojtyczka Changes are done, please take a look.

• Pycountry parity test

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1081,6 +1088,140 @@ def is_valid_national_id(column: str | Column, country: str = "US") -> Column:
...
+        raise InvalidParameterError(
+            f"Unsupported code_format for country code validation: '{code_format}'. Supported: [{supported}]."
+        )
+    col_str_norm, col_expr_str, col_expr = get_normalized_column_and_expr(column)
+    if normalized_format == "numeric":
+        # Numeric codes have no case, so case_sensitive is a no-op. Cast explicitly to string:
+        # without it, comparing a non-string (e.g. int) column against the string literals below
+        # makes Spark coerce the literals to the column's type instead, silently stripping leading
+        # zeros (e.g. "004" -> 4) and letting an unpadded value like 4 match a code that requires "004".
+        col_expr_compare = col_expr.cast("string")
```
mwojtyczka: For `code_format="numeric"` on an **integer** column, the value is cast to string without re-adding zero-padding, so a legitimately-typed int code like `4` (Afghanistan `004`) or `36` is compared as `'4'`/`'36'` and flagged invalid. This is now documented and tested as intended, but it makes the numeric check effectively unusable on integer-typed columns holding real ISO codes — a natural storage choice. Consider left-padding the cast for integer inputs, e.g. `F.lpad(col_expr.cast("string"), 3, "0")`, so `4` matches `004`. If the current behavior is deliberate, no change needed — just flagging that users will likely hit it as a surprise.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1081,6 +1088,140 @@ def is_valid_national_id(column: str | Column, country: str = "US") -> Column:
...
+        "alpha-3": _load_iso_codes("iso_3166_1_alpha_3.txt"),
+        "numeric": _load_iso_codes("iso_3166_1_numeric.txt"),
+    }
+
+
+# Precomputed once on first use and cached: the literal lists never change at runtime, so building
+# them per call (sorted() + one F.lit() per code) would repeat needless work on every check
+# evaluation. Only the alpha formats need a lowercase variant — numeric codes have no case and never
+# use it.
+@lru_cache(maxsize=1)
```
mwojtyczka: Minor: the cached literal helpers (`_iso_3166_1_literals_by_format` and `_iso_3166_1_literals_by_format_lower`) encode overlapping state — the lowercase variant duplicates the `sorted() + F.lit()` pipeline of the general builder, differing only by `.lower()` and the numeric exclusion. For fixed constant lists this could collapse to a single cached map keyed by `(format, case)`, or just cache the sorted code tuples and build the `F.lit` lists inline. Not blocking.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1081,6 +1088,140 @@ def is_valid_national_id(column: str | Column, country: str = "US") -> Column:
...
+
+    The large standard code lists are stored as data files rather than inline literals to keep them
+    readable and easy to regenerate. See the files under *databricks/labs/dqx/resources* for the
+    values and their authoritative sources.
+    """
+    # Read directly off the Traversable returned by files(): wrapping it in Path(str(...)) assumes a
+    # real filesystem path, which is not guaranteed under a zipped wheel (zipimport).
+    codes = frozenset((files("databricks.labs.dqx.resources") / resource_name).read_text(encoding="utf-8").split())
+    if not codes:
+        raise ComputationError(
```
mwojtyczka: `ComputationError` is used everywhere else in the codebase (`anomaly/core.py`, `anomaly/transformers.py`) for **Spark aggregate/statistic computation** failures. A missing or empty packaged resource file is a packaging/installation fault, not a computation failure — using `ComputationError` here conflates two unrelated failure classes (someone catching `ComputationError` around anomaly stats could unexpectedly swallow this). A packaging/environment-oriented `DQXError` subclass would classify it more accurately. Minor.

mwojtyczka (commented):
## Re-review — `is_valid_country_code`

Thanks for the follow-up. I verified the round-1 items are all addressed and re-ran the country tests against a live workspace (**8 passed**):

- numeric `case_sensitive` no-op — documented + tested
- integer-column behavior — documented + tested
- alpha-3 case-insensitive — new test added
- case-insensitive `code_format` — documented
- `_load_iso_codes` empty/missing guard — added
- literals cached + lazy load — done

I also validated a concern I nearly raised — that caching `F.lit` Column objects via `lru_cache` could break across Spark sessions. It does **not**: a Spark Connect `Column` wraps only an `Expression` with no session handle, and a cross-session reuse test passed. So the caching is safe — no change needed there.

Separately (cross-PR, not inline): this PR and #1368 both add an identical `_load_iso_codes` helper and `resources/__init__.py`, neither on `main` — whichever merges second will conflict. Worth landing the shared helper once and having the other import it.

Three minor, non-blocking items remain (none are correctness bugs):

1. **numeric check on integer columns** — documented behavior means int-typed numeric codes lacking zero-padding are always flagged; worth considering `lpad` for int inputs. (inline)
2. **four overlapping `lru_cache` helpers** — could collapse to fewer. (inline)
3. **`ComputationError` semantics** — a missing packaged resource is a packaging fault, not a compute error. (inline)

tests/unit/test_iso_3166_1_freshness.py
```diff
@@ -0,0 +1,24 @@
...
+
+def test_iso_3166_1_codes_match_pycountry():
+    """Guards against the packaged ISO 3166-1 resource files drifting from the upstream standard data.
+
+    Runs as part of the unit suite, which is also executed nightly (.github/workflows/nightly.yml),
+    so a pycountry upgrade whose country data no longer matches
+    src/databricks/labs/dqx/resources/iso_3166_1_{alpha_2,alpha_3,numeric}.txt is caught automatically
+    rather than silently validating against a stale code list.
+    """
+    expected_alpha_2 = frozenset(country.alpha_2 for country in pycountry.countries)
```
mwojtyczka: **This makes pycountry the source of truth, which contradicts the reconcile-against-ISO intent.** The test asserts the committed lists *exactly* equal `pycountry.countries`, but the `check_funcs.py` comment says pycountry is only a convenience and the lists must be reconciled against the official ISO 3166-1 list before committing. `pycountry` is pinned `~=26.2.16`, which permits `26.x` data updates; if a future pycountry release adds/removes/renumbers a country (or includes transitional/exceptional reservations the curated list intentionally excludes), this exact-equality test red-builds even when the committed list is *more* correct than pycountry. Consider asserting a direction/subset (packaged codes ⊆ pycountry, or a documented allow-list of intentional differences) instead of exact equality, or pin pycountry exactly so the data can't drift under you.

tests/unit/test_iso_3166_1_freshness.py
```diff
@@ -0,0 +1,24 @@
+from importlib.resources import files
+
+import pycountry
+
+
+def _read_packaged_codes(resource_name: str) -> frozenset[str]:
```
mwojtyczka: `_read_packaged_codes` reimplements `_load_iso_codes`'s reader verbatim (`frozenset((files(...) / name).read_text(encoding="utf-8").split())`). If the production reader ever changes (encoding, split rules, comment/blank-line handling), the test keeps validating the old logic and silently diverges. Import and exercise the real loader — e.g. `_iso_3166_1_codes_by_format()` (or `_load_iso_codes`) — so the test covers what actually runs.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1081,6 +1088,140 @@ def is_valid_national_id(column: str | Column, country: str = "US") -> Column:
...
+        # zeros (e.g. "004" -> 4) and letting an unpadded value like 4 match a code that requires "004".
+        col_expr_compare = col_expr.cast("string")
+        allowed = _iso_3166_1_literals_by_format()[normalized_format]
+    elif case_sensitive:
+        col_expr_compare = col_expr
+        allowed = _iso_3166_1_literals_by_format()[normalized_format]
+    else:
+        col_expr_compare = to_lowercase(col_expr)
+        allowed = _iso_3166_1_literals_by_format_lower()[normalized_format]
+    condition = F.when(col_expr.isNotNull(), ~col_expr_compare.isin(*allowed)).otherwise(F.lit(None))
```
mwojtyczka: Minor / consistency: the `F.when(col_expr.isNotNull(), ~...isin(...)).otherwise(F.lit(None))` wrapper is redundant — `~col_expr_compare.isin(*allowed)` already evaluates to NULL for NULL input (SQL null propagation), and `make_condition` treats NULL as a pass. `condition = ~col_expr_compare.isin(*allowed)` suffices. The identical simplification was just applied to `has_valid_string_case` in #1347, so keeping the wrapper here is inconsistent.

mwojtyczka (commented):
Re-review of the current head. The updates since my last pass are solid — zip-safe resource read, lazy `@lru_cache` loading (so a resource problem no longer breaks the whole module import), the dead numeric-lower literals removed, the perf fixture now uses a realistic valid/invalid mix, and the isin perf caveat is documented. A few **new** items on the freshness test and one consistency nit below; my earlier open threads (cross-PR dup, numeric-int, cached-literal overlap, ComputationError) still stand separately.

mwojtyczka (approved):
LGTM

mwojtyczka:
**Merge coordination with the sibling PR (#1368 ↔ #1369).** Both this PR and its sibling add an identical public `load_iso_codes` helper, the `databricks/labs/dqx/resources/__init__.py` package marker, the `importlib.resources` import, and the `MissingResourceError` error type — none of which are on `main`. Whichever merges second will hit a conflict on these shared additions.

The two have also diverged slightly (currency uses `_iso_4217_literals(fmt, lower)`; country uses a `(format, case)`-keyed cache), so the conflict will need a manual resolution rather than a clean auto-merge. Suggested options for the maintainer:
- Merge one, then rebase the other and resolve the shared-helper conflict, or
- Extract the shared scaffolding (`load_iso_codes`, `resources/__init__.py`, `MissingResourceError`, freshness-test pattern) into a small precursor PR that both rebase onto.

Flagging so the merge order is a deliberate choice. All code-level review threads here are resolved.

mwojtyczka:
## Consolidation merge pushed

I merged current `main` (which now includes the merged #1368 `is_valid_currency_code`) into this branch and resolved the conflicts by **consolidating the shared ISO-code machinery** — head is now `3aa465f3`, and the PR shows `MERGEABLE`.

**What was consolidated (no duplication):**
- Both checks now share the single, zip-safe `load_iso_codes` reader from `main`; this branch's earlier private `_load_iso_codes` copy (which used `Path(str(files(...)))`, not zip-safe) is gone.
- Removed the duplicate `MissingResourceError` the auto-merge produced. `ComputationError` is pre-existing on `main` (used by `anomaly/`), unrelated to this work.
- Single `pycountry` dev-dependency line covering both freshness tests.
- Docs, `all_row_checks.yaml`, and the integration/perf/unit tests keep both checks side by side; I collapsed the duplicated schema fragments the raw merge produced and fixed one mid-`DQRowRule` split.

**Validated locally:** no conflict markers; both checks register and share `load_iso_codes`; country (249 alpha-2) and currency (178) resources load; 117 unit tests pass incl. both ISO freshness tests; black/ruff/mypy clean.

---

## Review notes (post-merge)

The check itself is solid and thoroughly tested (alpha-2/alpha-3/numeric, case-insensitive, numeric int-column caveat, nulls, empty string). A few follow-ups, none blocking:

1. **Freshness test asserts exact equality with `pycountry`** (`test_iso_3166_1_freshness.py`). The docstring says the packaged lists are the authority "reconciled against official ISO", with pycountry "as a convenience" — but exact equality makes pycountry the source of truth. With the `~=26.2.16` pin allowing 26.x updates, a pycountry release whose country/`.numeric` set differs turns the nightly unit run red even when the packaged list correctly matches ISO. Consider subset (`packaged <= pycountry`) + an explicit allow-list for intentional divergence, or an exact (`==`) pin so upgrades are deliberate. Same pattern applies to the currency freshness test on `main`.
2. **pycountry attribute access** (`country.alpha_2/.alpha_3/.numeric`) isn't defensive — a future release with an entry missing an attribute raises `AttributeError` at collection instead of a clear drift message. `getattr(..., None)` + filter would degrade gracefully. (All entries have all three today.)
3. **Two divergent literal-cache idioms now sit adjacently**: this branch's `_iso_3166_1_literals()` (whole-dict, `maxsize=1`) vs `main`'s parameterized `_iso_4217_literals(fmt, lower)` (`maxsize=None`). Both correct, but the inconsistency is a maintenance snag.
4. **`is_valid_country_code` and `is_valid_currency_code` bodies are near-identical copies** — the validation/normalize/compare/message logic differs only in the standard label. A shared private helper (e.g. `_is_valid_iso_code(...)`) would finish the consolidation the shared reader started, so a fix to one can't drift from the other.

Items 3 and 4 are the natural next step of the same consolidation — worth doing here or as a fast follow-up so the two ISO checks stay in lockstep.

src/databricks/labs/dqx/check_funcs.py

PR #1370, opened 2026-07-23
Add has_no_gaps_per_time_window dataset-level check

### LINKED ISSUE #1366: [FEATURE]: New dataset-level check to identify gaps in time-series data

### Problem statement

Users wnat to be able to detect whether datasets contain data for every day. This is to detect gaps in the time series (for example, no data for 15/07 while 14/07 and 16/07 are present). Since a missing date does not correspond to any existing row, it cannot be flagged through a row-level check. This can be implemented today using `sql_query` but it requires a custom logic, e.g.
```
- criticality: error
  check:
    function: sql_query
    arguments:
      # Build the full daily calendar between the table's min and max date,
      # then keep only the days that have NO matching row in the data.
      query: |
        WITH bounds AS (
          SELECT min(event_date) AS min_d, max(event_date) AS max_d
          FROM {{ input_view }}
        ),
        calendar AS (
          SELECT explode(sequence(min_d, max_d, interval 1 day)) AS expected_date
          FROM bounds
        ),
        actual AS (
          SELECT DISTINCT event_date FROM {{ input_view }}
        )
        SELECT
          c.expected_date,
          true AS condition
        FROM calendar c
        LEFT ANTI JOIN actual a ON c.expected_date = a.event_date
      input_placeholder: input_view
      merge_columns:
        - expected_date          # the synthesize"
      condition_column: condition
      name: audience_missing_dates
      msg: "Missing date in time series (no rows for this day)"
```

The existing built-in check `is_data_fresh_per_time_window` cannot be used here since it only looks at days that already have rows and checks whether each has enough records. A completely missing day has no rows at all, so the check never sees it. Detecting absent dates needs a check that builds the expected calendar and compares it to what's present.

### Proposed Solution

Build new dataset-level check, e.g. `**has_no_gaps_per_time_window**`

Key insight: a dataset-level check receives the whole DataFrame and returns a condition per existing row. We can't attach an error to the missing row, but we can detect the gap from the surrounding rows and attach the violation to a real row next to it (the boundary row before the gap).

Approach (native dataset-level check, no sql_query needed):
1. Take the distinct dates and order them.
2. Use a window lead() to find the next present date for each date.
3. Where datediff(next_date, current_date) > 1, a gap exists.
4. Join back so rows on the boundary date (14/07) are flagged, e.g. "gap after 2025-07-14: next data is 2025-07-16".

The function will take `window_minutes` as parm to control the grain (days/hours/minutes).

Caveats:
- Anchoring to the data's own rows finds only interior gaps. Trailing gaps (no data for the most recent N days up to today) need an upper bound anchored to current_date() / an expected end date.
- The message lands on the boundary date, not the missing date itself - inherent, since the missing row doesn't exist.

### Additional Context

_No response_

SreeramaYeshwanthGowd:
## Changes

Adds a new dataset-level check `has_no_gaps_per_time_window(column, window_minutes, group_by=None, trailing_gap=False, curr_timestamp=None)` that detects gaps in a time series, i.e. time windows of a given size that contain no rows at all between windows that do (for example no data for 2025-07-15 while 2025-07-14 and 2025-07-16 are present). A missing window cannot be flagged with a row-level check because it has no row.

This implements the native dataset-level approach described in the issue: distinct values of the column are bucketed into fixed `window_minutes` windows, a window `lead()` finds the next present window, and a gap is flagged wherever the next present window starts more than one window after the current one. Since a missing window has no row to attach a violation to, the gap is reported on every row in the last present window before the gap.

**Grouped detection.** When `group_by` is provided, gaps are detected independently within each group (for example per device or session), which is the common case for IoT or clickstream data. When omitted, the whole column is treated as a single series.

**Interior and trailing gaps.** By default only interior gaps are detected. Set `trailing_gap=True` to also flag the last present window when it ends more than one window before the current time — so missing recent data (for example no rows reported today) is caught even at the tail of the series, where there is no later row to anchor it to. The anchor comes from `curr_timestamp` (defaulting to `current_timestamp()`), bucketed onto the same absolute-time (UTC-aligned) grid as the event windows; pass an explicit `curr_timestamp` if you need local-day anchoring. Null values are ignored and pass with no violation.

### Linked issues

Resolves #1366

### Tests

- [ ] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [x] added performance tests

- Unit tests in `tests/unit/test_dataset_checks.py` cover the `window_minutes` validation, the `group_by` type check, and the `curr_timestamp`-without-`trailing_gap` guard.
- Integration tests in `tests/integration/test_dataset_checks.py` cover: an interior gap (missing day between two present days), consecutive days without a gap, same-window duplicates, null values, multiple gaps, a sub-day window, a non-UTC/DST-crossing timezone, empty and single-row inputs, `group_by` (per-device gaps), and the trailing-gap variants (flagged, within-one-window not flagged, disabled by default, and grouped).
- `has_no_gaps_per_time_window` is also added to `tests/resources/all_dataset_checks.yaml` and exercised by the "all checks" integration test.
- A performance benchmark (`test_benchmark_has_no_gaps_per_time_window`) is added under `tests/perf/`, mirroring `test_benchmark_is_data_fresh_per_time_window`.

### Documentation and Demos

- [ ] added/updated demos
- [x] added/updated docs

`docs/dqx/docs/reference/quality_checks.mdx` gets a reference-table entry and YAML plus programmatic (`DQDatasetRule`) usage examples.


ghanse:
Thanks for your contribution. IMO we need to determine what "no gaps" means. I see a few options:

## Option A - All Fixed Intervals Contain Data
Every `gap_interval` in some fixed grid has data. This makes the boundaries relative to the absolute clock time.

For this case we can use a grid approach: generate every interval between `min(ts)` and `max(ts)`, then `LEFT ANTI JOIN` the occupied buckets. Any grid interval with no matching event is a gap.

```python
from pyspark.sql import DataFrame


def detect_gaps_grid(df: DataFrame, period_secs: int) -> DataFrame:
    """Build the full interval grid between min/max ts and anti-join occupied buckets."""
    bounds = df.select(
        F.min("event_ts").cast("long").alias("min_ts"),
        F.max("event_ts").cast("long").alias("max_ts"),
    ).first()
    min_ts, max_ts = bounds["min_ts"], bounds["max_ts"]

    grid_start = (min_ts // period_secs) * period_secs
    grid_end = (max_ts // period_secs) * period_secs
    n_points = (grid_end - grid_start) // period_secs + 1

    grid = (
        spark.range(0, n_points)
        .withColumn("bucket_ts", F.lit(grid_start) + F.col("id") * F.lit(period_secs))
        .select("bucket_ts")
    )

    occupied = (
        df.withColumn(
            "bucket_ts",
            (F.col("event_ts").cast("long") / F.lit(period_secs)).cast("long") * F.lit(period_secs),
        )
        .select("bucket_ts")
        .distinct()
    )

    # grid intervals with no matching event == gaps
    return grid.join(occupied, on="bucket_ts", how="left_anti")
```

## Option B - No Time Gaps Between Consecutive Points
Every consecutive timestamp is within `gap_limit` of the next timestamp. This makes the boundaries relative to the event time.

The naive way is a window sort with `lag()`, but `Window.orderBy` with no `partitionBy` forces a **global single-partition sort** that does not scale.

For this case we can reduce the number of points before sorting. Because a gap larger than `gap_limit` can never lie *inside* a single bucket of width `gap_limit`, we can first collapse events to occupied buckets with a **parallel aggregation** (keeping `min_ts`/`max_ts` per bucket), then `lag()` over the much smaller set of occupied buckets. This yields the exact same gaps as the naive window approach, but the sort is over occupied buckets rather than raw events, so it scales with the grid rather than the data volume.

```python
from pyspark.sql import DataFrame
from pyspark.sql.window import Window


def detect_gaps_bucket_reduce(df: DataFrame, period_secs: int) -> DataFrame:
    """Exact inter-event gap detection: parallel bucket aggregation, then lag over buckets."""
    buckets = (
        df.withColumn(
            "bucket",
            (F.col("event_ts").cast("long") / F.lit(period_secs)).cast("long"),
        )
        .groupBy("bucket")
        .agg(
            F.min(F.col("event_ts").cast("long")).alias("min_ts"),
            F.max(F.col("event_ts").cast("long")).alias("max_ts"),
        )
    )

    w = Window.orderBy("bucket")
    return (
        buckets.withColumn("prev_max_ts", F.lag("max_ts").over(w))
        .withColumn("gap_secs", F.col("min_ts") - F.col("prev_max_ts"))
        .where(F.col("prev_max_ts").isNotNull() & (F.col("gap_secs") > F.lit(period_secs)))
        .select(
            F.col("prev_max_ts").cast("timestamp").alias("prev_ts"),
            F.col("min_ts").cast("timestamp").alias("event_ts"),
            "gap_secs",
        )
    )
```

For real IoT/clickstream data, gap detection is usually done **per entity** (e.g. `device_id`/`session_id`); adding a `partitionBy` on that key lets Option B partition the window naturally and scale even further.

### Performance note
I benchmarked these approaches at scales from 1K to 100M rows with different grid periods. These were my findings:
- The naive window approach degrades at scale due to the full sort
- The grid and bucket-reduce approaches both scale well

ghanse (changes requested):
Left an overall comment. I think we need to do 2 things:

1. Define the meaning of "no time gaps"
2. Refactor the implementation to perform better on large datasets

SreeramaYeshwanthGowd:
Thanks @ghanse, I've pinned the semantics, checked the scaling properly, and added the per entity path.

Semantics. Committing to Option A (fixed grid, absolute clock): a gap is a fixed grid interval between two present intervals with no data, measured against the absolute time grid rather than the spacing between consecutive events. It keeps the current meaning, matches the sibling is_data_fresh_per_time_window which also buckets on a fixed epoch anchored F.window grid, and matches the documented intent of window_minutes as the data grain.

I dug into the ordered window. The key point is that the distinct() already collapses rows to the distinct window count before the sort, so the sort scales with the number of occupied windows, not the row count. For realistic data (100M rows over months or years) that is a few thousand to about 1.7M windows, which sorts fast. I benchmarked lead() against the grid anti join approach on serverless, and lead() is faster or on par for realistic distinct window counts, including at 100M rows across every grid period.

The anti join only pulls ahead once the distinct window count is past roughly 10M, so I kept the simpler lead() approach and its message that names the next present window.

Added an optional group_by. Gaps are detected independently within each entity and the ordered window partitions by the group key.

Tests added for sub day windows, multiple and multi window gaps, trailing gap, no gap, single row, empty, and per entity.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -2575,6 +2520,115 @@ def apply(df: DataFrame) -> DataFrame:
...
+    fixed time windows of *window_minutes* (a fixed grid aligned to absolute time), and a gap is
+    flagged whenever the next present window starts more than one window after the current one. Gaps
+    are therefore measured against this fixed absolute-time grid, not the elapsed time between
+    consecutive events.
+
+    When *group_by* is provided, gaps are detected independently within each group (for example per
+    device or session) and the work partitions by the group key, which is the common case for IoT or
+    clickstream data. When it is omitted, the whole column is treated as a single series.
+
+    Only interior gaps are detected. A trailing gap (no data for the most recent windows up to the
```
mwojtyczka: The check only detects interior gaps which is a good default. Can you extend this function with an option to enable training gap (e.g. `trailing_gap: bool | None = False`? we can anchor to the current timestamp

The reason is that if data is missing, checking against the current timestamp becomes important when using this validation to ensure that data is present for every day, not only at the boundary.

tests/integration/test_apply_checks.py
```diff
@@ -67,6 +67,32 @@ def test_apply_checks_and_split_on_empty_checks(ws, spark):
...
+            criticality="error",
+            check_func=check_funcs.has_no_gaps_per_time_window,
+            column="event_ts",
+            check_func_kwargs={"window_minutes": 1440},
+        ),
+    ]
+
+    good, bad = dq_engine.apply_checks_and_split(test_df, checks)
+
+    assert sorted(row["val"] for row in good.select("val").collect()) == [2, 3]
```
mwojtyczka: this is a smoke type of test so it's probably ok but it is generally better to validate against a complete DataFrame.

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: **Optional trailing-gap detection against the current time.** Right now only interior gaps are found: for the last present window `lead()` is null, so if a daily feed stops after 2025-07-20 and today is 2025-07-28, the week-long tail outage is silently missed.

The default (interior-only) is good, but please add an opt-in that anchors the trailing edge to the current timestamp — mirroring the `curr_timestamp: Column | None` parameter already used by the sibling `is_data_fresh_per_time_window`. Sketch:

```python
def has_no_gaps_per_time_window(
    column: str | Column,
    window_minutes: int,
    group_by: list[str | Column] | None = None,
    check_current_window: bool = False,          # default off -> current behavior
    curr_timestamp: Column | None = None,
) -> tuple[Column, Callable]:
    ...
    if curr_timestamp is None:
        curr_timestamp = F.current_timestamp()
```

Inside `apply`, when `check_current_window` is set, treat the current-timestamp window as a virtual trailing boundary: for the last present window per group, compute the elapsed windows to `curr_timestamp` and flag a gap when it exceeds one window (`F.coalesce(next_window_start_col, <curr window start>)`). That keeps the fixed-grid semantics and reuses the sibling's established pattern rather than adding a bespoke flag.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -2575,6 +2520,115 @@ def apply(df: DataFrame) -> DataFrame:
...
+        A tuple of:
+            - A Spark Column representing the gap condition.
+            - A closure that applies the gap detection and adds the necessary condition columns.
+
+    Raises:
+        InvalidParameterError: if *window_minutes* is not a positive integer, or if *group_by* is not
+            a list.
+    """
+    if window_minutes is None or window_minutes <= 0:
+        raise InvalidParameterError("window_minutes must be a positive integer")
```
mwojtyczka: The docstring and message promise a *positive integer*, but this only rejects `None`/`<= 0`. A float (`1440.5`) or bool passes here and then fails opaquely inside `F.window(safe_col_expr, "INTERVAL 1440.5 MINUTES")` at apply time, and `gap_seconds = window_minutes * 60` becomes a float. Add a type guard so it fails fast with the documented error, e.g.:

```python
if not isinstance(window_minutes, int) or isinstance(window_minutes, bool) or window_minutes <= 0:
    raise InvalidParameterError("window_minutes must be a positive integer")
```

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -2575,6 +2520,115 @@ def apply(df: DataFrame) -> DataFrame:
...
+        df = df.withColumn(window_start_col, F.col(interval_col).start)
+
+        # For each present window (within each group) find the next present window over the ordered
+        # windows. Partitioning the window by the group key keeps per-group gap detection independent
+        # and lets the work scale across partitions for per-entity data.
+        distinct_windows = df.filter(col_expr.isNotNull()).select(*group_cols, window_start_col).distinct()
+        ordered_windows = (
+            Window.partitionBy(*group_cols).orderBy(window_start_col)
+            if group_cols
+            else Window.orderBy(window_start_col)
```
mwojtyczka, line 2996: When `group_by` is omitted, `Window.orderBy(window_start_col)` has no `partitionBy`, so this is a global single-partition sort (the concern @ghanse raised). Your benchmark shows `distinct()` keeps the occupied-window count small enough for realistic grids that this is fine — worth a one-line comment recording that assumption here so a future maintainer doesn't reintroduce a fine-grain misuse that collapses to one partition at tens of millions of distinct windows.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -2575,6 +2520,115 @@ def apply(df: DataFrame) -> DataFrame:
...
+        # A gap exists when the next present window starts more than one window after the current one.
+        gap_seconds = window_minutes * 60
+        gaps = gaps.withColumn(
+            condition_col,
+            F.when(
+                F.col(next_window_start_col).isNotNull()
+                & ((F.unix_timestamp(next_window_start_col) - F.unix_timestamp(window_start_col)) > F.lit(gap_seconds)),
+                True,
+            ).otherwise(False),
+        )
```
mwojtyczka, line 3017: `F.when(<bool expr>, True).otherwise(False)` wraps an already-boolean expression. Assign it directly:

```python
gaps = gaps.withColumn(
    condition_col,
    F.col(next_window_start_col).isNotNull()
    & ((F.unix_timestamp(next_window_start_col) - F.unix_timestamp(window_start_col)) > F.lit(gap_seconds)),
)
```

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -2575,6 +2520,115 @@ def apply(df: DataFrame) -> DataFrame:
...
 
+@register_rule("dataset")
+def has_no_gaps_per_time_window(
+    column: str | Column, window_minutes: int, group_by: list[str | Column] | None = None
+) -> tuple[Column, Callable]:
+    """Checks whether a time-series column has gaps, i.e. time windows that contain no rows at all
+    between windows that do (for example no data for 2025-07-15 while 2025-07-14 and 2025-07-16 are
+    present).
+
+    A missing window has no row to attach a violation to, so the gap is reported on the boundary row
```
mwojtyczka: "the boundary row that precedes it (the last row before the gap)" reads as singular, but the implementation flags **every** row in the last present window — your own `test_has_no_gaps_per_time_window` has two rows in the 2025-07-14 window and both are flagged. Suggest rewording to "every row in the last present window before the gap" to avoid surprise about violation counts.

tests/integration/test_dataset_checks.py
```diff
@@ -3216,6 +3217,200 @@ def test_is_data_fresh_per_time_window_check_entire_dataset(spark: SparkSession,
...
 
 
+def _gap_violation_message(window_start: str, next_window_start: str) -> str:
+    return (
+        f"Gap in time series: no data between the window starting at {window_start} "
+        f"and the next present window starting at {next_window_start}"
+    )
+
+
+def test_has_no_gaps_per_time_window(spark: SparkSession, set_utc_timezone):
```
mwojtyczka, line 3298: Every gap test runs under `set_utc_timezone`, so non-UTC behavior is unverified. `F.window` aligns buckets to the UTC epoch, so a 1440-minute (daily) window start does not fall on local midnight in a non-UTC session, and the reported window-start strings shift accordingly. Please add at least one test under a non-UTC session timezone (ideally crossing a DST boundary) to pin the intended bucketing and message rendering.

mwojtyczka (commented):
## Code review — `has_no_gaps_per_time_window`

Solid PR — clear semantics writeup, good test matrix (sub-day windows, multiple gaps, single row, empty, per-entity), and the response to @ghanse's scaling question is well-reasoned (distinct() collapses to occupied-window count before the sort, and the benchmark justifies keeping `lead()` over the anti-join for realistic grids). Committing to Option A (fixed absolute-time grid) to match the sibling `is_data_fresh_per_time_window` is the right call.

One feature gap worth addressing before merge, plus a few smaller items.

### Feature gap — optional trailing-gap detection against the current time
1. **Only interior gaps are detected; a trailing outage up to "now" is never reported.** When the last present window is followed by nothing, `lead()` returns null for it, so a feed that simply stops producing data raises zero violations. The default (interior-only) is fine, but there should be an **opt-in** to anchor the trailing edge against the current timestamp — the sibling `is_data_fresh_per_time_window` already establishes exactly this seam with its `curr_timestamp: Column | None` parameter. Suggested shape (default off, preserves current behavior): (inline)

### Correctness / robustness
2. **`window_minutes` type isn't validated** — only `<= 0` is checked, so a float or bool passes and fails opaquely inside `F.window("INTERVAL 1440.5 MINUTES")` instead of raising the documented `InvalidParameterError`. (inline)

### Efficiency
3. **Unpartitioned global sort when `group_by` is omitted** — `Window.orderBy(...)` with no `partitionBy` is a single-partition sort. Your benchmark justifies it for realistic occupied-window counts; a code comment recording that assumption would stop a future maintainer reintroducing fine-grain misuse. (inline)

### Tests / docs
4. **All gap tests force UTC** (`set_utc_timezone`) — `F.window` aligns buckets to the UTC epoch, so daily windows don't land on local midnight in a non-UTC session and the message strings shift. Add a non-UTC (ideally DST-crossing) test to pin the intended rendering. (inline)
5. **`F.when(<bool>, True).otherwise(False)`** — the inner expression is already boolean; assign it directly. (inline)
6. **Docstring says "the boundary row" (singular)** but every row in the boundary window is flagged (your own test shows 2 same-window rows both flagged). Reword to "every row in the last present window before the gap". (inline)

mwojtyczka (changes requested):
Generally looking good - request to add optional trailing-gap detection against the current time, and fix some other minor comments

SreeramaYeshwanthGowd:
@mwojtyczka @ghanse Changes are done, please take a look.

• Trailing gap option
• Window type guard
• Scaling note added
• Boolean check simplified
• Docstring wording fixed
• Non UTC DST test
• Full dataframe test

mwojtyczka (approved):
LGTM

docs/dqx/docs/reference/quality_checks.mdx
```diff
@@ -1852,6 +1852,7 @@ You can also define your own custom dataset-level checks (see [Creating custom c
...
...                                                                                                                                                                                                                                                                                                                                | `column`: timestamp or date column (can be a string column name or a column expression); `window_minutes`: size of the time window in minutes that defines the expected data grain (for example 1440 for daily); `group_by`: optional list of columns or column expressions to detect gaps independently within each group                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   |
```
ghanse: Looks like we're missing argument documentation for `trailing_gap` and `curr_timestamp` here.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -2575,6 +2575,154 @@ def apply(df: DataFrame) -> DataFrame:
...
+        # and lets the work scale across partitions for per-entity data. When group_by is omitted,
+        # there is no partitionBy and the sort below is a single partition, but distinct() has already
+        # collapsed rows to the occupied-window count, so the sort scales with that count, not with
+        # row count; this holds for realistic grids (thousands to low millions of occupied windows).
+        distinct_windows = df.filter(col_expr.isNotNull()).select(*group_cols, window_start_col).distinct()
+        ordered_windows = (
+            Window.partitionBy(*group_cols).orderBy(window_start_col)
+            if group_cols
+            else Window.orderBy(window_start_col)
+        )
```
ghanse: This will fail if `group_cols` is a `Column` expression like `F.upper("region")`.

We should use `get_columns_as_strings(group_by, allow_simple_expressions_only=True)` to get the column names, then use the column names list through the `apply` method here. This should resolve any issues with unresolved identifiers.

We should also add an integration test where `group_by` is a Spark Column expression.

ghanse (changes requested):
Left a few comments. I think `group_by` is the biggest gap right now.

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: Fixed in 545a5b7e. group_by is now resolved once via get_columns_as_strings(group_by, allow_simple_expressions_only=True) and the resulting column-name list is used consistently for partitionBy, select, and the join, so a Column expression group key like F.col("region") no longer produces a rendered string that breaks the join. Added integration test test_has_no_gaps_per_time_window_group_by_column_expression covering group_by=[F.col("device")].

Note: allow_simple_expressions_only=True accepts simple references but rejects complex expressions like F.upper("region") with InvalidParameterError, matching how the is_aggr_* dataset checks treat group_by.

docs/dqx/docs/reference/quality_checks.mdx
mwojtyczka: Fixed in 545a5b7e. The reference-table entry now documents trailing_gap and curr_timestamp, phrased to match the sibling is_data_fresh_per_time_window entry.

ghanse (approved):
LGTM

Issue #1343, opened 2026-07-18
[FEATURE]: Profile classification support


IvannKurchenko:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Problem statement

Current profiling mechanism generates profiles and hence check numerical types that might contradict each other. For instance: suppose there is a table with categorical column `cat` of values 1-5. DQX would generate both checks including `is_in` which treats `cat` column as categorical and `min_max`, which treats `cat` column more as a measurement. Having both checks makes a little sense and having just `is_in` and other categorical checks would be preferred.

### Proposed Solution

Introduce `ProfileType` enumeration that reflects column profile type / category, such as:
```python
from pydantic import BaseModel
from pyspark.sql.types import DataType, NumericType

class ProfileType(BaseModel):
    name: str
    key: str
    priority: int
    spark_type: type[DataType]


DEFAULT_PROFILE_TYPES = [
    ProfileType(
        name="IDENTIFIER",
        key="id",
        priority=1,
        spark_type=NumericType,
    ),
    ProfileType(
        name="CATEGORY",
        key="category",
        priority=2,
        spark_type=NumericType,
    ),
    ProfileType(
        name="MEASUREMENT",
        key="measurement",
        priority=3,
        spark_type=NumericType,
    ),
]
```
`priority` value is used when profiles of different types are generated so to choose only one of them. This way the library can generate coherent checks. Introduce new optional `type` property to `DQProfile` of type `ProfileType`. Profiles with `None` types will are always preserved for backward compatibility - meaning never evicted, but log warning that type is supposed to be specified.

ghanse:
It's an interesting question whether this should be the responsibility of DQX or the calling code. I see some benefits for us to do it:

1. We can offer a better-curated set of profiled rules.
2. We can improve the profiler performance by skipping computations for rules that will be ignored or superseded.

There are a few challenges we should consider:
1. We need this to be easy to implement as new profiling rules are introduced. We can modify the profiler registry concept to require this profile type.
2. Depending on the column type, multiple profile types may need to be treated with equivalent priority.
3. If we allow this, users may ask us to allow custom priority order in the future.

@IvannKurchenko would you like to work on this? If not I can take it.

IvannKurchenko:
@ghanse I'd like to work on this, please.

Issue #1067, opened 2026-03-12
[FEATURE]: Extend profiler with additional metrics


mwojtyczka:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Problem statement

The currently list of metrics is limited to basics such as min, max, stdev etc.

### Proposed Solution

Include additional statistics beyond minimum and maximum values, such as the 10th percentile, along with the percentage of occurrences within each bucket or range.

mwojtyczka:
@passionworkeer we recently refactored the profiler and it will also be much easier to extend it now. If you still plan to complete this please let us know. For now we removed your assignment to free up the task

IvannKurchenko:
@mwojtyczka May I work on this?

mwojtyczka:
@IvannKurchenko feel free to work on this

IvannKurchenko:
@mwojtyczka Thank you for review and merging the PR. Shall this ticket be closed or anything else remains in to do?

PR #1384, opened 2026-07-29
Refactor profiler column metrics into an extensible registry

IvannKurchenko:
## Changes

Sets up an extension point for the profiler by moving inline column-metric aggregation in `DQProfiler._profile` into a registry-based system. Users can now register their own column metric functions via the `@register_profile_column_metric()` decorator, mirroring the existing `register_rule` / `register_profile_builder` extension patterns.

The three existing metrics (`count_non_null`, `count_distinct`, `empty_count`) are moved into this registry with no behavioural change. `_build_column_metrics` builds the aggregation from whatever is registered.

**Rationale for landing the refactoring without new metrics**

The profiling pipeline is: `column metrics → profile builder → check`. Adding a new metric only adds value once a builder consumes it and generates a check. Two paths were considered for #1067:

1. **Add percentile metrics (p10/p90) with a matching profile builder** that emits `is_aggr_not_less_than` / `is_aggr_not_greater_than` checks. But these checks are mostly useful for measurement data (revenue, sales amount, latency, temperature) and not meaningful for keys or categorical columns. Applying them indiscriminately would generate false positives. Selective, purpose-aware application is tracked in #1343.
2. **Land the refactoring only, expose the extension point, and let users register the metrics they need.** This keeps the profiler flexible without shipping metrics that don't yet have a purpose-fit builder.

This PR takes option 2. New built-in metrics can be added later, together with the specific builder that consumes them, once the classification work in #1343 makes selective application safe.

**What changed**

- New `PROFILE_COLUMN_METRIC_REGISTRY` and `register_profile_column_metric` decorator in `profiler/profiler_column_metrics.py`
- Existing metrics (`count_non_null`, `count_distinct`, `empty_count`) moved into the registry
- `DQProfiler._profile` refactored: inline aggregation extracted into `_build_column_metrics`, which iterates the registry
- `is_text` helper moved from `profile_builder.py` to `profiler/common.py` (now used across modules)

### Linked issues

Relates to #1067, related to #1343

### Tests

- [x] added unit tests
- [ ] manually tested
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

Unit tests cover: registry (register, overwrite, function-name key); built-in metric functions across column types; `is_text` helper; `_build_column_metrics` (alias correctness, `count_null` derivation, summary merge, empty DataFrame, `None`-returning metrics). Existing integration tests already cover the refactored aggregation path end-to-end.

### Documentation and Demos

- [x] added/updated docs
- [ ] added/updated demos

New sections in `data_profiling.mdx` guide and `profiler.mdx` reference showing how to register custom metrics; `dqx-profile-and-generate/SKILL.md` updated with the extension point.


mwojtyczka (commented):
Reviewed the profiler column-metrics registry refactor. The extension-point design is reasonable, but the wiring has a **blocking crash** plus a few correctness regressions in `_build_column_metrics` — details inline.

src/databricks/labs/dqx/profiler/profiler.py
```diff
@@ -442,34 +442,38 @@ def _profile(
...
-                metrics["empty_count"] = column_df.filter(F.col(column_label) == "").count()
-            else:
-                metrics["empty_count"] = 0
+            metrics = self._build_column_metrics(column_df, column_label, field, summary_stats, total_count)
 
             self._build_profiles_for_column(column_df, field_name, field_type, metrics, opts, dq_rules)
 
         self._add_llm_primary_key_for_dataframe(df, dq_rules, summary_stats, opts)
 
+    def _build_column_metrics(
```
mwojtyczka, line 462: **Blocking:** `_build_column_metrics` is missing the `self` parameter, but it's called as `self._build_column_metrics(column_df, column_label, field, summary_stats, total_count)`. `self` binds to `column_df`, every argument shifts by one, and Python raises `TypeError: _build_column_metrics() takes 5 positional arguments but 6 were given` on the **first column of every profiling run**. Add `self` as the first parameter:
```python
def _build_column_metrics(self, column_df, column_label, field, summary_stats, total_count):
```

src/databricks/labs/dqx/profiler/profiler.py
```diff
@@ -442,34 +442,38 @@ def _profile(
...
+        field_metric_aggregations = []
+        for metric_name, metric_function in PROFILE_COLUMN_METRIC_REGISTRY.items():
+            metric_col = metric_function(field, column_label)
+            if metric_col is not None:
+                field_metric_aggregations.append(metric_col.alias(metric_name))
+
+        field_aggregation_row = column_df.agg(*field_metric_aggregations).first()
+        field_aggregation_stats = field_aggregation_row.asDict() if field_aggregation_row else {}
+        field_summary_stats = summary_stats.get(field.name, {})
+        base_metrics = {**field_summary_stats, **field_aggregation_stats}
```
mwojtyczka: **Correctness — `count` metric is lost.** The old `_profile` set `metrics["count"] = total_count` for every column. Here `count` only survives if `summary_stats[field.name]` carries it — which is true only for columns `DataFrame.summary()` covers (numeric), and there it is the **non-null** count, not `total_count`. For text/date/boolean/timestamp columns there is no `count` key, so downstream `profiler_metrics.get("count", 0)` returns `0` in `make_null_or_empty_profile` / `make_min_max_profile`, breaking the null-ratio denominator and total-count logic. Suggest explicitly seeding `metrics["count"] = total_count` here.

**Correctness — `summary_stats` is no longer mutated.** Previously `metrics` was an alias of `summary_stats[field_name]`, so count/count_null/count_non_null/count_distinct/empty_count (and the written-back min/max) landed in `summary_stats`, which is the first value returned by the public `profile()` API. This function returns a fresh dict that's never assigned back into `summary_stats`, so those per-column metrics silently disappear from the return value (and `_add_llm_primary_key_for_dataframe`, which reads `summary_stats` afterwards, no longer sees the written-back min/max).

src/databricks/labs/dqx/profiler/profiler.py
```diff
@@ -442,34 +442,38 @@ def _profile(
...
+        metrics = {**base_metrics, "count_null": total_count - base_metrics["count_non_null"]}
```
mwojtyczka: **Correctness — `KeyError` risk.** `total_count - base_metrics["count_non_null"]` assumes `count_non_null` is always present. But the registry is now user-extensible: if someone overwrites/removes the `count_non_null` metric via `register_profile_column_metric`, or if every registered metric returns `None` for a column (empty aggregation), `base_metrics` has no `count_non_null` key and this raises `KeyError`, crashing profiling. Use `base_metrics.get("count_non_null", 0)` or validate the required metric is registered.

src/databricks/labs/dqx/profiler/profiler.py
```diff
@@ -419,7 +419,7 @@ def _profile(
         df_cols: list[T.StructField],
         dq_rules: list[DQProfile],
         opts: dict[str, Any],
-        summary_stats: dict[str, Any],
+        summary_stats: dict[str, int | float],
```
mwojtyczka: **Type hint is factually wrong.** `summary_stats` is a mapping of column-name → a per-column **dict** of metrics (values include `datetime`, `None`, `str`, `float`). Narrowing it to `dict[str, int | float]` misdescribes the structure (downstream `summary_stats.get(field.name, {})` expects a dict), and mypy under `make lint` should reject it. Revert to `dict[str, Any]` (or a precise `dict[str, dict[str, Any]]`).

src/databricks/labs/dqx/profiler/profiler_column_metrics.py
```diff
@@ -0,0 +1,60 @@
...
+    return wrapper
+
+
+@register_profile_column_metric()
+def empty_count(field: T.StructField, column_label: str) -> Column | None:
+    """
+    Profiling column metric for empty count. Applicable for text columns only, otherwise returns literal `0` for
+    backward compatibility.
+    """
+    return F.count_if(F.col(column_label) == "") if is_text(field.dataType) else F.lit(0)
```
mwojtyczka, line 51: **Minor (efficiency/trade-off):** for non-text columns `empty_count` returns `F.lit(0)`, so a constant column is added to every non-text column's aggregation instead of being skipped. `_build_column_metrics` already omits `None`-returning metrics, so returning `None` would avoid the no-op projection — but then the `empty_count` key would be absent for non-text columns, which downstream code may rely on. If the key must always exist, the `F.lit(0)` is fine; just flagging the trade-off.

tests/integration/test_profiler.py
```diff
@@ -161,6 +161,39 @@ def test_profiler_timestamp_ntz_column(spark, ws):
     assert min_max_profiles[0].parameters["max"] is not None
 
 
+def test_profiler_column_metrics_flow_into_generated_profiles(spark, ws):
```
mwojtyczka, line 193: This test (and `test_profiler_high_null_ratio_column_skips_is_not_null`) call the public `profile()` API, which routes through `_profile → self._build_column_metrics` and would hit the missing-`self` `TypeError` before reaching any assertion. Since the PR opened with these tests present, it looks like the integration suite wasn't run locally (the checklist has "added unit tests" but not "manually tested"/"integration tests", and the unit tests only exercise the metric functions in isolation, never `_build_column_metrics`). Worth running `make integration -k profiler` before the next push — it would have caught the crash.

mwojtyczka (changes requested):

src/databricks/labs/dqx/profiler/profiler_column_metrics.py
IvannKurchenko: Thanks for spotting. `F.lit(0)` was left for backward compatibility. Please, see previous implementation: https://github.com/databrickslabs/dqx/pull/1384/changes#diff-78698e2495ede28f8e5fd2dd9003ea0eaa8ae060337f319a0dab1a805539e940L464-L467. However happy to return `None` as well to prioritise efficiency.

tests/integration/test_profiler.py
IvannKurchenko: 😬 Oh, sorry about this. Re-run integration tests locally - passed everything except LLM's due to environment restrictions: 

Image

src/databricks/labs/dqx/profiler/profiler.py
IvannKurchenko: My bad. Fixed.

src/databricks/labs/dqx/profiler/profiler.py
IvannKurchenko: Thank you, resolved.

src/databricks/labs/dqx/profiler/profiler.py
IvannKurchenko: Thank you, resolved.

src/databricks/labs/dqx/profiler/profiler.py
IvannKurchenko: Thank you, resolved.

src/databricks/labs/dqx/profiler/profiler_column_metrics.py
```diff
@@ -0,0 +1,60 @@
...
+    Expected signature of the function is as follows:
+    (field,column_label) -> Column | None
+    where:
+        - field: struct field of the profiling column
+        - column_label: name of the column that is present in the dataframe to be aggregated
+    The function may return `None` if aggregation is not applicable.
+    """
+
+    def wrapper(metric_func: Callable) -> Callable:
+        PROFILE_COLUMN_METRIC_REGISTRY[metric_func.__name__] = metric_func
```
mwojtyczka: The registry is keyed by `metric_func.__name__`, but the sibling extension points this mirrors — `register_rule(rule_type)` and `register_profile_builder(profile_type)` — take an **explicit name string**. Two custom metric functions with the same `__name__` (e.g. both named `count_non_null` in different modules, or wrapped/`functools.partial` funcs) will silently overwrite each other in `PROFILE_COLUMN_METRIC_REGISTRY` with no error.

Also, since no argument is captured, the parameterless decorator-factory layer (the required `()`) adds nothing over a plain `def register_profile_column_metric(metric_func):`. Consider either taking an explicit `name: str` (consistent with the siblings, and lets callers disambiguate) or dropping the factory layer.

src/databricks/labs/dqx/profiler/common.py
```diff
@@ -3,6 +3,25 @@
 from decimal import Decimal
 from typing import Any
 
+from pyspark.sql import types as T
+
+# Type alias for annotations; use TEXT_TYPES for isinstance() checks.
+TextType = T.CharType | T.StringType | T.VarcharType
+TEXT_TYPES: tuple[type[TextType], ...] = (T.CharType, T.StringType, T.VarcharType)
```
mwojtyczka, line 10: `is_text()` was moved here, but the identical `TextType` / `TEXT_TYPES` definitions still live in `profile_builder.py` (lines 35-36), which `profiler.py` continues to import from. The same configuration is now defined in two modules and can drift. Since `profile_builder.py` already imports `is_text` from this module, it could import `TEXT_TYPES` / `TextType` from here too and drop its own copies (DRY — per the AGENTS.md guideline).

src/databricks/labs/dqx/profiler/profiler_column_metrics.py
```diff
@@ -0,0 +1,60 @@
...
+from collections.abc import Callable
+
+from pyspark.sql import Column
+from pyspark.sql import functions as F
+from pyspark.sql import types as T
+
+from databricks.labs.dqx.profiler.common import is_text
+
+
+DQProfileColumnMetricFunc = Callable[[T.StructField, str], Column | None]
```
mwojtyczka, line 11: `DQProfileColumnMetricFunc = Callable[[T.StructField, str], Column | None]` is defined but not used to type the decorator: `register_profile_column_metric` returns a bare `Callable` and `wrapper` takes/returns a bare `Callable`. Using the alias in these signatures lets mypy/basedpyright verify that registered functions match the metric signature. (Note `register_profile_builder` has the same looseness, so this is optional consistency.)

src/databricks/labs/dqx/profiler/profiler_column_metrics.py
```diff
@@ -0,0 +1,60 @@
...
+        PROFILE_COLUMN_METRIC_REGISTRY[metric_func.__name__] = metric_func
+        return metric_func
+
+    return wrapper
+
+
+@register_profile_column_metric()
+def empty_count(field: T.StructField, column_label: str) -> Column | None:
+    """
+    Profiling column metric for empty count. Applicable for text columns only, otherwise returns literal `0` for
```
mwojtyczka: Docstring uses backticks around `0` (and `None` on line 28). AGENTS.md forbids backticks around object names in docstrings — they break API-doc rendering; use italics (*0*, *None*) instead.

mwojtyczka (commented):
Follow-up review of the head commit. The six earlier threads are all addressed — resolved them. A few new points on the extension point and conventions:

IvannKurchenko:
Hello, @mwojtyczka! Thanks for a review. The previous feedback has been addressed. Would it be possible to have another round?

tests/unit/test_profiler_column_metrics.py
```diff
@@ -0,0 +1,46 @@
...
+@pytest.mark.parametrize("column_type", [T.StringType(), T.CharType(10), T.VarcharType(50)])
+def test_empty_count_returns_column_for_text_types(column_type):
+    field = T.StructField("col", column_type)
+    assert empty_count(field, "col") is not None
+
+
+@pytest.mark.parametrize("column_type", [T.IntegerType(), T.DoubleType(), T.LongType(), T.DateType()])
+def test_empty_count_returns_literal_zero_for_non_text_types(column_type):
+    field = T.StructField("col", column_type)
+    assert empty_count(field, "col") is not None
```
mwojtyczka: **Test coverage — this assertion can't detect the regression its name implies.** Both `test_empty_count_returns_column_for_text_types` (line 28) and `test_empty_count_returns_literal_zero_for_non_text_types` (here) assert only `empty_count(...) is not None`. Since `empty_count` returns a `Column` in *both* branches (`F.count_if(...)` for text, `F.lit(0)` for non-text), `is not None` is trivially true either way — so if the non-text branch were changed from `F.lit(0)` to `F.count_if(...)`, this test would still pass. The "returns literal zero for non-text" backward-compat guarantee the PR calls out is never actually exercised. Suggest evaluating the metric against a small Spark DataFrame and asserting the non-text result equals `0` (or at minimum asserting the two branches differ).

src/databricks/labs/dqx/profiler/profiler.py
```diff
@@ -439,44 +439,56 @@ def _profile(
...
+        if is_text(field_type) and trim_strings:
+            column_df = column_df.select(F.trim(F.col(column_label)).alias(column_label))
+        return column_df, column_label
+
+    def _build_column_metrics(
+        self,
+        column_df: DataFrame,
+        column_label: str,
+        field: T.StructField,
+        summary_stats: dict[str, Any],
```
mwojtyczka: **Altitude — one responsibility split across two methods.** `_build_column_metrics` receives the whole `summary_stats` dict but only reads its own entry (`summary_stats.get(field.name, {})`), while the matching write `summary_stats[field.name] = metrics` lives in the caller `_profile` (line 445). A reader must hold both methods in mind to follow how `summary_stats[field.name]` evolves. Passing only the per-field entry in (e.g. `field_summary_stats=summary_stats.get(field.name, {})`) and letting `_profile` own the single assignment of the returned dict would make `_build_column_metrics` a pure function of its inputs.

mwojtyczka (commented):
A few findings from a follow-up review pass (test-coverage and design-depth). None are correctness blockers.

tests/integration/test_profiler.py
```diff
@@ -190,6 +190,39 @@ def test_profiler_timestamp_precision_and_rounding(
     assert min_max_profiles[0].parameters == expected_parameters
 
 
+def test_profiler_column_metrics_flow_into_generated_profiles(spark, ws):
```
mwojtyczka, line 223: Please add a test that verifies the custom metrics flow — register a new custom metric via register_profile_column_metric(...), run the profiler, and assert the custom metric's computed value appears in the returned summary_stats. I verified it live env and it seems to work fine but we need a test

mwojtyczka (changes requested):
Thanks for the fixes. The previous issues are resolved now. Left some additional small comments - mainly missing test coverage

src/databricks/labs/dqx/profiler/profiler.py
```diff
@@ -439,44 +439,56 @@ def _profile(
...
+
+        field_aggregation_stats: dict[str, Any] = {}
+        if field_metric_aggregations:
+            field_aggregation_row = column_df.agg(*field_metric_aggregations).first()
+            if field_aggregation_row:
+                field_aggregation_stats = field_aggregation_row.asDict()
+
+        metrics: dict[str, Any] = {**field_summary_stats, **field_aggregation_stats}
+        metrics["count"] = total_count
+        metrics["count_null"] = total_count - metrics.get("count_non_null", 0)
```
mwojtyczka: **count_null can raise TypeError from a user-registered metric.** `total_count - metrics.get("count_non_null", 0)` subtracts a value taken straight from the registry aggregation. `register_profile_column_metric` explicitly supports overwriting keys (it logs *"Overwriting profile column metric registered as ..."*), so a user metric registered under `count_non_null` whose aggregation evaluates to SQL NULL on some column (e.g. `F.max(...)` over an all-null/empty `column_df`) lands `None` in `metrics` via `asDict()`, and `int - None` raises, aborting profiling. Consider coalescing the value to 0 before the subtraction.

src/databricks/labs/dqx/profiler/profiler.py
```diff
@@ -439,44 +439,56 @@ def _profile(
...
+                field_metric_aggregations.append(metric_col.alias(metric_name))
+
+        field_aggregation_stats: dict[str, Any] = {}
+        if field_metric_aggregations:
+            field_aggregation_row = column_df.agg(*field_metric_aggregations).first()
+            if field_aggregation_row:
+                field_aggregation_stats = field_aggregation_row.asDict()
+
+        metrics: dict[str, Any] = {**field_summary_stats, **field_aggregation_stats}
+        metrics["count"] = total_count
```
mwojtyczka, line 499: **Altitude: `count`/`count_null` are hardcoded special cases coupled to the literal key `count_non_null`, sitting outside the very registry this PR introduces.** Overwriting the built-in `count_non_null` metric (a supported, warning-logged registry operation) silently makes `count_null = total_count - 0 = total_count`. Every downstream builder reads these with `.get(key, 0)` (e.g. `make_null_or_empty_profile`, the `make_min_max_profile` null-guard), so they then compute wrong null ratios and emit/suppress the wrong `is_not_null`/`min_max` rules with no error. The derived counts should either be first-class registry metrics or be protected from redefinition, rather than trusting a string key the extension point can overwrite.

src/databricks/labs/dqx/profiler/profiler.py
```diff
@@ -439,44 +439,56 @@ def _profile(
...
+        for metric_name, metric_function in PROFILE_COLUMN_METRIC_REGISTRY.items():
+            metric_col = metric_function(field, column_label)
+            if metric_col is not None:
+                field_metric_aggregations.append(metric_col.alias(metric_name))
+
+        field_aggregation_stats: dict[str, Any] = {}
+        if field_metric_aggregations:
+            field_aggregation_row = column_df.agg(*field_metric_aggregations).first()
+            if field_aggregation_row:
+                field_aggregation_stats = field_aggregation_row.asDict()
```
mwojtyczka: **Metric values reaching `asDict()` can be `None`, and consumers assume numbers.** Once a registered metric's aggregation yields SQL NULL, `field_aggregation_row.asDict()` puts `None` into `metrics`. Consumers do arithmetic on these keys with only a missing-key guard — e.g. `make_null_or_empty_profile` (`profile_builder.py:190`) computes `empty_count / total_count`, so a metric registered under `empty_count` that evaluates to NULL gives `None / total_count` -> TypeError. `.get(key, 0)` guards absent keys but not present-but-`None` values, which the new registry makes reachable. Consider dropping/coalescing NULL aggregation results here before they enter the metrics dict.

tests/integration/test_profiler.py
```diff
@@ -190,6 +195,64 @@ def test_profiler_timestamp_precision_and_rounding(
...
+    assert not [p for p in profiles if p.name == "is_not_null" and p.column == "sparse"]
+
+
+def test_profiler_uses_registered_custom_column_metric(spark, ws):
+    # Verifies the register_profile_column_metric extension point end-to-end:
+    # a user-registered metric is executed against each column during profiling and its
+    # aggregated value is exposed under the registered key in the returned summary_stats.
+    metric_key = "p50"
+
+    @register_profile_column_metric(metric_key)
```
mwojtyczka, line 262: **Test mutates the process-global `PROFILE_COLUMN_METRIC_REGISTRY`, and the profiler runs every registered metric against every column.** Registration here (and in the unit tests) is global with only a `try/finally` cleanup. Under `pytest-xdist`, or if a prior test errors outside its `finally`, a leaked registration is executed for all columns in unrelated profiling tests, altering their `summary_stats` and causing order-dependent failures. Consider a fixture that snapshots and restores the registry around each test rather than relying on manual pop().

tests/unit/test_profiler_column_metrics.py
```diff
@@ -0,0 +1,47 @@
...
+        assert "custom_metric_key" in PROFILE_COLUMN_METRIC_REGISTRY
+        assert PROFILE_COLUMN_METRIC_REGISTRY["custom_metric_key"] is _test_metric
+    finally:
+        PROFILE_COLUMN_METRIC_REGISTRY.pop("custom_metric_key", None)
+
+
+@pytest.mark.parametrize("column_type", [T.StringType(), T.CharType(10), T.VarcharType(50)])
+def test_empty_count_returns_count_if_expression_for_text_types(column_type):
+    field = T.StructField("col", column_type)
+    assert str(empty_count(field, "col")) != str(F.lit(0))
```
mwojtyczka: **Assertion on `str(Column)` tests an implementation detail.** `str(empty_count(field, "col")) == str(F.lit(0))` depends on the `Column.__repr__`/`__str__` format, which is not a stable API and varies across Spark / Spark Connect versions. This can pass vacuously or break spuriously on a Spark upgrade without any behavior change, and it never actually exercises the `count_if` semantics. Prefer asserting on the evaluated aggregation result against a small DataFrame (that would be an integration test) or on the returned type.

docs/dqx/docs/guide/data_profiling.mdx
```diff
@@ -715,3 +715,59 @@ Combining either limit with `sample_by_column` may filter out values and lead to
...
+
+<Tabs>
+  <TabItem value="Python" label="Python" default>
+    ```python
+    from pyspark.sql import Column
+    from pyspark.sql import functions as F
+    from pyspark.sql import types as T
+    from databricks.labs.dqx.profiler.profiler_column_metrics import register_profile_column_metric
+
+    @register_profile_column_metric("percentile_10")
```
mwojtyczka, line 737: **Doc example registers a metric as an import-time side effect, with no opt-in or teardown note.** The `@register_profile_column_metric("percentile_10")` decorator mutates the global `PROFILE_COLUMN_METRIC_REGISTRY` the moment this module/cell is imported, so simply importing or pasting the example permanently adds `percentile_10` to every subsequent `profile()` call in the process (extra per-column aggregation, altered `summary_stats`). Worth a sentence noting this global, persistent effect and how to unregister, so users who copy it to "try it out" aren't surprised.

IvannKurchenko: Make sense, thanks. I added de-registration method also for `deregister_profile_builder`.

src/databricks/labs/dqx/profiler/profiler.py
IvannKurchenko: Thank you. `count_non_null` moved to direct calculations instead of registration for safety

IvannKurchenko:
@mwojtyczka May I ask for another review round, please?

src/databricks/labs/dqx/profiler/profiler.py
mwojtyczka: Reopening — moving `count_non_null` to a direct inline aggregation doesn't actually close this. In `_build_column_metrics` the registry loop still does `metric_col.alias(metric_name)` for **every** registered metric, so a user metric registered under the key `count_non_null` is added to the *same* `.agg()` as a second column aliased `count_non_null`. `Row.asDict()` keeps the **last** value for duplicate field names, so the user's value — not the inline `F.count(...)` — is what lands in `metrics["count_non_null"]`.

- If that metric evaluates to SQL `NULL` it is dropped by the `None`-filter, and `metrics["count_null"] = total_count - metrics["count_non_null"]` raises `KeyError`.
- If it returns a non-count value, `count_null` is silently wrong.

The comment on lines 471–473 claims a protection the code doesn't provide. To actually protect it, compute `count_non_null` in a separate `.agg()`, or reserve/reject the `count_non_null` key at registration time so the loop can't shadow it.

src/databricks/labs/dqx/profiler/profiler.py
```diff
@@ -439,44 +439,64 @@ def _profile(
...
+        field_name = field.name
+        field_type = field.dataType
+
+        column_df = df.select(field_name).dropna()
+        column_label = column_df.columns[0]
+        if is_text(field_type) and trim_strings:
+            column_df = column_df.select(F.trim(F.col(column_label)).alias(column_label))
+        return column_df, column_label
+
+    def _build_column_metrics(
```
mwojtyczka, line 467: `_prepare_column_df` and `_build_column_metrics` take `self` but use no instance state (only their arguments and the module-level `PROFILE_COLUMN_METRIC_REGISTRY`). The sibling helpers in this class — `get_columns_or_fields`, `_build_options_from_list`, `_get_fields`, `_do_cast` — are all `@staticmethod`. For consistency, make these two `@staticmethod` as well.

docs/dqx/docs/reference/profiler.mdx
```diff
@@ -124,6 +124,40 @@ The `DQDltGenerator` class creates Delta Live Tables expectation statements from
...
+```
+
+### Built-in column metrics
+
+The following metrics are available to all profile builders:
+
+| Metric key | Applicable types | Description |
+|---|---|---|
+| `count_non_null` | All | Non-null value count |
+| `count_null` | All | Null value count |
```
mwojtyczka: `count_non_null` and `count_null` are listed here alongside `count_distinct` / `empty_count`, but unlike those two they are **not** in `PROFILE_COLUMN_METRIC_REGISTRY` — `count_non_null` is computed inline and `count_null` is derived. A reader seeing them in the same "Built-in column metrics" table next to the `deregister_profile_column_metric` docs may reasonably call `deregister_profile_column_metric("count_non_null")` expecting to disable it; that is a silent no-op and the metric is still computed on every run. Consider noting that these two are always-on / not deregisterable.

docs/dqx/docs/guide/data_profiling.mdx
```diff
@@ -715,3 +715,83 @@ Combining either limit with `sample_by_column` may filter out values and lead to
...
+            name="min_max",
+            column=column_name,
+            description=f"Lower bound set to 10th percentile ({p10})",
+            parameters={"min": p10},
+        )
+    ```
+  </TabItem>
+</Tabs>
+
+<Admonition type="warning" title="Registration has global side effects">
```
mwojtyczka, line 793: This "Registration has global side effects" admonition is near-identical to the one at line 743 (only the *metric* / *builder* nouns differ). Edits to the shared caveat then have to be made in two places and can drift — consider a single shared note or a cross-reference.

src/databricks/labs/dqx/profiler/profiler.py
mwojtyczka: This one is still not fixed. It's the same underlying issue as the one above

mwojtyczka (changes requested):
2 comments are still not fixed. Left a few additional small comments

tests/unit/test_profiler_column_metrics.py
mwojtyczka: Verified on HEAD: the weak `is not None` assertions are replaced by `test_profiler_empty_count_computed` (integration), which evaluates against a real DataFrame and asserts text→`2` / non-text→`0` — so the non-text backward-compat guarantee is now exercised end-to-end. Minor residual: the exact `F.lit(0)`→`F.count_if(...)` swap on the int column wouldn't be caught (`intcol == ''` casts to NULL → count `0` either way), but that's not really achievable given empty-string semantics. Resolving.

docs/dqx/docs/reference/profiler.mdx
```diff
@@ -124,6 +124,52 @@ The `DQDltGenerator` class creates Delta Live Tables expectation statements from
...
+**Registry-managed built-ins:**
+
+| Metric key | Applicable types | Description |
+|---|---|---|
+| `count_distinct` | All | Distinct non-null value count |
+| `empty_count` | Text only | Empty-string count; `0` for non-text |
+
+Both are registered via `@register_profile_column_metric(...)` at import time and can be deregistered like any custom metric.
+
+Spark's `DataFrame.summary()` additionally contributes `mean`, `stddev`, `min`, `25`, `50`, `75`, and `max` for numeric columns.
```
mwojtyczka: **[docs correctness] The `summary()` percentile keys are wrong.** This lists them as `25`, `50`, `75`, but the profiler stores them under Spark's raw `DataFrame.summary()` labels: `_summarize` does `metric = row_dict["summary"]` (profiler.py:603-609), and `summary()` labels percentiles as `25%`, `50%`, `75%`. So a builder author following this table with `profiler_metrics.get("50")` always gets `None` — the correct keys are `25%` / `50%` / `75%`. Fix the doc (or normalize the keys when parsing `summary()` in `_process_row`).

mwojtyczka (commented):
Re-review at head `863ecf27` — all previously-flagged issues verified resolved. One doc-correctness finding inline. (The other item from the pass — `empty_count` returning `F.lit(0)` for non-text columns — is the intentional backward-compat trade-off already discussed here, so no action needed.) Resolving the review threads that are now fixed.

docs/dqx/docs/reference/profiler.mdx
mwojtyczka: Fixed in 0b94233f — the reference table now documents the literal `25%` / `50%` / `75%` keys that `DataFrame.summary()` emits.

docs/dqx/docs/guide/data_profiling.mdx
mwojtyczka: The builder admonition already cross-references this one for the shared caveat ("See [Registration has global side effects] above…", L796), which is the cross-reference suggested here. The two blocks otherwise document different functions (metric vs builder) and each needs its own name + deregister example, so they can not be fully merged. Resolving.

mwojtyczka:
## Docs feedback — "Extending the profiler with custom column metrics" (`docs/dqx/docs/guide/data_profiling.mdx`)

A few wording/completeness issues in the new section:

### 1. Say "decorator" and drop the misleading "no extra cost"
> Use `register_profile_column_metric` to add your own per-column metrics.
> Metrics are computed once per column before any profile builder runs, so they are available to every builder **at no extra cost**.

`register_profile_column_metric` is a decorator (the example uses it as `@register_profile_column_metric(...)`), so it should read "the `register_profile_column_metric` **decorator**". More importantly, **"at no extra cost" is misleading** — it directly contradicts the warning admonition a few lines below, which says *"Leaving unused metrics registered adds work to every profiling run and can slow it down."* Each additional registered metric is an extra aggregation on every run. The "once per column, shared across builders" point is worth keeping (builders don't recompute it), but not framed as free. Suggested:

> Use the `register_profile_column_metric` decorator to add your own per-column metrics. Each metric is computed once per column and shared across every profile builder (so builders never recompute it) — but each additional registered metric adds an aggregation to **every** profiling run (see the warning below).

### 2. `profile_column_metric_type` is used as an undefined term
> The *profile_column_metric_type* passed to the decorator becomes the key under which the value is available inside profile builders.

`profile_column_metric_type` is the decorator's argument name in code, but a reader only sees a bare string (`"percentile_10"`) in the example — the term appears out of nowhere and is never tied to it. Suggest connecting the two:

> The string passed to the decorator's *profile_column_metric_type* argument (e.g. `"percentile_10"` in the example above) becomes the key under which the value is exposed to profile builders.

### 3. Document the function interfaces, not just examples
The section shows examples but never states the **contract** each function must satisfy, so a reader can't write one without reverse-engineering the sample. Please add the input/output signatures (both are code-verified below):

**Custom column metric** — `@register_profile_column_metric(profile_column_metric_type)`
- Receives: *field* (`StructField` — the column's schema field) and *column_label* (`str` — the column name in the DataFrame).
- Returns: a PySpark aggregation `Column`, or `None` to skip the metric for that column type.

**Custom profile builder** — `@register_profile_builder(profile_type)`
- Receives: *df* (`DataFrame`), *column_name* (`str`), *column_type* (`DataType`), *profiler_metrics* (`dict[str, Any]` — the column-level stats computed by the profiler, keyed by metric type, including any custom metrics), and *profiler_options* (`dict[str, Any]` — profiler configuration).
- Returns: a `DQProfile` (a candidate rule), or `None` to emit nothing for that column.

The reference page (`docs/dqx/docs/reference/profiler.mdx`) already describes the metric-function interface but not the builder's; the guide describes neither. Both interfaces belong in the guide, and points 1–2 above apply to the reference page's parallel wording too.

mwojtyczka:
One more wording nit in the same section:

> Combine this with `register_profile_builder` to generate rules based on the metric:

The profiler produces *candidate* rules for review, not final rules (the sampling admonition earlier says "Manually review and tune rules generated from profiles"). Suggest:

> Combine this with `register_profile_builder` to generate data quality rule suggestions based on the metric:

mwojtyczka:
…and as with the custom-metric example, this builder example needs a brief explanation of the function interface right next to it — a reader shouldn't have to reverse-engineer it from the sample. The code-verified contract (also listed in point 3 of my earlier comment):

**Custom profile builder** — `@register_profile_builder(profile_type)`
- Receives: *df* (`DataFrame`), *column_name* (`str`), *column_type* (`DataType`), *profiler_metrics* (`dict[str, Any]` — the column-level stats the profiler computed, keyed by metric type, including any custom metrics you registered), and *profiler_options* (`dict[str, Any]` — profiler configuration).
- Returns: a `DQProfile` (a data quality rule suggestion), or `None` to emit nothing for that column.

mwojtyczka:
Addressed all the docs and testing feedback above directly on this branch (commits `3103199` and `a88902e`):

**Docs** (`guide/data_profiling.mdx`, `reference/profiler.mdx`):
- Call `register_profile_column_metric` / `register_profile_builder` decorators.
- Dropped the misleading "at no extra cost"; now notes each registered metric adds an aggregation to every profiling run.
- Tied *profile_column_metric_type* to the concrete decorator argument (e.g. `"percentile_10"`).
- Documented the metric- and builder-function interfaces (inputs/outputs) inline next to each example.
- Framed generated profiles as data quality rule suggestions.
- Cross-referenced the reference page's Custom Column Metrics section from the guide (bidirectional).

**Tests**:
- New integration test exercising the full extension path (custom column metric → custom `register_profile_builder` → generated `DQProfile` via `profile()`), which previously had no coverage; also asserts the built-in `count_distinct` metric.
- Builder-registry snapshot fixture for isolation.
- Unit test for the column-metric overwrite-and-warn path.

mwojtyczka (approved):
LGTM - made some small doc changes and increased code coverage

PR #1387, opened 2026-07-29
Fix NULL group matching in time-window gap checks

### LINKED ISSUE #1386: [BUG]: has_no_gaps_per_time_window silently drops violations for NULL group keys

- [x] I searched all open and closed issues and pull requests, including comments.

Related items are not duplicates:

- #1354 / #1356 fixed null-unsafe result reattachment in `sql_query`, grouped aggregate comparisons, and grouped aggregate outlier checks.
- #1370 added `has_no_gaps_per_time_window` after that fix and introduced a new ordinary equality join that does not use the null-safe helper.

### Current Behavior

`has_no_gaps_per_time_window(..., group_by=[...])` correctly computes a gap for a group whose key contains `NULL`, then silently loses the computed condition while joining the per-window result back to the input.

The function retains nullable grouping keys in `distinct_windows`, partitions by them, and computes `condition_col`. It then uses:

```python
joined = df.join(gaps, on=[*group_by_names, window_start_col], how="left")
```

Spark equality joins do not match `NULL = NULL`. Therefore, when any group-key component is null, the result row does not reattach and the affected input row passes without a violation.

### Expected Behavior

Gaps should be detected independently for every group, including groups with one or more null key components. A computed violation must be attached to the matching boundary row using null-safe equality.

### Steps To Reproduce

Run on current `main` (`632f094ee59583b6a5674f47fd40245b9f1606cd`) with a Spark session:

```python
from datetime import date

from databricks.labs.dqx.check_funcs import has_no_gaps_per_time_window

df = spark.createDataFrame(
    [
        (None, date(2025, 7, 14), 1),
        (None, date(2025, 7, 16), 2),
    ],
    "device string, event_date date, val int",
)

condition, apply = has_no_gaps_per_time_window(
    column="event_date",
    window_minutes=1440,
    group_by=["device"],
)

apply(df).select("device", "event_date", condition).show(truncate=False)
```

The `2025-07-14` row should report the missing `2025-07-15` window. Instead, its condition is null and it passes.

As a control, replacing `None` with `"A"` correctly reports the violation.

### Impact

This creates silent false negatives for per-device, session, tenant, region, or composite-key time-series quality checks whenever any grouping component is null. Missing data windows can therefore pass a quality gate without an exception or skipped-check signal.

### Affected Code

- `src/databricks/labs/dqx/check_funcs.py`: `has_no_gaps_per_time_window.apply`, result join near line 3020.
- `tests/integration/test_dataset_checks.py`: grouped gap tests cover only non-null keys.

### Suggested Fix

Reuse the existing `_join_results_on_null_safe_columns` helper added by #1356 and add a focused integration regression for a violating null group key. The `gaps` DataFrame is already unique per group/window key, satisfying the helper precondition.

### Cloud

AWS

### Operating System

macOS

### Environment

- DQX revision: `main@632f094ee59583b6a5674f47fd40245b9f1606cd`
- PySpark: 4.0.0 via Databricks Connect
- Python: 3.12.13
- Java: OpenJDK 17.0.19

mattfaltyn:
## Changes

- Reattach `has_no_gaps_per_time_window` results with the existing null-safe result-join helper.
- Preserve gap violations for groups with null key components without changing the public API or non-null behavior.
- Add a focused Spark integration regression for a null-key group with an interior gap.

### Linked issues

Fixes #1386

Related: #1354, #1356

### Tests

- [x] manually tested — focused regression passed with stock PySpark 4.0.0
- [ ] added unit tests — the affected behavior requires Spark
- [x] added integration tests
- [ ] added end-to-end tests — no workflow or external-resource behavior changed
- [ ] added performance tests — the change reuses the existing result-join helper at one call site

Validation performed:

- `make fmt`
- `make lint`
- `make test` — 1,549 passed
- Isolated stock-PySpark 4.0.0 smoke reproduction — null-key boundary row flagged, following row passed

The workspace-backed focused integration test cannot run locally without `DATABRICKS_HOST`; upstream integration CI provides the authoritative run.

### Documentation and Demos

- [ ] added/updated demos — not applicable
- [ ] added/updated docs — the fix restores existing documented behavior without changing the API
- [ ] added/updated agent skills — no public API or usage contract changed


mattfaltyn:
Local validation is green (`make fmt`, `make lint`, `make test` with 1,549 passed, plus the focused regression on stock PySpark 4.0.0). Could a maintainer please approve the fork workflows or create the standard `fork-test/pr-1387` mirror so the workspace integration and acceptance suites can run?

mwojtyczka:
Reviewed the null-safe join fix — looks correct and well-targeted. I pushed one commit (`192014ad`) adding a complementary test.

**Review summary (LGTM):**
- The bug is real: the old `df.join(gaps, on=[*group_by_names, window_start_col], how="left")` uses `=` semantics, so a NULL `group_by` key never matched (`NULL = NULL` → false) and gaps in NULL-keyed groups were silently never reported. Switching to `_join_results_on_null_safe_columns` (joins via `eqNullSafe`) fixes it.
- The helper's three documented preconditions all hold here: join keys are non-empty (`window_start_col` is always present, even without `group_by`); `gaps` is unique per key (derived from `distinct_windows`); and the result columns are uuid-suffixed, disjoint from `df`. Column ordering/output is unchanged.
- **Audited the rest of the codebase for the same bug class** — no other function needs this fix:
  - The aggregate `group_by` family (`is_aggr_*`, `has_no_aggr_outliers`, `aggr_matches_dataset`) already merges results through `_join_results_on_null_safe_columns`.
  - `compare_datasets` already uses `_match_rows` with `null_safe_row_matching=True`.
  - `is_unique` uses `Window.partitionBy` (not a join), which groups NULLs together — no join-NULL issue.
  - `foreign_key`'s `==` join is intentional (a NULL FK must *not* match — correct FK semantics), not an instance of this bug.

**Added test:** `test_has_no_gaps_per_time_window_group_by_mixed_null_and_named_keys`. The PR's existing test covers an all-NULL batch (proving `NULL<=>NULL` matches). The new test complements it with a batch that mixes a NULL-key group and named groups, asserting the null-safe join keeps each group's gaps isolated — the NULL-device group and device A each get their own interior gap flagged, device B (no gap) stays clean, with no cross-group contamination. It passes against a live workspace.

mwojtyczka (approved):
LGTM

PR #1399, opened 2026-07-31
Preserve fractional seconds in timestamp profiles

### LINKED ISSUE #1398: [BUG]: Timestamp profiling truncates fractional seconds when rounding is disabled

### Current Behavior

Profiling a `TimestampType` or `TimestampNTZType` column with fractional seconds and `options={"round": False}` truncates the learned min/max bounds to whole seconds.

For example, a source maximum of:

```text
2024-01-01T00:00:00.999999+00:00
```

is stored in the `min_max` profile as:

```text
2024-01-01T00:00:00+00:00
```

The generated `is_in_range` rule therefore rejects the source maximum that was used to generate the rule.

The timestamp fallback in `_make_min_max_profile_without_outlier_removal` formats aggregate values with `yyyy-MM-dd HH:mm:ss` and parses them with `%Y-%m-%d %H:%M:%S`, dropping fractional seconds before `_round_value` sees the values.

This is distinct from #475/#483: those fixed `round=False` option propagation and `datetime.max` overflow in the previous profiler implementation. The current fallback was introduced later by #1059.

### Expected Behavior

When `round` is disabled, timestamp min/max profiles preserve the source precision so the generated inclusive range accepts every profiled source value.

### Steps To Reproduce

```python
import datetime

from pyspark.sql import Row, SparkSession
from pyspark.sql import types as T

from databricks.labs.dqx.profiler.profile_builder import make_min_max_profile

spark = SparkSession.builder.master("local[1]").getOrCreate()
source_max = datetime.datetime(2024, 1, 1, 0, 0, 0, 999999)
df = spark.createDataFrame([Row(event_time=source_max)])

profile = make_min_max_profile(
    df,
    "event_time",
    T.TimestampType(),
    {"count_non_null": 1, "min": None, "max": None},
    {"remove_outliers": False, "round": False},
)

assert profile is not None
assert profile.parameters["max"].replace(tzinfo=None) == source_max
```

The final assertion currently fails because `profile.parameters["max"]` has `microsecond == 0`.

### Cloud

AWS

### Operating System

macOS

```shell
AssertionError
source max:  2024-01-01T00:00:00.999999
profile max: 2024-01-01T00:00:00
```

mattfaltyn:
## Changes

- preserve microsecond precision when timestamp min and max aggregates are converted to profile values
- document why Spark uses six fractional digits to match Python `%f` parsing
- cover `TimestampType` and `TimestampNTZType` with rounding enabled and disabled

### Linked issues

Resolves #1398

### Tests

- [x] manually tested
- [ ] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

Local validation:

- `make test`, 1,558 passed
- Black, Ruff, and mypy passed
- Pylint passed with 10.00 out of 10
- `git diff --check` passed
- targeted integration test collection passed for all four timestamp type and rounding combinations

The targeted integration tests require `DATABRICKS_HOST` to execute and that environment was not available locally.

### Documentation and Demos

- [ ] added or updated demos
- [ ] added or updated docs
- [ ] added or updated agent skills

No documentation or demo changes are needed because this restores the documented behavior that unrounded min and max profiles retain their source values.


src/databricks/labs/dqx/profiler/profile_builder.py
```diff
@@ -531,18 +531,18 @@ def _make_min_max_profile_without_outlier_removal(
...
+                F.date_format("min_value", "yyyy-MM-dd HH:mm:ss.SSSSSS").alias("min_value"),
+                F.date_format("max_value", "yyyy-MM-dd HH:mm:ss.SSSSSS").alias("max_value"),
             )
         aggregates = agg_df.collect()[0].asDict()
         if not aggregates or aggregates.get("min_value") is None:
             logger.info(f"Can't get min/max for field {column_name}")
             return None
         if isinstance(column_type, (T.TimestampType, T.TimestampNTZType)):
-            min_value = datetime.datetime.strptime(aggregates["min_value"], "%Y-%m-%d %H:%M:%S").replace(
+            min_value = datetime.datetime.strptime(aggregates["min_value"], "%Y-%m-%d %H:%M:%S.%f").replace(
```
mwojtyczka: **Correctness dependency (low risk):** this pairs `date_format(...'.SSSSSS')` with `strptime('%Y-%m-%d %H:%M:%S.%f')`, which only works because Spark's `SSSSSS` pattern always emits a 6-digit fraction (e.g. `.000000` for whole seconds). If a runtime ever rendered a whole-second timestamp without the trailing fraction, this `strptime` would raise `ValueError` and profiling would crash for that column rather than degrade.

Verified low-risk: the pre-existing `round=False` integration test (`test_profiler.py`, column `s1.ns1` with `datetime(2023, 1, 6, 10, 0, 11)`) exercises this exact whole-second path and asserts the parsed result, so the `.000000` emission is regression-guarded. A one-line comment noting the dependency on `SSSSSS` always producing a fraction would make the coupling explicit for future readers.

tests/integration/test_profiler.py
```diff
@@ -140,25 +140,32 @@ def test_profiler_is_in_large_table_few_distinct_values(spark, ws):
...
+    )
 
     min_max_profiles = [p for p in profiles if p.name == "min_max" and p.column == "created_at"]
     assert len(min_max_profiles) == 1
-    assert min_max_profiles[0].parameters["min"] is not None
-    assert min_max_profiles[0].parameters["max"] is not None
+    assert min_max_profiles[0].parameters == {
+        "min": datetime(2024, 1, 1, 0, 0, 0, 123456, tzinfo=timezone.utc),
+        "max": datetime(2024, 12, 31, 23, 59, 59, 654321, tzinfo=timezone.utc),
+    }
```
mwojtyczka: **Test-coverage gap:** the rewritten test asserts fractional-second preservation only with `round=False`. There is no assertion that the default `round=True` path still rounds a *fractional*-second timestamp to midnight — the branch most users hit, and the one the `.SSSSSS` change flows through before `_round_datetime` zeroes microseconds. `test_profiler_rounding_midnight_behavior` uses whole-second inputs, so it wouldn't detect microsecond leakage if a future refactor of `_round_datetime` stopped zeroing `microsecond`.

Consider adding a `round=True` case with a fractional-second input asserting the min/max round to midnight (microsecond=0), to lock down the default-path behavior for sub-second timestamps.

mwojtyczka (commented):
Reviewed the timestamp fractional-seconds fix. The change is correct and well-scoped — verified that `date_format(...'.SSSSSS')` and `strptime('%f')` are symmetric, that the default `round=True` path is unaffected (`_round_datetime` zeroes microseconds regardless), and that a pre-existing `round=False` whole-second integration test already guards the `.000000` emission. Two low-severity, non-blocking notes inline.

mwojtyczka (commented):
Generally looking good, left small comments

mwojtyczka (changes requested):

tests/integration/test_profiler.py
mattfaltyn: Thanks. I added `round=True` coverage for fractional values across both `TimestampType` and `TimestampNTZType`.

src/databricks/labs/dqx/profiler/profile_builder.py
mattfaltyn: Good point. I added a comment documenting that `SSSSSS` always emits six fractional digits, including `.000000`, to match `%f`.

mattfaltyn:
@mwojtyczka Thanks for the review. I addressed both comments and resolved the threads. The PR is ready for another look.

mwojtyczka (approved):
LGTM - i pushed some additional fixes

Issue #1073, opened 2026-03-13
[BUG]: [CONFIG_NOT_AVAILABLE] Configuration pipelines.id is not available.


FlavioDiasPs:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

dqx version: 0.13.0

When running DQX on a Databricks Serverless (Spark Connect) cluster, save_summary_metrics (and any method decorated with @telemetry_logger) logs 3 gRPC ERROR messages per call. The errors come from is_dlt_pipeline() in telemetry.py calling spark.conf.get('pipelines.id', None). On Spark Connect, the server rejects that config key with [CONFIG_NOT_AVAILABLE] instead of returning None. The try/except in is_dlt_pipeline() catches it so nothing actually breaks, but PySpark's gRPC client logs the exception at ERROR level before it reaches the catch, which pollutes logs and confuses users into thinking something went wrong.

### Expected Behavior

No ERROR logs should be emitted when checking for DLT pipeline context on non-DLT environments. is_dlt_pipeline() should handle Spark Connect clusters gracefully without triggering gRPC error logging.

### Steps To Reproduce

Use a Databricks Serverless compute cluster (Spark Connect)
Run any DQX method that goes through log_dataframe_telemetry, e.g.:

from databricks.labs.dqx.engine import DQEngine
from databricks.labs.dqx.metrics_observer import DQMetricsObserver
from databricks.sdk import WorkspaceClient

engine = DQEngine(WorkspaceClient(), observer=DQMetricsObserver())
engine.apply_checks_by_metadata_and_split(df, checks)

Observe 3x gRPC ERROR log entries per call

### Cloud

AWS

### Operating System

macOS

### Relevant log output

```shell
{"ts": "2026-03-10 14:26:49,358", "level": "ERROR", "logger": "pyspark.sql.connect.client.logging", "msg": "GRPC Error received", "exception": {"class": "_InactiveRpcError", "msg": "<_InactiveRpcError of RPC that terminated with:\n\tstatus = StatusCode.INTERNAL\n\tdetails = \"[CONFIG_NOT_AVAILABLE] Configuration pipelines.id is not available. SQLSTATE: 42K0I\"\n>"}}
```

mwojtyczka:
Closing this for now. Please re-open if you still see the issue.

Issue #1001, opened 2026-01-19
[BUG]: ATTEMPT_ANALYSIS_IN_PIPELINE_QUERY_FUNCTION when using with SDP


lucasarruda-ciandt:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

I have a Spark Declarative Pipeline (4.1.0) using DLT's expectations.
I'm working on a DQX (0.12.0) replacement proof-of-concept, however as soon as I call any DQEngine method such as apply_checks_by_metadata or apply_checks inside a table definition I get:

```
pyspark.errors.exceptions.base.PySparkException: [ATTEMPT_ANALYSIS_IN_PIPELINE_QUERY_FUNCTION] Operations that trigger DataFrame analysis or execution are not allowed in pipeline query functions. Move code outside of the pipeline query function.
```

The rest of the stacktrace gave me useful hints:

```
     ^^^^^^^^^^^^^^
.venv/lib/python3.12/site-packages/pyspark/pipelines/api.py:388: in temporary_view
    outer(query_function)
.venv/lib/python3.12/site-packages/pyspark/pipelines/api.py:372: in outer
    registry.register_flow(
.venv/lib/python3.12/site-packages/pyspark/pipelines/spark_connect_graph_element_registry.py:119: in register_flow
    df = flow.func()
         ^^^^^^^^^^^
src/pipeline.py:183: in preprocessing
    return dq_engine.apply_checks_by_metadata(df, [])
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
.venv/lib/python3.12/site-packages/databricks/labs/dqx/telemetry.py:66: in wrapper
    return func(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
.venv/lib/python3.12/site-packages/databricks/labs/dqx/engine.py:547: in apply_checks_by_metadata
    log_dataframe_telemetry(self.ws, self.spark, df)
.venv/lib/python3.12/site-packages/databricks/labs/dqx/telemetry.py:92: in log_dataframe_telemetry
    log_telemetry(ws, "streaming", str(df.isStreaming).lower())
                                       ^^^^^^^^^^^^^^
../../../.pyenv/versions/3.12.3/lib/python3.12/functools.py:995: in __get__
    val = self.func(instance)
          ^^^^^^^^^^^^^^^^^^^
[...]

PR #1401, opened 2026-07-31
Make telemetry best-effort: never throw, dedup per process, bounded timeout (#1355)

### LINKED ISSUE #1355: Telemetry makes a blocking per-check control-plane call guarded only by DatabricksError — SDK retry TimeoutError escapes and terminates streaming jobs; no opt-out
**Current Behavior**

`log_telemetry()` in `src/databricks/labs/dqx/telemetry.py` records usage by copying the `WorkspaceClient` config, appending a user-agent extra, recreating a `WorkspaceClient`, and synchronously calling `ws.clusters.select_spark_version()` (GET `/api/2.1/clusters/spark-versions`), discarding the response. It guards only with `except DatabricksError`.

This has three compounding problems:

1. **Call volume / placement.** `log_telemetry` is invoked from `_create_results_array` once **per check** (engine.py: `log_telemetry(self.ws, "check", check.check_func.__name__)`), plus via `@telemetry_logger` on every `apply_checks*` entry point and via `log_dataframe_telemetry` (streaming flag, DLT flag, one call per input table/path). In Structured Streaming `foreachBatch` usage this means many blocking control-plane GETs per micro-batch, forever, per job.
2. **Insufficient exception guard.** When the control plane returns retryable errors (e.g. 503), the databricks-sdk `ApiClient` retries for its configured window (default 300s) and then raises a **builtin `TimeoutError`** (`raise TimeoutError(f"Timed out after {timeout}") from last_err` in `databricks/sdk/retries.py`). `TimeoutError` is **not** a `DatabricksError`, so it escapes the `except DatabricksError` guard and propagates into user code. (`log_dataframe_telemetry`'s docstring says "This function is designed to never throw exceptions" — this shows it can; #1001 was another instance via `df.isStreaming`.)
3. **No opt-out.** There is no environment variable, config flag, or engine option to disable telemetry.

**What we observed in production (Azure, 2026-07-21):** the regional control plane returned 503 `TemporarilyUnavailable` for `/api/2.1/clusters/spark-versions` for ~45 minutes. Every DQX telemetry call blocked for the full 300s retry window and then raised `TimeoutError`, which propagated through `foreachBatch` and terminated **three production Structured Streaming jobs**, repeatedly, until their job-level retries were exhausted. Even before the crash, each micro-batch was stalled by serial 300-second telemetry retries. Verified the relevant code is identical from v0.9.3 through v0.15.0 and current `main`.

**Expected Behavior**

Telemetry must never affect the data path: it should never throw (any exception), never block a micro-batch for minutes, and be possible to disable. Concrete suggestions (any subset would help):

1. Broaden the guard to `except Exception` (log at debug) — minimal fix for the crash.
2. Bound the call: on the copied config set a small `retry_timeout_seconds` / HTTP timeout (a few seconds) so a control-plane brownout cannot stall micro-batches.
3. De-duplicate: cache per process per `(key, value)` so the ping fires once, not once per check per micro-batch — this also reduces load on the `/clusters/spark-versions` endpoint itself.
4. Consider the ucx/blueprint pattern — `databricks.sdk.config.with_user_agent_extra(...)` piggybacks telemetry on API calls the library already makes, needing no dedicated request.
5. Add an opt-out env var, e.g. `DQX_DISABLE_TELEMETRY`, mirroring the Databricks CLI's `DATABRICKS_CLI_DISABLE_TELEMETRY`.

Related prior reports of telemetry side effects: #1001, #1073. The original feature request #378 itself noted the extra SDK call "should be used carefully".

**Steps To Reproduce**

1. Build a `DQEngine` with any checks and apply them inside `foreachBatch` of a Structured Streaming query.
2. Make `/api/2.1/clusters/spark-versions` return a retryable error (easiest deterministic repro — monkeypatch the SDK call):

```python
from unittest.mock import patch
from databricks.sdk.errors import TemporarilyUnavailable

# simulate SDK retry exhaustion (what the SDK raises after its 300s window on repeated 503s)
def exhausted(*a, **kw):
    raise TimeoutError("Timed out after 0:05:00") from TemporarilyUnavailable("...")

with patch("databricks.sdk.service.compute.ClustersExt.select_spark_version", exhausted):
    dq_engine.apply_checks_by_metadata(df, checks)  # raises TimeoutError, fails the batch
```

3. Observe the `TimeoutError` escape `log_telemetry`'s `except DatabricksError` and fail the streaming query. In a live environment the same happens whenever the control plane has a 503 episode longer than the SDK retry window.

**Cloud:** Azure — **Operating System:** Linux

**Relevant log output**

```
File ".../databricks/labs/dqx/engine.py", line 580, in _create_results_array
    log_telemetry(self.ws, "check", check.check_func.__name__)
File ".../databricks/labs/dqx/telemetry.py", line 32, in log_telemetry
    ws.clusters.select_spark_version()
File ".../databricks/sdk/retries.py", line 71, in wrapper
    raise TimeoutError(f"Timed out after {timeout}") from last_err
TimeoutError: Timed out after 0:05:00
(root cause: repeated 503 TemporarilyUnavailable from GET /api/2.1/clusters/spark-versions)
```

mwojtyczka:
## Changes

Fixes #1355. Telemetry could terminate Structured Streaming jobs: `log_telemetry` made a **blocking per-check** control-plane call (`select_spark_version`, GET `/api/2.1/clusters/spark-versions`) guarded only by `except DatabricksError`. On a control-plane brownout the SDK retried for its ~300s window and then raised a builtin **`TimeoutError`**, which is *not* a `DatabricksError` — so it escaped the guard, propagated through `foreachBatch`, and failed the batch, **once per micro-batch, forever**.

This makes telemetry strictly best-effort — it can never throw, never block for long, and never fire more than once per signal — implementing the two agreed items from the issue thread plus a small bounded-timeout:

- **Never throw.** Broaden the guard `except DatabricksError` → `except Exception` (debug-logged). Also wrap `log_dataframe_telemetry`'s whole body so a raising `df.isStreaming` (the #1001 failure mode) can't propagate either, honoring its "never throw" docstring.
- **Deduplicate per process.** Cache each `(key, value)` so the ping fires **at most once per signal per process**, not once per check per micro-batch. Telemetry is a user-agent *adoption* signal (which features/checks/streaming/DLT are used), not an invocation counter, so dedup preserves exactly the signal we care about while removing the repeated blocking calls.
- **Bound the call.** Cap `retry_timeout_seconds` / `http_timeout_seconds` (5s) on the copied config so a brownout gives up in seconds instead of stalling the caller for minutes.

No background threads, no opt-out env var, and no `with_user_agent_extra` piggyback — see the "not done" tradeoffs below (all consistent with the [issue discussion](https://github.com/databrickslabs/dqx/issues/1355#issuecomment-)).

### Tradeoffs (deliberate)

- **Dedup marks a signal "sent" on *attempt*, not only on success.** This guarantees at most one control-plane call per signal per process, so a brownout can't re-stall every micro-batch. The cost: if the *very first* attempt for a signal hits a transient failure, that one signal is not retried for the life of the process (silently lost). Accepted — bounding the stall matters more than a single best-effort signal.
- **The dedup cache is process-global and hard-bounded (FIFO).** Most keys are naturally bounded (check names, `streaming`/`dlt`, `workflow`), but `input_table`/`input_path` add one small hashed entry per *distinct* input, so a very long-lived driver reading an ever-growing set of tables could otherwise grow it without limit. The cache is an `OrderedDict` capped at 10,000 entries, evicting oldest-first when over capacity. The cap is far above realistic distinct-input counts (thousands ≈ low-single-digit MB), so eviction only triggers for pathological workloads and merely lets an evicted signal be re-sent once later — harmless for best-effort telemetry.
- **No lock on the dedup cache.** Dict membership / insert / `popitem` are atomic under the GIL, so the cache can never be corrupted; a concurrent check-then-add race can at most fire one duplicate ping, and since eviction only runs when the cache is over capacity (never near-empty) a concurrent eviction can at most drop one extra entry — both harmless for a best-effort call, so a lock isn't worth the machinery.
- **Equal timeouts (`retry` == `http` == 5s).** Leaves no room for a retry within the window, so a slow-but-healthy response (>5s tail) is dropped. Intentional: the goal is to give up fast, and 5s is ample for a healthy control plane.

**Not done (per the issue discussion):** bounded-timeout-only tuning knob, `with_user_agent_extra` piggyback (the streaming path makes no SDK calls to piggyback on, so it would send *no* telemetry for exactly the usage we want to see), and a telemetry opt-out env var (we rely on adoption telemetry to improve DQX).

### Linked issues

Resolves #1355. Related prior reports of telemetry side effects: #1001, #1073.

### Tests

- [ ] manually tested
- [x] added unit tests
- [ ] added integration tests
- [ ] added end-to-end tests

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -892,6 +892,7 @@ redefining-builtins-modules = ["six.moves", "past.builtins", "future.builtins", 
 "tests/unit/test_anomaly_scoring_run.py" = "protected-access"
 "tests/integration_anomaly/test_anomaly_ai_explanation.py" = "protected-access"
 "tests/unit/test_checks_storage.py" = "protected-access"
+"tests/unit/test_telemetry.py" = "protected-access"
 
 # This plugin is used to generate correct links in README showed on PyPi
 [tool.hatch.metadata.hooks.fancy-pypi-readme]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -892,7 +892,6 @@ redefining-builtins-modules = ["six.moves", "past.builtins", "future.builtins", 
 "tests/unit/test_anomaly_scoring_run.py" = "protected-access"
 "tests/integration_anomaly/test_anomaly_ai_explanation.py" = "protected-access"
 "tests/unit/test_checks_storage.py" = "protected-access"
-"tests/unit/test_telemetry.py" = "protected-access"
 
 # This plugin is used to generate correct links in README showed on PyPi
 [tool.hatch.metadata.hooks.fancy-pypi-readme]
--- a/src/databricks/labs/dqx/telemetry.py
+++ b/src/databricks/labs/dqx/telemetry.py
@@ -4,6 +4,7 @@ import functools
 import logging
 import hashlib
 from io import StringIO
+from collections import OrderedDict
 from collections.abc import Callable
 from pyspark.sql import DataFrame, SparkSession
 from databricks.sdk import WorkspaceClient
@@ -21,11 +22,19 @@ _TELEMETRY_TIMEOUT_SECONDS = 5
 # already sent means the ping fires at most once per signal instead of once per check per micro-batch
 # — bounding both the load on the control plane and the number of times a brownout can stall a caller.
 #
-# No lock: set membership/add are atomic under the GIL and add is idempotent, so the cache can never be
-# corrupted. The only thing a lock would add is preventing a rare check-then-add race under concurrent
-# callers (e.g. overlapping foreachBatch micro-batches) from firing one duplicate ping — harmless for a
-# best-effort, timeout-bounded call, so it isn't worth the extra machinery.
-_sent_telemetry: set[tuple[str, str]] = set()
+# Most signals are naturally bounded (a fixed set of check names, streaming/dlt/workflow flags), but
+# input_table / input_path add one hashed entry per *distinct* input, so a very long-lived process that
+# reads an ever-growing set of tables could accumulate unboundedly. Cap the cache and evict oldest-first
+# (FIFO via OrderedDict) so memory is hard-bounded; the realistic ceiling is far below the cap, so
+# eviction only ever kicks in for pathological workloads and merely allows an evicted signal to be
+# re-sent once later — harmless for best-effort telemetry.
+#
+# No lock: dict membership/set/popitem are atomic under the GIL. A concurrent check-then-add race can at
+# most fire one duplicate ping, and since eviction only runs when the cache is over capacity (never near
+# empty) a concurrent eviction can at most drop one extra entry — both harmless for a best-effort call,
+# so a lock isn't worth the machinery.
+_TELEMETRY_CACHE_MAX_SIZE = 10_000
+_sent_telemetry: "OrderedDict[tuple[str, str], None]" = OrderedDict()
 
 
 def reset_telemetry_cache() -> None:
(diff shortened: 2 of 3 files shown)


ghanse (approved):
LGTM. My only concern is with the unboundedness. It's probably an edge case, but we could use something like an `OrderedDict` and pop the oldest item if we run into limits.

mwojtyczka:
> LGTM. My only concern is with the unboundedness. It's probably an edge case, but we could use something like an `OrderedDict` and pop the oldest item if we run into limits.

good idea, I implemented this and capped at 10k

PR #1402, opened 2026-08-01
Fix set_utc_timezone fixture to actually apply UTC via tzset

### LINKED ISSUE #1391: [BUG]: set_utc_timezone test fixture doesn't call time.tzset(), so it's a no-op on non-UTC dev machines

### Current Behavior

`tests/conftest.py::set_utc_timezone` sets `os.environ["TZ"] = "UTC"` so Spark timestamp
handling is consistent regardless of the host environment, but never calls `time.tzset()`
afterward. On glibc, the C library resolves the local timezone at process start (or at the
last `tzset()` call); setting the `TZ` environment variable alone does not change
`time.localtime()` / `time.tzname` for the already-running process. On any dev machine or CI
runner whose OS default timezone is not already UTC, the fixture is effectively a no-op.

### Expected Behavior

The fixture should actually force the process to observe UTC for the duration of the test, so
integration tests using `datetime(...)` (not `date(...)`) on a `TimestampType` column
serialize and render consistently regardless of the host's local timezone.

### Steps To Reproduce

On a machine whose OS timezone is not UTC (e.g. `Asia/Kolkata`):

```python
import os
import time

print(time.tzname)          # e.g. ('IST', 'IST')
os.environ["TZ"] = "UTC"
print(time.tzname)          # still ('IST', 'IST') -- unchanged
time.tzset()
print(time.tzname)          # now correctly ('UTC', 'UTC')
```

Concretely, this makes `tests/integration/test_dataset_checks.py::test_has_no_gaps_per_time_window_sub_day_window`
(and any other integration test using `datetime(...)` on a `TimestampType` column) fail on a
non-UTC host whenever pytest is invoked without `TZ=UTC` already set in the parent shell
before the Python process starts, even with the `set_utc_timezone` fixture active, since the
fixture's `os.environ` mutation happens after the interpreter has already cached the local
timezone.

### Affected Code

- `tests/conftest.py`: `set_utc_timezone` fixture.

### Suggested Fix

Call `time.tzset()` right after setting the environment variable, and again after restoring
it on teardown, guarded for platforms where it isn't available:

```python
import time

@pytest.fixture
def set_utc_timezone():
    os.environ["TZ"] = "UTC"
    if hasattr(time, "tzset"):
        time.tzset()
    yield
    os.environ.pop("TZ")
    if hasattr(time, "tzset"):
        time.tzset()
```

`time.tzset()` is POSIX-only (not available on Windows), hence the `hasattr` guard, which
also preserves today's behavior there.

### Cloud

AWS

### Operating System

Linux

```
Not cloud-specific -- reproduces on any non-UTC local dev machine or self-hosted CI runner
regardless of cloud provider. Observed on a Linux sandbox with OS timezone Asia/Kolkata.
```

SreeramaYeshwanthGowd:
## Changes

- Call `time.tzset()` in the `set_utc_timezone` fixture right after setting `TZ` and again after clearing it on teardown, guarded by `hasattr(time, "tzset")`.
- Add a regression test that forces a non-UTC starting timezone before invoking the fixture, so the test is host-agnostic and would fail without the fix regardless of the CI runner's default timezone.

### Linked issues

Fixes #1391

### Tests

- [x] manually tested - regression test fails against the pre-fix fixture (`AssertionError: assert 'EST' == 'UTC'`) and passes against the fix
- [x] added unit tests
- [ ] added integration tests - not needed, the fix is a pure stdlib change with no Spark-dependent behavior
- [ ] added end-to-end tests - no workflow behavior changed
- [ ] added performance tests - not applicable

Validation performed:

- `make fmt`
- `make lint`
- `make test`

The Spark-dependent tests that rely on `set_utc_timezone` (e.g. `tests/integration/test_dataset_checks.py`) can't be run locally without `DATABRICKS_HOST`; upstream integration CI provides the authoritative run.

### Documentation and Demos

- [ ] added/updated docs - internal test fixture only, no public API changed
- [ ] added/updated agent skills - no public API or usage contract changed


tests/unit/test_conftest_fixtures.py
```diff
@@ -0,0 +1,24 @@
...
+    if original is None:
+        os.environ.pop("TZ", None)
+    else:
+        os.environ["TZ"] = original
+    if hasattr(time, "tzset"):
+        time.tzset()
+
+
+def test_set_utc_timezone_overrides_a_non_utc_host(force_non_utc_start, set_utc_timezone):
+    assert time.tzname[0] == "UTC"
```
mwojtyczka, line 25: **Test can pass vacuously if tzdata is missing.** glibc silently falls back to UTC for any zone it cannot resolve (confirmed: an unknown `TZ` yields `time.tzname == ('UTC', 'UTC')`). If the runtime lacks the tz database, the "non-UTC start" is already UTC and this test passes **without exercising the fix at all** — a false green. Assert the precondition so the test stays honest, e.g. have `force_non_utc_start` assert `time.tzname[0] != "UTC"` (or `pytest.skip`) right after its `tzset()`, before `yield`.

tests/conftest.py
```diff
@@ -196,8 +196,13 @@ def set_utc_timezone():
     Set the timezone to UTC for the duration of the test to make sure spark timestamps    are handled the same way regardless of the environment.
     """
     os.environ["TZ"] = "UTC"
+    if hasattr(time, "tzset"):
+        # Without this, glibc keeps using the host's cached timezone (POSIX-only, hence the guard)
+        time.tzset()
     yield
     os.environ.pop("TZ")
```
mwojtyczka: **Teardown discards a pre-existing `TZ`.** `os.environ.pop("TZ")` removes any `TZ` the environment had before the fixture rather than restoring it, and would `KeyError` if `TZ` were ever cleared mid-test. Your new `force_non_utc_start` already models the correct pattern; since this PR is editing the fixture anyway, consider aligning:

```python
original = os.environ.get("TZ")
os.environ["TZ"] = "UTC"
if hasattr(time, "tzset"):
    time.tzset()
yield
if original is None:
    os.environ.pop("TZ", None)
else:
    os.environ["TZ"] = original
if hasattr(time, "tzset"):
    time.tzset()
```

Optional — reasonable to leave out of scope.

tests/conftest.py
```diff
@@ -196,8 +196,13 @@ def set_utc_timezone():
     Set the timezone to UTC for the duration of the test to make sure spark timestamps    are handled the same way regardless of the environment.
```
mwojtyczka: Minor: stray run of spaces in the docstring (`spark timestamps    are handled`). Not introduced by this PR, but it sits right above the edit — a one-character cleanup while you're here. Skip if you prefer to keep the diff focused.

mwojtyczka (commented):
## Code Review — Fix `set_utc_timezone` fixture to actually apply UTC via `tzset`

**Overview.** The fixture set `os.environ["TZ"] = "UTC"` but never called `time.tzset()`, so under glibc the C library kept its cached timezone and the process's local time never actually changed. This PR adds `time.tzset()` after setting and after clearing `TZ` (guarded by `hasattr(time, "tzset")`), plus a host-agnostic regression test that forces `America/New_York` before invoking the fixture.

The core fix is **correct and well-targeted** — verified locally that without `tzset()` the change is inert and with it `time.tzname` flips to `('UTC', 'UTC')`. The `hasattr` guard is the right idiom (`time.tzset` is POSIX-only); CI unit jobs run on `linux-ubuntu-latest`, so it's belt-and-suspenders but appropriate.

**Strengths**
- Minimal, surgical fix for a real latent bug affecting every timestamp-sensitive test relying on this fixture.
- The regression test forces a non-UTC start rather than depending on the runner default — genuinely host-agnostic.
- `force_non_utc_start` correctly saves/restores the original `TZ`.

**Findings** — see inline comments:
1. Regression test can pass *vacuously* if tzdata is missing (glibc silently falls back to UTC for an unresolvable zone) — the one finding that materially affects whether the test keeps guarding the fix.
2. `set_utc_timezone` teardown discards a pre-existing `TZ` instead of restoring it (pre-existing; the PR is already editing this fixture).
3. Minor stray whitespace in the docstring one line above the edit.

**Conventions / coverage / security.** Unit-test placement is correct (no Spark/workspace, fast). No performance or security implications — `tzset()` is negligible, no untrusted input or I/O.

**Verdict:** solid, correct fix. Approve once finding #1 is addressed or consciously waived; #2 and #3 are optional polish.

mwojtyczka (changes requested):
Thanks for the PR! generally looking good, left a few comments

SreeramaYeshwanthGowd:
Thanks @mwojtyczka, addressed both findings: force_non_utc_start now skips when tzdata cannot produce a non UTC start, and set_utc_timezone properly restores the original TZ instead of dropping it, with a shared helper factored out.

tests/unit/test_conftest_fixtures.py
```diff
@@ -0,0 +1,23 @@
...
+
+@pytest.fixture
+def force_non_utc_start():
+    """Force a non-UTC starting timezone regardless of the host's actual default."""
+    original = os.environ.get("TZ")
+    os.environ["TZ"] = "America/New_York"
+    if hasattr(time, "tzset"):
+        time.tzset()
+    if time.tzname[0] == "UTC":
+        pytest.skip("Host's tzdata doesn't resolve America/New_York; glibc fell back to UTC")
```
mwojtyczka: **Prefer removing the skip entirely over fixing it.** Two problems here: (1) a `pytest.skip()` means that on a host without tzdata this test is silently sidestepped rather than exercising the fix — a green suite no longer guarantees the tzset fix works. (2) The skip is raised during setup *before* `yield`, so the post-yield `restore_tz_env(original)` never runs and `TZ=America/New_York` (plus its `tzset`) leaks into every later test in the same worker.

Both go away if you force the non-UTC start with a **POSIX TZ string** instead of an IANA name — glibc parses POSIX TZ (`std offset dst`) directly, with no tz database, so it never falls back to UTC and no skip is needed:

```python
@pytest.fixture
def force_non_utc_start():
    """Force a non-UTC starting timezone regardless of the host's actual default."""
    original = os.environ.get("TZ")
    os.environ["TZ"] = "EST5EDT"  # POSIX TZ string: parsed by glibc without tzdata
    if hasattr(time, "tzset"):
        time.tzset()
    yield
    restore_tz_env(original)
```

Verified locally: `TZ=EST5EDT` → `time.tzname == ('EST', 'EDT')` even on a box with no tz database, so `test_set_utc_timezone_overrides_a_non_utc_host` always runs and always starts non-UTC. (If you want to keep exercising a DST boundary the assertion doesn't currently rely on that, so the POSIX string is sufficient.)

tests/conftest.py
```diff
@@ -190,14 +190,28 @@ def override_cluster_id(debug_env):
     return debug_env
 
 
+def restore_tz_env(original: str | None) -> None:
+    """Restore the TZ environment variable to its pre-fixture value and reapply it via tzset."""
+    if original is None:
+        os.environ.pop("TZ", None)
+    else:
+        os.environ["TZ"] = original
+    if hasattr(time, "tzset"):
```
mwojtyczka, line 199: Minor (DRY): the `set TZ` + `if hasattr(time, "tzset"): time.tzset()` apply pattern now appears in three places — `set_utc_timezone` and `restore_tz_env` here, and `force_non_utc_start` in `test_conftest_fixtures.py`. Per the repo's DRY rule (extract at 3+ occurrences), a small `apply_tz(value: str | None)` helper alongside `restore_tz_env` would keep the POSIX guard in one place. Non-blocking.

mwojtyczka (commented):
## Re-review — after the pushed fixes

The three earlier findings are addressed: teardown now restores the pre-existing `TZ` (via the new `restore_tz_env` helper), the docstring whitespace is gone, and the vacuous-tzdata case is handled. Two follow-ups from the new code — see inline.

The headline one: the vacuous-tzdata case was handled by adding a `pytest.skip()`, but **a skipped test isn't the goal here** — a green run should mean the fix was actually exercised, not silently sidestepped on a host without tzdata. And the skip has a concrete bug (teardown leak) on top of that. Both problems disappear if `force_non_utc_start` uses a **POSIX TZ string** (e.g. `EST5EDT`) instead of the IANA name `America/New_York`: glibc parses POSIX TZ strings directly with no tz database, so it always resolves to a non-UTC zone and the test always runs. Verified locally — `TZ=EST5EDT` → `tzname=('EST','EDT')` even with no tzdata.

tests/unit/test_conftest_fixtures.py
mwojtyczka: Thanks for handling the tzdata case. One step further though: a skip means this test is silently sidestepped on a host without tzdata rather than exercising the fix, which we do not want — a green run should mean the fix ran. Using a POSIX TZ string (e.g. EST5EDT) instead of the IANA name removes the tzdata dependency entirely, so no skip is needed and the test always runs. Details and a suggested diff in my new review comment on the fixture.

mwojtyczka (changes requested):

SreeramaYeshwanthGowd:
Thanks @mwojtyczka, switched force_non_utc_start to the POSIX TZ string EST5EDT as suggested, which removes the tzdata dependency and the skip entirely, and fixed the leak that came with it. Also consolidated the tz set and restore logic into one apply_tz helper now.

mwojtyczka (approved):
LGTM

PR #1403, opened 2026-08-01
Add is_valid_language_code row-level check (ISO 639)

### LINKED ISSUE #1385: [FEATURE]: New check for language codes (ISO 639)

### Problem statement

https://en.wikipedia.org/wiki/ISO_639 is a widely used standard for encoding language codes (alpha-2, e.g. "en", and alpha-3, e.g. "eng"). Although we have `is_in_list`, using it for a list this size is tedious. The list is clearly defined so this check can be generated and reused directly.

### Proposed Solution

Add a dedicated check validating a column against the ISO 639 language code list, supporting both the alpha-2 and alpha-3 representations.

### Additional Context

SreeramaYeshwanthGowd:
## Changes

- Add `is_valid_language_code(column, code_format="alpha-2", case_sensitive=True)`, validating a column against ISO 639 language codes: alpha-2 (ISO 639-1, e.g. `en`) or alpha-3 (ISO 639-3, e.g. `eng`).
- Reuses the existing `_is_valid_iso_code`/`_ISO_CODES_BY_STANDARD` shared registry (the same mechanism as `is_valid_country_code`/`is_valid_currency_code`), adding a new `_ISO_639` entry rather than a parallel implementation.
- Packaged code lists at `resources/iso_639_1_alpha_2.txt` (184 codes) and `resources/iso_639_3_alpha_3.txt` (7927 codes), loaded via the existing `load_iso_codes` helper.
- The alpha-3 list was reconciled directly against the live SIL ISO 639-3 registry rather than shipped as a raw pycountry snapshot, since the pinned pycountry version lags SIL by a handful of newly-assigned and retired codes. The freshness test records this exact, intentional divergence as an explicit allow-list (per the existing freshness-test convention for future intentional divergences) so it stays correct going forward and self-corrects once pycountry catches up.
- Legacy ISO 639-2 bibliographic codes that differ from the terminology code (e.g. `ger` for German, vs `deu`) are not accepted; documented in the docstring.

### Linked issues

Fixes #1385

### Tests

- [x] manually tested - live Spark: alpha-2, alpha-3, and case-insensitive behavior all verified directly, including that the 10 reconciled codes behave correctly (7 newly-added SIL codes accepted, 3 retired codes rejected)
- [x] added unit tests
- [x] added integration tests
- [x] added performance tests

Validation performed:

- `make fmt`
- `make lint`
- `make test`
- Live integration tests (serverless): `test_col_is_valid_language_code_alpha_2`, `_alpha_3`, `_case_insensitive` all passed
- Freshness test (`test_iso_639_freshness.py`) passed against pinned pycountry 26.2.16, with the SIL-vs-pycountry divergence independently re-verified against SIL's live registry data

### Documentation and Demos

- [x] added/updated docs


src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1388,6 +1409,49 @@ def is_valid_currency_code(
...
+    Every code still recognized by the registration authorities is accepted, including deprecated
+    *alpha-2* codes not yet withdrawn from circulation (e.g. *sh* for Serbo-Croatian). ISO 639 codes
+    are conventionally lowercase; *case_sensitive* compares against the codes as registered.
+
+    By default the comparison is case-sensitive; pass *case_sensitive* as False to accept values in
+    any case. *code_format* matching itself is case-insensitive (*"ALPHA-3"*/*"Alpha-2"* are also
+    accepted). Null values will pass the check with no violation reported.
+
+    For best performance with large lists in general, prefer the *foreign_key* check function; the
+    *alpha-2* list (184 codes) is small, but the *alpha-3* list (~7930 codes) is large enough that
```
mwojtyczka: Cosmetic: the docstring says the alpha-3 list is *"~7930 codes"*, but the packaged `iso_639_3_alpha_3.txt` is exactly 7927. I would remove the actual numbers from the docstring to reduce maintenance.

tests/resources/all_row_checks.yaml
```diff
@@ -514,3 +514,11 @@
     arguments:
       column: col_currency
       code_format: alphabetic
+
+# is_valid_language_code check
+- criticality: error
+  check:
+    function: is_valid_language_code
+    arguments:
+      column: col_language
```
mwojtyczka, line 523: Note (no change needed): `col_language` here only uses valid values (`en`/`de`), so this all-checks path is a smoke test that the check runs end-to-end rather than a positive/negative assertion — consistent with how `col_country`/`col_currency` are handled. The real positive/negative coverage lives in `tests/integration/test_row_checks.py`.

mwojtyczka (approved):
## Code Review — Add `is_valid_language_code` row-level check (ISO 639)

**Overview.** Adds `is_valid_language_code(column, code_format="alpha-2", case_sensitive=True)` validating against ISO 639-1 (alpha-2, 184 codes) and ISO 639-3 (alpha-3, 7927 codes), as a clean extension of the shared `_is_valid_iso_code` / `_ISO_CODES_BY_STANDARD` registry — no parallel implementation — plus packaged resource files, a freshness test, full unit/integration/perf coverage, and docs.

**This is a well-executed PR.** Approving; the notes below are cosmetic.

### Verification performed
- **Resource files:** alpha-2 = 184, alpha-3 = 7927; both sorted, lowercase, no duplicates.
- **Freshness reconciliation is real and exact:** reproduced the expected set against pycountry 26.2.16 — packaged alpha-2 equals pycountry's set, and packaged alpha-3 equals `(pycountry_alpha3 ∪ 7 SIL-added) − 3 SIL-retired` = 7927, zero missing/stale diff. The 7 added codes (`dyl`…`zhk`) are present and the 3 retired (`yol,mrd,shl`) are absent.
- **Pattern parity:** error messages, `_iso_literals` caching, numeric-format exclusion, and the perf benchmark idiom all match the country/currency siblings.

### Strengths
- **DRY:** extends the registry (`_ISO_639`, `"language"` kind) rather than duplicating logic; the `_IsoStandard` design forces supplying the message noun.
- **Freshness test is a model of its kind:** exact two-directional equality, an explicit allow-list for the *intentional* SIL divergence with a dated rationale, and `getattr(..., None)` so a future pycountry attribute rename surfaces as a readable mismatch rather than `AttributeError`.
- **Docstring** documents the non-obvious semantics: alpha-2 is not a subset of alpha-3, ISO 639-2 bibliographic codes rejected, deprecated-but-live codes (`sh`) accepted.
- **Coverage** is complete: unit (auto-name, format case-insensitivity, all three error paths), integration (both formats incl. the `ger` bibliographic case, case-insensitive, null passthrough), perf benchmark, and both all-checks paths (YAML + classes) plus streaming.

### Correctness / conventions / security
- Null passthrough via SQL null propagation (no `isNotNull()` needed) is correct; case-insensitive path lowercases both sides; `"numeric"` correctly rejected with the shared error (unit-tested).
- `@register_rule("row")` present, returns a `Column` via `make_condition`, full type hints, Google-style docstring. Resource loading via `importlib.resources` (`files()`) — zip-safe.
- No untrusted input, SQL construction, ReDoS surface, or I/O beyond packaged read-only resources.

### Minor suggestions (non-blocking) — see inline

**Verdict: Approve.** Correct, well-tested, faithful to conventions; the freshness reconciliation is verified accurate to the code.

tests/integration/test_row_checks.py
```diff
@@ -2537,6 +2538,116 @@ def test_col_is_valid_currency_code_numeric_int_column(spark):
...
+        [None],
+        [violation("en")],
+        [violation("ger")],
+        [violation("xxx")],
+        [violation("")],
+        [None],
+    ]
+    expected = spark.createDataFrame(checked_data, checked_schema)
+
+    assertDataFrameEqual(actual, expected)
```
mwojtyczka, line 2615: **Suggestion — make the SIL reconciliation self-verifying (non-blocking).** The PR description reports that the 7 SIL-added codes are accepted and the 3 SIL-retired codes are rejected, verified manually. Right now that behavior is only covered *transitively*: the freshness test pins the resource-file contents and this check reads that file. A tiny behavioral assertion would turn the headline reconciliation claim into a regression guard that fails loudly if the resource file (or the loader) ever regresses, independent of the freshness test:

```python
def test_col_is_valid_language_code_reconciled_sil_codes(spark):
    # SIL-assigned codes pycountry 26.2.16 does not yet carry -> must be accepted;
    # SIL-retired codes pycountry still carries -> must be rejected.
    test_df = spark.createDataFrame([["dyl"], ["zhk"], ["yol"], ["shl"]], "a: string")
    actual = test_df.select(is_valid_language_code("a", code_format="alpha-3"))

    def violation(value: str) -> str:
        return f"Value '{value}' in Column 'a' is not a valid ISO 639 language code"

        [[None], [None], [violation("yol")], [violation("shl")]],
        "a_is_not_a_valid_language_code: string",
    )
```

Cheap to add and it converts the "manually tested" line in the PR body into an automated check.

mwojtyczka (approved):
LGTM

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: Fixed in 41a9b242 — dropped the exact counts (both the 184 and the ~7930) from the docstring entirely so it cannot drift from the packaged resource files, which stay the source of truth.

PR #1404, opened 2026-08-01
Add is_valid_subdivision_code row-level check (ISO 3166-2)

### LINKED ISSUE #1390: [FEATURE]: New check for country subdivision codes (ISO 3166-2)

### Problem statement

https://en.wikipedia.org/wiki/ISO_3166-2 is a widely used standard for encoding country subdivisions (states, provinces, regions), e.g. "US-CA" for California. Although we have `is_in_list`, using it for a list this size is tedious. The list is clearly defined so this check can be generated and reused directly.

### Proposed Solution

Add a dedicated check validating a column against the ISO 3166-2 subdivision code list. Subdivision codes are scoped per country, so the check should account for that (e.g. optionally validating consistency against a country column) rather than treating the list as flat.

### Additional Context

_No response_

SreeramaYeshwanthGowd:
## Changes

- Add `is_valid_subdivision_code(column, case_sensitive=True, country_column=None)`, validating a column against ISO 3166-2 country subdivision codes (e.g. `US-CA`, `GB-ENG`, `DE-BY`).
- Unlike ISO 3166-1/ISO 4217, ISO 3166-2 has only one code representation, so this check does not reuse the `_is_valid_iso_code` code_format-branching helper; it has its own dedicated loader (`_iso_3166_2_codes`) built on the same `load_iso_codes` mechanism.
- Packaged code list at `resources/iso_3166_2.txt` (5,046 codes), with a freshness test against pycountry.
- Addresses the per-country scoping the issue called out: every code already embeds its country's ISO 3166-1 alpha-2 prefix, so a plain membership check against the full list already rejects a subdivision suffix paired with the wrong country (e.g. `US-BY` is not a registered code, even though `US` and `BY` are each valid elsewhere in the list).
- Added an optional `country_column` parameter for datasets that store country and subdivision in separate columns, to additionally cross-validate consistency between them. Null-safe in both directions: a null `country_column` value passes (nothing to cross-check), and a null subdivision value passes per the existing convention.

### Linked issues

Fixes #1390

### Tests

- [x] manually tested - live Spark: default (case-sensitive), case-insensitive, and `country_column` behavior all verified directly, including the country-mismatch, invalid-code, and both null-value cases
- [x] added unit tests
- [x] added integration tests
- [x] added performance tests

Validation performed:

- `make fmt`
- `make lint`
- `make test`
- Live integration tests (serverless): `test_col_is_valid_subdivision_code`, `_case_insensitive`, `_with_country_column` all passed
- Freshness test (`test_iso_3166_2_freshness.py`) passed against pinned pycountry 26.2.16

### Documentation and Demos

- [x] added/updated docs


src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1388,6 +1388,106 @@ def is_valid_currency_code(
...
+        condition = code_is_invalid
+        message = F.concat_ws(
+            "",
+            F.lit("Value '"),
+            col_expr.cast("string"),
+            F.lit(f"' in Column '{col_expr_str}' is not a valid ISO 3166-2 subdivision code"),
+        )
+    else:
+        _, country_expr_str, country_expr = get_normalized_column_and_expr(country_column)
+        country_expr_compare = country_expr if case_sensitive else to_lowercase(country_expr)
```
mwojtyczka, line 1495: **`case_sensitive` silently governs the country cross-check too.** With the default `case_sensitive=True`, a valid `column="US-CA"` paired with `country="us"` is flagged as a country mismatch, because `prefix_expr "US" != country_expr_compare "us"` is True. A user may reasonably expect `case_sensitive` to control only the *code's* validity, not the country comparison. It's a defensible design (they opted into case sensitivity), but the coupling is undocumented — at minimum, note it in the docstring / reference table, or consider always case-folding the prefix-vs-country comparison independently of `case_sensitive`.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1388,6 +1388,106 @@ def is_valid_currency_code(
...
+        prefix_expr = F.split(col_expr_compare, "-").getItem(0)
+        # != yields NULL (not True) when either side is NULL, so a null country_column passes.
+        condition = code_is_invalid | (prefix_expr != country_expr_compare)
+        message = F.concat_ws(
+            "",
+            F.lit("Value '"),
+            col_expr.cast("string"),
+            F.lit(f"' in Column '{col_expr_str}' is not a valid ISO 3166-2 subdivision code for country '"),
```
mwojtyczka, line 1503: **Misleading message when an invalid code is paired with a NULL `country_column`.** For `column="US-ZZ"` (invalid), `country=NULL`: `condition = code_is_invalid | (prefix != NULL)` = `True | NULL` = True, so the row is correctly flagged — but `concat_ws` drops the NULL `country_expr`, producing `"...is not a valid ISO 3166-2 subdivision code for country '' in Column 'country'"`. The empty-quotes country reads as if the country were blank rather than absent. This branch is also untested (the country_column test covers null-code and valid-code+null-country, but not invalid-code+null-country). Consider `coalesce`-ing the country value or branching the message when the country is null.

tests/integration/test_row_checks.py
```diff
@@ -2537,6 +2538,109 @@ def test_col_is_valid_currency_code_numeric_int_column(spark):
...
+    checked_data = [
+        [None],
+        [violation("US-CA", "GB")],
+        [violation("US-ZZ", "US")],
+        [None],
+        [None],
+    ]
+    expected = spark.createDataFrame(checked_data, checked_schema)
+
+    assertDataFrameEqual(actual, expected)
```
mwojtyczka, line 2669: **Missing test: `country_column` combined with `case_sensitive=False`.** The three subdivision tests cover default, case-insensitive (no country), and country_column (default case) — but never the combination, which is exactly the lowercased prefix-vs-country path affected by the `case_sensitive` coupling noted on `check_funcs.py`. A regression in that path would pass CI. Worth one small case: e.g. `is_valid_subdivision_code("a", case_sensitive=False, country_column="country")` with `a="us-ca"`, `country="us"` expected to pass.

mwojtyczka (commented):
## Code Review — Add `is_valid_subdivision_code` (ISO 3166-2)

**Overview.** Adds a row-level `is_valid_subdivision_code(column, case_sensitive=True, country_column=None)` check validating ISO 3166-2 subdivision codes (e.g. `US-CA`), with an optional `country_column` cross-check that the subdivision's country prefix matches a separate country column. Packaged resource file + freshness test + unit/integration/perf coverage + docs, following the established `is_valid_country_code`/`currency_code`/`language_code` pattern.

**Well-implemented overall.** Verified: the freshness test matches pycountry 26.2.16 exactly (5046 == 5046), the null-handling in the cross-check is correct (`!=` yields NULL so a null `country_column` passes), and the sibling conventions (`@register_rule("row")`, `make_condition`, lazy cached loaders, `foreign_key` performance note) are all followed. Findings below are correctness-edge / coverage, not crashes — see inline.

**Minor note (non-blocking):** `_iso_3166_2_codes` / `_iso_3166_2_literals` duplicate the shape of the shared `_iso_*_codes_by_format` / `_iso_literals` infra. Because ISO 3166-2 has a single format (no `code_format`) and the `country_column` cross-check is genuinely unique, I don't think forcing it into `_ISO_CODES_BY_STANDARD` / `_is_valid_iso_code` would be cleaner — but be aware null-handling or message-format changes now need to be mirrored across both paths.

**Verdict:** approve-worthy once the inline items are considered; none block merge.

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: I think this is fine. I would just leave a docstring comment

mwojtyczka (changes requested):
Thank you for the PR. Generally looking good. Left a few comments

pyproject.toml
```diff
@@ -147,7 +147,7 @@ test = [
     "fastapi>=0.119.0,<1.0",
     "dbt-core~=1.11.6",
     "dbt-databricks~=1.11.6",
-    "pycountry~=26.2.16",  # used only to test that the packaged ISO 3166-1 and ISO 4217 code lists stay up to date
+    "pycountry~=26.2.16",  # used only to test that the packaged ISO 3166-1, ISO 3166-2, and ISO 4217 code lists stay up to date
```
mwojtyczka: I would drop this comment, no need to maintain this. Developers can check where the pycountry is used in the code

SreeramaYeshwanthGowd:
Thanks @mwojtyczka, addressed all four: documented that case sensitive also governs the country cross check, fixed the message so a null country renders as the word null instead of blank quotes, dropped the pyproject comment, and added the two missing test cases you flagged.

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: Fixed in 8f93f8e0b — a null country now renders as 'null' via F.when(country_expr.isNull(), F.lit("null")) rather than empty quotes. Resolving.

tests/integration/test_row_checks.py
mwojtyczka: Addressed in 8f93f8e0b — test_col_is_valid_subdivision_code_case_insensitive_with_country_column now covers the country_column + case_sensitive=False combination. Resolving.

pyproject.toml
mwojtyczka: You kept the comment and extended it to list ISO 3166-2 in 8f93f8e0b rather than dropping it — that's fine, your call; it was only a nit. Resolving.

pyproject.toml
```diff
@@ -147,7 +147,7 @@ test = [
     "fastapi>=0.119.0,<1.0",
     "dbt-core~=1.11.6",
     "dbt-databricks~=1.11.6",
-    "pycountry~=26.2.16",  # used only to test that the packaged ISO 3166-1, ISO 4217, and ISO 639 code lists stay up to date
+    "pycountry~=26.2.16",  # used only to test that the packaged ISO 3166-1, ISO 3166-2, ISO 4217, and ISO 639 code lists stay up to date
```
mwojtyczka:
```suggestion
    "pycountry~=26.2.16",
```

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: Fixed in 8f93f8e0b — a null country now renders as 'null' via F.when(country_expr.isNull(), F.lit("null")) rather than empty quotes. Resolving.

tests/integration/test_row_checks.py
mwojtyczka: Addressed in 8f93f8e0b — test_col_is_valid_subdivision_code_case_insensitive_with_country_column now covers the country_column + case_sensitive=False combination. Resolving.

pyproject.toml
mwojtyczka: You kept the comment and extended it to list ISO 3166-2 rather than dropping it — that's fine, your call; it was only a nit. Resolving.

mwojtyczka (approved):
LGTM

PR #1405, opened 2026-08-02
Add ref_tables attribute to sql_query check 

### LINKED ISSUE #1224: [FEATURE]: Improve sql_query dataset-level check with reference tables

### Problem statement

To create a dataset-level check that references another table, we need to pass a python dict to the apply_checks_by_metadata() method or pre-join the reference tables into the input DataFrame before applying checks. If we are working with yaml files, you need to touch and modify our code for every input table. 

### Proposed Solution

Create an additional argument for the `sql_query` check so that we can pass our reference tables, just as we do in other checks (such as `foreign_key`, `compare_datas`, etc.), allowing us to define quality checks using only YAML files. This will also make working with parameterized notebooks easier.

### Additional Context

_No response_

STEFANOVIVAS:
## Changes

Add a new ref_tables: dict[str, str] | None = None argument to sql_query, mapping placeholder name → fully-qualified table name (mirrors the shape of ref_dfs, but with table names instead of DataFrames):  

      - criticality: error
        check:
          function: sql_query
          arguments:
            input_placeholder: sensor
            ref_tables:
              sensor_specs: catalog1.schema1.sensor_specs   # resolved via spark.table() at apply time
            merge_columns: [sensor_id]
            query: |
              SELECT sensor_id, MAX(...) AS condition
              FROM {{ sensor }} sensor
              LEFT JOIN {{ sensor_specs }} specs ON sensor.sensor_id = specs.sensor_id
              GROUP BY sensor_id

Implementation (src/databricks/labs/dqx/check_funcs.py):
1. Add ref_tables param to sql_query()'s signature and docstring.
2. In apply(): for each (name, table) in ref_tables, resolve via spark.table(table), then register as a temp view exactly like the existing ref_dfs loop does today.
3. Raise InvalidParameterError if a name appears in both ref_tables and the runtime ref_dfs dict (ambiguity, same style as the existing "both ref_df_name and ref_table provided" check) — and if a ref_tables key collides with input_placeholder.
4. Extend _validate_sql_query_params() to validate ref_tables is a dict[str, str] with non-empty string keys/values when provided.

Tests:
- Unit (tests/unit/test_dataset_checks.py): invalid ref_tables type/entries, name collision with ref_dfs, collision with input_placeholder.
- Integration (tests/integration/test_apply_checks.py, alongside the existing test_apply_checks_with_sql_query_and_ref_df tests): a real table registered via ref_tables alone (no ref_dfs), and a mixed case using both ref_tables and ref_dfs for different names in one query.

Docs (docs/dqx/docs/reference/quality_checks.mdx):
- Update the sql_query argument list (~line 1905) and the "Providing Reference DataFrames programmatically" section (~line 2551) to document ref_tables, with a YAML-only example that needs no apply_checks(ref_dfs=...) call at all.

### Linked issues

Resolves #1224 

### Tests

- [ ] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

### Documentation and Demos

- [x] added/updated demos
- [x] added/updated docs


docs/dqx/docs/reference/quality_checks.mdx
```diff
@@ -2561,6 +2561,31 @@ The reference DataFrames are used in selected Dataset-level checks:
...
+            ref_df_key: catalog1.schema1.ref_table
+          merge_columns:
+            - col1
+            - col2
+          query: |
+            SELECT col1, col2, SUM(col3) = 0 AS condition
+            FROM {{ input_view }} input
+            JOIN {{ ref_df_key }} ref ON input.col1 = ref.ref_col1
+            GROUP BY col1, col2
+    ```
```
ghanse, line 2584: Minor suggestion for clarity:

```suggestion
    ```yaml
    - criticality: error
      check:
        function: sql_query
        arguments:
          input_placeholder: input_view
          ref_tables:
            ref_table_key: catalog1.schema1.ref_table
          merge_columns:
            - col1
            - col2
          query: |
            SELECT col1, col2, SUM(col3) = 0 AS condition
            FROM {{ input_view }} input
            JOIN {{ ref_table_key }} ref ON input.col1 = ref.ref_col1
            GROUP BY col1, col2
    ```
```

docs/dqx/docs/reference/quality_checks.mdx
```diff
@@ -3264,6 +3289,27 @@ The reference DataFrames are used in selected Dataset-level checks:
...
+        "merge_columns": ["col1", "col2"],
+        "query": """
+            SELECT col1, col2, SUM(col3) = 0 AS condition
+            FROM {{ input_view }} input
+            JOIN {{ ref_df_key }} ref ON input.col1 = ref.ref_col1
+            GROUP BY col1, col2
+        """,
+      },
+    )
+    ```
```
ghanse, line 3312: Same applies here:

```suggestion
    ```python
    DQDatasetRule(
      criticality="error",
      check_func=sql_query,
      check_func_kwargs={
        "input_placeholder": "input_view",
        "ref_tables": {"ref_table_key": "catalog1.schema1.ref_table"},
        "merge_columns": ["col1", "col2"],
        "query": """
            SELECT col1, col2, SUM(col3) = 0 AS condition
            FROM {{ input_view }} input
            JOIN {{ ref_table_key }} ref ON input.col1 = ref.ref_col1
            GROUP BY col1, col2
        """,
      },
    )
    ```
```

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -4682,24 +4695,84 @@ def _apply_dataset_level_sql_check(
...
+        ref_tables: Mapping of placeholder name to fully qualified table name, supplied as a check argument.
+        spark: The active SparkSession used to read tables named in *ref_tables*.
+
+    Returns:
+        A single dictionary combining both sources, keyed by placeholder name.
+
+    Raises:
+        InvalidParameterError: If a name is present in both *ref_dfs* and *ref_tables*.
+    """
+    all_ref_dfs = dict(ref_dfs or {})
```
ghanse, line 4716: Any reason to include `dict(...)` here?

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1960,7 +1972,8 @@ def apply(df: DataFrame, spark: SparkSession, ref_dfs: dict[str, DataFrame]) ->
         filtered_df.createOrReplaceTempView(unique_input_view)
         replacements = {input_placeholder: unique_input_view}
 
-        for ref_name, ref_df in (ref_dfs or {}).items():
+        all_ref_dfs = _resolve_sql_query_ref_dfs(ref_dfs, ref_tables, spark)
+        for ref_name, ref_df in all_ref_dfs.items():
```
ghanse, line 1976: This simplifies the implementation, but might make the usage more challenging. Users need to specify and reference the table by its view name rather than referencing the table directly.

I think we should reconsider this.

ghanse (commented):
Looks good overall! Left a few comments.

src/databricks/labs/dqx/check_funcs.py
STEFANOVIVAS: Agree! Maybe it is more suitable to programmatic approach check definition rather than metadata approach.I just want to keep the backward compatibility.

@ghanse and @mwojtyczka upon reviewing the documentation again, I verified that it is already possible to specify the table name directly within the query itself, without needing to reference a DataFrame or view. Therefore, I see that my primary motivation for opening this issue no longer exists. I am not sure what happened, but I apologize.

mwojtyczka (commented):
## Code Review — Add `ref_tables` to the `sql_query` check

**Overview.** Adds an optional `ref_tables` argument to `sql_query`, mapping placeholder names to fully qualified table names that are read via the Spark catalog and merged with the runtime `ref_dfs`. This lets a `sql_query` check reference tables entirely from YAML/JSON metadata without passing DataFrames programmatically. Includes a new `_resolve_sql_query_ref_dfs` helper, validation in `_validate_sql_query_params` (malformed dict, empty keys/values, collision with `input_placeholder`), unit + integration tests, and docs.

**Clean, well-tested PR.** The validation is thorough, the SQL-safety re-check on the resolved query is preserved, and the collision guard against `input_placeholder` is a good catch. No crash bugs. Findings below are edge-consistency and minor cleanups — see inline.

**Inherent timing note (no action needed):** the `ref_tables`/`input_placeholder` collision is validated eagerly at rule-build time, but the `ref_dfs`-vs-`ref_tables` collision can only be caught at apply time (when `ref_dfs` is known). That asymmetry is unavoidable; just flagging that the two collision errors surface at different moments for anyone validating configs up front.

**Verdict:** approve-worthy; the one item worth a real look is the `ref_dfs`/`input_placeholder` shadowing (inline on the apply loop).

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1960,7 +1972,8 @@ def apply(df: DataFrame, spark: SparkSession, ref_dfs: dict[str, DataFrame]) ->
...
             ref_name_unique = f"{ref_name}_{unique_str}"
             ref_df.createOrReplaceTempView(ref_name_unique)
             replacements[ref_name] = ref_name_unique
```
mwojtyczka, line 1979: **A `ref_dfs` key equal to `input_placeholder` silently shadows the input view.** Here `replacements` starts as `{input_placeholder: unique_input_view}`, then this loop overwrites `replacements[ref_name]` for every entry in `all_ref_dfs` — which includes `ref_dfs`. So `apply_checks(df, checks, ref_dfs={"input_view": other_df})` (default placeholder) makes `{{ input_view }}` resolve to `other_df` instead of the actual input — silent wrong results, no error. The PR adds exactly this collision guard for `ref_tables` (`input_placeholder in ref_tables`) but not for the merged `ref_dfs`. This is partly pre-existing (`ref_dfs` predates this PR), but since you're adding the symmetric guard anyway, consider generalizing it: reject `input_placeholder` appearing as a key in the merged `all_ref_dfs` inside `_resolve_sql_query_ref_dfs`.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -4682,24 +4695,84 @@ def _apply_dataset_level_sql_check(
...
+        InvalidParameterError: If a name is present in both *ref_dfs* and *ref_tables*.
+    """
+    all_ref_dfs = dict(ref_dfs or {})
+    for ref_name, ref_table in (ref_tables or {}).items():
+        if ref_name in all_ref_dfs:
+            raise InvalidParameterError(
+                f"Reference '{ref_name}' was provided both in the 'ref_tables' argument and in the "
+                "'ref_dfs' dictionary supplied when applying checks. Please provide only one to avoid ambiguity."
+            )
+        all_ref_dfs[ref_name] = spark.table(ref_table)
```
mwojtyczka, line 4723: **`ref_tables` is resolved via raw `spark.table()`, bypassing the table readers the rest of DQX uses.** Two existing helpers already cover this:

- **`_get_ref_df(...)` (`check_funcs.py:4303`)** — the close sibling, already used by four dataset checks in this same file (`foreign_key` etc. at lines 1899/2647/2783/3199). Same `str -> DataFrame` shape you use here, but its table branch guards with a DQX `InvalidParameterError` instead of letting a bad name surface as a raw Spark `AnalysisException`. Lowest-effort consistency fix: resolve `ref_tables` entries the same way.
- **`read_input_data(spark, input_config)` / `get_reference_dataframes` (`io.py`)** — the richer path the engine uses (`engine.py:914,1045`). `get_reference_dataframes` is literally `{name: read_input_data(spark, cfg) for name, cfg in reference_tables.items()}` over the workflow-level `reference_tables: dict[str, InputConfig]` (`config.py:187`) — i.e. the exact reference-table use case `ref_tables` reinvents. Adopting it would additionally give you `format`, `options`, storage-path, and streaming support, none of which `dict[str, str]` can express.

Recommendation: use `_get_ref_df` so a missing/invalid table raises a DQX error rather than `AnalysisException`.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -4682,24 +4695,84 @@ def _apply_dataset_level_sql_check(
...
     if not is_sql_query_safe(query):
         raise UnsafeSqlQueryError(
             "Provided SQL query is not safe for execution. Please ensure it does not contain any unsafe operations."
         )
 
+    if ref_tables is not None:
+        if not isinstance(ref_tables, dict):
+            raise InvalidParameterError("'ref_tables' must be a dictionary mapping names to table names.")
+
+        invalid_ref_tables = [
```
mwojtyczka, line 4759: Minor simplification: `invalid_ref_tables` materializes a list of `(name, table)` tuples only to test its truthiness. `any(...)` expresses the same validation without allocating the discarded list:

```python
if any(
    not isinstance(k, str) or not k or not isinstance(v, str) or not v
    for k, v in ref_tables.items()
):
    raise InvalidParameterError(...)
```

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -4682,24 +4695,84 @@ def _apply_dataset_level_sql_check(
...
+    Args:
+        ref_dfs: Dictionary of reference DataFrames supplied when applying checks.
+        ref_tables: Mapping of placeholder name to fully qualified table name, supplied as a check argument.
+        spark: The active SparkSession used to read tables named in *ref_tables*.
+
+    Returns:
+        A single dictionary combining both sources, keyed by placeholder name.
+
+    Raises:
+        InvalidParameterError: If a name is present in both *ref_dfs* and *ref_tables*.
```
mwojtyczka, line 4714: Docstring nit: the `Raises` section lists only `InvalidParameterError` (the both-provided collision), but this function also calls `spark.table(ref_table)`, which raises `AnalysisException` for a missing/invalid table. Worth documenting that failure mode so a maintainer isn't surprised by it (ties into the `spark.table()` consistency note above).

mwojtyczka (changes requested):

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: If users want to use dataframes, this change is still useful. It only extends the existing capabilities, without changing it

Yes, you can provide the table directly in the query and use variable substitution to replace e.g. catalog or schema: https://databrickslabs.github.io/dqx/docs/guide/additional_configuration/#defining-default-variables-for-substitution

mwojtyczka:
Closing the PR and the issue: the sql query is already flexible enough. Users can create views if needed or use parametrization

PR #1408, opened 2026-08-03
test: guard public check function parameter order

### LINKED ISSUE #952: [CHORE]: Improve Tests to prevent parameter position breaking-changes

### Problem statement

**Context:**
When adding new optional parameters to existing public API functions, we need to ensure backward compatibility is maintained. New parameters should be added at the end to avoid breaking existing code that uses positional arguments.
**Gap:**
No automated tests verify that function signatures maintain backward compatibility when parameters are added or reordered.

### Proposed Solution

Add unit tests that verify backward compatibility for public API functions when their signatures change:
Positional arguments continue to work as before
Parameters are received in the correct positions
New parameters work correctly with keyword arguments
Use inspect.signature() to assert parameter order

Scope:
1. Applies to all public check functions in check_funcs.py
2. Especially important for frequently-used functions like is_aggr_*, is_in_range, is_unique, etc.
3. Add tests when modifying function signatures

### Additional Context

_No response_

AtomicGlance:
## What changed

- added an explicit parameter-order contract for every core check in `check_funcs.py`
- added coverage that fails when a registered core check is missing from the contract

## Why

Reordering public parameters breaks callers that use positional arguments. This makes those changes visible during review.

Closes #952

## Checks

- 52 focused tests passed
- Black, Ruff, and mypy passed
- changed-file pylint: 10.00/10
- the full unit suite reached 100%; three unrelated Windows/sandbox failures remained (TTL edge case, networked CLI smoke test, and a temporary log-file lock)


tests/unit/test_check_func_signatures.py
```diff
@@ -0,0 +1,149 @@
...
+    "has_no_gaps_per_time_window": ("column", "window_minutes", "group_by", "trailing_gap", "curr_timestamp"),
+    "has_valid_schema": ("expected_schema", "ref_df_name", "ref_table", "columns", "strict", "exclude_columns"),
+    "is_valid_json": ("column",),
+    "has_json_keys": ("column", "keys", "require_all"),
+    "has_valid_json_schema": ("column", "schema"),
+}
+
+
+def test_signature_contract_covers_all_registered_check_functions():
+    core_check_functions = {
```
mwojtyczka: **The coverage test does not actually cover *all* registered checks.** 32 checks are silently excluded: all 31 geo checks plus the PII check

Can you please extend the PR to cover these as well?

tests/unit/test_check_func_signatures.py
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    assert set(EXPECTED_PARAMETER_ORDER) == core_check_functions
+
+
+@pytest.mark.parametrize(("function_name", "expected_order"), EXPECTED_PARAMETER_ORDER.items())
+def test_registered_check_function_parameter_order(function_name: str, expected_order: tuple[str, ...]):
+    check_function = getattr(check_funcs, function_name)
+
+    assert tuple(inspect.signature(check_function).parameters) == expected_order
```
mwojtyczka: **This asserts parameter names/order but not parameter *kind*.** `tuple(inspect.signature(fn).parameters)` yields only names, so a change that keeps names/order but makes a leading parameter keyword-only — e.g. `def is_not_null(*, column)` — still passes here, even though it breaks positional callers (`is_not_null(col)` → `TypeError`). Since the PR's rationale is specifically "reordering public parameters breaks callers that use positional arguments," consider also asserting each parameter's `kind` (e.g. capture `(name, kind)` pairs, or assert the leading params remain `POSITIONAL_OR_KEYWORD`) so keyword-only/positional-only conversions are caught too.

tests/unit/test_check_func_signatures.py
```diff
@@ -0,0 +1,149 @@
+import inspect
+
+import pytest
+
+from databricks.labs.dqx import check_funcs
+from databricks.labs.dqx.rule import CHECK_FUNC_REGISTRY
+
+
```
mwojtyczka, line 13: Minor: the hand-maintained contract guards against *drift* of existing entries well, but for a brand-new check a contributor can add both the function and a matching (equally-wrong) contract entry and both tests still pass — the intended order is never independently captured. 

Please update the agent instrucitons (AGENTS.md) for this.

tests/unit/test_check_func_signatures.py
```diff
@@ -0,0 +1,149 @@
+import inspect
```
mwojtyczka, line 3: Nit: most sibling test files open with a short module docstring summarizing what they cover (e.g. `test_pydantic_serialization.py`, `test_action_destinations_log.py`). A one-liner here ("Parameter-order contract for public check functions; guards positional-arg compatibility — see #952") would match the convention.

mwojtyczka (commented):
Nice, focused guard for #952 — I verified the 51-entry contract matches the real signatures exactly and the coverage assertion matches the current core set. A few notes to make the guarantee match the PR's stated goal ("every core check", "reordering public parameters breaks positional callers"). Details inline.

tests/unit/test_check_func_signatures.py
mwojtyczka: added

mwojtyczka (changes requested):
Thank you for the PR. Generally looking good. I left a few comemnts.
I updated agent instructions so that we verify if new checks are added tot he signature verification.

tests/unit/test_check_func_signatures.py
AtomicGlance: Covered in d6db2a16. The contract now imports the core, geo, and PII modules, includes all 83 registered checks, and compares the expected names with both the registry and the resolved functions.

tests/unit/test_check_func_signatures.py
AtomicGlance: Covered in d6db2a16. The assertion now compares (name, kind) pairs. Positional-or-keyword remains the default, with the existing keyword-only tail of has_no_aggr_outliers recorded explicitly.

mwojtyczka (approved):
LGTM. I pushed a small refactor to resolve the check functions once at import time and reuse the mapping across the parametrized cases

PR #1419, opened 2026-08-04
Treat is_in_list check function values as column expressions, not literals

### LINKED ISSUE #1393: [BUG]: some checks do not support column expressions in values when created from metadata

### Current Behavior

When creating checks from metadata, the `column` param is consistently wrapped in an expression to ensure it is interpreted as a Column when appropriate (e.g. via calls to [_get_column_expr](https://github.com/databrickslabs/dqx/blob/632f094ee59583b6a5674f47fd40245b9f1606cd/src/databricks/labs/dqx/check_funcs.py#L4394)).

For other supplementary/comparison values, most check functions are passing through [get_limit_expr](https://github.com/databrickslabs/dqx/blob/632f094ee59583b6a5674f47fd40245b9f1606cd/src/databricks/labs/dqx/check_funcs.py#L4296) to perform similar automatic resolution (i.e. unless the value is a digit or a datetime, it is interpreted as a column expression, so column references are automatically resolved and literal values must be quoted to suit).

However, the `*_in_list` check functions do not align to this pattern, where if the `allowed` values aren't already Column type, they are interpreted as literals:

- `is_not_null_and_is_in_list`: https://github.com/databrickslabs/dqx/blob/632f094ee59583b6a5674f47fd40245b9f1606cd/src/databricks/labs/dqx/check_funcs.py#L376
- `is_in_list`: https://github.com/databrickslabs/dqx/blob/632f094ee59583b6a5674f47fd40245b9f1606cd/src/databricks/labs/dqx/check_funcs.py#L434
- `is_not_in_list`: https://github.com/databrickslabs/dqx/blob/632f094ee59583b6a5674f47fd40245b9f1606cd/src/databricks/labs/dqx/check_funcs.py#L491

This isn't a major problem when checks are created programmatically (as users can wrap allowed values with `F.col()` or `F.expr()` explicitly to avoid the issue), but when they are created via metadata (e.g. when loaded from yaml definitions), the issue arises where the values are always being interpreted as literals, with no way to force them to be interpreted as column expressions.

The docs also contain a problematic example, e.g. in [here](https://databrickslabs.github.io/dqx/docs/reference/quality_checks/#converting-dqx-classes-to-metadata) it provides a programmatic declaration:
```python
DQRowRule(
    criticality="warn",
    check_func=check_funcs.is_not_null_and_is_in_list,
    column=F.col("c"),
    check_func_kwargs={"allowed": ["a", F.col("d")]},
  )
```
and then provides a declarative metadata equivalent in [here](https://databrickslabs.github.io/dqx/docs/reference/quality_checks/#converting-metadata-to-dqx-classes) as:
```yaml
  - name: c_is_null_or_is_not_in_the_list
    criticality: warn
    check:
      function: is_not_null_and_is_in_list
      arguments:
        column: c
        allowed:
        - a
        - d
```
though due to this bug, these are not equivalent (e.g. the programmatic version correctly will compare column `c` to the column `a` and the column `d`, but the declarative version would compare the column `c` to the literal string `"a"` and the literal string `"d"`)

### Expected Behavior

Allowed values in the `*_in_list` functions should be handled consistently as with other functions - e.g. they are considered a column expression by default, and literal values should require explicit quoting.

### Steps To Reproduce

e.g. define a check in yaml as:

```yaml
- name: example_allowed_column_interpretated_as_literal
  criticality: error
  check:
    function: is_in_list
    arguments:
      column: some_column
      allowed:
        - some_column
```

this would be expected to always pass, as some_column should always match against itself, but instead it ends up being interpreted as 

```python
F.col("some_column").isin(F.lit("some_column"))
```

abhyuday1203:
## Changes
Replace `F.lit(item)` with `get_limit_expr(item)` in `is_in_list`, `is_not_in_list`,
and `is_not_null_and_is_in_list` so that string values in allowed/forbidden lists
are resolved as column expressions by default, consistent with other check functions.

## ⚠️ BREAKING CHANGE!

The `allowed` / `forbidden` arguments of `is_in_list`, `is_not_in_list`, and
`is_not_null_and_is_in_list` are now resolved the same way as the limits of the
comparison checks (via `get_limit_expr`):

- a **bare string** is interpreted as a **column expression** (not a string literal),
- a numeric string such as `"3"` is parsed as a number,
- an ISO-date string such as `"2024-01-01"` is parsed as a date.

To match a **string literal**, the value must now be single-quoted (e.g. `'value'`)
or passed as `F.lit("value")`.

### Migration

Existing checks that relied on bare strings being treated as literals must quote them:

```yaml
# Before (previously matched the literal strings)
- check:
    function: is_in_list
    arguments:
      column: status
      allowed: [ACTIVE, INACTIVE]

# After (quote to keep literal semantics)
- check:
    function: is_in_list
    arguments:
      column: status
      allowed: ["'ACTIVE'", "'INACTIVE'"]
```

```python
# Programmatic API
DQRowRule(check_func=checks.is_in_list, column="status", allowed=["'ACTIVE'", "'INACTIVE'"])
# or
DQRowRule(check_func=checks.is_in_list, column="status", allowed=[F.lit("ACTIVE"), F.lit("INACTIVE")])
```

Unquoted string allowlists now resolve as column references and will either raise an
`AnalysisException` (no such column) or silently compare against another column's value.
Profiler-generated `is_in_list` checks are unaffected — the generator now quotes string
values automatically.

### Linked issues
Closes #1393

### Tests
- [x] manually tested

### Documentation and Demos
- [ ] added/updated demos
- [x] added/updated docs


mwojtyczka:
> All commits in PR should be signed ('git commit -S ...'). See https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

@abhyuday1203 please sign your commits. We cannot accept the commits otherwise: https://databrickslabs.github.io/dqx/docs/dev/contributing/

mwojtyczka (commented):
Thanks for tackling this — the direction is right. Routing `*_in_list` values through `get_limit_expr` makes them consistent with the rest of the comparison-check family (`is_equal_to`, `is_not_equal_to`, `is_not_less_than`, `is_not_greater_than`, `is_in_range`, `is_not_in_range`, `is_data_fresh`, `has_no_outliers`), which already interpret a bare string as a column expression and require quoting for literals. This closes the exact gap #1393 describes: metadata/YAML authors currently have no way to make an allowed value a column reference. Accepting the breaking change here is reasonable.

A few things to address before merge:

**Correctness / edge cases** — `get_limit_expr` doesn't only do "string → column expr"; it also special-cases digit strings (`"3"` → integer `3`) and ISO-date-shaped strings (`"2024-01-01"` → `to_date(...)`). Fine for range/comparison checks, but for `is_in_list` over a string column it means `["3", "2024-01-01"]` silently stop being string-membership values. The PR note doesn't cover these cases.

**Tests (blocking)** — This inverts the default interpretation of string values in three registered check functions with no test updates. Existing tests pass unquoted string literals and will now fail at query time (e.g. `tests/integration/test_row_checks.py` 694/697/701; digit/nested-list resources). Per AGENTS.md, please update the literal-string tests to the new quoting convention and add a positive test proving a bare string now resolves as a column reference.

**Docs (blocking)** — PR self-identifies as breaking with the docs checkbox unchecked. The exact example #1393 called out (`docs/dqx/docs/reference/quality_checks.mdx` allowed lists) still shows unquoted literals. Please update the docs + add a CHANGELOG/migration note covering the new default and the digit/date subtlety.

**Minor** — the identical comprehension is now copy-pasted in all three functions; a small shared helper would keep future edge-case fixes from drifting.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -373,7 +373,7 @@ def is_not_null_and_is_in_list(column: str | Column, allowed: list, case_sensiti
     if not allowed:
         raise InvalidParameterError("allowed list must not be empty.")
 
-    allowed_cols = [item if isinstance(item, Column) else F.lit(item) for item in allowed]
+    allowed_cols = [item if isinstance(item, Column) else get_limit_expr(item) for item in allowed]
```
mwojtyczka: Blocking (tests + docs): this inverts the default meaning of string values. Bare strings in `allowed` now resolve as **column references** (`get_limit_expr("a")` → `F.expr("a")`), not literals — so existing checks/tests that pass unquoted literals (e.g. `is_not_null_and_is_in_list("a", ["str1"])`) break at query time. Edge cases via `get_limit_expr`: digit strings (`"3"` → integer `3`) and ISO-date strings (`"2024-01-01"` → `to_date`) also change. Please update the existing literal-string tests, add a positive test for the new column-ref behavior, and update the docs/CHANGELOG with the quoting requirement.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -431,7 +431,7 @@ def is_in_list(column: str | Column, allowed: list, case_sensitive: bool = True)
...
```
mwojtyczka: Same change in `is_in_list`. Concrete breakage: `tests/integration/test_row_checks.py` passes unquoted literals like `["banned","suspended","deleted"]` and `["admin","root"]`, which now resolve as column references → `AnalysisException` (no such column) or a silent wrong comparison if a same-named column exists. Note the digit case too: `all_row_checks.yaml` uses integer values (fine), but a string `"3"` would become integer `3`. Consider extracting the shared `[item if isinstance(item, Column) else get_limit_expr(item) for item in ...]` into one helper across the three functions.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -488,7 +488,7 @@ def is_not_in_list(column: str | Column, forbidden: list, case_sensitive: bool =
     if not forbidden:
         raise InvalidParameterError("forbidden list must not be empty.")
 
-    forbidden_cols = [item if isinstance(item, Column) else F.lit(item) for item in forbidden]
+    forbidden_cols = [item if isinstance(item, Column) else get_limit_expr(item) for item in forbidden]
```
mwojtyczka: Same for `is_not_in_list` (`forbidden`). This is the most safety-sensitive of the three: a forbidden-list check that silently reinterprets `"banned"` as a column reference could **stop flagging** forbidden values (or error), which is a security-relevant regression for blocklist use cases. Needs the same test + docs coverage.

mwojtyczka (changes requested):

src/databricks/labs/dqx/check_funcs.py
abhyuday1203: Thanks for the review! I've updated the existing literal-string tests to use the new quoting convention, added the requested positive integration test for the new column-reference behavior and I'm working through the docs/CHANGELOG updates.

Before I push the remaining changes, I wanted to validate the new integration test locally. Right now I'm blocked because tests/integration/test_row_checks.py requires a Databricks workspace (DATABRICKS_HOST). I looked through the acceptance workflow, and it seems to rely on the protected tool environment and workspace credentials.

For external contributors, is there a preferred way to run these integration tests locally, or is this something that's expected to be validated by the GitHub acceptance workflow?

src/databricks/labs/dqx/check_funcs.py
abhyuday1203: I've updated the existing literal-string tests and added the requested positive integration test for the new behavior. I also agree that the repeated get_limit_expr conversion should be shared instead of duplicated across the three functions, so I'll clean that up before pushing.

src/databricks/labs/dqx/check_funcs.py
abhyuday1203: So Yeah, I've updated the tests for the new quoting behavior here as well and I'm making the corresponding docs/CHANGELOG updates before pushing.

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: The docs cover how to run the tests by con tributors: https://databrickslabs.github.io/dqx/docs/dev/contributing/#running-integration-tests-and-code-coverage

The CI tests can only be triggered by contriburos. We will do that once you are ready with a follow up review.

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: don't make changelog changes, we are doing this as part of a release process

src/databricks/labs/dqx/check_funcs.py
abhyuday1203: Thanks for the guidance.. I've now completed the requested updates:
**-Updated the existing literal-string integration tests to use the new quoting convention.
-Added a positive integration test covering the new bare-string column-reference behavior.
-Updated the relevant documentation examples to reflect the new quoting requirement.
-Verified the updated unit and integration tests locally.**

src/databricks/labs/dqx/check_funcs.py
abhyuday1203: Done! I extracted the repeated get_limit_expr conversion into a shared _get_limit_exprs helper so the conversion logic is no longer duplicated across the three *_in_list checks.

src/databricks/labs/dqx/check_funcs.py
abhyuday1203: Thanks for the heads up! Understood. I updated the documentation only and left the changelog unchanged.

mwojtyczka (commented):
The quoting change in `check_funcs.py` is fine in isolation, but two upstream **generators** that emit bare-string allowlists were not updated, so features that work today will silently break after this merges:

**Profiler-generated `is_in_list` rules break** — `src/databricks/labs/dqx/profiler/generator.py:263`. `dq_generate_is_in` returns `"allowed": params["in"]` verbatim (bare distinct strings like `['active','inactive','pending']`). After this change those resolve as column references via `get_limit_expr`, so the profile → generate → apply round trip raises `AnalysisException` or silently mismatches. The generator needs to quote its output.

**Data-contract enum constraints break the same way** — the contract generator emits `is_in_list` with bare enum strings (e.g. `['active','inactive','maintenance','faulty']`, asserted in `tests/unit/test_datacontract_generator.py:573`). Those now misbehave as column references; enum values must be quoted too.

Inline comments below cover the remaining points (type coercion, None/empty regression, docs inconsistency, redundant guard).

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -348,6 +348,10 @@ def normalize_segment(segment: Column) -> Column:
     return make_condition(condition, message, f"{col_str_norm}_has_invalid_{case}_string_case")
 
 
+def _get_limit_exprs(values: list[Any]) -> list[Column]:
+    return [item if isinstance(item, Column) else get_limit_expr(item) for item in values]
```
mwojtyczka: **Silent type coercion beyond the documented change.** Routing list items through `get_limit_expr` doesn't only resolve bare strings as column references — it also parses numeric- and ISO-date-looking strings into non-string types. `allowed=['2025-01-01']` against a string column becomes `F.to_date(...)` (a DATE), so `isin` never matches and valid rows get flagged; `'123'` becomes an integer via `F.expr('123')`. The PR note only warns about column-reference resolution, not this coercion — worth documenting explicitly.

mwojtyczka: **Regression for `None` / empty-string entries.** `get_limit_expr(None)` raises `MissingParameterError`, so `allowed=[None, 'x']` now fails at build time (previously produced `F.lit(None)`). `allowed=['']` becomes `F.expr('')`, which fails Spark analysis (previously matched the empty string via `F.lit('')`). Neither case is covered by the added tests — consider a guard or explicit regression tests.

mwojtyczka: **Redundant guard.** `get_limit_expr` already returns `limit` unchanged when `isinstance(limit, Column)`, so `item if isinstance(item, Column) else get_limit_expr(item)` duplicates a branch that must stay in sync with `get_limit_expr`. This simplifies to `return [get_limit_expr(item) for item in values]`.

docs/dqx/docs/reference/quality_checks.mdx
```diff
@@ -4685,8 +4685,8 @@ Converting a list of DQX dictionary representations (metadata format loaded from
           arguments:
             column: c
             allowed:
-            - a
-            - d
+            - "'a'"
+            - "'d'"
```
mwojtyczka: This metadata example was updated to quote the values (`"'a'"` / `"'d'"`), but the paired class-based example above (`check_func_kwargs={"allowed": ["a", F.col("d")]}`, ~line 4637) still passes a bare `'a'`. After this change the class example resolves `'a'` as a column reference — contradicting the new behavior. The two examples are now inconsistent; the class one should use `"'a'"` (or `F.lit("a")`).

src/databricks/labs/dqx/check_funcs.py

docs/dqx/docs/reference/quality_checks.mdx
mwojtyczka: fixed

mwojtyczka (approved):
LGTM

abhyuday1203:
@mwojtyczka Hey Marcin, I was actually still working on the remaining review comments after hours when I saw your new commits. I was working through the generator and test changes you pointed out. Thanks for taking care of the remaining pieces and getting it to LGTM!

PR #1422, opened 2026-08-04
Decouple AI rule generation from Spark session

### LINKED ISSUE #1095: [FEATURE]: Refactor AI-Assisted Rules Generation to remove dependency on Spark

### Problem statement

Currently AI-assisted rules generation depends on spark session to fetch table properties

### Proposed Solution

Use workspace client instead of spark to fetch table metadata.

### Additional Context

After that refactor the App to use dqx core in the app.

ghanse:
## Changes

This PR decouples AI-assisted rule generation from Spark to allow use without an active Spark session. A small `get_table_column_metadata` gets column metadata for any Unity Catalog tables using the Databricks SDK.

Spark sessions are lazily created when required (e.g. for reading data from file paths or inferring primary keys using AI).

### Linked issues

Resolves #1095 

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

### Documentation and Demos

- [ ] added/updated demos
- [x] added/updated docs


src/databricks/labs/dqx/io.py
```diff
@@ -15,6 +15,7 @@
 STORAGE_PATH_PATTERN = re.compile(r"^(/|s3:/|abfss:/|gs:/)")
 # catalog.schema.table or schema.table or database.table (backticks allow special chars like hyphens)
 TABLE_PATTERN = re.compile(r"^(?:(?:`[^`]+`|[a-zA-Z0-9_]+)\.)?(?:`[^`]+`|[a-zA-Z0-9_]+)\.(?:`[^`]+`|[a-zA-Z0-9_]+)$")
+UC_TABLE_PATTERN = re.compile(r"^[a-zA-Z0-9_]+\.[a-zA-Z0-9_]+\.[a-zA-Z0-9_]+$")
```
mwojtyczka: `UC_TABLE_PATTERN` only accepts `[a-zA-Z0-9_]`, so it rejects valid backtick-quoted 3-level UC tables (e.g. `` `my-catalog`.schema.table `` or ``main.`my-schema`.tbl``). Those are legitimate UC tables the `tables.get` API can resolve, but they'll fail this match and fall through to the Spark path in `generator._get_schema_info` — re-introducing the Spark session the PR is trying to avoid. The existing `TABLE_PATTERN` right above already handles backticks; consider reusing/extending that logic (or stripping backticks) so special-char UC names still take the SDK path. The new `test_uc_table_pattern_rejects_non_uc_locations` currently codifies this as intended — worth reconsidering.

src/databricks/labs/dqx/utils.py
```diff
@@ -680,6 +680,30 @@ def get_table_primary_keys(table: str, spark: Any) -> set[str]:
...
+        table: Fully qualified table name (e.g. *catalog.schema.table*).
+
+    Returns:
+        A JSON string containing the column metadata with columns wrapped in a "columns" key.
+
+    Raises:
+        NotFound: If the table does not exist or is not accessible.
+    """
+    table_info = workspace_client.tables.get(table)
+    columns = [{"name": col.name or "", "type": col.type_text or ""} for col in (table_info.columns or [])]
```
mwojtyczka: `col.type_text` is the UC DDL type text, which can differ in case/format from Spark's `field.dataType.simpleString()` used by `get_column_metadata` (e.g. nested types like `ARRAY<INT>`/`STRUCT<...>` vs `array<int>`/`struct<...>`). The docstring says the two produce the same shape and are 'interchangeable', but the *type values* can differ for the same table depending on which path runs, making the LLM prompt context (and thus generated rules) non-deterministic across UC vs Spark. Consider normalizing (e.g. lowercase, or map to Spark's simpleString form) so both paths emit identical type strings.

mwojtyczka (commented):
Nice refactor — decoupling rule generation from Spark and reading UC schemas via the SDK is the right shape, and splitting PK detection into `DQLLMPrimaryKeyEngine` cleanly isolates the one path that genuinely needs Spark. Good test coverage too. A few things worth addressing before merge (most are correctness/consistency around the UC-vs-Spark routing):

- **Special-char UC tables silently force Spark** (`UC_TABLE_PATTERN`). A valid backtick-quoted 3-level table (`` `my-catalog`.schema.table ``) is rejected by the pattern and falls through to the Spark path — re-introducing the Spark dependency the PR removes, for exactly the hyphenated/special-char names UC allows. The existing `TABLE_PATTERN` already handles backticks.
- **`type_text` vs `simpleString()` divergence**: the SDK and Spark paths format types differently, so the docstring claim that they're interchangeable isn't quite true — same table, different LLM context.
- **Routing by omission**: non-UC locations (two-level HMS names, views, invalid strings) fall to Spark implicitly rather than being explicitly classified/validated.
- **`DQLLMPrimaryKeyEngine.__init__` eagerly builds Spark**, inconsistent with the lazy-property pattern introduced for `DQGenerator`.
- Minor: duplicated configurator/dspy-context boilerplate across the two engines; and one PK test asserts an internal `final_status` value rather than the documented result contract.

Details inline.

src/databricks/labs/dqx/profiler/generator.py
```diff
@@ -158,6 +174,23 @@ def generate_dq_rules_ai_assisted(
...
+        Unity Catalog tables are resolved through a Databricks workspace client. Other locations
+        (e.g. storage paths, files, or temporary views) are read using a Spark session.
+
+        Args:
+            input_config: Input config providing the input data location.
+
+        Returns:
+            A JSON string containing the column metadata with columns wrapped in a "columns" key.
+        """
+        if UC_TABLE_PATTERN.match(input_config.location):
```
mwojtyczka: Routing is by omission: only `UC_TABLE_PATTERN` is checked, and everything else (two-level HMS name like `default.users`, a bare view name, or a malformed location) silently goes to the Spark path. For a caller expecting Spark-free operation, a two-level table unexpectedly spins up a session; a truly invalid location defers to `read_input_data`'s generic `InvalidConfigError` rather than being validated here. Consider explicitly classifying the location (UC 3-level → SDK; storage path / 2-level table → Spark; otherwise raise a clear error) instead of an implicit else-fallback.

src/databricks/labs/dqx/llm/llm_pk_engine.py
```diff
@@ -0,0 +1,63 @@
...
+    ):
+        """
+        Initializes the primary key detection engine.
+
+        Args:
+            model_config: Configuration for the LLM model.
+            spark: Optional Spark session. If None, a new session is created.
+            detector: Optional primary key detector. If None, one is created using *spark*.
+        """
+        self.spark = SparkSession.builder.getOrCreate() if spark is None else spark
```
mwojtyczka: `__init__` eagerly calls `SparkSession.builder.getOrCreate()`, so merely constructing `DQLLMPrimaryKeyEngine` requires Spark — inconsistent with the lazy `spark` property you introduced for `DQGenerator` in this same PR. Since the detector can be injected (and the metadata classification path may not need a live session), consider making Spark lazy here too, so construction doesn't force a session before any data scan actually happens.

src/databricks/labs/dqx/llm/llm_pk_engine.py
```diff
@@ -0,0 +1,63 @@
...
+        self._configurator = LLMModelConfigurator(model_config)
```
mwojtyczka, line 38: Minor (DRY): `self._configurator = LLMModelConfigurator(model_config)` plus the `with dspy.settings.context(lm=self._configurator.create_lm())` wrapper are now duplicated between `DQLLMEngine` and `DQLLMPrimaryKeyEngine`. The split was to isolate Spark, but the LM-configuration mechanics are copy-pasted; a small shared base/mixin would keep token-handling changes from having to be applied in two places.

tests/unit/test_llm_pk_engine.py
```diff
@@ -0,0 +1,54 @@
...
+
+def test_primary_key_engine_reports_metadata_failure_without_raising():
+    """An unusable Spark session surfaces as a failed result, not an exception."""
+    engine = DQLLMPrimaryKeyEngine(model_config=LLMModelConfig(model_name="test-model"), spark=Mock())
+
+    result = engine.detect_primary_keys_with_llm("catalog.schema.orders")
+
+    assert result["success"] is False
+    assert result["table"] == "catalog.schema.orders"
+    assert result["final_status"] == "metadata_error"
```
mwojtyczka: This asserts `result['final_status'] == 'metadata_error'`, which couples the test to `LLMPrimaryKeyDetector`'s internal failure taxonomy rather than the engine's documented contract (return a failed result with `success`/`table`, don't raise). If the detector's error classification is refactored, this breaks even though the engine behavior is unchanged. Consider asserting on the documented keys (`success is False`, `table == ...`) and dropping the internal `final_status` check.

mwojtyczka (changes requested):

mwojtyczka (approved):
LGTM

PR #1428, opened 2026-08-05
feat(app): DQX Studio dqlake integration

### LINKED ISSUE #1296: [BUG]: sql_query cross-table rule generated by DQX Studio fails with `UNRESOLVED_COLUMN: condition`

### Current Behavior

The Cross-table rules tab in DQX Studio generates a `sql_query` check payload that is missing two properties required by the `sql_query` engine function. Running the generated payload without modification fails immediately with an `UNRESOLVED_COLUMN` error.

### Expected Behavior

The `sql_query` engine function requires **both** of the following regardless of mode. This needs to be added in the DQX Cross-table rules tab:

1. A boolean `condition` column in the query result (`True = violation`). This is hardcoded via `F.col(condition_column)` (defaulting to `"condition"`) on both the row-level and dataset-level code paths. The HAVING pattern returns only violating rows with no `condition` column — neither code path handles this shape.

2. `merge_columns` in the arguments for row-level checks. Without it, the engine falls back to dataset-level mode, which still requires a `condition` column and changes semantics so that all rows are flagged together rather than individually.

### Steps To Reproduce

1. Open DQX Studio → **Cross-table rules** tab.
2. Enter or generate a cross-table SQL check using the HAVING pattern (e.g., the placeholder example).
3. Export/copy the generated check payload.
4. Execute via `DQEngine.apply_checks_by_metadata(df, checks)`.

### Generated payload (broken)

```python
{
    "check": {
        "function": "sql_query",
        "arguments": {
            "query": (
                "SELECT o.order_id, o.total, SUM(li.amount) AS line_total "
                "FROM catalog.schema.orders o "
                "JOIN catalog.schema.line_items li ON li.order_id = o.order_id "
                "GROUP BY o.order_id, o.total "
                "HAVING o.total != line_total"
            )
        }
    },
    "name": "orders_total_matches_line_items",
    "criticality": "warn"
}
```

## Workaround A — Row-level (recommended, preserves per-row granularity)

Add `TRUE AS condition` to the SELECT and add `merge_columns` to arguments. Rows absent from the HAVING result receive `NULL` from the left join, which Spark treats as a pass.

```python
{
    "function": "sql_query",
    "arguments": {
        "query": (
            "SELECT o.order_id, TRUE AS condition "
            "FROM catalog.schema.orders o "
            "JOIN catalog.schema.line_items li ON li.order_id = o.order_id "
            "GROUP BY o.order_id, o.total "
            "HAVING o.total != SUM(li.amount)"
        ),
        "merge_columns": ["order_id"]
    }
}
```

## Workaround B — Dataset-level (all-or-nothing aggregate)

Collapse to a single aggregate row with an explicit boolean condition. No `merge_columns` needed, but all rows are flagged together.

```python
{
    "function": "sql_query",
    "arguments": {
        "query": (
            "SELECT (ROUND(SUM(iv.total), 2) != ROUND(SUM(li.amount), 2)) AS condition "
            "FROM {{ input_view }} iv "
            "JOIN catalog.schema.line_items li ON li.order_id = iv.order_id"
        )
    }
}
```

## Proposed Fix

The UI should be updated to do at least one of the following:

- Update the SQL query placeholder and hint text to use the `(expr) AS condition` pattern instead of HAVING.
- Expose a `merge_columns` field in the form for row-level checks.
- Alternatively, wrap the user-supplied HAVING query in an outer query that adds the `condition` column automatically before serializing the payload, e.g.:

```sql
SELECT key_col,
       (key_col IN (SELECT key_col FROM (<user_query>))) AS condition
FROM {{ input_view }}
```

## Environment

- DQX Studio version: `v0.14.0`
- `databricks-labs-dqx` version: *(confirm from `pip show databricks-labs-dqx`)*

### Cloud

AWS

### Operating System

Windows

```shell
[UNRESOLVED_COLUMN.WITH_SUGGESTION] A column, variable, or function parameter
with name `condition` cannot be resolved. Did you mean one of the following?
[`o.total`, `line_total`, `o.order_id`]. SQLSTATE: 42703
```

berrybluecode:
## Summary

This branch brings **DQX Studio** from the thinner Studio shell on `main` to the full dqlake-parity product (~84 non-merge commits, ~757 files). Relative to `main`, it adds the governed rule lifecycle, monitored-table / collection workflows, equal-rule-weight scoring, Results/Genie, RBAC, Marketplace, demo seed, and a complete Studio docs section.

### Platform & governance
- **Rules Registry** — reusable rules with draft → pending → approved lifecycle, versioning, revoke, import/export, and ODCS/contract import
- **Monitored tables & Collections (Table Spaces)** — apply rules, pin versions, schedules, approval queues, steward ownership
- **RBAC & object permissions** — roles, entitlements, object-level grants, steward grant UX
- **Lakebase / BYO warehouse** — Postgres OLTP path updates, external SQL warehouse support, deploy/ops docs

### Quality results
- **Equal-rule-weight DQ score** — row-level pass rates vs table-level binary verdicts; score cache + UC metric views
- **Results UI** — table/collection/rule/global breakdowns, quarantine samples, SQL-pushdown failed-row filters
- **Ask Genie** — Genie space provisioning + in-app chat over results

### Authoring & AI
- Built-in / visual builder / custom SQL (including **cross-table** and **row vs table-level**)
- AI rule suggestion, AI SQL write/improve (with granularity context), native rule tests
- Apply-on-tag (governed tags), pass thresholds, Labels/tags fixes
- **Rules Marketplace** — curated reusable packs with import + already-added UX

### Docs, demo & polish
- New **Studio docs** (start-here, authoring, monitoring, running, governance, what’s new) including quality-score and cross-table guides
- Built-in **demo seed** content and many Studio UX/perf/bug-bash fixes

## Resolves

#1296 

## Test plan
- [x] Fresh deploy from this branch (Lakebase + warehouse); app starts, migrations apply
- [x] Rules Registry: create (built-in / visual / SQL), submit, approve, revoke, edit-in-place revision, import/export
- [x] Cross-table authoring (FK, visual join, SQL row-level + table-level); apply and run on a monitored table
- [x] AI suggest + AI SQL Write/Improve with Applies-to toggle; 
- [x] Monitored table + Collection: apply rules, schedule, Run now, Drafts & Review approve/reject/revoke
- [x] Results: DQ score for mixed row/table-level rules; breakdowns; failed-record drill-down; Ask Genie
- [x] Permissions: role gating + object grants / steward; Marketplace pack import
- [x] Smoke Studio docs locally (`docs/dqx`) — especially quality-score, cross-table, permissions
- [x] Rebase/merge `main` and resolve conflicts (branch is ~88 ahead / ~30 behind)


docs/dqx/docs/guide/data_profiling.mdx
```diff
@@ -16,7 +15,6 @@ The generated data quality rules (checks) candidates can be used as input for th
 In addition, the Lakeflow (DLT) generator can generate native Lakeflow Pipelines expectations.
 
 You can run profiling and quality checks generation using the following approaches:
```
mwojtyczka, line 18: why did we remove this?

docs/dqx/docs/guide/best_practices.mdx
```diff
@@ -127,7 +127,6 @@ For "gold" layer tables, you may want to define different sets of rules dependin
 ### Workflows vs Embedded Usage
 
 DQX offers different execution modes. Choose the one that best fits your use case:
-* [DQX Studio](/docs/guide/dqx_studio) for a no-code experience (**recommended for no-code**). A browser-based UI to author, review, run, and monitor quality rules on already persisted data. Easiest to adopt for less technical users, with built-in approval workflows and run history.
```
mwojtyczka, line 130: Why is this removed, i find this useful

docs/dqx/docs/guide/index.mdx
```diff
@@ -23,25 +25,20 @@ For more details, see the [Installation Guide](/docs/installation/).
 | Quality checking type | Integration with processing pipelines | Description                                                                                                                                                                                                    |
 |---------------------- | ------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | **In-transit**        | Code-level only                       | DQX allows data quality to be validated on the fly while the data is being processed, before it is written to storage. This requires DQX to be used as a library and integrated directly into user pipelines.  |
-| **At-rest**           | Code-level or No-code (DQX Studio or Workflows) | DQX enables data quality checking on existing data stored in tables. For a no-code experience, [DQX Studio](/docs/guide/dqx_studio) (a Databricks App) is recommended; alternatively, install DQX in the workspace as a tool to deploy the no-code workflows.                                      |
```
mwojtyczka, line 26: why are we removing it. We need a guidance on when to use what

docs/dqx/docs/guide/quality_checks_apply.mdx
```diff
@@ -6,12 +6,10 @@ import Admonition from '@theme/Admonition';
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 import useBaseUrl from '@docusaurus/useBaseUrl';
-import { AvailableSinceVersion, FeatureTags } from '@site/src/components/FeatureTags';
 
 # Applying Quality Checks
 
 You can apply quality checks to your data using the following approaches:
```
mwojtyczka, line 13: Same as before

docs/dqx/docs/studio/whats-new/index.mdx
```diff
@@ -0,0 +1,118 @@
+---
+sidebar_position: 1
+title: What's new
+---
+
+# What's new in DQX Studio
```
mwojtyczka: Please remove whats-new. We are using versionining now (by tagging pages and sections in the docs) to descibe what has changed and in which version instead of maintaining separate pages

docs/superpowers/notes/dqlake-reference.md
```diff
@@ -0,0 +1,96 @@
+# dqlake reference — components to port into DQX admin settings
```
mwojtyczka: Why do we have to keep the md files from superwpoers?

mwojtyczka (commented):
Reviewed with a **scoped** pass focused on the highest-risk hand-written backend (low-code SQL compiler, registry lifecycle, scheduler, permissions/RBAC, results/scoring). Given the size (~198k additions across 757 files, incl. a 21k-line generated `api.ts` and 173 UI `.tsx`), a full line-by-line review of everything in one sitting isn't feasible — **the UI, generated clients, docs, and lockfiles were not reviewed here.**

The SQL-safety discipline across the backend is genuinely strong: `is_sql_query_safe` + `UnsafeSqlQueryError` are applied widely, demo SQL uses `quote_ident`/`escape_sql_string` behind a `_require_known_table` allowlist, and `permissions_service` deliberately matches on verified `principal_id` (not the spoofable SCIM display name). `dq_results_service` is pure aggregation with correct division guards. No confirmed authorization bypass found.

**Findings worth addressing (details inline):**
1. Low-code compiler emits dotted column names and join table/column identifiers **raw**; `is_sql_query_safe` allows SELECT/subqueries, so a crafted identifier can smuggle a subquery past the gate (authenticated + self-scoped, so medium — but identifiers should be quoted/validated).
2. `approve()` does a read-modify-write on `version` with no optimistic lock — concurrent approvals collide and corrupt the `dq_rule_versions` snapshots the materializer resolves against.
3. LIKE operators don't escape `%`/`_`, so `contains '100%'` matches unintended rows — a correctness bug in the compiled predicate.
4. `seed_builtin_rule` skips the `_validate_definition_sql_safety` gate the other write paths apply (low severity today; defense-in-depth gap).
5. Unguarded `row[1]` in score-run collection degrades the score-cache refresh on an unexpected row shape.

**Process note (not a code finding):** an 88-commit, 757-file PR is very hard to review safely as one unit — splitting future Studio increments (e.g. registry, results/scoring, scheduler, RBAC as separate PRs) would materially improve review quality and reduce merge risk.

app/src/databricks_labs_dqx_app/backend/lowcode_compile.py
```diff
@@ -0,0 +1,691 @@
...
+    with the ``{{input_view}}`` marker so it is unambiguous against a joined table
+    that shares the column name: ``{{input_view}}.{{col}}``. The library
+    substitutes ``{{input_view}}`` with the unique input-view name and the app
+    materializer substitutes ``{{col}}`` with the real column, yielding
+    ``<view>.<col>``. Only valid inside the QUERY TEXT (predicate / join ON), never
+    for ``merge_columns`` (which the library passes verbatim to a bare-name join).
+    Joined-table (dotted) columns are already qualified and stay raw.
+    """
+    if "." in column:
+        return column
```
mwojtyczka: `_ref` returns any column containing `.` **raw** (verbatim), and `compile_joins_to_sql` similarly inserts `target_table` / `joined_column` raw (lines ~516, ~525). The only downstream defense is `is_sql_query_safe`, which by design blocks destructive keywords (DROP/DELETE/…) but **allows SELECT and subqueries**. So an authenticated rule author whose visual-builder AST carries a joined-table name or dotted `column_ref` like `` t WHERE 1=1 UNION SELECT secret FROM other -- `` (or a correlated subquery) compiles to a body that passes the safety gate and runs at materialization against whatever the run principal can read. Scope is limited (authenticated, self-authored rule, self-scoped grants), but consider validating identifiers (snake_case / `quote_ident`) rather than emitting them raw — defense in depth against the SELECT-allowing gate.

app/src/databricks_labs_dqx_app/backend/services/registry_service.py
```diff
@@ -0,0 +1,1240 @@
...
+        """
+        rule = self._get(rule_id)
+        if rule is None:
+            raise RuntimeError(f"Registry rule not found: {rule_id}")
+        self._check_transition(rule.status, "approved")
+        prev_status = rule.status
+        rule.status = "approved"
+        rule.version += 1
+        rule.updated_by = user_email
+        self._update(rule)
```
mwojtyczka: `approve()` reads `rule.version` (via `_get` above), does `rule.version += 1`, then `_update` + `_write_version_snapshot`, with no compare-and-set on the current version/status. Two approvers acting on the same pending rule concurrently both read N, both write N+1 and both emit a vN+1 snapshot; the rule ends at N+1 instead of N+2 and `dq_rule_versions` holds duplicate/conflicting N+1 rows, which the materializer's frozen-snapshot resolution (`_get_current_snapshots`, matched on `(rule_id, version)`) then reads ambiguously. Consider a conditional update (`WHERE version = N AND status = 'pending_approval'`) or a row lock so the increment is atomic.

app/src/databricks_labs_dqx_app/backend/lowcode_compile.py
```diff
@@ -0,0 +1,691 @@
...
+def _row_sql(left: str, operator: str, value: object, qualify: bool = False) -> str:
+    op = operator
+    if op in _COMPARISON_OPS:
+        return f"{left} {op} {_value_sql(value, qualify)}"
+    if op == "equals":
+        return f"{left} = {_value_sql(value, qualify)}"
+    if op == "not equals":
+        return f"{left} != {_value_sql(value, qualify)}"
+    if op == "contains":
+        return f"{left} LIKE '%{_like_literal(value)}%'"
```
mwojtyczka: `contains` compiles to `{left} LIKE '%{_like_literal(value)}%'`, and `_like_literal` (line ~244) only doubles single quotes — it does **not** escape the LIKE metacharacters `%` and `_`. So a rule authored as `column contains '100%'` becomes `LIKE '%100%%'`, whose trailing `%` is a wildcard, silently matching values it shouldn't (e.g. `'1000abc'') and mis-classifying rows as pass/fail. Same applies to `does not contain` / `starts with` / `ends with`. Consider escaping `%`, `_` (and `\`) in `_like_literal` and appending an `ESCAPE '\'` clause.

app/src/databricks_labs_dqx_app/backend/services/registry_service.py
```diff
@@ -0,0 +1,1240 @@
...
+            steward=steward,
+            is_builtin=True,
+            source="builtin",
+            created_by=user_email,
+            created_at=now,
+            updated_by=user_email,
+            updated_at=now,
+        )
+        rule.fingerprint = compute_registry_rule_fingerprint(rule)
+        self._insert(rule)
```
mwojtyczka, line 682: `seed_builtin_rule` inserts directly at `status='approved'` without calling `_validate_definition_sql_safety`, which both `create_rule` and `update_draft` apply before persisting. Low severity today (seed source is trusted, `mode='dqx_native'`), but if a seed definition ever carries an unsafe SQL body it lands as an approved rule and only fails later at materialization (`render_check`), leaving a permanently-unusable approved row. Worth running the same gate here for consistency/defense-in-depth.

app/src/databricks_labs_dqx_app/backend/services/scheduler_service.py
```diff
@@ -344,6 +636,915 @@ def _advance_after_failure(self, name: str, cfg: dict[str, Any], now: datetime,
...
+        )
+        rows = self._sql.query(sql)
+
+        fqns: set[str] = set()
+        for row in rows:
+            run_id = row[0] if row else None
+            if not run_id:
+                continue
+            self._pending_score_runs.pop(run_id, None)
+            fqn = row[1]
```
mwojtyczka: `row[0]` is guarded (`if not run_id: continue`) but `row[1]` is then read with no `len(row) > 1` / None check. A query result row shaped unexpectedly (fewer columns, or NULL fqn) raises IndexError/TypeError and aborts the score-cache refresh tick. It's caught by the outer handler and retried, so impact is a transient failed refresh rather than a crash — but it silently degrades score freshness. A length/None guard (mirroring the `run_id` check) would make it robust.

app/src/databricks_labs_dqx_app/backend/lowcode_compile.py
```diff
@@ -0,0 +1,691 @@
...
+    if op == "does not contain":
+        return f"{left} NOT LIKE '%{_like_literal(value)}%'"
+    if op == "starts with":
+        return f"{left} LIKE '{_like_literal(value)}%'"
+    if op == "ends with":
+        return f"{left} LIKE '%{_like_literal(value)}'"
```
mwojtyczka: **Fix for the LIKE `%`/`_` escaping bug.** Using Spark's `contains`/`startswith`/`endswith` builtins takes the operand as a plain string literal (quoted/escaped by `_quote`), so `%` and `_` are treated literally and there's no `ESCAPE`-clause subtlety.

⚠️ This module is a byte-for-byte port of `ui/lib/lowcodeCompile.ts` — please mirror the same change there so the client and server compilers stay in lock-step (and add/adjust the unit test for `contains '100%'`).

```suggestion
    if op == "contains":
        return f"contains({left}, {_quote(value)})"
    if op == "does not contain":
        return f"NOT contains({left}, {_quote(value)})"
    if op == "starts with":
        return f"startswith({left}, {_quote(value)})"
    if op == "ends with":
        return f"endswith({left}, {_quote(value)})"
```

mwojtyczka (commented):
Adding the two **confirmed** findings as one-click suggestions. The other three findings from my earlier review are design calls (identifier quoting, `approve()` locking) that are yours to make, so I've left those as discussion.

app/src/databricks_labs_dqx_app/backend/services/registry_service.py
```diff
@@ -0,0 +1,1240 @@
...
+            user_metadata=dict(user_metadata or {}),
+            steward=steward,
+            is_builtin=True,
+            source="builtin",
+            created_by=user_email,
+            created_at=now,
+            updated_by=user_email,
+            updated_at=now,
+        )
+        rule.fingerprint = compute_registry_rule_fingerprint(rule)
```
mwojtyczka, line 681: **Fix for the missing SQL-safety gate on seed.** Run the same `_validate_definition_sql_safety` check the `create_rule`/`update_draft` write paths apply, so a seeded built-in rule can never persist an unsafe body at `status='approved'`.

```suggestion
        self._validate_definition_sql_safety(rule.mode, rule.definition)
        rule.fingerprint = compute_registry_rule_fingerprint(rule)
```

app/src/databricks_labs_dqx_app/backend/routes/v1/monitored_tables.py
```diff
@@ -0,0 +1,1525 @@
...
+# ------------------------------------------------------------------
+# Profiling (READ-ONLY — reuses dq_profiling_results, never writes here)
+# ------------------------------------------------------------------
+
+
+@router.get(
+    "/{binding_id}/profile",
+    response_model=MonitoredTableProfileOut,
+    operation_id="getMonitoredTableProfile",
+    dependencies=[require_role(*_ALL_ROLES)],
```
mwojtyczka, line 501: **Possible IDOR — confirm intent.** This endpoint (and `GET /{binding_id}/versions/{version}/checks` @545, `GET /{binding_id}/pending-applications` @750) gate on `require_role(*_ALL_ROLES)` only, with no `perms.require_object(ObjectType.MONITORED_TABLE.value, binding_id, ...)` in the body. Every *mutating* sibling on the same binding does call `require_object` — delete@377, owner@416, schedule@458, run@602, applied-rules@667/821/905/943/981. So an authenticated low-privilege user can `GET` the profiling result, the frozen `checks_json` (column mappings + rule logic), and staged pending applications for **any** `binding_id`, including tables/collections they hold no object grant on.

If the ACL model intends object grants to govern **reads** too, these three need the same `require_object` (with a read/`VIEW` privilege). If reads are deliberately role-gated only (object grants govern writes), this is by design — but the asymmetry with the mutation siblings is worth an explicit decision either way.

app/src/databricks_labs_dqx_app/backend/services/data_product_service.py
```diff
@@ -0,0 +1,1107 @@
...
+                cast(ScheduleKind, row[12])
+                if len(row) > 12 and row[12] in get_args(ScheduleKind)
+                else SCHEDULE_KIND_DEFAULT
+            ),
+            steward_display_name=row[13] if len(row) > 13 else None,
+        )
+
+    @staticmethod
+    def _opt_str(value: str | None) -> str:
+        return f"'{escape_sql_string(value)}'" if value else "NULL"
```
mwojtyczka: **SQL string break-out via unvalidated free text.** `_opt_str` (here) and the `name` interpolation at line 332 / `_assert_name_available` at 943 embed user-supplied `name` / `description` / `steward_display_name` via `escape_sql_string`. Per this repo's own `sql_utils` contract, `escape_sql_string` doubles single quotes but **deliberately does not escape backslashes**, and is only safe for values already validated by `validate_fqn`/`validate_object_id`. These are unvalidated free text (create() only checks non-empty), so a value ending in `\` turns the doubled closing `''` into an escaped quote on the Delta string-literal path and breaks out. `escape_sql_string_strict` is the backslash-safe variant that exists for exactly this. Suggestion switches `_opt_str`:

```suggestion
    def _opt_str(value: str | None) -> str:
        return f"'{escape_sql_string_strict(value)}'" if value else "NULL"
```

Also update the two direct `name` sites (line 332 `escape_sql_string(product.name)` and line 943 `escape_sql_string(name)`) to `escape_sql_string_strict`, and add `escape_sql_string_strict` to the `sql_utils` import at line 72. (`product_id` stays on the plain variant — it's `validate_object_id`-guarded.)

app/src/databricks_labs_dqx_app/backend/services/data_product_service.py
```diff
@@ -0,0 +1,1107 @@
...
+            LookupError: *product_id* does not exist.
+            RuntimeError: *binding_id* does not exist (when adding a new member).
+            BindingNotApprovedError: the binding is not approved (when adding
+                a new member) — maps to HTTP 400 at the route.
+        """
+        if self._fetch_product(product_id) is None:
+            raise LookupError(f"Data product not found: {product_id}")
+        e_pid = escape_sql_string(product_id)
+        e_bid = escape_sql_string(binding_id)
+        rows = self._sql.query(
```
mwojtyczka: **add_member TOCTOU.** This SELECT-then-INSERT (`else` branch ~469-480) has no unique constraint or lock behind it, so two concurrent `add_member` calls for the same `(product_id, binding_id)` both see zero rows and both INSERT, duplicating the membership — which then makes `member_table_fqns()` fan the same table to the scheduler twice, and violates the upsert-by-binding_id contract in the docstring (line 424). A `UNIQUE (product_id, binding_id)` constraint (catch the violation as a no-op) or a row lock would make it correct.

app/src/databricks_labs_dqx_app/backend/services/data_product_service.py
```diff
@@ -0,0 +1,1107 @@
...
+        """Delete a data product and its members.
+
+        Raises:
+            LookupError: *product_id* does not exist.
+        """
+        e = escape_sql_string(product_id)
+        rows = self._sql.query(f"SELECT product_id FROM {self._products_table} WHERE product_id = '{e}'")  # noqa: S608
+        if not rows:
+            raise LookupError(f"Data product not found: {product_id}")
+        self._sql.execute(f"DELETE FROM {self._members_table} WHERE product_id = '{e}'")  # noqa: S608
```
mwojtyczka: **Non-transactional cascade delete.** The member delete (413) and product delete (414) are separate statements with no surrounding transaction; a crash/connection drop between them leaves member rows whose parent product is gone. Those orphans are skipped defensively elsewhere but never cleaned up, so they accumulate in `dq_data_product_members`. Wrap both in a transaction, or rely on a DB `ON DELETE CASCADE`.

app/src/databricks_labs_dqx_app/backend/services/genie_space_service.py
```diff
@@ -0,0 +1,1789 @@
...
+        return True
+    except Exception as e:
+        # Best-effort resilience contract: a failed PATCH must never break
+        # startup — the existing space still answers with its old config and
+        # the un-persisted hash makes the next startup retry.
+        logger.warning(f"Genie space update skipped: {type(e).__name__}: {e}")
+        return False
+
+
+def ensure_dq_genie_space(
```
mwojtyczka, line 1697: **Defense-in-depth nit (not injectable).** `catalog`/`schema` flow from config through here into ~50 `quote_object_fqn(...)` sites. This is **safe** — `quote_ident` doubles embedded backticks, so a crafted value stays inert identifier text (a `'; DROP...` inside backticks is not injection). The only improvement is to `validate_identifier(catalog)` / `validate_identifier(schema)` at this entry point so a misconfigured value fails fast with a clear error instead of emitting ~50 malformed-but-inert statements. Low priority; noting for completeness since a parallel scan flagged it as injection (it isn't).

mwojtyczka (commented):
Second, deeper pass — this time the execution/enforcement layers (materializer, `sql_utils`, the routes RBAC layer, and the data-product / genie / monitored-table services). Overall the SQL-safety foundation holds up very well: `sql_utils` is carefully reasoned (the backslash / `escape_sql_string_strict` split, quote-aware comment stripping), the **materializer re-runs `is_sql_query_safe` on the SUBSTITUTED text** (so slot substitution can't re-introduce a forbidden keyword — this closes the re-injection angle on my earlier low-code finding), `dq_results` validates FQNs and strict-escapes failed-row facets, and `monitored_table_service` came back clean.

New findings below (details inline). The two worth blocking on:

1. **Possible IDOR** — three binding-scoped **read** endpoints on `monitored_tables` gate on role only and skip the `perms.require_object(...)` check that *every* mutating sibling on the same binding enforces. If object-level read scoping is intended, a low-privilege user can read another binding's profile / frozen checks / staged applications. Flagged as **confirm-intent** rather than definitely-a-bug — it depends on whether object grants are meant to govern reads or only writes.
2. **SQL string break-out via data-product free-text** — `name` / `description` / `steward_display_name` are unvalidated user text embedded via `escape_sql_string`, which by this repo's own contract does NOT escape backslashes and is only safe for validated values. A trailing `\` breaks out of the literal. Suggestion inline switches these to `escape_sql_string_strict`.

Plus two lower-severity data-product items (add_member TOCTOU duplicate; non-transactional cascade delete) and a genie defense-in-depth nit (catalog/schema unvalidated at entry — NOT injectable, since `quote_ident` doubles backticks).

mwojtyczka (changes requested):
Generally going in the right direction. I left some comments. 

Addional general comments:
* Simplify the documentation and reduce the number of screenshots so that we are able to maintain it in the future
* Remove superpower md files
* Don't remove existing references to DQX studio
* Mark DQX Studio as Beta in docs
* Remove `whats-new` section, and move it to PR description. We are using versioning now. Given this is first Beta version we don't need to communicate changes

OGordon100:
## DQX Studio — What's new

_These notes previously lived on the in-docs "What's new" page (now removed). Captured here for the release notes._

### Redesigned app & new documentation
- **A ground-up app redesign** — cleaner layout and consistent patterns across authoring, monitoring, results, and governance.
- **Substantially faster and more fluid** — lists, tables, and results load quicker and navigation feels smoother.
- **New documentation** — DQX is now documented as two tracks: DQX Core (the Python package) and DQX Studio (the no-code app).

### Authoring
- **Rules Registry** — a library of reusable, versioned rule definitions applied across many tables; each rule can become many checks.
- **Low-code rule builder** — author rules visually: pick a built-in check or write a custom condition with a plain-language builder, AND/OR chaining, filters, and joins.
- **Suggest rules with AI** — point the app at a table and it proposes a tailored, reviewable set of checks.
- **Profiling** — scan a table to measure the shape of its data and get suggested candidate rules.
- **Severity & dimensions** — every rule gets a customizable Low/Medium/High/Critical priority and a quality dimension.
- **Rule versioning** — rules are versioned; each check can track the latest approved version or be pinned.
- **Import & export** — import from DQX YAML or ODCS data contracts, bulk-import across many tables, and export back out.
- **Marketplace** — browse a catalogue of open-source pre-built rule packs, filter by industry/region, try before importing.

### Monitoring
- **Monitored tables** — apply rules to a Unity Catalog table with per-table severity/threshold overrides and bulk registration.
- **Table overview & sample data** — schema, sample-data preview, and an ask-a-question box.
- **Quality thresholds** — require a share of rows to pass (e.g. 95%) with per-rule/per-column overrides and clear breach indicators.
- **Governed tags & auto-mapping** — tag a rule's column with a Unity Catalog governed tag and it applies automatically to any column carrying that tag.
- **Collections (data products)** — group related monitored tables into a product you can view, run, and schedule together.

### Running & reviewing
- **Runs & scheduling** — run on demand or hourly/daily/weekly/custom, with pause/resume and a full runs history.
- **Results dashboard** — read the DQ score and trend, break down by dimension/severity/rule/column, and drill to failing records.
- **Ask Genie** — ask questions about results in plain English and get an answer with a chart.
- **Comments** — discuss and hand off findings with a comment thread on results.

### Governance & security
- **Roles & entitlements** — map Databricks workspace groups to Viewer / Rule Author / Rule Approver / Admin.
- **Four-eyes approval** — draft → submit → review → approve, with a diff and comments.
- **Object-level permissions** — View, Modify, Apply, and Execute on a specific rule, table, or collection.
- **Data access respects Unity Catalog** — the app reads as you; you only see what you already have access to.

### Setup & administration
- **Admin Settings** — one place to manage entitlements, governed tags, data retention, the AI endpoint, and the built-in demo.
- **Built-in demo** — deploy sample tables, rules, collections, and past runs to explore every feature against realistic data.

app/src/databricks_labs_dqx_app/ui/CLAUDE.md
ghanse, line 1: We aligned on using `AGENTS.md` with a thin `CLAUDE.md` that links there, I think it's better practice.

app/CLAUDE.md
ghanse, line 1: We aligned on using `AGENTS.md` with a thin `CLAUDE.md` that links there, I think it's better practice.

ghanse (changes requested):
Simple suggestion:

We should add some small updates to `AGENTS.md`, `docs_authoring`, or other contributor guidelines about contributing to DQX Studio. We now have separate documentation, code, issue tags, etc.

app/src/databricks_labs_dqx_app/backend/services/score_service.py
```diff
@@ -0,0 +1,64 @@
...
+        if input_row_count <= 0 or not check_metrics:
+            return None
+        dataset_names = set(dataset_check_names)
+        rule_scores: dict[str, list[float]] = defaultdict(list)
+        for metric in check_metrics:
+            failed = metric.error_count + metric.warning_count
+            if metric.check_name in dataset_names:
+                check_score = 0.0 if failed > 0 else 1.0
+            else:
+                check_score = 1.0 - failed / input_row_count
```
mwojtyczka: **Row-check score is never clamped to `[0, 1]`** (low confidence).

`check_score = 1.0 - failed / input_row_count` can go negative when a check's aggregated `failed` count (`error_count + warning_count`) exceeds the run's `input_row_count`. That's possible if two physical checks collapse to one `check_name` in the breakdown, or a filtered run's `input_row_count` is smaller than the captured failure count. A negative `check_score` drags the rule mean, table score, and `compute_product_score` mean below 0 — a value the UI/metric-view present as a 0..1 rate. It would also break parity with the SQL metric view if that side clamps and this doesn't.

Suggest clamping, e.g. `check_score = max(0.0, 1.0 - failed / input_row_count)`.

mwojtyczka (commented):
## Code review — automated pass

**Scope.** This PR is large (666 files, ~190k lines). Only one file touches DQX Core (`contract_rules_generator.py`); the rest is the DQX Studio app under `app/`. A full line-by-line pass over all ~200 code files wasn't feasible, so this reviewed the highest-signal surfaces: the core change, the SQL-safety layer, SQL-pushdown row filters, the scheduler/cron + retention logic, the scoring math, RBAC/permissions, the low-code SQL compiler, demo datagen SQL, and i18n locale sync.

**Verified sound (not bugs).**
- SQL-injection layer (`validate_fqn`/`quote_ident`/`escape_sql_string`, new `strip_sql_line_comments`) — the comment stripper can only over-reject, never hide a forbidden keyword from `is_sql_query_safe`.
- RBAC — `RULE_APPROVER` bypasses object grants, but the run route is independently gated by `require_role(*CAN_RUN_ROLES)`, so an approver can't escalate to running rules.
- Core `contract_rules_generator.py` — provenance keys are filtered out before `user_metadata.update()`; no double-apply on owner resolution.
- i18n — all 5 locales have identical 2807-key sets.
- Retention sweep, product/table score None-filtering, and demo SQL escaping — all sound.

**One finding (low-confidence robustness gap):** unclamped row-check score in `score_service.py` — see inline comment.

No solid correctness bug was found within the reviewed subset. The next place to look would be the ~180 unreviewed app files: large services (`genie_space_service.py`, `materializer.py`, `registry_service.py`, `models.py`) and the Postgres migrations.

<sub>Automated review via Isaac Review / code-review.</sub>

.superpowers/sdd/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
+!.gitignore
```
mwojtyczka: remove superpowers

app/scripts/post_deploy_lakebase_migrations.py
```diff
@@ -0,0 +1,117 @@
...
+(the same identity that ran ``databricks bundle deploy``). When the deployer
+owns the schema objects, pending migrations apply cleanly; the subsequent app
+startup ``run_all()`` is then a no-op.
+
+Usage (from ``app/``):
+
+    uv run python scripts/post_deploy_lakebase_migrations.py -p <profile> -t <target>
+"""
+
+from __future__ import annotations
```
mwojtyczka: Please remove all `from __future__ import annotations` from the code

app/src/databricks_labs_dqx_app/backend/services/vector_store.py
```diff
@@ -0,0 +1,264 @@
+"""Vector Search auto-provisioning — Rules Registry Phase 7F.
```
laurencewells: Can we remove Vector Store, adds more infra complexity and costs for customers
Recommendation. Pull from this PR and look at PG Vector or In RAM fuzzy match

app/src/databricks_labs_dqx_app/backend/services/score_service.py
mwojtyczka: fixed

.superpowers/sdd/.gitignore
mwojtyczka: fixed

app/scripts/post_deploy_lakebase_migrations.py
mwojtyczka: fixed

mwojtyczka:
## Review comments — resolution summary

Went through the review comments on this PR. Most were **already resolved by later commits on the branch** (verified directly against the current code); a few needed **new fixes**, pushed in `c36a0391`.

### Newly fixed (`c36a0391`)

- **DQ score clamp** — `score_service.py`: row-check score is now clamped to `[0, 1]` (`max(0.0, 1.0 - failed / input_row_count)`), so an aggregated failure count exceeding `input_row_count` can no longer drive the rule / table / product scores negative.
- **Dead code** — `lowcode_compile.py`: removed the unused `_like_literal` (string matching already uses the Spark `contains`/`startswith`/`endswith` builtins).
- **`from __future__ import annotations`** removed from `build_app.py`, `post_deploy_lakebase_migrations.py`, `test_build_app.py` (repo policy — keep annotations eager).
- **Superpowers scratch** removed (`.superpowers/sdd/.gitignore`).
- Added regression tests for the score clamp and for the already-fixed findings below.

### Already resolved on the branch (verified, no change needed)

- **`registry_service.py` approve() race** → `_update_approve_cas` performs a conditional update (`WHERE version = N AND status = …`) + re-read, raising on a concurrent modification.
- **`registry_service.py` seed SQL-safety** → `seed_builtin_rule` now calls `_validate_definition_sql_safety`, same as `create_rule`/`update_draft`.
- **`lowcode_compile.py` LIKE `%`/`_`** → uses Spark builtins; dotted identifiers validated via `_quote_ident_path` / `validate_identifier`.
- **`data_product_service.py`** → free text escaped with `escape_sql_string_strict`; `add_member` is race-safe (`UNIQUE (product_id, binding_id)` + conflict-catch + re-read of the winner's id); `delete` runs in a transaction on Postgres (member-first order on Delta).
- **`scheduler_service.py` row guard** → the score-refresh loop already length-guards `row[1]`/`row[2]`.
- **`genie_space_service.py`** → `validate_identifier(catalog)` / `validate_identifier(schema)` at the entry point (fail-fast; the ~50 quoting sites were already injection-safe).
- **Docs** → `whats-new/` already removed (versioning is done via page/section tags); the "why was this removed?" / "need guidance on when to use what" items are present in the current guide pages.
- **IDOR (`monitored_tables.py`)** → the three flagged GET reads (`/profile`, `/versions/{version}/checks`, `/pending-applications`) now all call `require_object(..., Privilege.SELECT)`.

### Open decision — `monitored_tables.py` reads

A few reads remain role-gated only (no `require_object`): `GET /{binding_id}` (base detail), `/{binding_id}/versions`, `/{binding_id}/tag-suggestions`, `/{binding_id}/profile/suggestions`. Worth a conscious call on whether these should also be object-grant gated (`SELECT`) — the base `GET /{binding_id}` is the direct mirror of the mutating siblings that *are* gated.

### Note

The `add_member` UPSERT refactor considered here was **reverted** — the committed code was already race-safe, and the refactor introduced a primary-key overwrite regression on concurrent conflicts.

<sub>Summary posted after addressing the review feedback; fixes in commit `c36a0391`.</sub>

mwojtyczka (approved):
LGTM - pending remval of Vector Store

app/src/databricks_labs_dqx_app/backend/services/rule_embeddings.py
```diff
@@ -0,0 +1,273 @@
...
+                return False
+            self._upsert(rule.rule_id, rule.version, text, embedding)
+        except Exception:
+            logger.warning("Failed to embed rule %s (non-fatal)", rule.rule_id, exc_info=True)
+            return False
+
+        self._upsert_vector_search_index(rule.rule_id, text, embedding)
+        return True
+
+    def iter_embeddings(self) -> list[tuple[str, list[float]]]:
```
vb-dbrks, line 255: `iter_embeddings()` re-runs the select and re-parses every stored vector from json on each call, and `_retrieve_per_column` calls it once per column - so a wide table does the full read + parse once per column inside a single request.

`get_rule_embeddings_service` is a per-request `Depends` factory, so memoising the parsed corpus on the instance would be per-request by construction - no cross-request staleness to reason about. writes would need to clear it so a publish in the same request doesn't read a stale copy.

hoisting this out of the per-column loop looks like the single biggest win here, and it's independent of the VS question.

app/src/databricks_labs_dqx_app/backend/services/rule_suggester.py
```diff
@@ -0,0 +1,656 @@
...
+
+        Falls back to a single table-level query when the column list is empty
+        (no profile yet) so behaviour degrades to the old path rather than
+        returning nothing.
+        """
+        if not columns:
+            return self._retriever.retrieve(self._build_query_text(table_fqn, columns), self._top_k)
+
+        best_by_rule: dict[str, RetrievedRule] = {}
+        for column in columns:
```
vb-dbrks: `_resolve_columns` returns every UC column with no cap, and each iteration's `retrieve()` makes its own `embed_text()` call - so embedding round trips scale 1:1 with table width. on a wide table that's the dominant latency rather than the scan.

the per-column retrieval itself looks right, the docstring reasoning about column-specific rules missing a global top-K is convincing. it's the round trips that look avoidable - `serving_endpoints.query` takes `input` as an array and the response is a list of elements carrying `index`, so several query texts can go per call. `embed_text` returns on the first element in `data`, so today that shape isn't used.

would bound it rather than just collapse it though - FMAPI endpoints have per-request payload and token limits, so one unbounded batch would fail on a very wide table where this currently just crawls. chunked batches, or a cap on how many columns take part in retrieval, would both fix the scaling without swapping one unbounded thing for another.

app/src/databricks_labs_dqx_app/backend/services/rule_embeddings.py
```diff
@@ -0,0 +1,273 @@
...
+            value_cols={
+                "rule_version": rule_version,
+                "embed_text": text,
+                "embedding": json.dumps(embedding),
+                "model": model,
+                "updated_at": RawSql("current_timestamp()"),
+            },
+        )
+
+    def _upsert_vector_search_index(self, rule_id: str, text: str, embedding: list[float]) -> None:
```
vb-dbrks: +1 to @laurencewells on removing the vector store - one extra bit to catch when it goes: publish currently writes to the VS index as well as postgres, so otherwise we're left with a write path that has no reader.

app/src/databricks_labs_dqx_app/backend/dependencies.py
```diff
@@ -255,13 +372,284 @@ async def get_rules_catalog_service(
...
+        read_columns=_build_column_reader(obo_ws, obo_sql),
+    )
+
+
+async def get_rule_embeddings_service(
+    sql: Annotated[OltpExecutorProtocol, Depends(get_sp_oltp_executor)],
+    sp_ws: Annotated[WorkspaceClient, Depends(get_sp_ws)],
+    app_settings: Annotated[AppSettingsService, Depends(get_app_settings_service)],
+) -> RuleEmbeddingsService:
+    """Create a RuleEmbeddingsService routed at the OLTP executor + SP credentials.
```
vb-dbrks: this says the SP client is used "mirroring the `AIGateway` split-auth rationale", but `AIGateway`'s docstring says the opposite for its own calls - every model call runs with the caller's OBO client, "never the app's service principal", because generation is user-facing and request-scoped. query embedding sits on exactly that path, so the citation points the other way.

two things worth picking deliberately rather than inheriting - SP embedding calls don't go through the kill switch or the per-user rate limit `AIGateway` enforces, and they're what makes the serving endpoint ACL grant in `vector_store._grant_serving_endpoint_access` necessary at all.

if SP is deliberate, could the docstring say why in its own terms? startup backfill genuinely has no user in scope which would justify it, but that argument doesn't extend to per-query embedding.

app/src/databricks_labs_dqx_app/backend/services/rule_suggester.py
```diff
@@ -0,0 +1,656 @@
...
+        candidate_rules: list[RegistryRule],
+        columns: list[ColumnMeta],
+        table_fqn: str,
+        user_email: str,
+    ) -> list[dict[str, Any]]:
+        candidates_payload = [
+            {
+                "rule_id": rule.rule_id,
+                "name": get_rule_name(rule.user_metadata) or rule.rule_id,
+                "description": get_rule_description(rule.user_metadata) or "",
```
vb-dbrks, line 613: published rule descriptions and column comments both reach the judge prompt so they're an injection surface. looks contained - `_post_process` re-validates every mapping against the real columns and declared slots, so worst case is a plausible but wrong suggestion rather than anything unsafe. that reasoning is only derivable by reading both methods though - a line in the class docstring would save the next reader the trip.

app/src/databricks_labs_dqx_app/backend/services/rule_suggester.py
```diff
@@ -0,0 +1,656 @@
...
+from databricks_labs_dqx_app.backend.services.tag_mapping_service import family_for_type as _family_for_type
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_TOP_K = 20  # Per-column retrieval unions per column; cap = max(top_k, top_k*3) in _retrieve_per_column
+
+# NL "describe a rule" match path: retrieve against the user prompt (not
+# column-built queries), then run the same mapping judge for staging.
+DEFAULT_MATCH_TOP_K = 5
+MIN_MATCH_SCORE = 0.45  # Cosine similarity floor for a hit to count as a match
```
vb-dbrks: this floor is specific to whichever embedding model is configured, and it only governs `match_from_query` rather than the per-column `suggest()` path. worth saying both out loud so it doesn't get re-tuned blind or assumed to apply everywhere.

```suggestion
MIN_MATCH_SCORE = 0.45  # Cosine floor for match_from_query hits only; re-tune per embedding model
```

app/src/databricks_labs_dqx_app/backend/services/rule_embeddings.py
```diff
@@ -0,0 +1,273 @@
...
+        Best-effort by construction: a read failure or a malformed stored
+        vector never raises — it is logged and skipped so retrieval degrades
+        to "no candidates" rather than a 500.
+
+        Returns:
+            A list of ``(rule_id, embedding)`` tuples. Rows with an empty,
+            non-list, or unparseable embedding are omitted.
+        """
+        try:
+            rows = self._sql.query_dicts(f"SELECT rule_id, embedding FROM {self._table}")  # noqa: S608
```
vb-dbrks: AGENTS.md asks us not to silence lints. this one looks genuinely safe - the interpolated name comes from `sql.fqn`, never user input - so it's the missing justification rather than the suppression that stands out.

```suggestion
            # Table name comes from sql.fqn (internal), never user input.
            rows = self._sql.query_dicts(f"SELECT rule_id, embedding FROM {self._table}")  # noqa: S608
```

vb-dbrks (changes requested):
had a look at the AI/retrieval path rather than the whole PR - the rules registry suggester specifically. nothing here contradicts the approval, it's mostly perf plus one docstring thing.

main one is the per-column fan-out in the suggester - each column triggers its own embedding call and its own full read+parse of `dq_rule_embeddings`, so cost scales with table width inside a single request. that's independent of the vector search discussion and imo the biggest win available here.

rest is smaller - the dual write to the VS index wants removing alongside the provisioner, and the SP vs OBO rationale in `dependencies.py` contradicts what `AIGateway` documents for itself.

app/tests/test_rule_suggester.py
```diff
@@ -0,0 +1,739 @@
+"""Tests for ``services/rule_suggester.py`` — Rules Registry Phase 4C.
+
+Exercises the full pipeline with fakes (no Databricks, no Vector Search, no
```
vb-dbrks, line 3: coverage here is genuinely good... 38 tests across the degradation paths, and TestPostProcessing properly treats the judge output as untrusted (mapping to a nonexistent column dropped, multi slot completeness enforced, unknown rule_id ignored). no complaints on any of that.

the gap is that the judge is a fake returning a canned response, so what's proven is "if the model says X, we handle X correctly". nothing establishes that a real embedding endpoint plus a real judge actually suggests sensible rules for a real table.

practical consequence is there's no baseline... if someone tunes the system prompt, swaps the embedding model, or changes top_k, we can't tell whether suggestions got better or worse, the suite stays green either way. the system prompt in particular isn't asserted on anywhere, so it can be edited freely with nothing catching it.

would a small golden set be worth adding? even 20ish tables with hand labelled expected rules, run against a real endpoint behind an integration marker, would give precision and recall numbers and make every later change to the suggester measurable rather than a judgement call. happy to help put the fixture together if that's useful.

app/src/databricks_labs_dqx_app/backend/dependencies.py
berrybluecode: Fixed with the stronger option — split auth, not just a docstring tweak.

Query embeds (embed_texts / retrieval): use the caller’s OBO client when present (same identity model as AIGateway)
Writes / backfill (embed_and_store): always use the SP (no user on startup/publish)
Docstring in get_rule_embeddings_service rewritten so it no longer cites AIGateway incorrectly
Startup/demo still construct with SP only (user_ws=None); request DI injects both.

ghanse (commented):
This looks good overall, a few findings below:

### SQL Parameterization and Escaping
Parameterized SQL statements are built using Python f-strings and called through `_sql.execute(...) and `_sql.query(...)`. Any future safety mechanisms we introduce will require updates across every f-string.

We could add a small query-builder to handle the common CRUD statements. For example:

```
# current:
self._sql.execute(
  f"UPDATE {self._table} SET pinned_version = {value} WHERE id = '{e}'"
)

# suggested:
self._sql.update(  # builds and executes the UPDATE statement
  self._table, set={"pinned_version": value}, where={"id": rule_id}
)
```

### Code Quality
A few minor code quality nits. Verify if these need fixing:

- There are several very large files with >1,200 lines of code
- There are several instances of linter silencing
- There are a lot of `except Exception` blocks that only log

app/tests/test_rule_suggester.py
vb-dbrks: resolved now with the new eval harness

app/src/databricks_labs_dqx_app/backend/services/rule_suggester.py
```diff
@@ -0,0 +1,686 @@
...
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_TOP_K = 20  # Per-column retrieval unions per column; cap = max(top_k, top_k*3) in _retrieve_per_column
+
+# Bound how many columns get their own retrieval query. Embedding round trips
+# (even batched) and prompt size still grow with width; beyond this we keep
+# UC order and drop the rest from retrieval only — the judge still sees the
+# full column list for mapping.
+MAX_RETRIEVAL_COLUMNS = 64
```
vb-dbrks: I measured what this costs us. On the 90 column table in the eval fixture, raising this to 90 takes recall from 0.750 to 1.000 and does not cost precision. Reproduced across two runs.

It also does not protect the thing it looks like it protects. This caps the retrieval queries, but the judge prompt still receives every column, so the max_tokens exhaustion I described in the review happens regardless of what this is set to.

Given embedding calls are batched now, I think the original reason for the cap is largely gone. I would raise it a lot or drop it.

app/src/databricks_labs_dqx_app/backend/services/rule_suggester.py
```diff
@@ -0,0 +1,686 @@
...
+                "description": get_rule_description(rule.user_metadata) or "",
+                "dimension": get_rule_dimension(rule.user_metadata),
+                "severity": get_rule_severity(rule.user_metadata),
+                "slots": [{"name": slot.name, "family": slot.family} for slot in rule.definition.slots],
+            }
+            for rule in candidate_rules
+        ]
+        columns_payload = [
+            {"name": column.name, "type": column.type, "family": column.family, "comment": column.comment}
+            for column in columns
```
vb-dbrks, line 622: This is where the prompt becomes unbounded. Every candidate rule is offered against every column, so the response grows with table width while max_tokens stays fixed at 2048. Past about fifty columns it truncates mid JSON and the user gets nothing.

Bounding what goes in here is what fixes the truncation properly. Raising max_tokens just moves the cliff further out.

app/src/databricks_labs_dqx_app/backend/services/rule_suggester.py
```diff
@@ -0,0 +1,686 @@
...
+            )
+            retrieval_columns = columns[:MAX_RETRIEVAL_COLUMNS]
+
+        query_texts = [self._build_column_query_text(table_fqn, column) for column in retrieval_columns]
+        best_by_rule: dict[str, RetrievedRule] = {}
+        for hits in self._retriever.retrieve_many(query_texts, self._top_k):
+            for hit in hits:
+                existing = best_by_rule.get(hit.rule_id)
+                if existing is None or hit.score > existing.score:
+                    best_by_rule[hit.rule_id] = hit
```
vb-dbrks, line 572: This is the one point from my earlier review that is still open, and I have changed my mind about why it matters.

We build a query per column, so right here we know which column earned which rule. Then we key the union on rule_id, drop the column, and the judge re-derives the mapping from the full column list.

I originally argued this would cause mis mapping. Having built the eval I do not think that holds ... the judge maps columns well, recall was 1.000 on six of eight tables and the stronger models took none of the seven wrong column decoys. So I am withdrawing that concern.

What I would raise instead is that this is exactly the information we would need to bound the judge prompt. If RetrievedRule carried the column, we could offer the judge the pairs retrieval actually found rather than the full cross product, which bounds the prompt by construction. Not asking for it in this PR, but it is the clean fix for the truncation rather than a bigger budget.

vb-dbrks (changes requested):
I built a golden set eval for the rule suggester and ran it live on field-eng against real endpoints. Numbers below. Short version is that I think the default model choice is a problem, and there is a bug behind it that means some users get nothing at all.

This replaces my earlier review ... everything from that one was addressed, thanks for the quick turnaround.

## The default is broken on wide tables

AI_ENDPOINT_NAME_DEFAULT is databricks-gpt-5-4-nano, and get_ai_enabled returns True when unset, so a fresh deploy has AI on and pre-pointed at nano.

On a 90 column table nano returns zero suggestions. available comes back False with "AI judge returned an unparsable response". Reproduced three times out of three.

The cause is that the judge's output size scales with column count while max_tokens stays at a fixed 2048. I truncated a synthetic table to different widths to find where it breaks:

| columns | 10 | 20 | 30 | 45 | 64 | 90 |
|---|---|---|---|---|---|---|
| completion_tokens | 1174 | 1599 | 1344 | 1831 | 2048 capped | 2048 capped |
| usable | yes | yes | yes | yes | **NO** | **NO** |

So anything past roughly fifty columns loses everything. With thousands of tables in a real estate I don't think 100 column event and feature tables are an edge case.

It is also mis-reported. AIGateway only raises its budget specific error when the content is empty and finish_reason says length. This endpoint returns non empty truncated JSON and leaves finish_reason unset, so we fall through to the generic parse error and the user is told the response was unparsable rather than that the answer did not fit. That sends whoever debugs it in the wrong direction.

## And it is not just nano

I scored six candidate judges against 48 labelled answers, embeddings on gte-large-en throughout:

| judge | precision | recall | f1 | suggestions | decoys taken | 90 col table | secs |
|---|---|---|---|---|---|---|---|
| gpt-5-4-nano (current default) | 0.456 | 0.854 | 0.594 | 90 | 3 of 7 | **FAILS** | 45 |
| gpt-5-4-mini | | | | | | **FAILS** | |
| gpt-5-4 | 0.478 | 0.896 | 0.623 | 90 | 0 | ok | 55 |
| claude-haiku-4-5 | **0.618** | 0.875 | **0.724** | **68** | 2 of 7 | ok | **45** |
| claude-sonnet-4-5 | 0.595 | **0.917** | 0.721 | 74 | 1 of 7 | ok | 81 |
| claude-opus-4-7 | 0.469 | **0.938** | 0.625 | 96 | 0 | ok | 87 |

gpt-5-4-mini fails the wide table the same way nano does, so stepping up one size in that family does not fix it. Only the full gpt-5-4 survives there.

Two things I did not expect. A more capable model makes over suggestion worse, not better ... opus returned 96 suggestions for 48 correct answers, the most of any model, because it follows the "apply universal checks broadly" instruction more thoroughly. So over suggestion is a prompt problem and we cannot buy our way out of it with a bigger model. And haiku beats opus on the number a reviewer actually feels, best precision and fewest suggestions to wade through.

The decoys column is the one I care about most. Those are seven plausible but wrong column bindings I planted, things like the email format rule on a free text note field, or the uniqueness rule on a foreign key that legitimately repeats. Every one of them passes all our structural checks, because each is a real column of the right family with all slots filled, so post process cannot catch them and a reviewer might well approve them.

## What I am asking for

1. Fix the output budget so it scales with column count, and fix the error path so truncation does not report as unparsable.
2. Raise the default. I would go with claude-sonnet-4-5 ... recall 0.917, only one wrong column binding, survives wide tables. Haiku has better precision and half the latency, but it took two of the seven decoys, and I would rather trade some speed than ship a default that is twice as likely to produce a wrong binding a reviewer approves.
3. MAX_RETRIEVAL_COLUMNS, see the inline comment. It costs 25% recall and does not bound the thing that is actually breaking.

Raising the default alone is not enough, because the endpoint is an admin dropdown ... anyone who picks nano reinstates a silently broken deploy. Both need doing.

## Caveats

One run per model. From repeat runs, recall and the wide table pass or fail are stable, but precision drifts about 0.02 to 0.04, so please do not pick between haiku and sonnet on precision alone. gpt-5-4-mini failing is a single observation where nano's is three.

Read the precision column as a floor rather than a score. My answer key labels the rules a table must have, while the prompt tells the judge to apply universal checks to every column a reasonable analyst would consider required, so defensible suggestions get counted against it. Recall is the number to trust.

## The eval is in this PR

I pushed it to this branch. tests/test_rule_suggester_eval.py is the deterministic half, 24 tests, runs inside make app-test in about three seconds with no workspace and no tokens, and it holds a committed baseline so a retrieval or post process change shows up as a number. tests/ai_eval is the live half behind make app-ai-eval, gated on an env var so it cannot fire by accident. Everything above is reproducible from it, and the model comparison is one command.

Happy to walk anyone through it, and happy to be told the fixture is wrong somewhere ... the answer key is a judgement call in places and I would rather argue about it with the harness in front of us.

vb-dbrks (approved):
LGTM. There needs to be a follow up PR in the future to fail the harness tests on regression with the quality of AI suggestions.

laurencewells:
LGTM

OGordon100:
@ghanse I'm not sure about query editing - we are effectively building a custom library around every conceivable SQL function (which behind the scenes will only be building a SQL query by some form of addition) - that will get very limited for more complex queries very quickly, and make readability/debugging difficult.

I'd suggest leaving as-is.

ghanse (approved):

PR #1429, opened 2026-08-05
Report MCP server integration test coverage

### LINKED ISSUE #1360: [FEATURE]: MCP Server follow up

### Problem statement

Use coverage report from the MCP server path.
Add bundle deployment.

### Proposed Solution

Follow up release

### Additional Context

_No response_

vb-dbrks:
## Changes

Reports code coverage for the MCP server for the first time, adds adoption telemetry for its tools, and fixes several bugs found while doing so.

The MCP integration suite drives the server in a **deployed Databricks App** and the runner in a **serverless job** — both remote — so pytest-side `coverage` measured nothing and the `mcp` flag was never uploaded. Coverage is now collected where the code runs and shipped back through a UC volume.

### 1. Remote coverage collection

A test-only wheel (`tests/integration_mcp/coverage_bootstrap/`) ships a `.pth`. Python execs `*.pth` files at interpreter start, so the tracer is live *before* `server.app` / `dqx_mcp_runner.runner` are imported and module-level lines are measured. **Installing the wheel is the only switch** — it exists solely on coverage-enabled deploys, so nothing ships to users and no runtime module imports `coverage`.

| | How the wheel is installed | How data returns |
|---|---|---|
| **Runner job** | the `dev-coverage` bundle target adds it to `runner_environment_dependencies` | job containers have a `/Volumes` FUSE mount → file copy |
| **App** | `ci_deploy.sh` appends it to `requirements.txt` for the deploy only (trap-restored) | Apps have **no** volume mount → Files API upload |

Configuration is derived from what each runtime already has, so there is no new bundle plumbing: the **runner** reads the `--results-volume` / `--run-id` args its task already receives (a serverless job environment cannot set env vars — the bundle schema's `Environment` allows only `base_environment`/`client`/`dependencies`/`environment_version`/`java_dependencies`), and the **app** reads its existing `DQX_CATALOG` / `DQX_TMP_SCHEMA` config env. Data lands in `<results_volume>/coverage`, which already grants read/write to both service principals.

The app's final flush comes from a graceful stop (`ws.apps.stop_and_wait`, called by the test before teardown) landing in an `atexit` hook, plus a checkpoint thread that bounds loss if the platform's ~15s SIGTERM budget is exceeded. `Coverage(sigterm=True)` is deliberately not used: its handler re-raises SIGTERM via `os.kill()` right after saving, killing the process before `atexit` runs.

Reporting is scoped to `mcp-server/` only. Without that the report also included the test process's own modules (`conftest.py`, the test module, `tests/constants.py`), which no other DQX suite reports and which dragged the repo total down by tens of percent. Every other suite pins its tree the same way — the integration and anomaly workflows generate a `.coveragerc` with `source = ../../src`.

Data files are scoped to the run that produced them. The coverage directory is derived from the catalog plus a fixed schema, so concurrent CI runs share it; only files this process downloaded are deleted, and older ones are skipped rather than reaped, so a run still checkpointing cannot lose its data to another run's teardown.

### 2. MCP tool telemetry

Each tool invocation records a `dqx_mcp/<tool_name>` User-Agent extra on one cheap call to the workspace's **own** control plane, which already logs every request it receives. Nothing is sent to a DQX-owned endpoint, so no data, metadata or telemetry leaves the customer's account.

This makes MCP usage separable from direct library use: the runner job installs real DQX and already emits `engine/*` signals, which are otherwise indistinguishable from a notebook user calling the same function.

Semantics match `src/databricks/labs/dqx/telemetry.py` — each `(key, value)` sent at most once per process, marked on *attempt* so a brownout cannot re-stall later calls, 5s retry/http caps, a FIFO-bounded dedup cache, and a broad `except` so telemetry can never surface in a tool response. It is an adoption signal, not an invocation counter. Reimplemented rather than imported because that module imports pyspark at module scope and the server depends on neither pyspark nor DQX; it is pure stdlib, so no new dependency and no lockfile change. `DQX_MCP_DISABLE_TELEMETRY` turns it off.

Two implementation notes worth review:

- **The ping runs on a daemon thread.** Called inline it added an HTTP round-trip to the first call of every tool — up to the 5s timeout on an unhealthy workspace, before the tool body ran. Measured on a live deployment after the change, the first `get_workflow` call took 0.51s against 0.59–0.65s for the deduped calls, so the ping is off the request path. A regression test asserts a hung control plane costs under a second.
- **Registration is wrapped once**, so a tool added later is instrumented by construction rather than by remembering to annotate it. Verified on a live deployment that all 12 tools keep their names, descriptions and input schemas.

The signal also carries `dqx/<version>`, which the telemetry pipeline reads into its `release_version` column. The library sets that token at import; the MCP server does not import DQX, so without it every row would be filtered out by the dashboard's `release_version IS NOT NULL` predicate. It is sourced from a `DQX_VERSION` app config value the bundle sets from the same pin the runner installs, so the reported version is the one that actually runs the checks. A test asserts the two match.

### 3. Bugs fixed

- **The stale-view sweeper never listed anything.** `SHOW VIEWS IN <catalog>.<schema>` is rejected at runtime with `CROSS_CATALOG_SCHEMA_REFERENCE_NOT_SUPPORTED` — `SHOW` resolves the schema against the session's current catalog, and the sweep runs statelessly against a shared warehouse with no session to set one on. The sweep caught the error, logged a warning and returned 0, which is indistinguishable from "no stale views". Now queries `information_schema.views`, which takes the catalog as the first level of the name. This one bug suppressed 46 lines of coverage: everything downstream of the listing — including all of `drop_view` — never executed.
- **The schema match was case-sensitive.** Unity Catalog folds identifiers to lower case in `information_schema`, so a non-lowercase schema matched nothing and the sweep silently returned 0 again. Now compares with `lower(table_schema)`, restoring the case-insensitive behaviour `SHOW VIEWS` had.
- **A short TTL could delete a live run's input.** The sweep runs on the job-submission path, so it sees the temp view of the job being submitted. The TTL is now clamped up to a 900s floor inside the server, so no configuration can delete a view whose job may still be queued.
- **`staged_*` job inputs leaked.** Excluding them from the sweep stopped deleting a queued job's input, but nothing else reclaimed them, so every inline-contract call left a file on the results volume permanently. They now age out at 24h — far longer than any plausible queue wait — and unrecognised files are left alone.
- **`get_run_result` leaked a run's existence.** Polling a run the app service principal cannot read surfaced `PERMISSION_DENIED`, which confirms the run exists. Run ids are guessable integers, so that is the disclosure the ownership guard exists to prevent. It now returns the same structured `not_found` as an unknown id, with a warning logged so an operator can still distinguish a misconfigured ACL from a stale id.
- **`DQX_VERSION` could disable telemetry silently.** The value is interpolated into a User-Agent header, and an interior newline survives `.strip()`; a header containing one is rejected outright by the HTTP client. It is now validated against a version-shaped pattern and falls back to a placeholder, which still satisfies the pipeline's `release_version IS NOT NULL` predicate.
- **`execute_sql` dropped an explicit empty `parameters` list.** Now tests `is not None`, so a caller-supplied empty list is honoured.
- **A production deploy could ship the test-only coverage wheel.** `sync.include` was `.build/**`. Only `dev-coverage` *builds* the bootstrap wheel, but `.build/` is never cleaned, so a wheel left there by an earlier coverage deploy was synced by the next plain `dev` deploy and published by `publish_extra_wheels` — test instrumentation reaching a user deploy off nothing but a stale local file. Observed on a real `dev` deploy. The base list now names the runner wheel explicitly and `dev-coverage` appends the bootstrap wheel (`sync.include` appends across targets rather than replacing, unlike an environment spec).
- **Telemetry never sent anything at all.** `_send_async` claimed the dedup slot before spawning the ping thread, then `log_telemetry` claimed the same `(key, value)` inside that thread, found it taken, and returned before putting the header on the wire — so the feature was inert on the production path while every unit test passed (they called `log_telemetry` directly with a fresh cache, or patched `_send_async`/`Thread`). Now synchronous, matching `src/databricks/labs/dqx/telemetry.py`, with `log_telemetry` as the sole claimant. The lock is kept because — unlike the library — MCP serves requests concurrently on anyio worker threads.
- **Job-backed tools stalled every agent mid-workflow.** `get_run_result` polled once and told the caller to "call again after a short wait", but an LLM agent has no sleep primitive: it called straight back, its own loop detection saw a repeated no-progress action, and it stopped to ask the user. Reproduced against a live deployment on *every* job-backed tool. `get_run_status` now blocks up to 30s (2s→8s backoff), returning as soon as the run is terminal — chosen against the ~60s idle-request timeout of clients and the Apps front door, and the fact that each in-flight poll holds an anyio worker thread. Env-configurable and clamped to 55s.
- **A run's rightful owner could be told it does not exist.** The permission branch matched the substring `"permission"` as well as the structured `PERMISSION_DENIED` code, so an unrelated error merely mentioning permissions (e.g. *"insufficient permissions to attach warehouse"*) was swallowed into `not_found`, hiding the real cause. Now gated on the error code alone; the submitter/caller check is what actually prevents the disclosure.
- **`DQX_SWEEP_TTL_SECONDS` silently drove result-file deletion.** `sweep_stale_result_files` defaulted its TTL to `_VIEW_TTL_SECONDS`, so a knob documented as tuning the *view* sweep also governed `<run_id>.json` files, and the view TTL's 900s floor leaked across to files it was never reasoned about. Separate `DQX_RESULT_TTL_SECONDS`, same 3600s default.
- **Telemetry spawned a thread per tool call.** The dedup check ran *inside* the spawned thread, so a tool that had already reported still paid for a thread and an SDK client on every later call, and concurrent first calls to the same tool could each fire a ping. The `(key, value)` slot is now claimed under a lock *before* the thread is created.

### 4. Test and CI changes

- **`mcp.yml` merges and uploads to Codecov** under flag `mcp`, mirroring `anomaly.yml`, and passes `-m` to `coverage report` so the CI log lists the missing lines.
- **CI installs DQX built from the repo; production keeps the published pin.** The runner installed `databricks-labs-dqx` from PyPI, so a PR breaking `src/databricks/labs/dqx/` left this suite green — it tested the last *release*, not the code under review. The `dev-coverage` target now installs the in-repo wheel. The `[datacontract]` extra cannot ride on a wheel path: the PEP 508 `pkg[extra] @ file:///…` form is valid for pip (verified) but the Databricks CLI rejects it while normalising dependency paths, so `datacontract-cli` is listed explicitly with a test keeping it in sync with `pyproject.toml`.
- **`mcp.yml` runs on every PR** (dropped the `paths:` filter), like `anomaly` and `acceptance`: the MCP server exercises the DQX library end to end, so a change anywhere in the library or shared plumbing can break it. The old filter was also stale — it claimed the runner installs the in-repo DQX wheel, but the runner installs a pinned published release.
- **New integration assertions**, each covering a path the tool interface can actually reach:
  - the stale-view sweeper reaping an orphan (a view's age comes from its `v_<epoch>_<hex>` name, so a test can plant one that is already past the TTL floor instead of waiting)
  - an inline contract being staged to the results volume as `staged_<uuid>` for the runner to read
  - a run belonging to another job returning `not_found` with no payload
  - all three `generate_rules_from_contract` sources (volume path, inline text, workspace path via export+base64) plus its either/or errors
  - table-name validation rejecting unqualified, over-qualified, backtick-breakout and whitespace forms
  - a caller-unreadable path being refused before any job is submitted
  - a failed run reporting the task's own error rather than the generic job-level message
  - a check on a missing column being reported as *skipped* rather than failed
- **Sweep thresholds are env-overridable.** Production defaults are unchanged; the `dev-coverage` target sets the interval to 0 so one run exercises both sweepers. Only the thresholds vary — the sweep logic is identical everywhere.
- **Bundle guards.** The `dev-coverage` target must repeat the base env and dependency lists (DAB target overrides replace a list rather than merging into it), so a test asserts it stays a superset of the base and that only one DQX version is pinned across the file. A further test asserts the base `sync.include` never wildcards `.build` nor names the test-only wheel.
- **Telemetry concurrency guards.** One test asserts five calls to the same tool spawn one thread; another puts eight callers on a barrier and asserts a single ping. Both fail against the previous code (5 threads, 8 pings).

### Tests

Validated end to end against a live workspace (deploy → every tool → flush → download → merge), not only locally. Separately deployed to field-eng under the plain `dev` target and driven over the real Apps front door against purpose-built demo tables — which is how the `.build/**` sync bug above was found, and how the agent stall and the inert telemetry were caught (both are invisible to unit tests).

Live evidence for the two riskiest changes: polling a pending run took `30070ms`/`30114ms` and returned early at `9092ms` once the job finished (versus ~600ms before, which is what made the agent give up); and the deployed runner job resolves the in-repo DQX wheel with `datacontract-cli` alongside it and **no** published pin competing with it. The full integration suite passes in 2013s of its 2700s budget.

- [x] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

Also verified: the built wheel carries `dqx_mcp_coverage.pth` at the site-packages **root** (inside the package it would never be exec'd); an isolated-venv run proves the tracer records a **module-level** line, i.e. it starts before import; `bundle validate` shows `dev` with no bootstrap wheel and no sweep overrides vs `dev-coverage` with both; the workspace upload→export→base64 round-trip is byte-identical; the sweeper fix drops a planted 2h-old view and keeps a fresh one; and the telemetry User-Agent satisfies both dashboard predicates.

`make fmt`, `make mcp-check` (0 errors) and `make mcp-test` (196 passed) exit clean, both bundle targets validate, and `mcp-server/uv.lock` is unchanged.

### Notes for reviewers

- Running `mcp` on every PR adds ~30 min of CI and deploys a real app per PR — the same cost `anomaly`/`acceptance` already pay. Happy to reinstate a corrected filter instead.
- Some paths remain uncovered because they are unreachable through the tool interface, and are left rather than contorting tests to reach them: the runner's malformed-principal and non-3-part-table rejections (both values are produced server-side or by the Apps front door, so a test cannot supply a bad one), the cross-user identity check (the front door sets the identity header), and the CORS fallback that only fires when `DATABRICKS_HOST` is unset. `telemetry.py` is at 97% under unit tests; the counted figure is lower because only integration coverage is uploaded.
- The reported percentage is branch coverage, so it is not comparable to a line-only figure — partial branches count as misses.

### Documentation and Demos

- [ ] added/updated demos
- [x] added/updated docs

`tests/integration_mcp/coverage_bootstrap/README.md` documents the mechanism, the local recipe, and two things not to "simplify" (why a `.pth` and not `sitecustomize.py`; why never `sigterm=True`).

The MCP guide is updated for the current Genie Code UI: the flow moved from **Settings → MCP Servers** to **Customizations → MCP servers → Add Server**, which opens a panel offering four server types — a Databricks App belongs in one specific section and nothing in the UI signals that, so the steps say which. Both affected screenshots are retaken, and the stale **Refresh tools list** reference is gone (that button no longer exists). A 3-minute narrated walkthrough is added (4.6MB, 7% of the source recording), with the screenshot steps collapsed behind a details block since they cover the same ground — the approval-mode and expected-polling notes stay visible, because both stop a user thinking something is broken.


mwojtyczka (commented):
Excellent PR — collecting coverage where the code actually runs (remote app + serverless job) via a test-only `.pth` bootstrap is a clean solution, the concurrency/atexit ordering in the bootstrap is carefully reasoned, and the new integration assertions are real (IDOR, skipped-vs-failed, contract sources) rather than line-padding. The production sweeper fix (only sweep `*.json`, plus the TTL floor) is a genuine bug catch with unit tests both halves.

I verified the one change that looked risky and it's fine: dropping `safe_schema = _validate_sql_identifier(schema)` from `sweep_stale_views` is **not** an injection/NameError regression — `drop_view` independently re-validates every FQN part against `_SAFE_IDENTIFIER_RE` and quotes it before any SQL, and the `information_schema` query interpolates only the validated+quoted `safe_catalog` while binding the schema as a parameter.

Findings are mostly low-severity (details inline):
1. **information_schema schema match is case-sensitive** — `WHERE table_schema = :schema_name` binds the raw value, but UC stores those names lowercased, so a non-lowercase schema arg lists nothing (the old `SHOW VIEWS` was case-insensitive). Works today with the fixed `dqx_mcp_tmp`.
2. **`staged_*` inputs now leak** — correctly no longer swept by age, but nothing else sweeps them, so they accumulate forever.
3. **Coverage prune can delete a concurrent run's live file** — the `since-1h` cutoff is more aggressive than the "earlier run" justification (coverage-only, low sev).
4. **dev-coverage duplicates the base env/deps lists** — the DQX pin is now in three places with nothing enforcing sync.
5. Minor: `execute_sql` drops an explicit empty `parameters=[]`.

mcp-server/server/utils.py
```diff
@@ -582,35 +588,79 @@ def stage_bytes_to_results_volume(content: bytes, suffix: str = "") -> str:
...
+        # rejected outright with CROSS_CATALOG_SCHEMA_REFERENCE_NOT_SUPPORTED ("Run 'USE CATALOG'
+        # first"), because SHOW resolves the schema against the session's current catalog only. This
+        # runs statelessly against a shared warehouse, so there is no session to set a catalog on —
+        # information_schema takes the catalog as the first level of the name and needs no USE.
+        # Imported here, not at module scope: this module is imported by the app at startup and
+        # keeps SDK types out of its import graph (see the other local `from databricks.sdk`
+        # imports).
+        from databricks.sdk.service.sql import StatementParameterListItem
+
+        rows = execute_sql(
```
mwojtyczka, line 683: **Case-sensitivity regression (works today, latent).** `WHERE table_schema = :schema_name` binds the raw `schema` value, but Unity Catalog stores identifiers in `information_schema` folded to lower case, so an equality against a non-lowercase schema name matches nothing and the sweep silently returns 0 (which looks exactly like "no stale views"). The previous `SHOW VIEWS IN <catalog>.<schema>` resolved the schema identifier case-insensitively. The only caller passes the fixed lowercase `dqx_mcp_tmp`, so it's fine now — but it depends on every caller pre-normalizing. Suggest `WHERE lower(table_schema) = :schema_name` binding `schema.lower()` (or lowercasing the arg) to preserve the old semantics.

mcp-server/server/utils.py
```diff
@@ -644,6 +699,8 @@ def sweep_stale_result_files(ws: Any, ttl_seconds: int = _VIEW_TTL_SECONDS) -> i
     for entry in entries:
         if getattr(entry, "is_directory", False):
             continue
+        if not (entry.path or "").endswith(".json"):
+            continue  # staged job inputs — see the docstring
```
mwojtyczka: **Correct fix, but `staged_*` inputs now leak.** Excluding non-`.json` files rightly stops deleting a queued job's staged input by age — but no other sweeper covers `staged_*`, so every `generate_rules_from_contract`-with-inline-content call leaves a `staged_<uuid>.yaml` on the results volume forever, and the volume grows unbounded. Consider sweeping `staged_*` too, with a much longer TTL than any plausible queue wait (so a pending job's input is safe) rather than excluding them from cleanup entirely.

tests/integration_mcp/conftest.py
```diff
@@ -300,11 +387,122 @@ def wait(self, run_id: int, *, timeout: float = 300.0, interval: float = 4.0) ->
...
+
+    Files older than *since* (less a clock-skew tolerance) belong to an earlier run — concurrent runs
+    each write their own and nothing else prunes the directory — so they are deleted rather than
+    downloaded, keeping the volume from growing without bound and keeping a prior run's data out of
+    this run's merged report. A delete that loses a race with a concurrent writer is ignored: this is
+    opportunistic housekeeping, never a correctness dependency.
+    """
+    downloaded: list[str] = []
+    stale = 0
+    cutoff = since - _COVERAGE_CLOCK_SKEW_TOLERANCE_SECONDS
```
mwojtyczka, line 522: **Prune is more aggressive than the "earlier run" justification.** This deletes any `.coverage*` file older than `since - 1h`. The safety argument is "the volume is per-run in CI," but `_coverage_dir()` defaults to `/Volumes/{CATALOG}/dqx_mcp_tmp/mcp_results/coverage` — keyed on the catalog + fixed schema, not `name_prefix` — so two runs can share it. A second run whose deploy began >1h earlier but is still checkpointing would have its live file deleted mid-run by this run's teardown, silently losing that run's coverage. Coverage-only and low-likelihood, but safer to prune by the coverage filename's own run-id/pid, or to skip deleting files this process didn't just download.

mcp-server/databricks.yml
```diff
@@ -31,6 +31,42 @@ variables:
   # implicitly) or `bundle deploy` fails with "User is not authorized to use this service principal".
   runner_service_principal_id:
     default: "00000000-0000-0000-0000-000000000000"
+  # The app's runtime environment, as a complex variable so the dev-coverage target can extend it
+  # (DAB target overrides don't merge into a list). The default is what every real deploy gets.
+  app_environment:
```
mwojtyczka, line 68: **List duplication with no sync guard.** Because DAB target overrides replace a list rather than merging, the `dev-coverage` target (below) re-lists the whole `app_environment` and `runner_environment_dependencies` base contents to add one entry each. So `databricks-labs-dqx[datacontract]==0.15.0` is now pinned in **three** places in this file, and any change to the production default that isn't mirrored into `dev-coverage` means CI silently deploys/tests a configuration users never get (or the DQX pin drifts between prod and CI). A comment flags it but nothing enforces it — a small unit test asserting the two lists share the same base (like the wheel-filename pin test in the sibling PR #1430) would catch drift.

mcp-server/server/utils.py
```diff
@@ -290,6 +295,7 @@ def execute_sql(ws: Any, query: str, warehouse_id: str) -> list[dict[str, Any]]:
         statement=query,
         warehouse_id=warehouse_id,
         wait_timeout="30s",
+        **({"parameters": parameters} if parameters else {}),
```
mwojtyczka: Minor: `if parameters else {}` treats an explicit empty list `[]` the same as `None`, silently dropping the kwarg. Harmless today (the one caller always passes a non-empty list), but `if parameters is not None` states the actual intent — send bindings when the caller provided them, even if the list is empty.

mwojtyczka (changes requested):
Generally looking good, left a few comments. Also code cov should be green

mcp-server/server/utils.py
vb-dbrks: Confirmed and fixed.

mcp-server/server/utils.py
vb-dbrks: Took your suggested shape... staged_* now ages out at 24h, longer than any plausible queue wait so a pending job's input is still safe

tests/integration_mcp/conftest.py
vb-dbrks: addressed this now.

mcp-server/databricks.yml
vb-dbrks: Added the guard as suggested.

mcp-server/server/utils.py
vb-dbrks: Agreed, changed to is not None.

mwojtyczka (commented):
Automated code review — 4 findings (2 confirmed, 2 for maintainer judgment). Details inline.

mcp-server/server/telemetry.py
```diff
@@ -0,0 +1,175 @@
...
+            # Imported here, not at module scope: keeps SDK construction off the import path and
+            # avoids a circular import (utils imports nothing from this module).
+            from .utils import get_sp_client_for_telemetry
+
+            log_telemetry(get_sp_client_for_telemetry(), TELEMETRY_KEY, tool_name)
+        except Exception as e:  # never let instrumentation break anything
+            logger.debug(f"Telemetry skipped for {tool_name}: {e}")
+
+    try:
+        threading.Thread(target=_run, name=f"dqx-mcp-telemetry-{tool_name}", daemon=True).start()
```
mwojtyczka: **efficiency (confirmed):** `_send_async` spawns a new daemon thread on *every* tool call, and the `(key, value)` dedup lives in `log_telemetry` which runs *inside* the thread. That contradicts the docstring's claim that "the dedup cache still bounds this to one thread per distinct signal per process" — a thread is created for every request even after the signal was already sent, and N concurrent first-invocations of the same tool all pass the `if (key, value) in _sent_telemetry` check before any of them sets it, firing N control-plane pings instead of one.

Impact is low (MCP is low-QPS), but the guarantee the comment promises doesn't hold. Checking the dedup cache *before* spawning the thread would honor it.

tests/integration_mcp/conftest.py
```diff
@@ -300,11 +390,144 @@ def wait(self, run_id: int, *, timeout: float = 300.0, interval: float = 4.0) ->
...
+
+def _download_coverage_files(ws: WorkspaceClient, coverage_dir: str, since: float) -> list[str]:
+    """Download this run's ``.coverage*`` data files from the UC volume dir into the repo root.
+
+    Only files this run produced are downloaded, and only files it downloaded are deleted. That
+    coupling matters because the coverage directory is genuinely shared: it is derived from the
+    catalog plus a fixed schema, not from ``name_prefix``, so every concurrent CI run writes into the
+    same place. Deleting purely by age could therefore reap a *live* file belonging to a run that
+    started earlier and is still checkpointing — losing that run's coverage silently, which is the
+    exact failure this whole mechanism exists to prevent.
```
mwojtyczka: **conventions (confirmed):** This docstring asserts the coverage directory is "genuinely shared: it is derived from the catalog plus a fixed schema, not from `name_prefix`, so every concurrent CI run writes into the same place." But `_coverage_dir` builds the path from the per-run `tmp_schema`, and `ci_deploy.sh` derives `TMP_SCHEMA` per run as `${NAME_PREFIX}_tmp` where `NAME_PREFIX` carries a fresh uuid (`mcp-dqx-it-<hex>`). So each concurrent CI run actually gets its own `/Volumes/<cat>/<unique_schema>/mcp_results/coverage`.

The elaborate age-skip + download-only-what-you-delete coupling guards against a cross-run collision that can't occur under the real layout — worth either correcting the docstring or simplifying the logic, so a future maintainer doesn't trust a false invariant.

.github/workflows/mcp.yml
```diff
@@ -7,14 +7,13 @@
 # acceptance action exactly like the other integration workflows.
 name: mcp
 
+# Runs on every PR (no `paths:` filter), like the anomaly and acceptance suites: the MCP server
+# exercises the DQX library end-to-end, so a change anywhere in the library, its packaging, or the
+# shared test/CI plumbing can break it. Running always also means the `mcp` Codecov flag is uploaded
+# for every commit, so those files never drop out of a report for want of a carryforward rule.
```
mwojtyczka, line 13: **altitude (for maintainer judgment):** Removing the `paths:` filter makes this suite deploy a real Databricks app + runner job + volumes on *every* PR (1h timeout budget), including doc-only or otherwise-unrelated PRs, purely to keep the Codecov `mcp` flag fresh.

The stated benefit (a carryforward-free flag on every commit) could instead be achieved with a Codecov carryforward rule, keeping this expensive live-deploy suite path-scoped. Worth confirming the per-commit cost was intended rather than scoping it to mcp/library changes.

tests/integration_mcp/conftest.py
```diff
@@ -300,11 +390,144 @@ def wait(self, run_id: int, *, timeout: float = 300.0, interval: float = 4.0) ->
...
+        # last_modified is epoch MILLIS on this API (files.get_metadata returns an HTTP date string
+        # instead — do not mix them). An absent or non-numeric value is treated as current, so
+        # missing metadata can never cost this run its data.
+        last_modified = getattr(entry, "last_modified", None)
+        if isinstance(last_modified, (int, float)) and last_modified / 1000 < cutoff:
+            skipped += 1
+            continue
+        body = ws.files.download(path).contents
+        dest = _REPO_ROOT / name
+        dest.write_bytes(body.read() if body is not None else b"")
```
mwojtyczka: **correctness (plausible):** When `ws.files.download(path).contents` is `None` (a branch this line itself anticipates), `dest.write_bytes(b"")` creates a zero-byte `.coverage.*` file in the repo root. `coverage combine` will then hit a file that isn't a valid SQLite coverage database and raise — the workflow's `|| true` masks the abort, but the combine of the remaining valid files is lost, silently understating the `mcp` flag. Consider skipping the write (or deleting the empty dest) when the body is `None`.

mcp-server/server/telemetry.py
vb-dbrks: Confirmed and fixed in 4a85393 — the docstring was describing an invariant the code didn't enforce.

`_claim_signal()` now does the check-and-insert under a lock, and `_send_async` calls it *before* creating the thread. So a tool that has already reported costs nothing on later calls (no thread, no client construction), and concurrent first calls produce exactly one ping.

Two regression tests, both of which fail against the previous code:
- `test_no_thread_is_spawned_once_the_signal_has_been_sent` — 5 calls, 1 thread (was 5)
- `test_concurrent_first_calls_send_exactly_one_signal` — 8 threads on a barrier, 1 ping (was 8)

tests/integration_mcp/conftest.py
vb-dbrks: Confirmed — you traced it correctly, the docstring was stale. `_coverage_dir` takes the per-run `tmp_schema` (`<name_prefix>_tmp`, and `name_prefix` carries a fresh uuid), so under CI each run does own its directory.

Corrected the docstring in 4a85393 rather than dropping the logic, because two cases can still put a foreign file there:
- an explicit `DQX_MCP_COVERAGE_DIR` pointing several runs at one directory (that override exists precisely to share it)
- a re-run against a surviving schema whose previous `bundle destroy` didn't complete

Both are off the normal path, so the docstring now says so instead of claiming a collision that can't occur under CI.

tests/integration_mcp/conftest.py
vb-dbrks: I tried to reproduce the abort and couldn't — `coverage combine` tolerates a zero-byte data file rather than raising. Tested on 7.4.4 (the version pinned in `uv.lock`, so what CI runs) and 7.6.10, with two valid files plus one empty:

```
=== combine ===
Combined data file .coverage.run0
Combined data file .coverage.run1
Combined data file .coverage.zero
combine exit=0
=== report ===
m.py    4   0   100%      <- both valid files still merged
report exit=0
```

So the valid files aren't lost and the flag isn't understated. It looks like coverage treats an empty file as "no data yet" rather than a corrupt database.

Guarded it anyway in 4a85393, since writing a file we know is invalid has no upside: an empty body now logs and `continue`s, leaving the remote file in place so the run that owns it can still flush. Happy to be corrected if you've seen it abort on a version I didn't try.

.github/workflows/mcp.yml
vb-dbrks: Fair challenge, and the cost is real — worth stating what it is so you can make the call.

Checked the neighbouring suites: neither `anomaly.yml` nor `acceptance.yml` carries a `paths:` filter, so both already deploy live workspace resources on every non-draft, non-fork PR. Running `mcp` the same way is the existing convention rather than a new cost centre, which is why I dropped the filter rather than fixing it.

The coverage argument is secondary and I'd not defend it alone — as you say, a carryforward rule handles that. The primary reason is that the old filter was **wrong**, not just narrow: it claimed the runner installs the in-repo DQX wheel, but the runner installs a pinned published release, so the paths it listed didn't correspond to what actually breaks the suite. A change to shared CI plumbing or the acceptance harness breaks `mcp` without touching anything the filter matched.

So the options are a correct-but-narrower filter plus a carryforward rule, or unfiltered like its neighbours. I went with the latter for consistency; happy to add a corrected filter if you'd rather hold the ~30 min. Your call — I don't have a strong preference.

vb-dbrks:
Pushed `c852509f` addressing the four findings above, plus one bug that manual deployment turned up.

**Findings**
1. Telemetry thread-per-call — fixed, with two regression tests that fail against the previous code.
2. Coverage-dir docstring — corrected; the directory is per-run as you traced.
3. `paths:` filter — left unfiltered for consistency with `anomaly`/`acceptance` (neither has a filter), but the old filter was factually wrong rather than just narrow. Details inline; happy to go either way.
4. Zero-byte `combine` abort — couldn't reproduce on 7.4.4 or 7.6.10 (`combine` tolerates it and still merges the valid files). Guarded anyway. Evidence inline.

**New bug, found by deploying rather than by reading**

`sync.include` was `.build/**`. Only `dev-coverage` *builds* the coverage bootstrap wheel — but `.build/` is never cleaned, so a wheel left there by an earlier coverage deploy was synced by the next plain `dev` deploy and published by `publish_extra_wheels`. I hit this on a real `dev` deploy in field-eng: the app logged `Published auxiliary wheel …coverage_bootstrap…` on a target that is supposed to have no test instrumentation at all.

That directly contradicted the docstring's claim that production deploys are a no-op "because the glob finds nothing" — true of the build, not of the sync. The base list now names the runner wheel explicitly and `dev-coverage` appends the test wheel; a test asserts the base list never wildcards `.build` nor names the test wheel.

Worth noting `sync.include` **appends** across targets rather than replacing, unlike an environment spec or a complex variable — the opposite of the constraint documented elsewhere in this file.

mcp-server/server/telemetry.py
```diff
@@ -0,0 +1,198 @@
...
+    Args:
+        ws: WorkspaceClient whose config carries the signal.
+        key: telemetry key (e.g. ``dqx_mcp``).
+        value: telemetry value (e.g. the tool name).
+    """
+    if _telemetry_disabled():
+        return
+    # Claiming here as well as in _send_async keeps this function correct when called directly; the
+    # claim is idempotent, so the second check is a cheap no-op on the wrapped path.
+    if not _claim_signal(key, value):
```
mwojtyczka, line 128: **Telemetry is never sent on the production (wrapped) path — double dedup-claim early-returns.**

`with_telemetry.wrapper` → `_send_async(tool_name)` claims the slot synchronously at line 182 via `_claim_signal(TELEMETRY_KEY, tool_name)` (returns `True`, inserts the tuple). The daemon thread's `_run` then calls `log_telemetry(client, TELEMETRY_KEY, tool_name)`, which hits this line and claims the **same** `(key, value)` again → `_claim_signal` returns `False` → `return`. So `with_user_agent_extra(...)` / `client.clusters.select_spark_version()` never run, and the User-Agent signal — the whole point of the feature — never goes on the wire.

The comment just above ("a cheap no-op on the wrapped path") is inaccurate: it's an early return that skips the send, not a no-op. Unit tests miss it because they call `log_telemetry` directly with a fresh cache, or patch `_send_async` / `threading.Thread`.

Fix option: have `_send_async` not claim before spawning (let `log_telemetry` be the sole claimant), or pass a flag so the thread path skips the second claim. Whichever you pick, add a test that drives the real wrapped path (no patching of `_send_async`/`Thread`) and asserts the send occurs exactly once.

mcp-server/server/telemetry.py
```diff
@@ -0,0 +1,198 @@
...
+            # Imported here, not at module scope: keeps SDK construction off the import path and
+            # avoids a circular import (utils imports nothing from this module).
+            from .utils import get_sp_client_for_telemetry
+
+            log_telemetry(get_sp_client_for_telemetry(), TELEMETRY_KEY, tool_name)
+        except Exception as e:  # never let instrumentation break anything
+            logger.debug(f"Telemetry skipped for {tool_name}: {e}")
+
+    try:
+        threading.Thread(target=_run, name=f"dqx-mcp-telemetry-{tool_name}", daemon=True).start()
```
mwojtyczka: **Consequence of the double-claim (see comment on line 125):** because `log_telemetry` always early-returns on the wrapped path, this thread + the `WorkspaceClient` constructed in `_run` do pure wasted work on every first-per-tool call — a thread spawned and an SDK client built only to reach a `log_telemetry` that returns before sending. Fixing the double-claim resolves this too.

mcp-server/server/utils.py
```diff
@@ -793,6 +888,14 @@ def get_run_status(run_id: int) -> dict[str, Any]:
         error_code = getattr(e, "error_code", "") or ""
         if error_code in ("RESOURCE_DOES_NOT_EXIST", "INVALID_PARAMETER_VALUE") or "does not exist" in str(e).lower():
             return _run_not_found(run_id)
+        # A run the app SP cannot see is also not_found, for the same reason the job_id mismatch below
+        # is: telling the caller "permission denied" confirms the run EXISTS, which is exactly the
+        # disclosure the ownership guard exists to prevent — run ids are guessable integers. The job
+        # ACL blocking the read first is defence in depth, not a licence to leak. Logged so an
+        # operator can still distinguish a misconfigured ACL from a genuinely stale run_id.
+        if error_code == "PERMISSION_DENIED" or "permission" in str(e).lower():
```
mwojtyczka: **Broad substring match can mask unrelated errors as `not_found`.** `"permission" in str(e).lower()` catches any `DatabricksError` whose message merely contains "permission", beyond the structured `PERMISSION_DENIED` code. A run the caller legitimately owns that hits an unrelated error whose text includes "permission" (e.g. "insufficient permissions to attach warehouse", or a transient dependency error) gets swallowed into `_run_not_found`, so the owner is told their existing run doesn't exist and the real error is hidden. Consider gating on the `error_code` alone (`PERMISSION_DENIED`), or narrowing the substring fallback to a more specific phrase.

tests/integration_mcp/conftest.py
```diff
@@ -300,11 +390,152 @@ def wait(self, run_id: int, *, timeout: float = 300.0, interval: float = 4.0) ->
...
+    for entry in ws.files.list_directory_contents(coverage_dir):
+        path = entry.path or ""
+        name = path.rsplit("/", 1)[-1]
+        if not path or not name.startswith(".coverage"):
+            continue
+        # last_modified is epoch MILLIS on this API (files.get_metadata returns an HTTP date string
+        # instead — do not mix them). An absent or non-numeric value is treated as current, so
+        # missing metadata can never cost this run its data.
+        last_modified = getattr(entry, "last_modified", None)
+        if isinstance(last_modified, (int, float)) and last_modified / 1000 < cutoff:
```
mwojtyczka, line 532: **Shared-directory coverage race.** The docstring calls out the explicit `DQX_MCP_COVERAGE_DIR` case where multiple runs share one dir. Here a file is downloaded (and then deleted at line 532) whenever `last_modified >= cutoff`. A concurrent run B's live checkpoint file is newer than run A's `cutoff`, so A downloads it — merging B's partial coverage into A's report — and then deletes B's file. The "only delete what we downloaded / skip older ones" mitigation only protects files *older* than A's start, not a concurrent run's newer file. Under the normal per-uuid `tmp_schema` layout runs are isolated, so this is limited to the shared-dir edge case, but worth guarding (e.g. tag files with a per-run id) or documenting as unsupported.

mcp-server/server/utils.py
```diff
@@ -624,11 +699,22 @@ def sweep_stale_views(
 
 
 def sweep_stale_result_files(ws: Any, ttl_seconds: int = _VIEW_TTL_SECONDS) -> int:
```
mwojtyczka: **Result-file TTL is coupled to the view TTL.** `sweep_stale_result_files` defaults `ttl_seconds` to `_VIEW_TTL_SECONDS`, and `_maybe_sweep_stale_views` calls it with the default. So `DQX_SWEEP_TTL_SECONDS` (and the 900s floor) intended to tune the *view* sweep also silently drives the `.json` result-file TTL — there's no independent knob, and the name `_VIEW_TTL_SECONDS` misleads a maintainer into thinking result files have their own threshold. Consider a separate `DQX_RESULT_TTL_SECONDS` (or at least renaming/commenting to make the shared default intentional).

mcp-server/databricks.yml
```diff
@@ -275,3 +312,66 @@ artifacts:
...
+  # test-only coverage bootstrap wheel, whose .pth starts a coverage tracer at interpreter start
+  # (see tests/integration_mcp/coverage_bootstrap). Kept out of `dev` deliberately —
+  # production/user deploys must not ship test instrumentation, and no production module is
+  # coverage-aware. Selected by scripts/ci_deploy.sh when DQX_MCP_COVERAGE_DIR is set.
+  dev-coverage:
+    # Target-scoped so a production (`dev`) deploy never builds it and does not need the repo's
+    # tests/ tree present.
+    artifacts:
+      dqx_mcp_coverage_bootstrap_wheel:
+        build: uv build ../tests/integration_mcp/coverage_bootstrap --wheel --out-dir ./.build
```
mwojtyczka: **Coverage bootstrap wheel is built twice on a dev-coverage deploy.** This `artifacts.dqx_mcp_coverage_bootstrap_wheel.build` runs `uv build ../tests/integration_mcp/coverage_bootstrap --wheel --out-dir ./.build` during `bundle deploy`, and `scripts/ci_deploy.sh` (line 94) runs the identical command manually just before. Wasted build work each CI deploy, and the two invocations can silently diverge if only one's flags are updated. Consider building it in one place only.

mwojtyczka (commented):
Automated code review (6 findings). Verified against the PR head by tracing the call paths and re-running the reproduction locally. The most important is the telemetry double-claim (finding 1): the feature appears to be dead on its only production path.

mcp-server/server/telemetry.py
```diff
@@ -0,0 +1,198 @@
...
+# the cap exists so a future high-cardinality signal cannot grow memory without limit.
+_TELEMETRY_CACHE_MAX_SIZE = 10_000
+_sent_telemetry: "OrderedDict[tuple[str, str], None]" = OrderedDict()
+
+# Guards _sent_telemetry. Tools are served concurrently, so an unguarded read-then-write lets N
+# simultaneous first invocations of the same tool all observe "not sent yet" and each fire a ping.
+_cache_lock = threading.Lock()
+
+# Namespaces the signal so MCP usage is separable from direct DQX library usage.
+TELEMETRY_KEY = "dqx_mcp"
```
mwojtyczka:
```suggestion
TELEMETRY_KEY = "mcp"
```

mcp-server/databricks.yml
```diff
@@ -42,6 +42,68 @@ variables:
...
+      # release; bump it when upgrading DQX. The [datacontract] extra can be applied here because
+      # this is a PyPI spec, not a wheel path.
+      #
+      # The [llm] extra (dspy) is intentionally NOT installed — the runner stays lean and
+      # deterministic, mirroring the DQX Studio task runner. As a result the
+      # generate_rules_from_contract tool is deterministic-only: it does not process the
+      # contract's free-text expectations (DQX's LLM "text rules" path). Enabling that is a
+      # follow-up: add the dspy/litellm deps from the DQX [llm] extra here, plumb an LLM
+      # endpoint config through to the runner, and re-expose the parameter in server/tools.py.
+      - databricks-labs-dqx[datacontract]==0.15.0
```
mwojtyczka: For CI really we need to build the wheel.

mcp-server/server/telemetry.py
vb-dbrks: Good catch — you're right, it never sent. Made it synchronous like the library instead, so `log_telemetry` is the only place that claims the dedup slot. Added a test that drives the real wrapped path with nothing patched.

mcp-server/server/telemetry.py
vb-dbrks: Fixed by the same change — no thread, no wasted client.

mcp-server/server/utils.py
vb-dbrks: Done, gated on the error code only.

mcp-server/server/utils.py
vb-dbrks: Done — added `DQX_RESULT_TTL_SECONDS`, same 3600 default so nothing changes in practice.

mcp-server/databricks.yml
vb-dbrks: Done, dropped the bundle artifact. `ci_deploy.sh` has to keep its copy because it needs the filename before `bundle deploy` runs.

mcp-server/server/telemetry.py
vb-dbrks: Done.

tests/integration_mcp/conftest.py
vb-dbrks: Agreed, and it's only reachable via an explicit `DQX_MCP_COVERAGE_DIR` pointing two runs at one directory — under the normal per-uuid schema each run owns its dir. Documented that case as unsupported in the docstring rather than adding per-run file tagging, since the coverage-only blast radius didn't seem worth the machinery. Happy to tag the files if you'd rather.

.github/workflows/mcp.yml
vb-dbrks: Correction to what I said above: I claimed this suite exercises the DQX library end to end, which wasn't true — the runner installs a published release, so a library change wasn't tested here at all. My justification was wrong.

Fixed the underlying thing instead: the `dev-coverage` (CI) target now installs DQX built from the repo source, while production keeps the published pin. So a library regression can actually fail this suite now, which is what I'd wrongly assumed already happened. That also makes running it unfiltered worth the cost — but shout if you'd still prefer a filter.

vb-dbrks:
Pushed `452be2fc` — five commits addressing your review, plus two bugs that only showed up once I drove a real deployment.

**Your findings**

| Finding | |
|---|---|
| Double-claim never sends | Fixed — telemetry is synchronous now, like the library |
| Wasted thread + client | Fixed by the same change |
| `"permission"` substring | Fixed, error code only |
| Result TTL coupling | Fixed, `DQX_RESULT_TTL_SECONDS` |
| Wheel built twice | Fixed, bundle artifact dropped |
| `TELEMETRY_KEY = "mcp"` | Done |
| Shared-dir coverage race | Documented as unsupported — happy to tag files if you'd rather |
| `paths:` filter | See my correction below |

**Found by deploying, not reading**

Every job-backed tool stalled the agent mid-workflow. `get_run_result` returned instantly with `running` and told the caller to wait — but an agent has no sleep primitive, so it called straight back, its loop detection saw no progress, and it stopped to ask the user. The server now absorbs the wait (30s, bounded well under the front door's idle timeout). Measured live: polls take ~30s and return early once the job finishes, and a 70s job completes in 3 back-to-back calls.

**The `paths:` correction**

I said this suite exercises the DQX library end to end. It didn't — the runner installed a published release, so a library change wasn't tested here at all. The `dev-coverage` target now installs the wheel built from the repo, so that's true going forward. Production keeps the pin.

One thing I couldn't express: `databricks-labs-dqx[datacontract] @ file:///Volumes/…whl` is valid PEP 508 and pip resolves the extra fine, but the CLI rejects it (`first path segment in URL cannot contain colon`). So `datacontract-cli` is listed explicitly, with a test keeping it in sync with `pyproject.toml`.

**Verification**

`make fmt`, `make mcp-test` (206 passed) and `make docs-build` clean; full integration suite passes in 2013s. Also raised the local pytest timeout — a passing run was using 1773s of an 1800s budget, so it was one slow call from going red. `uv.lock` unchanged.

**Still unproven, and worth saying:** I haven't demonstrated that a real `src/databricks/labs/dqx/` regression now fails this suite. That's the whole point of the source-wheel change, so I'd like to break a library function deliberately and confirm the suite goes red before you rely on it.

mwojtyczka (approved):
LGTM

PR #1430, opened 2026-08-05
Deploy the MCP server with native DAB grants; drop the setup job

### LINKED ISSUE #1360: [FEATURE]: MCP Server follow up

vb-dbrks:
## Changes

Aligns the MCP server's deployment with the pattern DQX Studio adopted in #1300: Unity Catalog grants are **native Databricks Asset Bundle declarations** instead of a post-deploy notebook job.

Installing previously took a pre-deploy shell script (to create the volume `workspace.artifact_path` pointed at, which had to pre-exist) plus a post-deploy `dqx_setup` notebook job that looked up the app's service principal at runtime — with retry, because it raced app provisioning — and issued ~17 GRANT/DDL statements. Both are now deleted:

- The tmp schema and the `mcp_results` / `dqx_artifacts` volumes are **bundle resources with native `grants:`**, and the runner job's ACL is a native `permissions:` block. The app SP is referenced declaratively as `${resources.apps.mcp-dqx.service_principal_client_id}`, so there is no runtime lookup.
- `workspace.artifact_path` is **removed**. `bundle deploy` uploads artifacts *before* it creates resources, so a volume used as the artifact path can never be bundle-managed — that ordering is the whole reason the pre-deploy script existed. Instead the **app publishes the runner wheel to the volume at startup** (`server/bootstrap.py`, hash-marker so an unchanged wheel is not re-uploaded) and the runner installs it from an absolute `/Volumes` path.
- **`make mcp-deploy` is one command** end to end, and `make mcp-destroy` is one again.

### Two deliberate divergences from Studio

**The app keeps `CAN_MANAGE_RUN`, not `CAN_MANAGE`.** Studio needs `CAN_MANAGE` because it rewrites the runner job's environment at runtime (`_update_job_wheels`). A spike confirmed a serverless `python_wheel_task` installs a wheel from an absolute `/Volumes` path, so the job never needs rewriting — and the app does not gain the ability to modify the job definition (including its `run_as`), which would let it escape the least-privilege boundary the dedicated runner SP exists to enforce. The cost is that the wheel filename is pinned in the bundle; a unit test asserts it matches `runner/pyproject.toml`, and the app warns at startup on drift.

**Catalog grants stay out of the bundle.** `resources.catalogs` does support `grants`, and I tested it: unbound, the deploy fails with `CATALOG_ALREADY_EXISTS`; bound **without** `lifecycle.prevent_destroy`, a routine `bundle destroy` deletes the user's catalog and everything in it (`catalog_name` is a free variable, so that could be `main`); bound **with** `prevent_destroy`, `bundle destroy` aborts entirely and removes nothing, so teardown would need an unbind step first. So `USE CATALOG` (and `CREATE SCHEMA` for the runner SP) are applied by a new idempotent `scripts/grant_catalog_prereqs.sh` — via the UC permissions API, so no SQL warehouse is needed. The reasoning is recorded in `databricks.yml` so it does not look overlooked.

The deploy still needs no second pass: it runs `bundle deploy` → catalog grants (resolving the app SP from the now-existing app) → `bundle run`. Granting before the app first starts is what lets it publish the wheel on its first attempt.

### Behaviour change

`bundle destroy` now removes the tmp schema and both volumes, including any result files in `mcp_results`. It previously left the schema in place. Per-user `dqx_mcp_<user>` output schemas are not bundle-managed and survive. This is called out in the uninstall docs.

There is no migration path because `mcp-server` has never shipped in a release (`v0.15.0` predates it). A pre-existing `dqx_mcp_tmp` schema makes the deploy fail with a clear error rather than adopting a schema it would later delete; `--var tmp_schema_name=` covers that case.

### Tests

Deployed on a live workspace following the rewritten docs verbatim, then torn down:

- one-command deploy succeeded; the three steps ran in the documented order with no setup job and no volume script
- the app SP was auto-resolved and granted before first start — **no restart step**
- the app published `dqx_mcp_runner-0.1.0-py3-none-any.whl` + its hash marker
- the runner job installed that wheel from the volume and completed (`TERMINATED SUCCESS`)
- native grants present on the schema and both volumes; catalog grants present for the users group, runner SP and app SP
- `make mcp-destroy` removed the app, job, schema and volumes — and left the catalog intact

Re-deployed a second time after the review fixes, with a **non-default users group**, to confirm them against real UC: `MANAGE` present on the schema for both SPs and a user-owned temp view actually dropped by the runner SP; the non-default group granted at both catalog and schema level; the wheel published and installed; teardown clean.

- [x] manually tested
- [x] added unit tests
- [ ] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

`mcp-server/tests/test_bootstrap.py` covers wheel discovery and selection, the hash-skip, upload, a swallowed failure and the drift warning, plus two pin guards: the bundle's wheel filename matches `runner/pyproject.toml`, and the runner job's `databricks-labs-dqx==<version>` matches `src/databricks/labs/dqx/__about__.py`. Both were verified to fail on a deliberate mismatch — nothing coupled the DQX pin to the repo version before, so a release would have silently left the runner on the previous one. `make fmt`, `make mcp-check` (0 errors), `make mcp-test` (168 passed) and `make docs-build` are green.

### Review follow-ups

All five findings from @mwojtyczka's review are addressed in `a7118661`, and the first was a genuine regression:

- **`MANAGE` restored on the tmp schema for both SPs.** The deleted `setup.py` granted `MANAGE` specifically so the SPs could DROP temp views; the native grants used `MODIFY`. The views are created by each *user* through their OBO token, so the user owns them and both SPs are non-owners — and UC only lets a non-owner delete an object with ownership or `MANAGE`, while `MODIFY` is data DML. Without it the runner's end-of-run drop and the app's stale-view sweeper fail on every tool call and orphan views. Verified on a live workspace: a view owned by my user was dropped by the runner SP after the run, leaving the schema empty.
- **`USERS_GROUP` now feeds the bundle as well as the grant script**, so a non-default group gets its schema grants too (previously it got `USE_CATALOG` while `USE_SCHEMA`/`CREATE_TABLE` stayed on the default). Confirmed by deploying with a non-default group.
- **The grant script fails on a real app-lookup error** instead of treating every failure as "app not deployed yet", which previously let a deploy report success with the app SP ungranted.
- **The runner wheel is selected by the pinned filename**, not `sorted()[-1]` — name order is not version order (`0.9.0` sorts after `0.10.0`), and an ambiguous `.build/` is now reported rather than resolved by guesswork. The test now uses `0.9.0` vs `0.10.0` so it fails if selection reverts to sorting.
- **Minor:** the CI wheel-wait records success in a flag rather than re-running `fs ls` (a transient error could fail a healthy deploy), and `DQX_WHEELS_VOLUME` derives from the volume resource instead of rebuilding the path from variables.

### Documentation and Demos

- [ ] added/updated demos
- [x] added/updated docs

The MCP install docs described the removed flow, so they would have misled anyone following them. Rewritten: the one-command deploy and why its order matters, a **genuinely portable no-`make`/Windows path** (Databricks CLI only, including `grants update catalog`), a new "Verify the deployment" section, an uninstall warning for the behaviour change above, a grants table marking which grants are native vs applied by the deploy, and troubleshooting rows for the new failure modes (wheel never published, version drift, and a stale-bundle-state engine fallback found during the manual test).

### Note for reviewers

This touches `mcp-server/databricks.yml` and `scripts/ci_deploy.sh`, which #1429 (MCP coverage) also touches. Whichever lands second needs a rebase — #1429's `dev-coverage` target installs the coverage wheel via a relative `./.build/...whl` path that no longer resolves once `artifact_path` is gone, and will need the absolute volume path instead.


mcp-server/databricks.yml
```diff
@@ -31,23 +37,89 @@ variables:
...
+      grants:
+        # Users create definer's-rights temp views here through their own OBO token.
+        - principal: ${var.users_group}
+          privileges: [USE_SCHEMA, CREATE_TABLE]
+        # The app reads through those views and drops stale ones (the in-app sweeper).
+        - principal: ${resources.apps.mcp-dqx.service_principal_client_id}
+          privileges: [USE_SCHEMA, SELECT, MODIFY]
+        # The runner job reads the views and writes results as its own least-privilege identity.
+        - principal: ${var.runner_service_principal_id}
+          privileges: [USE_SCHEMA, SELECT, MODIFY]
```
mwojtyczka: **Likely regression — please confirm the DROP path.** The removed `notebooks/setup.py` granted both SPs `GRANT MANAGE ON SCHEMA … dqx_mcp_tmp`, with an explicit comment that `MANAGE` (not ownership) is what lets a non-owner **DROP** a view another user created via their OBO token — and the old docs grants table said the same ("`MANAGE` on `dqx_mcp_tmp` lets the SPs drop OBO-created temp views"). These native grants use `MODIFY` instead. In UC, dropping a securable needs ownership or `MANAGE` on it (or its parent); table/schema `MODIFY` is data DML, not `DROP` of a differently-owned view. So the runner's end-of-run `DROP VIEW` of the caller's definer's-rights temp view and the app's stale-view sweeper would fail with a permission error and leak temp views into `dqx_mcp_tmp`. The PR also silently deleted the docs line that used to explain the `MANAGE` requirement, which makes me think the downgrade wasn't intentional. If the SPs still perform those drops, restore `MANAGE` for both principals here.

mcp-server/scripts/grant_catalog_prereqs.sh
```diff
@@ -0,0 +1,73 @@
...
+app = os.environ.get("DQX_APP_SP", "").strip()
+if app:
+    changes.append({"principal": app, "add": ["USE_CATALOG"]})
+print(json.dumps({"changes": changes}))
+PY
+)"
+
+echo "Applying catalog grants on ${DQX_MCP_CATALOG}:"
+echo "  ${CHANGES}"
+databricks grants update catalog "${DQX_MCP_CATALOG}" ${PROFILE_ARG[@]+"${PROFILE_ARG[@]}"} --json "${CHANGES}" >/dev/null
```
mwojtyczka, line 85: **Users-group principal is split across two grant sources.** This script grants the users group `USE_CATALOG` (from `$USERS_GROUP`, default `account users`), but its schema-level `USE_SCHEMA`/`CREATE_TABLE` on `dqx_mcp_tmp` now comes from the native bundle block (`${var.users_group}`). `make mcp-deploy` passes `USERS_GROUP` to this script but does **not** pass a matching `--var users_group` to the bundle unless the caller also puts it in `BUNDLE_VARS`. So `make mcp-deploy … USERS_GROUP=analysts` grants `analysts` `USE_CATALOG` while the schema grant stays on the default `account users` — a user in `analysts` but not `account users` can enter the catalog but can't create the temp views every data tool needs. The single `USERS_GROUP` knob should feed both the script and the bundle var (or the Makefile should append `--var users_group=` when `USERS_GROUP` is set).

mcp-server/scripts/grant_catalog_prereqs.sh
```diff
@@ -0,0 +1,73 @@
...
+
+PROFILE_ARG=()
+[ -n "${DATABRICKS_PROFILE:-}" ] && PROFILE_ARG=(--profile "$DATABRICKS_PROFILE")
+
+# Resolve the app's service principal when it wasn't supplied. The app SP is created with the app,
+# so this only succeeds after `bundle deploy` — which is exactly when the deploy flow calls this,
+# and why the install needs no second "grant, then restart" pass. A missing app (first run before
+# any deploy) is fine: the grants for users and the runner SP still apply.
+if [ -z "${APP_SP}" ] && [ -n "${APP_NAME}" ]; then
+  APP_SP="$(databricks apps get "${APP_NAME}" ${PROFILE_ARG[@]+"${PROFILE_ARG[@]}"} -o json 2>/dev/null \
```
mwojtyczka: **App-SP resolution collapses all failures to "skip the grant."** `databricks apps get … 2>/dev/null | python3 (except: print("")) … || true` makes a transient API error, an auth blip, or malformed JSON indistinguishable from "app not deployed yet." In `make mcp-deploy` this runs *after* the app exists, so an empty result there is a real error — but the script prints the "expected before first deploy" NOTE and continues, granting users + runner SP but not the app SP. The app then can't publish the runner wheel on first start and every data tool fails, while the deploy reported success. Consider failing (non-zero) when `APP_NAME` is set but resolution returns empty, so it surfaces at deploy time.

mcp-server/server/bootstrap.py
```diff
@@ -0,0 +1,136 @@
...
+    for parent in Path(__file__).resolve().parents:
+        if (parent / "requirements.txt").exists():
+            roots.extend([parent, parent / ".build"])
+            break
+    for root in roots:
+        if not root.is_dir():
+            continue
+        matches = sorted(root.glob("dqx_mcp_runner-*.whl"))
+        if matches:
+            return matches[-1]  # highest version if several builds are present
```
mwojtyczka: **`sorted()[-1]` is lexicographic, not semantic version order.** `test_prefers_highest_version_when_several_exist` asserts this picks the "highest version," but string sort disagrees with version order: `dqx_mcp_runner-0.10.0` sorts *before* `0.9.0`, and `0.2.0` sorts after `0.10.0`. If `.build/` ever holds more than one wheel, the wrong one can be selected; when it also differs from the pinned `DQX_RUNNER_WHEEL_FILENAME`, the app publishes a wheel the runner job won't install (it installs the pinned name by absolute path) and every data tool fails. Usually `.build/` has one wheel, but the code and test claim semantics they don't have — parse with `packaging.version`, or assert a single match and drop the "highest version" framing.

mcp-server/scripts/ci_deploy.sh
```diff
@@ -73,6 +109,26 @@ echo "::group::deploy + start app"
...
+# Fail loudly here rather than let the test fail later with a confusing error.
+WHEEL_DIR="dbfs:/Volumes/${DQX_MCP_TEST_CATALOG}/${TMP_SCHEMA}/dqx_artifacts"
+for _ in $(seq 1 36); do
+  if databricks fs ls "${WHEEL_DIR}" "${PROFILE_ARG[@]}" 2>/dev/null | grep -q 'dqx_mcp_runner-.*\.whl'; then
+    echo "runner wheel published under ${WHEEL_DIR}"
+    break
+  fi
+  sleep 5
+done
+if ! databricks fs ls "${WHEEL_DIR}" "${PROFILE_ARG[@]}" 2>/dev/null | grep -q 'dqx_mcp_runner-.*\.whl'; then
```
mwojtyczka: **Minor: the post-loop re-check can abort a healthy deploy.** The wait loop already `break`s on success, but this line runs a *fresh* `databricks fs ls | grep -q` to decide failure. Under `set -euo pipefail`, if that single final `fs ls` hits a transient error even though the wheel was found moments earlier, the `if !` treats it as "not published" and `exit 1` fails CI spuriously. Track success with a flag set on `break` and check that here instead of re-running `fs ls`.

mwojtyczka (commented):
Really clean PR — the move to native DAB grants is well-motivated, the two divergences from Studio (CAN_MANAGE_RUN, catalog grants out of band) are justified in-line, and the docs rewrite is thorough. The bootstrap wheel-publish is best-effort with a hash-skip and a drift guard, and the CI ordering is carefully reasoned. Findings below are mostly PLAUSIBLE (this is deploy/infra config I can't execute locally, so they're grounded in what the diff shows); the first is the one I'd confirm before merge.

1. **MANAGE → MODIFY on the tmp schema (likely regression).** The deleted `setup.py` granted the app/runner SPs `MANAGE` *specifically* so they could DROP OBO-user-owned temp views (its own comment + the old docs said so). The native grants use `MODIFY`, which in UC doesn't confer DROP on another principal's view — so the runner's end-of-run view drop and the app's stale-view sweeper would fail and orphan views. The docs line explaining the MANAGE requirement was also silently dropped.
2. **Users-group principal is split** between the script (USE_CATALOG) and the native block (USE_SCHEMA/CREATE_TABLE) — `make mcp-deploy USERS_GROUP=x` only feeds the script, so a non-default group gets catalog but not schema access.
3. **App-SP resolution swallows all errors** — a real failure looks identical to "app not deployed yet" and the deploy reports success without granting the app SP USE CATALOG.
4. **Wheel selection is a string sort, not semver** — the test asserts version-selection semantics the code doesn't have.
5. Minor: publish-path vs volume-resource derivation, and a CI wheel-wait re-check that can abort a healthy deploy under `set -e`.

Details inline.

mcp-server/databricks.yml
vb-dbrks: Good catch. restored manage for both SPs

mcp-server/scripts/grant_catalog_prereqs.sh
vb-dbrks: $USERS_GROUP now feeds both the script and the bundle var. fixed.

mcp-server/scripts/grant_catalog_prereqs.sh
vb-dbrks: Agreed. addressed now.

mcp-server/server/bootstrap.py
vb-dbrks: Agreed. it now prefers the pinned DQX_RUNNER_WHEEL_FILENAME (which is what the job installs anyway) and refuses to publish when there are several wheels and none matches...

mcp-server/scripts/ci_deploy.sh
vb-dbrks: fixed.

mcp-server/databricks.yml
mwojtyczka: Fixed in a7118661 — both SPs get MANAGE on the tmp schema again, with a comment explaining MANAGE (not MODIFY) is what permits a non-owner DROP. Verified on a live workspace. Resolving.

mcp-server/scripts/grant_catalog_prereqs.sh
mwojtyczka: Fixed in a7118661 — USERS_GROUP now feeds the bundle users_group var as well as the grant script (via an mcp_users_var guard), so both the catalog and schema grants land on the same group. Resolving.

mcp-server/scripts/grant_catalog_prereqs.sh
mwojtyczka: Fixed in a7118661 — the script now captures stderr+exit status and only skips on a genuine \"does not exist\" error; any other lookup failure exits 1. Resolving.

mcp-server/server/bootstrap.py
mwojtyczka: Fixed in a7118661 — find_runner_wheel now selects the pinned DQX_RUNNER_WHEEL_FILENAME, falls back to a sole match, and reports an ambiguous .build/ as None rather than guessing by sort. Test uses 0.9.0 vs 0.10.0 to catch a regression. Resolving.

mcp-server/scripts/ci_deploy.sh
mwojtyczka: Fixed in a7118661 — the wheel-wait records success in a WHEEL_PUBLISHED flag instead of re-running fs ls after the loop. Resolving.

mcp-server/server/bootstrap.py
```diff
@@ -0,0 +1,156 @@
...
+    ws = WorkspaceClient()
+    destination = f"{volume}/{wheel.name}"
+    try:
+        local_hash = compute_wheel_hash(wheel)
+        if local_hash == _read_remote_hash(ws, volume):
+            logger.info(f"Runner wheel unchanged ({local_hash[:12]}…); already published at {destination}")
+            return destination
+
+        with open(wheel, "rb") as handle:
+            ws.files.upload(destination, handle, overwrite=True)
```
mwojtyczka, line 144: **Non-blocking / design judgment.** On version drift (this branch's warning fires when `DQX_RUNNER_WHEEL_FILENAME` is set, the expected wheel is absent, and `find_runner_wheel` fell back to a sole differently-named match), the code warns but then still uploads to `{volume}/{wheel.name}` and writes `.wheels_hash` for it. Two consequences:

1. The runner job installs `{volume}/{expected}` by absolute path — which stays absent — so every data tool fails on library install regardless of the publish.
2. The hash marker now records the *wrong* wheel, so a later app restart with the same `.build/` short-circuits (`unchanged, already published`) and never re-uploads — the deploy can't self-heal without clearing the marker.

Publishing a wheel the job can't install is strictly worse than publishing none, and it poisons the marker. Consider returning `None` here on the `wheel.name != expected` mismatch (same as the `len(matches) > 1` ambiguity path already does). I see `test_warns_on_version_drift` + the comment at test line 117 make the publish-anyway behavior intentional, and the version-pin guard test keeps this unreachable on a real deploy — so this is a robustness nit, not a live bug. Flagging in case skip-on-drift is the cleaner contract.

mwojtyczka (commented):
Round 2 review. All five findings from the previous round are genuinely fixed and covered by tests — thanks for the thorough follow-up (MANAGE restored, USERS_GROUP feeds the bundle, the grant script distinguishes 'not found' from real errors, wheel selection is pinned-name-first, and the wheel-wait uses a success flag). I resolved those threads. The new DQX-version-pin guard test is a nice add and currently passes (repo and bundle are both 0.15.0).

One design-judgment point this round (non-blocking) — details inline.

mwojtyczka (approved):
LGTM - left one comment

PR #1436, opened 2026-08-07
Add is_valid_uuid row-level check

### LINKED ISSUE #1412: [FEATURE]: New check for UUID format validation

### Problem statement

UUIDs (RFC 9562, obsoletes RFC 4122) are a very common surrogate/dedup key format in data pipelines. DQX has no check to validate that a column's values conform to the canonical UUID format.

### Proposed Solution

Add a dedicated check validating a column against the canonical UUID format: 8-4-4-4-12 hyphenated hex groups, optionally verifying the version nibble and variant bits per RFC 9562.

### Additional Context

_No response_

SreeramaYeshwanthGowd:
## Changes

- Add `is_valid_uuid(column)`, validating a column against the canonical RFC 9562 UUID string form (8-4-4-4-12 hyphenated hex groups, case-insensitive) - obsoletes RFC 4122, same standard the issue references.
- Mirrors the existing pure pattern-match checks (`is_valid_email`, `is_valid_ipv4_address`): a new `DQPattern.UUID` regex member plus a thin `@register_rule("row")` wrapper around the shared `_matches_pattern` helper. No extra parameters.
- Also validates the version nibble (must be `1`-`8`) and variant bits (must be `8`/`9`/`a`/`b`) per RFC 9562, not just the 8-4-4-4-12 shape - addresses the issue's "optionally verifying the version nibble and variant bits" ask by making that verification part of the canonical match rather than a separate opt-in parameter.
- Consequence of that design choice: the all-zero Nil UUID and all-one Max UUID are **not** matched, since neither carries a valid version/variant. Documented explicitly in the docstring so it isn't a surprise; happy to add a relaxed/opt-out mode if a maintainer prefers accepting those two special values.
- Regex is anchored and fixed-width (no nested quantifiers or backtracking) - ReDoS-safe per the repo's regex-input guidance.

### Linked issues

Resolves #1412

### Tests

- [ ] manually tested - no live Databricks workspace available in this environment; the regex was instead hand-verified against all 18 test cases (6 valid, 12 invalid) directly in Python's `re` module before writing the Spark-based tests, so the Spark tests are checking the same logic that was already confirmed correct outside Spark
- [x] added unit tests
- [x] added integration tests
- [x] added performance tests

Validation performed:

- `make fmt`
- `make lint`
- `make test` 

### Documentation and Demos

- [x] added/updated docs


src/databricks/labs/dqx/check_funcs.py
```diff
@@ -111,6 +111,10 @@ class DQPattern(Enum):
     # 000/666/9xx (9xx covers ITINs), group 00, serial 0000. Anchored, fixed-width; ReDoS-safe.
     SSN_US = r"^(?!000|666|9\d{2})\d{3}([- ]?)(?!00)\d{2}\1(?!0000)\d{4}$"
 
+    # Canonical UUID form per RFC 9562: 8-4-4-4-12 hex groups; version nibble
+    # restricted to 1-8, variant bits to 8/9/a/b. Anchored, fixed-width; ReDoS-safe.
+    UUID = r"^[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-[1-8][0-9A-Fa-f]{3}-[89ABab][0-9A-Fa-f]{3}-[0-9A-Fa-f]{12}$"
```
mwojtyczka: **Strict version/variant validation is mandatory, but issue #1412 asks for it to be *optional*.** The issue's proposed solution: "validate ... the canonical UUID format: 8-4-4-4-12 hyphenated hex groups, **optionally** verifying the version nibble and variant bits per RFC 9562." This pattern hardcodes `[1-8]` version + `[89ABab]` variant into the match with no way to relax it. Two consequences (verified locally vs Python's `uuid.UUID`, which accepts all of these):
- **Nil UUID** `00000000-...-000000000000` and **Max UUID** `ffffffff-...` are rejected — both are RFC-defined and extremely common sentinels.
- Legacy **Microsoft/COM GUIDs** with a variant nibble in `c`–`f` are rejected.

For a check named `is_valid_uuid`, flagging values that every UUID library considers valid is a surprising, high-false-positive default. Suggest matching the issue: validate the 8-4-4-4-12 shape by default and gate version/variant checks behind an opt-in arg (e.g. `strict: bool = False`), or at minimum accept Nil/Max.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -1164,6 +1168,25 @@ def is_valid_national_id(column: str | Column, country: str = "US") -> Column:
...
+
+    Null values will pass the check with no violation reported.
+
+    Args:
+        column: column to check; can be a string column name or a column expression
+
+    Returns:
+        Column object for condition
+    """
+    return _matches_pattern(column, DQPattern.UUID)
```
mwojtyczka: Minor, and inherited (not introduced here): because `_matches_pattern` applies this via `rlike` (Java regex, no MULTILINE), the trailing `$` also matches just before a single trailing `\n`, so `"550e8400-e29b-41d4-a716-446655440000\n"` passes. Verified locally. Same latent behavior affects the sibling email/SSN/IPv4 patterns, so it's arguably out of scope for this PR — noting it because the new UUID check re-exposes it. A `\z` anchor (or `\A...\z`) would be strict; only worth doing if the team wants to fix the whole family.

docs/dqx/docs/reference/quality_checks.mdx
```diff
@@ -60,6 +60,7 @@ You can also define your own custom checks in Python (see [Creating custom check
...
+| `is_valid_uuid`                    | Checks whether the values in the input column are valid UUIDs (RFC 9562, canonical 8-4-4-4-12 hyphenated hex form).                                                                                                                                                                                                                                                                                                               | `column`: column to check (can be a string column name or a column expression)                                                                                                                                                                                                                                                                                                                                                                                                                                                                       |
```
mwojtyczka: Nit: the docs table places `is_valid_uuid` between `is_valid_email` and `has_valid_string_case`, but in check_funcs.py it's defined after `is_valid_national_id`. Sibling checks are otherwise grouped in parallel order across docs and source — worth aligning to avoid drift. Non-blocking.

mwojtyczka (commented):
## Code Review — Add `is_valid_uuid` row-level check

Clean implementation that mirrors the existing pattern-match checks (`DQPattern.UUID` + thin `@register_rule("row")` wrapper over `_matches_pattern`). The regex is correct for its stated strict intent, anchored and fixed-width (ReDoS-safe), and the PR description is commendably explicit about the Nil/Max UUID trade-off. Tests + docs included.

**Verified locally** against Python's `uuid.UUID`: the pattern rejects the Nil UUID, Max UUID, and any UUID whose variant nibble is `c`–`f` (e.g. legacy Microsoft/COM GUIDs), all of which `uuid.UUID` accepts.

**Verdict:** approve-worthy on quality; the mandatory-vs-optional strict validation is a product call worth settling first. Findings are inline — none are correctness defects in the strict interpretation, but #1 materially affects real-world false-positive rates.

mwojtyczka (changes requested):
Generally looking good, left a few comments

SreeramaYeshwanthGowd:
@mwojtyczka thanks for the review, all three points are addressed, and for the trailing newline I opened #1440 to fix the whole pattern family separately since it affects email, SSN and IPv4 too.

mwojtyczka (approved):
LGTM

PR #1442, opened 2026-08-08
Fix filtered is_unique violations

### LINKED ISSUE #1441: [BUG]: Filtered is_unique flags rows outside the filter

- [x] I searched open and closed issues and pull requests.

### Current behavior

`is_unique` with a `filter` / `row_filter` counts only matching rows, but applies the resulting duplicate count to every row in the same key partition.

For example, given three rows with the same key where two match the filter and one does not, all three rows are reported as duplicates. The excluded row can therefore be quarantined or rejected even though the rule does not apply to it.

```python
df = spark.createDataFrame(
    [("duplicate", True), ("duplicate", True), ("duplicate", False)],
    "key string, included boolean",
)

condition, apply = is_unique(["key"], row_filter="included")
apply(df).select("key", "included", condition).show()
```

The row where `included = false` incorrectly receives the same `not unique` violation as the two included rows.

### Expected behavior

Only rows matching the filter should be evaluated and reported. Rows outside the filter should never receive a violation from that rule, even when they share a key with matching duplicates.

### Root cause

The windowed count is correctly conditional on the filter, but the final condition checks only whether that count exceeds one. It does not also require the current row to match the filter.

### Proposed fix

Gate the duplicate condition with the existing filter condition and add a regression test covering included and excluded rows in the same key partition.

mattfaltyn:
## Changes

- require the current row to match `is_unique`'s filter before reporting a duplicate violation
- add a regression test where filtered and unfiltered rows share the same duplicate key

### Linked issues

Fixes #1441

### Tests

- [x] manually tested with standalone PySpark 3.5.5
- [ ] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

Validation performed:

- `.venv/bin/pytest tests/unit -q -n 0` — 1,674 passed
- `black --check` and `ruff check` on both changed files — passed
- full-project mypy — passed
- focused repository integration test collected successfully but requires `DATABRICKS_HOST`; the same regression scenario passed against local PySpark

### Documentation and Demos

- [ ] added/updated demos
- [ ] added/updated docs

No documentation change: this restores the documented filter semantics without changing the public API.


mwojtyczka (approved):
LGTM

PR #1450, opened 2026-08-11
Feature: Categorical values distribution

### LINKED ISSUE #1344: [FEATURE]: New check for categorical values distribution

### Problem statement

Current `is_in` check can verify whether column values are within allowed enumeration. However, in some cases more advanced verification is needed to check whether categorical value distribution is close to configured. Hypothetical example: sales deal states should be distributed approximately like: close won - 40%, close lost - 10%, open - 50%;

### Proposed Solution

Use [Total variation distance of probability measures](https://en.wikipedia.org/wiki/Total_variation_distance_of_probability_measures) measurement to compare two relative distributions between test and data with some threshold. Generate check via profiler if all conditions matches for `make_is_in_profile` profiler 

### Additional Context

_No response_

IvannKurchenko:
## Changes
New check: Categorical values distribution check at dataset level using TDV measurement.

### Linked issues

Resolves https://github.com/databrickslabs/dqx/issues/1344

### Tests

- [ ] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [x] added performance tests

### Documentation and Demos

- [x] added/updated demos
- [x] added/updated docs


src/databricks/labs/dqx/check_funcs.py
```diff
@@ -460,6 +461,313 @@ def is_in_list(column: str | Column, allowed: list, case_sensitive: bool = True)
...
+        # This avoids groupBy(col).collect() which would pull an unbounded number of distinct
+        # values to the driver on high-cardinality columns.
+        expected_keys_ordered = list(normalized_distribution.keys())
+        per_key_aliases = [f"__cnt_{i}" for i in range(len(expected_keys_ordered))]
+        aggregations = [
+            F.count(F.when(group_expr == F.lit(key), F.lit(1))).alias(alias)
+            for key, alias in zip(expected_keys_ordered, per_key_aliases)
+        ]
+        aggregations.append(F.count(group_expr).alias("__total"))
+        aggregate_row = filtered.agg(*aggregations).collect()[0]
```
mwojtyczka: **[correctness] Eager `.collect()` in the `apply()` closure breaks streaming.**

Every other dataset check (`is_unique`, `has_no_outliers`, `aggr_*`, `foreign_key`) keeps its closure lazy — only transformations, returning `df.withColumn(...)`. `is_in_distribution` is the only one that calls `filtered.agg(...).collect()` inside `apply`, which the executor invokes during `apply_checks`.

Failure: `apply_checks()` on a streaming DataFrame reaches `closure_func(df)` → `.collect()`, which raises `AnalysisException` ('Queries with streaming sources must be executed with writeStream.start()'). In batch it also forces an extra eager job mid-plan. Consider a lazy formulation (broadcast-join the expected distribution and compute TVD via aggregation expressions).

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -460,6 +461,313 @@ def is_in_list(column: str | Column, allowed: list, case_sensitive: bool = True)
...
+    _validate_distance(distance)
+
+    col_str_norm, col_expr_str, col_expr = get_normalized_column_and_expr(column)
+
+    unique_str = uuid.uuid4().hex
+    condition_col = f"__condition_{col_str_norm}_{unique_str}"
+    message_col = f"__message_{col_str_norm}_{unique_str}"
+
+    def apply(df: DataFrame) -> DataFrame:
+        column_type = df.schema[col_expr_str].dataType
```
mwojtyczka: **[correctness] `df.schema[col_expr_str]` KeyErrors for a `Column` expression input.**

`col_expr_str` is the *rendered expression* (e.g. `upper(v)`), not a schema field name. For `is_in_distribution(F.upper(F.col('v')), ...)`, `df.schema['upper(v)']` raises `KeyError` since the schema only has field `v`. The signature/docstring advertise column expressions, but only `F.col("value")` (name == field) is tested. Same pattern exists in `has_no_outliers`, so it's a shared latent gap rather than net-new — worth a test and/or resolving the type off the expression.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -460,6 +461,313 @@ def is_in_list(column: str | Column, allowed: list, case_sensitive: bool = True)
...
+            deviations = [abs(normalized_distribution[k] - actual_counts.get(k, 0) / total) for k in expected_keys]
+            deviations.append(abs(expected_residual - actual_residual))
+            tvd = 0.5 * math.fsum(deviations)
+
+            # math.isclose neutralises IEEE-754 noise at the boundary — otherwise a "supposed to
+            # pass" case with tvd mathematically equal to distance can trip when float rounding
+            # nudges the computed tvd just above distance (e.g. abs(0.15-0.2) == 0.05000000000000002).
+            # The docstring's "less than or equal to the given distance" already promises the
+            # boundary passes, so equality-within-precision must not fire the check.
+            is_violation = tvd > distance and not math.isclose(tvd, distance)
```
mwojtyczka: **[correctness] int keys vs byte/short/date columns rely on implicit Spark upcasting.**

For a `byte` column with `{1: 0.75, 2: 0.25}`, `F.lit(1)` is an INT literal compared to a BYTE column. Spark upcasts so the tests pass, but there's no explicit cast of the literal to the column type. Residual/total bucketing depends on this equality holding for *every* supported small-int/date type; a type where Spark equality semantics differ (or a future type addition) would silently miscount into the residual bucket rather than erroring. An explicit cast of the key literal to the column type would make this robust.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -460,6 +461,313 @@ def is_in_list(column: str | Column, allowed: list, case_sensitive: bool = True)
...
+            message_text = (
+                f"Column '{col_expr_str}' actual distribution deviates from expected by TVD={tvd:.6f}, "
+                f"which exceeds the allowed distance={distance}."
+            )
+
+        return df.withColumn(condition_col, F.lit(is_violation)).withColumn(message_col, F.lit(message_text))
```
mwojtyczka: **[altitude] Verdict baked as `F.lit(python_bool)` from the driver-side collect result.**

Because the TVD is computed on the driver from collected counts, `is_violation` is a Python bool turned into `F.lit`, and the whole dataset shares a single plan-baked verdict. This is a design consequence of the `.collect()` approach (finding on line 599) and is what prevents the check from participating in a lazy/streaming plan. Reformulating to compute TVD with aggregation expressions would remove both the collect and the literal-baking.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -460,6 +461,313 @@ def is_in_list(column: str | Column, allowed: list, case_sensitive: bool = True)
...
+
+    _validate_distribution_items(distribution)
+    _validate_distribution_values(distribution)
+    _validate_distribution_keys_types(distribution)
+    _validate_distribution_keys_collision(case_sensitive, distribution)
+
+
+def _validate_distribution_values(
+    distribution: dict[bool, float] | dict[str, float] | dict[int, float] | dict[datetime.date, float]
+):
```
mwojtyczka: **[conventions] Missing `-> None` return annotations.**

`AGENTS.md` (Type Hints): *"Every parameter and return value must be annotated. Enforced by mypy (make lint)."* `_validate_distribution_values`, `_validate_distribution_keys_collision`, `_validate_distribution_keys_types`, and `_validate_distribution_items` end in `):` with no `-> None`. `_validate_distribution` and `_validate_distance` in the same block do annotate `-> None`, so these four are inconsistent.

tests/resources/all_dataset_checks.yaml
```diff
@@ -264,4 +264,18 @@
       time_column: col5
       aggr_type: avg
       lookback_num_intervals: 2
-      warmup_num_intervals: 2
\ No newline at end of file
+      warmup_num_intervals: 2
```
mwojtyczka, line 267: **[test-coverage] Fixture uses `distance: 1.0` (the TVD maximum), so it can never fail.**

`is_violation = tvd > 1.0` is unreachable (TVD is bounded by 1). The 'apply every check' regression fixture therefore only verifies the check *runs* without error, never that a real distribution mismatch is caught end-to-end through the YAML/metadata path. A regression that inverts the comparison or mis-buckets residual mass would still pass. Consider a second fixture entry with a distribution + small `distance` that is expected to flag.

mwojtyczka (commented):
Automated code review (high-effort recall pass) for `is_in_distribution`. 6 findings inline, ranked most-severe first. Correctness items #1–#3 are marked *plausible* — verified by reading the sibling dataset-check closures rather than executing Spark locally.

This review was generated with assistance from Claude Code.

IvannKurchenko:
Hello, @mwojtyczka! Thanks for a review. The previous feedback has been addressed. Would it be possible to have another round?

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -510,6 +524,374 @@ def is_in_list(column: str | Column, allowed: list, case_sensitive: bool = True)
...
+        missing_array = F.array_sort(F.filter(F.array(*missing_key_names), lambda value: value.isNotNull()))
+        has_missing_column = F.size(missing_array) > 0
+        missing_message_column = F.concat(
+            F.lit(f"Column '{column_expression}' distribution is missing expected keys ["),
+            F.array_join(missing_array, ", "),
+            F.lit("] and impute=False."),
+        )
+        is_violation_column = (
+            F.when(total_column == 0, F.lit(False))
+            .when(has_missing_column, F.lit(True))
```
mwojtyczka: **[correctness] `impute=False` flags a 0.0-probability key as 'missing'.** `has_missing_column` is true whenever any expected key has a per-key count of 0, so a key whose expected probability is exactly `0.0` (which validation permits) is reported as a violation when it is legitimately absent. Example: `distribution={A: 0.0, B: 1.0}`, data all `B` -> 'missing expected keys [A]', yet with `impute=True` the same data yields TVD=0 and passes. The two modes disagree on identical data. Consider excluding expected-probability-0 keys from the missing-key set.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -510,6 +524,374 @@ def is_in_list(column: str | Column, allowed: list, case_sensitive: bool = True)
...
+        tvd_col = F.lit(0.5) * reduce(py_operator.add, deviation_terms)
+
+        # math.isclose (rel_tol=1e-09, abs_tol=0.0) as a Column expression: neutralises IEEE-754
+        # noise at the boundary so a "supposed to pass" case with tvd mathematically equal to
+        # distance can't trip when float rounding nudges the computed tvd just above distance
+        # (e.g. abs(0.15-0.2) == 0.05000000000000002). The docstring's "less than or equal to the
+        # given distance" already promises the boundary passes, so equality-within-precision must
+        # not fire the check.
+        distance_lit = F.lit(distance)
+        is_close_col = F.abs(tvd_col - distance_lit) <= F.lit(1e-9) * F.greatest(F.abs(tvd_col), F.abs(distance_lit))
```
mwojtyczka: **[correctness] `is_close` uses `abs_tol=0`, leaving the `distance=0` boundary unprotected against float noise.** The tolerance is `1e-9 * max(|tvd|, |distance|)`, which collapses to ~0 when `distance=0` and `tvd` is ~1e-16. For three equal `1/3` categories with a 1:1:1 sample, `expected_residual = 1 - fsum(1/3,1/3,1/3) = 1.1e-16` (L737) gives `tvd = 5.5e-17`; the tolerance is ~`5.5e-26`, so `~is_close` is True and `tvd > 0` -> every row is flagged despite a mathematically perfect match. `distance=0` is exercised throughout the tests. `math.isclose`'s default `abs_tol=0` is unsafe against zero — add a small absolute floor to the tolerance or special-case `distance=0`.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -510,6 +524,374 @@ def is_in_list(column: str | Column, allowed: list, case_sensitive: bool = True)
...
+    condition_column_name: str,
+    message_column_name: str,
+) -> DataFrame:
+    if impute:
+        is_violation_column = F.when(total_column == 0, F.lit(False)).otherwise(tvd_violation_column)
+        message_expression = F.when(total_column == 0, F.lit("")).otherwise(tvd_message_column)
+    else:
+        # Missing keys are those whose per-key count is zero; sort for stable output.
+        missing_key_names = [
+            F.when(F.col(alias) == 0, F.lit(repr(key))) for key, alias in zip(expected_keys, per_key_aliases)
```
mwojtyczka: **[minor] With `case_sensitive=False`, the 'missing keys' message echoes the case-normalized key.** `expected_keys` here comes from `normalized_distribution` (lowercased at L641-642), so `repr(key)` renders `'gold'` rather than the user's `'Gold'` — the violation text reports a key the user never wrote. Carry the original key alongside the normalized one for the message.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -510,6 +524,374 @@ def is_in_list(column: str | Column, allowed: list, case_sensitive: bool = True)
...
+
+    def apply(df: DataFrame) -> DataFrame:
+        # Resolve the type off the expression itself
+        column_type = _is_in_distribution_get_data_type(df, col_expr, col_expr_str)
+
+        filtered = df.filter(safe_filter_expr(row_filter)) if row_filter else df
+
+        is_string_column = isinstance(column_type, (types.StringType, types.CharType))
+        group_expr = col_expr
+        normalized_distribution: dict[Any, float] = distribution
```
mwojtyczka: **[conventions] `Any` in new type hints.** `dict[Any, float]` here (and `list[Any]` / `dict[Any, float]` on the helper signatures) — AGENTS.md Type Hints: *'Unknown types -> object or Protocol, not Any ... New code outside [anomaly/] must not introduce it.'* The supported key types are a bounded set; `object` or a bounded union satisfies the rule with identical behavior.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -15,6 +16,7 @@
 import pyspark.sql.functions as F
 from pyspark.sql import types
 from pyspark.sql import Column, DataFrame, SparkSession
+from pyspark.sql.types import DataType
```
mwojtyczka: **[minor] Redundant import.** `from pyspark.sql.types import DataType` duplicates the already-imported `types` module used everywhere else in this file (`types.StringType`, etc.). `_is_in_distribution_get_data_type` can return `types.DataType`; dropping this line keeps the file's single-import convention.

mwojtyczka (commented):
## Follow-up review pass

**Resolved (verified against HEAD — threads now closed):** the earlier `.collect()`/streaming (L599), driver-baked verdict (L646), `df.schema[expr]` KeyError (L565), implicit int upcast (L640), and missing `-> None` (L691) concerns are all addressed by the refactor to lazy conditional aggregation + explicit key casts. Marked those five threads resolved.

**Still open:** the YAML fixture thread (`all_dataset_checks.yaml:267`) is only *partially* addressed — the new second entry (TVD=0.5, distance=1.0) guards against a comparison *inversion*, but no fixture entry actually flags a mismatch end-to-end (both entries use `distance: 1.0` and pass by design). Left open so you can decide whether positive-detection belongs here or in the integration test.

**New findings (inline):** two correctness edge cases (0.0-probability key under `impute=False`; `is_close` `abs_tol=0` at `distance=0`) plus three minor items.

mwojtyczka (changes requested):
Left some additional comments, previous comments resolved except one only partilally addressed.

src/databricks/labs/dqx/check_funcs.py
IvannKurchenko: Thank you. Eventually I decided to remove `impute` parameter as it brings more problems then convenience.

src/databricks/labs/dqx/check_funcs.py
IvannKurchenko: Thanks, removed scaling

src/databricks/labs/dqx/check_funcs.py
IvannKurchenko: Removed imputation logic

IvannKurchenko:
@mwojtyczka May I ask for another review round, please? Your previous comments has been addressed.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -581,6 +597,315 @@ def is_in_list(column: str | Column, allowed: list, case_sensitive: bool = True)
...
+                _VALUES_DISTRIBUTION, {k.lower() if isinstance(k, str) else k: v for k, v in distribution.items()}
+            )
+
+        expected_keys_ordered = list(normalized_distribution.keys())
+        per_key_aliases = [f"__cnt_{i}" for i in range(len(expected_keys_ordered))]
+
+        # Explicit cast pins the literal to the column type so byte/short/date keys don't rely
+        # on implicit Spark upcasting — a future type addition with different equality semantics
+        # would otherwise silently miscount into the residual bucket instead of erroring.
+        key_lits = [F.lit(k).cast(column_type) for k in expected_keys_ordered]
```
mwojtyczka, line 725: **[correctness] No validation that distribution key types are compatible with the column type.**

`_is_in_distribution_validate_distribution` checks key *homogeneity* and supported key types, and `_is_in_distribution_get_data_type` checks the column type — but nothing verifies the two match. The explicit `F.lit(k).cast(column_type)` added here then silently maps an incompatible key to `NULL`: e.g. `is_in_distribution` on an `IntegerType` column with string keys `{"A": 0.5, "B": 0.5}` casts both keys to `NULL`, so `group_expr == key_lit` never matches, every value falls into the residual bucket, and TVD computes to `1.0` — flagging every row with a spurious `TVD=1.000000`. Under ANSI SQL mode the cast throws mid-aggregation instead.

Consider validating that the key type matches the resolved column type and raising `InvalidParameterError` (the docstring already promises keys must match the column type).

tests/resources/all_dataset_checks.yaml
```diff
@@ -264,4 +264,33 @@
...
+# exactly (TVD=0). distance=1.0 keeps the check safely passing on the perf fixture too,
+# where col10 carries arbitrary integer values.
+- criticality: error
+  check:
+    function: is_in_distribution
+    arguments:
+      column: col10
+      distribution:
+        2: 1.0
+      distance: 0.0
```
mwojtyczka, line 282: **[test-coverage] Fixture comment contradicts the `distance` value.**

The comment above says *"distance=1.0 keeps the check safely passing on the perf fixture too, where col10 carries arbitrary integer values"*, but this entry uses `distance: 0.0`. With `{2: 1.0}` / `distance: 0.0` against the perf fixture's arbitrary `col10` ints, TVD is well above 0, so the check flags essentially every row there — it does *not* pass safely. It only stays non-breaking because the perf test asserts row count, not violations.

Either restore `distance: 1.0` (matching the comment) or update the comment to reflect `0.0`.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -581,6 +597,315 @@ def is_in_list(column: str | Column, allowed: list, case_sensitive: bool = True)
...
+
+        total_col = F.col("__total")
+        deviation_terms = _is_in_distribution_get_deviation_terms(
+            normalized_distribution,
+            per_key_aliases,
+            total_col,
+        )
+        tvd_col = F.lit(0.5) * reduce(py_operator.add, deviation_terms)
+
+        # math.isclose (rel_tol=1e-09, abs_tol=0.0) as a Column expression: neutralises IEEE-754
```
mwojtyczka: **[minor] Comment describes the inverse of the implemented tolerance.**

The comment says this replicates `math.isclose (rel_tol=1e-09, abs_tol=0.0)`, but the code implements a fixed *absolute* tolerance: `F.abs(tvd_col - distance_lit) <= 1e-9` is effectively `abs_tol=1e-9, rel_tol=0`. `math.isclose` with `abs_tol=0.0` collapses to exact equality when `distance == 0` (its tolerance is `1e-9 * max(|tvd|, |distance|)`), which is the opposite of what is written here.

The absolute-tolerance behaviour is arguably the better choice (it actually protects the `distance=0` boundary), but the comment should describe it accurately.

mwojtyczka (changes requested):
Re-reviewed: previous comments are addressed, posted 1 bug and 2 nits regarding comments

src/databricks/labs/dqx/check_funcs.py
IvannKurchenko: Thank you for the comment. This has been addressed in dcd8dc7.

tests/resources/all_dataset_checks.yaml
IvannKurchenko: Thank you for the comment. This has been addressed in dcd8dc7.

src/databricks/labs/dqx/check_funcs.py
IvannKurchenko: Thank you for the comment. This has been addressed in dcd8dc7.

mwojtyczka (commented):
LGTM - will trigger tests and merge if all is green

mwojtyczka (approved):
Thank you for the fixes. LGTM - will trigger tests and merge if all is green

PR #1453, opened 2026-08-11
fix(check_funcs): support F.col("*") in aggregate checks with a filter

### LINKED ISSUE #1435: [BUG]: _is_aggr_compare does not support * column value when created with a filter applied

### Current Behavior

When defining a dataset level check with an aggregation condition over '*' (wildcard), if a filter is also applied, it results in:
```
[[INVALID_USAGE_OF_STAR_OR_REGEX](https://learn.microsoft.com/azure/databricks/error-messages/error-classes#invalid_usage_of_star_or_regex)] Invalid usage of '*' in expression `casewhen`. SQLSTATE: 42000
```

This is due to how the `apply` closure within `_is_aggr_compare` passes the column value (which is either `F.col("*")` or `F.expr("*")` at this point depending on whether the check was constructed programmatically or from metadata) into a `when` clause: https://github.com/databrickslabs/dqx/blob/e7b1109fcd670ff315bd6e42f9cda857280f7ab5/src/databricks/labs/dqx/check_funcs.py#L3376

(which is not allowed, as the wildcard cannot be expanded in that context).

### Expected Behavior

The check should function as expected.

Likely, the "*" corner case needs to be given special consideration, and substituted for a sentinel/placeholder value instead, when it is injected into the when clause.

### Steps To Reproduce

Any check which contains both an aggr function, a filter, and column '*', e.g.

```yaml
- filter: x = 'y'
  check:
    function: is_aggr_not_less_than
    arguments:
      aggr_type: count
      column: '*'
      limit: 100
```

mwojtyczka:
## Changes

Fix `INVALID_USAGE_OF_STAR_OR_REGEX` when a dataset-level aggregate check aggregates over `*` **and** a filter is applied and the check is constructed with `F.col("*")`.

The aggregate helpers substitute the `*` placeholder inside the `CASE WHEN` they build for a `row_filter` only when the aggregated column stringifies to `"*"`. That covers the string `"*"` and `F.expr("*")` forms (both normalize to `"*"`), but a programmatically constructed `F.col("*")` stringifies to `"unresolvedstar()"`, so the guard was skipped and the star landed in the `CASE WHEN ... THEN` clause, which Spark rejects.

### What the fix does

- Extracts shared helpers — `resolve_aggregate_column` (canonicalizes every `*` form to the `("", "*")` name pair) and `build_filtered_aggregate_input` (wraps the column in the row filter, substituting a non-null literal placeholder for `count(*)`).
- Uses them in **`_is_aggr_compare`** (`is_aggr_not_less_than` / `is_aggr_not_greater_than` / `is_aggr_equal` / `is_aggr_not_equal`), in **`has_no_aggr_outliers`** (identical bug on its documented `count(*)` path), and in **`aggr_matches_dataset`** — which also resolves the *reference* column through `resolve_aggregate_column`, so a `count(*)` comparison built with `F.col("*")` reports `'*'` consistently on both sides of the message.
- **Validates star/aggregate combinations** (`validate_star_aggregate`) at build time, two-tier:
  - *Native path* (unfiltered, and the `aggr_matches_dataset` reference side which uses `DataFrame.filter`): `count` and `count_distinct` both accept `*`; single-arg aggregates (sum/avg/…) are rejected with a clear error instead of failing later in Spark.
  - *Filtered-count placeholder path* (a `row_filter` on the checked/outlier side): `count` only — the `F.lit(1)` placeholder makes `count_distinct` collapse to 1 and other aggregates meaningless.

  This keeps existing `count(DISTINCT *)` usage working (no upgrade break) while still rejecting the silently-wrong filtered case. Comparison is case-sensitive, matching `getattr(F, aggr_type)`.
- Star detection matches only the exact rendered forms `{"*", "unresolvedstar()"}` (pinned by a unit test), so a real column literally named `unresolvedstar` isn't misclassified.
- Documents `InvalidParameterError` in the `is_aggr_*` and `has_no_aggr_outliers` docstrings.

The declarative repro from the issue (string `column: '*'` + filter) was already fixed by #1309; this closes the remaining programmatic `F.col("*")` gap, generalizes the fix across the aggregate check family, and centralizes the logic so future aggregate helpers inherit it.

### Linked issues

Closes #1435

### Tests

- [x] added regression tests
- `make lint` (black, ruff, mypy, pylint 10/10) and unit tests pass.
- **Unit** (`tests/unit/test_dataset_checks.py`, `tests/unit/test_has_no_aggr_outliers.py`): star-form canonicalization; two-tier `validate_star_aggregate` (count/count_distinct native vs count-only placeholder, case-sensitivity, single-arg rejection) across all three star forms and both `is_aggr_*` and the `aggr_matches_dataset` reference column.
- **Integration** — verified against a live workspace:
  - `test_is_aggr_count_star_column_expr_with_row_filter` — programmatic `F.col("*")` + `row_filter`.
  - `test_is_aggr_count_distinct_star_unfiltered` — `count(DISTINCT *)` runs end-to-end.
  - `test_is_aggr_star_column_expr_unfiltered_runs_natively` — `F.col("*")` `count`/`count_distinct` run natively (no filter), end-to-end.
  - `test_aggr_matches_dataset_count_distinct_star_runs_natively` — reference-side `count(DISTINCT *)` (`F.col("*")`) runs end-to-end.
  - `test_aggr_matches_dataset_star_as_column_expression` — `F.col("*")` reports `'*'` on both sides.
  - `test_apply_aggr_star_check_with_filter_by_metadata` — the exact declarative repro, asserted via `assert_df_equality`.
  - `test_has_no_aggr_outliers.py::test_count_star_aggr_type_with_row_filter` — `count(*)` outlier check with a `row_filter`.

### Documentation and Demos

- [ ] added/updated demos
- [x] added/updated docs


src/databricks/labs/dqx/check_funcs.py
```diff
@@ -2731,7 +2732,9 @@ def aggr_matches_dataset(
         ref_group_by_names = get_columns_as_strings(resolved_ref_group_by, allow_simple_expressions_only=True)
 
     ref_column = column if ref_column is None else ref_column
-    _, ref_col_str, ref_col_expr = get_normalized_column_and_expr(ref_column)
+    # Canonicalize the star the same way the checked side does (via _is_aggr_compare) so a count(*)
+    # comparison built with F.col("*") reports '*' consistently on both sides of the message. See #1435.
+    _, ref_col_str, ref_col_expr = resolve_aggregate_column(ref_column)
```
vb-dbrks, line 2794: `validate_star_aggregate` only runs on the checked column, via `_is_aggr_compare` — the ref side isn't guarded, so `aggr_matches_dataset(column="id", ref_column="*", aggr_type="sum")` gets past it and fails in Spark instead.

Not a correctness issue: the ref side uses a real `.filter()` rather than a `CASE WHEN`, so there's no placeholder to silently turn it into a count. Deliberate, or worth guarding for the error message?

vb-dbrks (commented):
Nice generalisation — centralising the star handling rather than patching each call site is the right shape, and pinning `unresolvedstar()` in a unit test is a good guard for something Spark could rename.

One docs thing, outside the diff so I can't leave a suggestion on it. `docs/dqx/docs/reference/quality_checks.mdx:2199` — the `has_no_aggr_outliers` example uses `column: '*'`, and the inline comment offers exactly the aggregates this PR now rejects:

```yaml
      column: '*'
      time_column: event_date
      aggr_type: count              # other types: sum, avg, min, max, stddev, etc.
```

With `column: '*'` every type in that list now raises `InvalidParameterError`, so the example contradicts the new validation. Suggest:

```yaml
      aggr_type: count              # '*' supports count only; other aggregates need an explicit column
```

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: Good call — guarded it rather than leaving it to Spark. `aggr_matches_dataset` now runs `validate_star_aggregate` on the reference column too (commit 2c666296), so `ref_column="*"` (or `F.col("*")`) with a non-count aggregate fails at build time with the clear `InvalidParameterError` instead of an opaque `INVALID_USAGE_OF_STAR_OR_REGEX` in Spark. The checked side keeps its guard via `_is_aggr_compare`.

mwojtyczka:
@vb-dbrks thanks for the review!

- **Docs example**: fixed in 4e02d744 — `docs/dqx/docs/reference/quality_checks.mdx` now reads `aggr_type: count  # '*' supports count only; other aggregates need an explicit column`, matching the new `validate_star_aggregate` behavior.
- **Reference-column guard**: addressed in 2c666296 (replied inline) — `aggr_matches_dataset` now validates the reference column too, so a star `ref_column` with a non-count aggregate fails at build time.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -4563,6 +4561,85 @@ def get_normalized_column_and_expr(column: str | Column) -> tuple[str, str, Colu
...
+    Args:
+        aggr_col_str: Canonicalized display name of the column (as returned by *resolve_aggregate_column*).
+        aggr_type: The aggregate function name.
+
+    Raises:
+        InvalidParameterError: If *aggr_col_str* is the star *"*"* and *aggr_type* is not *count*.
+    """
+    # Compare case-sensitively: aggregate names are resolved case-sensitively elsewhere (getattr(F, aggr_type)
+    # in _build_aggregate_expression), so only the exact lowercase "count" is a valid count(*).
+    if aggr_col_str == "*" and aggr_type != "count":
```
vb-dbrks: Thanks for the quick turnaround on both of those.

One more on this guard: `count(DISTINCT *)` is valid Spark, so this rejects a combination that works today. `count_distinct` takes varargs (`(col, *cols)`) so the star expands fine, unlike `sum`/`avg` which are single-arg and genuinely fail. Checked on local Spark 4.0.0 (same major as `databricks-connect~=17.3` bundles), 5 rows / 3 distinct:

```
count(*)                                  = 5
count_distinct(*)                         = 3      <- works today
count_distinct(F.col("*"))                = 3      <- works today
sum(*)                                    -> WRONG_NUM_ARGS: sum requires 1 parameters but the actual number is 3
avg(*)                                    -> WRONG_NUM_ARGS
count(CASE WHEN eu THEN 1 END)            = 3      <- the placeholder, correct for count
count_distinct(CASE WHEN eu THEN 1 END)   = 1      <- the placeholder, silently wrong
count(CASE WHEN eu THEN * END)            -> INVALID_USAGE_OF_STAR_OR_REGEX
```

So the silent-wrongness this guard exists to prevent is real, but only on the **filtered** path — that last-but-one line is exactly it. Unfiltered, `count_distinct(*)` never touches `F.lit(1)` and returns the right answer.

Net effect: an existing check like `{"column": "*", "aggr_type": "count_distinct"}` runs fine on the current release and raises `InvalidParameterError` at build time after this PR. Would gating the rejection on `row_filter` being set, or allowlisting `count_distinct` alongside `count`, keep the fix without the upgrade break?

Minor while you're here: the new `InvalidParameterError` isn't listed in the `Raises:` section of the four `is_aggr_*` docstrings or `has_no_aggr_outliers`.

mwojtyczka: Great catch — you're right that `count(DISTINCT *)` works today, so a blanket rejection was an upgrade break. Fixed in dc5337fb by making `validate_star_aggregate` two-tier:

- **Native path** (unfiltered, and the `aggr_matches_dataset` reference side which uses `DataFrame.filter`): `count` **and** `count_distinct` are allowed; single-arg aggregates (sum/avg/…) are rejected at build time with a clear error rather than failing later in Spark.
- **Filtered-count placeholder path** (a `row_filter` on the checked/outlier side): `count` only, since the `F.lit(1)` placeholder collapses `count_distinct` to 1 — exactly your `count_distinct(CASE WHEN … THEN 1 END) = 1` line.

So `{"column": "*", "aggr_type": "count_distinct"}` runs unchanged unfiltered (added an integration test), while filtered `count_distinct(*)` now raises a clear `InvalidParameterError` instead of silently returning 1. Also documented `InvalidParameterError` in the four `is_aggr_*` and `has_no_aggr_outliers` `Raises:` sections.

vb-dbrks (approved):
LGTM

ghanse (approved):

PR #1454, opened 2026-08-11
fix: reject a trailing newline in row-level pattern checks

### LINKED ISSUE #1440: [BUG]: Row-level pattern checks accept a trailing newline (rlike anchors)
## Current Behavior

The row-level pattern checks (`is_valid_email`, `is_valid_national_id`, `is_valid_uuid`, `is_valid_ipv4_address`) validate through `_matches_pattern`, which applies each `DQPattern` regex with Spark's `rlike` (Java regex semantics). The patterns anchor the end with `$`, and in Java regex `$` matches at the end of input or just before a final line terminator, so a value with a single trailing newline is reported valid.

Example: `is_valid_uuid` accepts `"550e8400-e29b-41d4-a716-446655440000\n"`.

## Expected Behavior

A trailing newline should make the value invalid. The checks should match the full string with no trailing line terminator.

## Steps to Reproduce

```python
from databricks.labs.dqx.check_funcs import is_valid_uuid

df = spark.createDataFrame([("550e8400-e29b-41d4-a716-446655440000\n",)], "a: string")
df.select(is_valid_uuid("a")).show(truncate=False)  # no violation reported
```

## Affected Code

The `^...$` anchors in the `DQPattern` enum members (`IPV4_ADDRESS`, `IPV4_CIDR_BLOCK`, `EMAIL_ADDRESS`, `SSN_US`, `UUID`, `UUID_STRICT`) in `src/databricks/labs/dqx/check_funcs.py`, consumed by `_matches_pattern`.

## Suggested Fix

Switch the family's anchors from `^...$` to `\A...\z` (Java regex start and end of input, no trailing newline tolerance) in one focused change, and add a trailing-newline negative case to each check's tests.

Surfaced during review of #1436; kept out of that PR to avoid changing the sibling checks' behavior.

mwojtyczka:
## Summary

Fixes #1440. The row-level pattern checks (`is_valid_ipv4_address`, `is_valid_email`, `is_valid_uuid`, `is_valid_national_id`, and the `is_ipv4_address_in_cidr` value path) validated through `DQPattern` regexes anchored with `^...$`. Under Java regex — which Spark `rlike` uses — `$` also matches just before a *final* line terminator, so an otherwise-valid value with a single trailing newline was reported valid.

For example, `is_valid_uuid` accepted `"550e8400-e29b-41d4-a716-446655440000\n"`.

## Fix

Switch the family's anchors from `^...$` to `\A...\z` (Java absolute start/end of input, no trailing-newline tolerance) for `IPV4_ADDRESS`, `IPV4_CIDR_BLOCK`, `EMAIL_ADDRESS`, `SSN_US`, `UUID`, and `UUID_STRICT`. In non-multiline mode `\A` is identical to `^` at the start, so the only behavioral change is at the end of the value.

**Scope of the behavioral change.** Java's `$` tolerates any single trailing *line terminator*, not just `\n`, so the set of values that changes verdict is wider than a lone newline. Measured on Spark 4.0.0 with `UUID`:

| trailing | before | after |
| --- | --- | --- |
| *(none)* | valid | valid |
| `\n` | valid | **invalid** |
| `\r` | valid | **invalid** |
| `\r\n` | valid | **invalid** |
| `\u0085` (NEL) | valid | **invalid** |
| `\u2028` (LS) | valid | **invalid** |
| `\n\n` | invalid | invalid |
| space, tab | invalid | invalid |

The `\r` row is the one worth calling out on upgrade: values loaded from CRLF-terminated files with an untrimmed last field carry a trailing `\r`, and those silently passed these checks before. They are now correctly reported as violations.

Two supporting changes:

- **Python `re` compatibility.** `is_ipv4_address_in_cidr` validates its `cidr_block` *argument* with Python's `re`, which does not understand `\z` (it raises `bad escape \z`) and spells the same anchor `\Z`. A small `_pattern_for_python_re` helper translates only the end anchor, so each pattern stays a single source of truth while both engines get the correct semantics.
- **No self-referential pattern.** Extracted the shared unanchored dotted-quad into `_IPV4_ADDRESS_BODY` so `IPV4_CIDR_BLOCK` no longer derives from `IPV4_ADDRESS`'s already-anchored value (the previous `IPV4_ADDRESS[:-1]` slice).

## Tests

- A parametrized `test_pattern_checks_reject_a_trailing_newline` pairs each check's valid value with the same value plus a trailing newline, asserting the newline value is now a violation while the base value stays valid.
- A trailing-newline CIDR string added to `test_ipv4_cidr_invalid_blocks_raise_error`, exercising the Python `re` argument-validation path.

Verified against a local Spark session: all six checks now reject a trailing newline, valid values still pass, and the CIDR argument validator accepts good CIDRs but rejects a trailing-newline one. `black`, `ruff`, `pylint` (10.00/10), and `mypy` all pass on the changed files.


vb-dbrks (approved):
LGTM. I've updated the PR description slightly to reflect the wider scope of the change: Java's `$` tolerates any single trailing line terminator, not just `\n`, so `\r` from CRLF sources is affected too. Measured table is in the body.

Verified on local Spark 4.0.0 — all six patterns reject the trailing terminators and no valid value regressed. The `_IPV4_ADDRESS_BODY` extraction keeps the capture groups intact, so `regexp_extract(cidr, IPV4_CIDR_BLOCK, 5)` still returns the prefix.

Non-blocking: `DQPattern` is public and documented, so `re.match(DQPattern.UUID.value, s)` in user code now raises `bad escape \z`. `_pattern_for_python_re` solves it internally but is private — worth exposing a Python-usable form (e.g. `DQPattern.UUID.python_pattern`) as a follow-up? Also `is_ipv4_address_in_cidr`'s column path isn't covered by the new tests; it does behave correctly, but via the `ipv4_msg_col.isNotNull()` short-circuit rather than the pattern.

tests/integration/test_row_checks.py
```diff
@@ -2261,6 +2261,42 @@ def test_col_is_valid_national_id_column_expr_and_lowercase_country(spark):
...
+
+    expected = spark.createDataFrame(
+        [
+            [None],  # base value stays valid
+            [f"Value '{newline_value}' in Column 'a' does not match pattern '{pattern_name}'"],
+        ],
+        f"{result_column}: string",
+    )
+
+    assertDataFrameEqual(actual, expected)
```
ghanse: Maybe parameterize the newline character to test that `"\r"`, `"\r\n"`, `"\u0085"`, and `"\u2028"` are also rejected.

tests/unit/test_row_checks.py
```diff
@@ -102,6 +102,14 @@ def test_col_is_ipv4_address_in_cidr_invalid_cidr_block():
         is_ipv4_address_in_cidr("a", cidr_block="invalid")
 
 
+def test_col_is_ipv4_address_in_cidr_trailing_newline_cidr_block():
+    # Regression for issue #1440: an otherwise-valid CIDR with a trailing newline must be rejected.
+    # Validated by re.match (the \Z anchor produced by _pattern_for_python_re), so this raises before
+    # any Spark call - keep it at the fast unit layer.
+    with pytest.raises(InvalidParameterError, match="is not a valid IPv4 CIDR block"):
+        is_ipv4_address_in_cidr("a", cidr_block="192.168.1.0/24\n")
```
ghanse: Same applies here

ghanse (approved):
LGTM. Left a small suggestion to broaden the test coverage.

tests/integration/test_row_checks.py
mwojtyczka: Done in 05b910f0. Parameterized over the full set of Java-regex line terminators — LF, CR, CRLF, NEL (U+0085), LS (U+2028), PS (U+2029) — via a shared `_TRAILING_LINE_TERMINATORS` list.

I verified on Spark 4.0.0 that the old `$` leaked exactly those six and nothing else (VT 0x0B, FF 0x0C, other controls, and a trailing space were already rejected by `$`), so that set is the complete regression surface for the `\A...\z` change. Re your `"\^E"`: I read that as NEL (U+0085, "next line"), which is in the set — a literal 0x05/ENQ isn't a Java line terminator, so `$` never leaked it and it's out of scope here.

tests/unit/test_row_checks.py
mwojtyczka: Done in 05b910f0. Same treatment on the unit CIDR-argument test: parameterized over the same six terminators. This path goes through Python's `re` (`_pattern_for_python_re` maps `\z`→`\Z`, absolute end), which rejects all six — so it stays a fast, Spark-free unit test.

PR #1455, opened 2026-08-11
fix: validate ref_row_filter against unsafe SQL in aggr_matches_dataset

mwojtyczka:
## Summary

`aggr_matches_dataset` applied its `ref_row_filter` argument via `ref_df.filter(ref_row_filter)` as a raw string — the one filter parameter in `check_funcs.py` that bypassed `safe_filter_expr` / `is_sql_query_safe`. Every sibling filter (the checked-side `row_filter`, `foreign_key`, `is_unique`, `sql_query`, `has_no_outliers`) routes user/templated SQL through `safe_filter_expr`, and CLAUDE.md requires validating such SQL and rejecting destructive predicates with `UnsafeSqlQueryError`.

## Scope / severity

To be precise about impact: `DataFrame.filter(str)` compiles a **WHERE predicate**, not a full statement, so a raw destructive `ref_row_filter` could not execute DDL — it surfaced as a raw Spark `ParseException`. This change is therefore a **consistency + defense-in-depth** fix, not a remote-code-execution fix:

- Aligns `ref_row_filter` with the documented `is_sql_query_safe` requirement and every sibling filter.
- A destructive predicate now raises a clear `UnsafeSqlQueryError` with **sanitized** logging, instead of leaking a raw parser error.
- Legitimate `SELECT` subquery filters remain allowed (that's `is_sql_query_safe`'s policy).

## Change

```python
ref_filtered_df = ref_df.filter(safe_filter_expr(ref_row_filter)) if ref_row_filter else ref_df
```

## Tests

Adds `test_aggr_matches_dataset_rejects_destructive_ref_row_filter`, mirroring the existing checked-side guard `test_has_no_outliers_rejects_destructive_row_filter`: it builds the check, invokes the apply closure with a destructive `ref_row_filter`, and asserts `UnsafeSqlQueryError`.

Verified against a local Spark session: an unsafe `ref_row_filter` now raises `UnsafeSqlQueryError` (previously a raw `ParseException`), while a safe filter still applies. `black`, `ruff`, `pylint` (10.00/10), and `mypy` pass on the changed module.

Found during code review of a separate branch; the gap exists on `main`, so this is filed independently.

### DIFF
--- a/src/databricks/labs/dqx/check_funcs.py
+++ b/src/databricks/labs/dqx/check_funcs.py
@@ -2777,7 +2777,10 @@ def aggr_matches_dataset(
             The DataFrame with additional condition and metric columns for upstream comparison.
         """
         ref_df = _get_ref_df(ref_df_name, ref_table, ref_dfs, spark)
-        ref_filtered_df = ref_df.filter(ref_row_filter) if ref_row_filter else ref_df
+        # Validate ref_row_filter with safe_filter_expr, like every other filter in this module: it is
+        # user/templated SQL, so a destructive predicate must raise UnsafeSqlQueryError (with sanitized
+        # logging) rather than reach ref_df.filter() as a raw string. SELECT subqueries stay allowed.
+        ref_filtered_df = ref_df.filter(safe_filter_expr(ref_row_filter)) if ref_row_filter else ref_df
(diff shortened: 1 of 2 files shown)


mwojtyczka:
Folding this into #1453 (fix/aggr-star-col-filter), which covers the same aggregate-checks-with-filters area. Reopening the change there.

PR #1461, opened 2026-08-12
Add UK and India national ID formats

### LINKED ISSUE #1413: [FEATURE]: Extend is_valid_national_id to support UK and India

### Problem statement

is_valid_national_id currently supports only the US Social Security Number format. Follow-up to #1265, where the per-country structure (country parameter, pattern dict keyed by ISO 3166 alpha-2 code) was explicitly built to support more countries incrementally.

### Proposed Solution

Add patterns for UK National Insurance Number (NINO) and India PAN, both pure format checks with no checksum, following the existing pattern convention (anchored, ReDoS-safe regex) and citing each country's official format source, matching how the US pattern is documented.

### Additional Context

Canada Social Insurance Number (SIN) was considered too, but its check digit uses an actual Luhn checksum rather than a fixed format, making it a different scope of work; tracking it separately rather than bundling it here.

AtomicGlance:
## Summary

Extend `is_valid_national_id` with format checks for:

- UK National Insurance numbers (`GB`), including HMRC-unallocated prefix exclusions.
- Indian Permanent Account Numbers (`IN`), with the fourth character restricted to supported holder-type codes.

The existing US behavior remains unchanged. Added Spark integration coverage for valid, invalid, and null PAN/NINO values, plus updated the quality-check reference.

Closes #1413

## Validation

- `.venv\Scripts\pytest.exe tests/unit/test_row_checks.py -q` — 71 passed
- `.venv\Scripts\ruff.exe check src/databricks/labs/dqx/check_funcs.py tests/unit/test_row_checks.py tests/integration/test_row_checks.py` — passed
- `.venv\Scripts\black.exe --check src/databricks/labs/dqx/check_funcs.py tests/unit/test_row_checks.py tests/integration/test_row_checks.py` — passed
- `git diff --check` — passed
- The targeted Spark integration test requires configured Databricks workspace credentials; the PR workflows are awaiting maintainer approval.


src/databricks/labs/dqx/check_funcs.py
```diff
@@ -117,6 +117,15 @@ class DQPattern(Enum):
...
     SSN_US = r"\A(?!000|666|9\d{2})\d{3}([- ]?)(?!00)\d{2}\1(?!0000)\d{4}\z"
+    # UK National Insurance Number: two-letter prefix, six digits, and an A-D
+    # suffix. Exclude prefixes that HMRC does not allocate.
+    NINO_GB = (
+        r"\A(?!(?:BG|GB|KN|NK|NT|TN|ZZ))(?!(?:[DFIQUV]))[A-Z]"
+        r"(?![DFIOQUV])[A-Z] ?\d{2} ?\d{2} ?\d{2} ?[ABCD]\z"
+    )
+    # Indian Permanent Account Number (PAN): five letters, four digits, and a
+    # final letter.
+    PAN_IN = r"\A[A-Z]{5}\d{4}[A-Z]\z"
```
ghanse: According to [Permanent account number](https://en.wikipedia.org/wiki/Permanent_account_number#Structure), the 4th digit should be a letter from a small list (A, B, C, F, G, H, L, J, P, and T are valid options).

tests/unit/test_row_checks.py
ghanse, line 1: I'd suggest reverting these changes and adding cases for PAN and NINO to `test_col_is_valid_national_id` integration test in `/tests/integration/test_row_checks.py`

ghanse (changes requested):
Looks good overall! Left a few small suggestions.

tests/unit/test_row_checks.py
AtomicGlance: Done in 02512c4d. I removed the PAN/NINO regex cases and the private-helper imports from the unit test, then added valid, invalid, and null cases for both formats to the existing Spark integration test.

src/databricks/labs/dqx/check_funcs.py
AtomicGlance: Good catch. I constrained the fourth PAN character to `[ABCFGHJLPT]` and added Spark integration coverage for representative valid holder types and an invalid `Z` value in 02512c4d.

AtomicGlance:
@ghanse @nehamilak-db I’ve addressed the requested PAN/NINO changes in 02512c4, including the integration coverage, and the signed-commit check is passing. Could you please re-review when convenient? The required CI workflows are currently waiting for maintainer approval.

tests/integration/test_row_checks.py
```diff
@@ -2298,6 +2298,68 @@ def violation(value: str) -> str:
...
+    schema_nino = "nino: string"
+    nino_df = spark.createDataFrame(
+        [
+            ["AB123456A"],
+            ["AB 12 34 56 A"],
+            ["BX586745C"],
+            ["DF123456A"],  # invalid first letter
+            ["BG123456A"],  # unallocated prefix
+            ["AB123456E"],  # invalid suffix
+            [None],
```
ghanse: Maybe add these cases to improve the coverage a bit?

- Empty string
- Line break character at the end of string
- Spaces before/after the string

tests/integration/test_row_checks.py
```diff
@@ -2298,6 +2298,68 @@ def violation(value: str) -> str:
...
+
+    schema_pan = "pan: string"
+    pan_df = spark.createDataFrame(
+        [
+            ["ABCPD1234F"],
+            ["AACTA1234A"],
+            ["ABCZD1234F"],  # Z is not a valid holder type
+            ["AB12E1234F"],  # letters and digits in the wrong positions
+            ["ABCPD12345"],  # final character must be a letter
+            [None],
```
ghanse: Same applies here as well.

tests/integration/test_row_checks.py
ghanse: I'm thinking it might be good to `@pytest.mark.parametrize` this for different types of IDs. We can define the `violation` closure, then load the appropriate example data and violations based on the type of ID being checked.

I think this will be more scalable as we add other national ID formats.

ghanse (changes requested):
Left a few comments on the tests. Otherwise looks good!

tests/integration/test_row_checks.py
AtomicGlance: Added in 24598158. The NINO table now covers an empty string, a trailing newline, and leading and trailing spaces.

tests/integration/test_row_checks.py
AtomicGlance: Added the same boundary cases for PAN in 24598158: an empty string, a trailing newline, and leading and trailing spaces.

tests/integration/test_row_checks.py
AtomicGlance: Refactored in 24598158. The integration test is now parameterized by country, pattern name, and case table, so another national ID format only needs a new parameter set.

AtomicGlance:
Addressed the latest test feedback in 24598158: the integration test is now parameterized across US, GB, and IN, with empty-string, trailing-newline, and surrounding-space cases for NINO and PAN. Ruff, Black, diff checks, and all 71 row-check unit tests pass. The targeted integration test still requires DATABRICKS_HOST locally, so the approved PR workflow will provide the Spark run. All three review conversations are resolved. @ghanse, could you take another look when convenient?

ghanse (approved):
LGTM!

PR #1466, opened 2026-08-13
Fix reference MapType dataset comparisons

### LINKED ISSUE #1465: [BUG]: compare_datasets does not skip reference-only MapType columns
## Current Behavior

`compare_datasets` skips a shared `MapType` column only when that type appears in the source DataFrame schema. If the source column has a comparable scalar type but the reference column with the same name is `MapType`, the column remains in `compare_columns` and DQX builds an `eqNullSafe` comparison between incompatible Spark types.

This makes the documented MapType exclusion directional: swapping the source and reference DataFrames changes a supported skip into a Spark analysis failure.

## Existing Issue Search

- [x] I searched issues, pull requests, issue comments, and pull-request review comments across all states. No existing report or active fix covers the reference-only MapType case.

## Expected Behavior

If either side of a shared column is `MapType`, `compare_datasets` should skip that column automatically, consistent with the API documentation:

> Map types (any column comparison on map type is skipped automatically).

## Steps to Reproduce

1. Create a source DataFrame with schema `id INT, value STRING`.
2. Create a reference DataFrame with schema `id INT, value MAP<STRING, STRING>`.
3. Apply `compare_datasets(columns=["id"], ref_columns=["id"], ref_df_name="reference")` with the reference DataFrame supplied through `ref_dfs`.
4. Materialize the resulting DataFrame.

The implementation derives `map_type_columns` only from `df.schema`, so `value` is retained and later compared using `eqNullSafe`. Reversing the schemas avoids the failure because the source-side map is detected and excluded.

## Suggested Fix

Build the excluded map-column set from both `df.schema` and `ref_df.schema`, then add mirrored integration cases covering a MapType present on only the source side and only the reference side.

## Environment

- DQX `main`: `16031b435e17fa66a6333c1f871390f59e599ca7`
- Latest release checked: `v0.16.0`
- Python: 3.12.13
- Cloud: All (the asymmetric schema inspection is not cloud-specific)
- Operating system: macOS

## Additional Context

The existing `test_dataset_compare_ref_as_table_and_skip_map_col` gives both DataFrames the same MapType column, so it covers only the currently working source-side exclusion. The issue is the untested asymmetric-schema case.

mattfaltyn:
## Changes

- Exclude shared columns from `compare_datasets` when either the source or reference schema declares them as `MapType`.
- Add mirrored integration coverage for source-only and reference-only MapType columns.

This keeps dataset comparison symmetric and restores the documented behavior that MapType comparisons are skipped automatically.

### Linked issues

Resolves #1465

### Tests

- [ ] manually tested
- [ ] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

Local validation:

- `make fmt`
- `make lint`
- `make test` — 2,327 passed

The focused integration cases require the repository's Databricks workspace test environment, which is not configured locally. Per the fork-contribution workflow, a maintainer must approve the standard workflows and trigger the integration and end-to-end suites.

### Documentation and Demos

- [ ] added/updated demos
- [ ] added/updated docs

No documentation change is needed: the existing `compare_datasets` documentation already states that any MapType column comparison is skipped automatically. This PR aligns the implementation with that contract.


ghanse (approved):
This looks good overall.

PR #1468, opened 2026-08-13
docs: add MCP server to README key capabilities

mwojtyczka:
## Changes

Two docs-related changes:

1. **README** — adds the **MCP server** to the "Key capabilities" list. The MCP (Model Context Protocol) server shipped in 0.16.0 ([#1252](https://github.com/databrickslabs/dqx/issues/1252)) but was never added to the README, so the capabilities list was incomplete. New bullet, placed next to DQX Studio (the other Databricks App surface).

2. **`deploy-docs.sh`** — make the gh-pages deploy script idempotent across repeat runs. Previously it only synced the `gh-pages` worktree to `origin/gh-pages` on first run (when the worktree was created); on later runs it committed on top of whatever the local `gh-pages` branch happened to be. If `origin/gh-pages` had advanced or a prior run left an un-pushed commit, the append-only push was rejected as non-fast-forward and left the local branch diverged. The script now resets the existing worktree to `origin/gh-pages` before syncing the build, so the deploy commit is always a clean fast-forward.

No functional changes to the library.

### DIFF
--- a/.github/scripts/deploy-docs.sh
+++ b/.github/scripts/deploy-docs.sh
@@ -130,6 +130,15 @@ if [[ ! -d "${WORKTREE_DIR}" ]]; then
       git commit --allow-empty -m "Initialize gh-pages"
     )
   fi
+else
+  # Worktree already exists. Reset it to origin/gh-pages so the new deploy
+  # commit is a clean fast-forward and the append-only push can't be rejected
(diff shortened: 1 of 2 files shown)


vb-dbrks (approved):
LGTM

ghanse (approved):
LGTM

PR #1471, opened 2026-08-14
Fix check-name escaping in check_metrics SQL literals

### LINKED ISSUE #1166: [FEATURE]: Add status to summary & check_metrics

### Problem statement

This was a very welcome change:  www.github.com/databrickslabs/dqx/pull/1097
However for dashboarding purposes we don't really care about the amount of violating rows. We care about the `status` (or you can call this `result`).

Now I simply do:
```python

    def _enrich_check_metrics(raw: str | None) -> str | None:
        """Add a status field to each DQX check_metrics entry.

        DQX returns: [{"check_name": ..., "error_count": N, "warning_count": N}]
        We enrich to: [..., {"status": "error"|"warn"|"passed"}]
        """
        if raw is None:
            return None
        checks = json.loads(raw)
        for check in checks:
            if check.get("error_count", 0) > 0:
                check["status"] = "error"
            elif check.get("warning_count", 0) > 0:
                check["status"] = "warn"
            else:
                check["status"] = "passed"
        return json.dumps(checks)


_enrich_check_metrics(metrics.get("check_metrics")
```

I'd be better as a default part of dqx. It could also allow this: https://github.com/databrickslabs/dqx/issues/1093
The status could be error, warn. pass it could be implemented for each test and for all tests together  even if 0 rows.

Getting:
```python
overall_status = metrics.get("status") # all checks: error, warn, pass
for single_result in metrics.get("check_metrics"):
      status = single_result.get('status') # error, warn, pass
```

Rows should be seen as extra, it's a great feature of dqx but overall the result is what matters and what we can act on.

### Proposed Solution

Add status to the `check_metrics`

### Additional Context

_No response_

### LINKED ISSUE #1474: Check names containing quotes or backslashes corrupt check_metrics (wrong counts, unparseable JSON)

### Current Behavior

`DQMetricsObserver` builds `check_metrics` by embedding the check name into a Spark SQL string literal, escaping only single quotes as `''`. Spark's parser runs with `spark.sql.parser.escapedStringLiterals` false, where the backslash is the escape character and ANSI `''` doubling is **not** honoured. Two failures follow, both silent:

**1. Single quotes and backslashes produce wrong counts.** The `''` pair is dropped outright rather than unescaped, so the `exists(_errors, x -> x.name = '...')` comparison looks for a different name than the one in `_errors` and never matches. The check is reported as passing.

A check named `it's_valid` is reported as `its_valid`; a check named `back\slash` is compared as `backslash`. Both come back with `error_count` 0 and no indication anything went wrong.

**2. Double quotes produce unparseable JSON.** `json.dumps` encodes them as `\"`, but the parser consumes the backslash, so the emitted metric is malformed and `json.loads` raises.

```
[{"check_name":"he said "hi"","error_count":0,"warning_count":0}]
```

Existing coverage does not catch either case: `test_get_metrics_with_checks_escapes_single_quotes` derives its expected value from `get_metrics` itself, so it asserts that the production code equals the production code and passes regardless of whether the escaping is correct.

### Expected Behavior

A check name round-trips byte-for-byte into `check_metrics`, and its error and warning counts are correct, for any name the rule schema accepts.

### Steps To Reproduce

```python
from databricks.labs.dqx.metrics_observer import DQMetricsObserver
import json

observer = DQMetricsObserver(name="obs")
engine = DQEngine(WorkspaceClient(), observer=observer)

checks = [{
    "name": "it's_valid",
    "criticality": "error",
    "check": {"function": "is_not_null", "arguments": {"column": "id"}},
}]

df = spark.createDataFrame([[None]], "id int")   # one violating row
checked_df, observation = engine.apply_checks_by_metadata(df, checks)
checked_df.count()

print(observation.get["check_metrics"])
# [{"check_name":"its_valid","error_count":0,"warning_count":0}]
#                 ^ quote dropped        ^ should be 1
```

Replacing the name with `he said "hi"` makes `json.loads(observation.get["check_metrics"])` raise `JSONDecodeError`.

### Cloud

AWS

### Operating System

macOS

```shell
json.decoder.JSONDecodeError: Expecting ',' delimiter: line 1 column 26 (char 25)
```

---

A fix is included in #1471, which rewrites this function for #1166: escape the backslash as `\\` and the single quote as `\'`, with unit tests pinning the emitted SQL and a parametrised integration test asserting the round-trip for plain, single-quote, double-quote, backslash and mixed names. Verified against serverless compute — 4 of the 5 name shapes fail before the change and all pass after. Happy to split it into its own PR if you would rather review it separately.

moomindani:
## Changes

Check names were embedded in Spark SQL string literals with only single quotes escaped, as `''`. Spark's parser runs with `spark.sql.parser.escapedStringLiterals` false, where the backslash is the escape character and ANSI `''` doubling is not honoured. Two silent failures followed:

- **A single quote or backslash produced wrong counts.** The `''` pair is dropped rather than unescaped, so the `exists()` comparison looked for a different name than the one in `_errors`, never matched, and the check was reported as passing with `error_count` 0. `it's_valid` was also reported as `its_valid`.
- **A double quote produced unparseable JSON.** `json.dumps` encodes it as `\"`, the parser consumed the backslash, and `json.loads` on the metric raised.

Fixed via a shared `_sql_literal_escape` helper applied to both the JSON-encoded name and the comparison literal.

### Scope change

This PR originally also added a `status` field to each `check_metrics` entry (#1166). That was dropped in review: `status` is a policy rather than a fact, and a fixed zero-tolerance rule baked into the metrics schema would compete with the alert conditions and custom metrics where customers already own the threshold. #1166 will be answered with a documentation example instead. The escaping fix is unrelated and stands on its own, so it stays here.

### Linked issues

Resolves #1474

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

The pre-existing `test_get_metrics_with_checks_escapes_single_quotes` builds its expectation from the same production helper it exercises, so it holds whatever the escaping does. Added two unit tests that pin the emitted SQL, plus a parametrised integration test asserting the round-trip for plain, single-quote, double-quote, backslash and mixed names.

Verified against serverless compute that the tests fail on unfixed code and pass on fixed code: four of the five name shapes fail before the change (one count mismatch, three `JSONDecodeError`), all five pass after.

### Documentation and Demos

- [ ] added/updated demos
- [ ] added/updated docs

No documentation change — the fix restores the documented behaviour rather than altering it.

### DIFF
--- a/app/tasks/src/dqx_task_runner/runner.py
+++ b/app/tasks/src/dqx_task_runner/runner.py
@@ -580,9 +580,18 @@ def _run_dryrun_sql_check(
         )
 
         # SQL checks treat each violation row as one error against a single
-        # logical "check" (criticality 'error' by convention).
+        # logical "check" (criticality 'error' by convention). ``status`` mirrors
+        # what ``DQMetricsObserver`` emits so both producers write one shape into
+        # the shared metrics table.
         synth_check_metrics = _json_dumps(
-            [{"check_name": str(check_name), "error_count": invalid_rows, "warning_count": 0}]
+            [
+                {
+                    "check_name": str(check_name),
+                    "error_count": invalid_rows,
+                    "warning_count": 0,
+                    "status": "error" if invalid_rows > 0 else "passed",
+                }
+            ]
         )
         observed = {
             "input_row_count": str(total_rows),
--- a/docs/dqx/docs/guide/summary_metrics.mdx
+++ b/docs/dqx/docs/guide/summary_metrics.mdx
@@ -30,7 +30,7 @@ DQX automatically captures the following built-in metrics for every data quality
 | `error_row_count`        | `int`                                           | Number of rows that failed error-level checks                       |
 | `warning_row_count`      | `int`                                           | Number of rows that triggered warning-level checks                  |
 | `valid_row_count`        | `int`                                           | Number of rows that passed all checks (have no errors and warnings) |
-| `check_metrics`          | `string` (JSON)                                 | Per-check breakdown of error and warning counts, stored as a JSON array of structs: `[{"check_name": "...", "error_count": N, "warning_count": N}, ...]`. Automatically included when checks are applied. |
+| `check_metrics`          | `string` (JSON)                                 | Per-check breakdown of error and warning counts plus a pass/fail status, stored as a JSON array of structs: `[{"check_name": "...", "error_count": N, "warning_count": N, "status": "..."}, ...]`. Automatically included when checks are applied. |
 
 ### Per-Check Metrics
 
@@ -46,9 +46,9 @@ The `check_metrics` value is a JSON-serialized array of structs, with one entry 
 
 ```json
 [
-  {"check_name": "id_is_not_null", "error_count": 5, "warning_count": 0},
-  {"check_name": "name_is_not_null_and_not_empty", "error_count": 0, "warning_count": 3},
-  {"check_name": "passenger_incorrect_count", "error_count": 0, "warning_count": 0}
+  {"check_name": "id_is_not_null", "error_count": 5, "warning_count": 0, "status": "error"},
+  {"check_name": "name_is_not_null_and_not_empty", "error_count": 0, "warning_count": 3, "status": "warn"},
+  {"check_name": "passenger_incorrect_count", "error_count": 0, "warning_count": 0, "status": "passed"}
 ]
 ```
 
@@ -56,6 +56,14 @@ Each entry contains:
 - `check_name` — the name of the check (either explicitly set via `name` in the rule definition, or auto-derived from the check function and arguments)
 - `error_count` — number of rows where this check triggered an error (`0` if no rows failed the check)
 - `warning_count` — number of rows where this check triggered a warning (`0` if no rows failed the check)
+- `status` — outcome of the check for this run: `error` if `error_count > 0`, `warn` if only `warning_count > 0`, otherwise `passed`
+
+<Admonition type="tip" title="Why status and not just counts">
+For dashboarding you usually care whether a check passed, not how many rows violated it. `status` is
+particularly useful for dataset-level checks: when an ungrouped dataset-level check fails, the result is
+attached to every input row, so `error_count` equals `input_row_count` and reads as though every row is
+individually bad. `status` gives you the single pass/fail signal instead.
+</Admonition>
 
 When persisted to the metrics table, `check_metrics` is stored as a single row with `metric_name = 'check_metrics'` and `metric_value` containing the JSON string. You can parse it with Spark's `from_json` or `json_tuple` functions for analysis:
 
@@ -64,10 +72,11 @@ SELECT
   run_id,
   check.check_name,
   check.error_count,
-  check.warning_count
+  check.warning_count,
+  check.status
 FROM main.analytics.dq_metrics
 LATERAL VIEW explode(
-  from_json(metric_value, 'array<struct<check_name:string,error_count:bigint,warning_count:bigint>>')
+  from_json(metric_value, 'array<struct<check_name:string,error_count:bigint,warning_count:bigint,status:string>>')
 ) AS check
 WHERE metric_name = 'check_metrics'
 ```
@@ -660,7 +669,7 @@ Summary metrics can be written and centralized in a delta table. The metrics tab
 | `checks_location`     | `STRING`              | Location where checks are stored (table name or file path), if known.                                                                                                                                                                                                                                                                                                    |
 | `rule_set_fingerprint`| `STRING`              | SHA-256 fingerprint of the rule set used for this run. Enables correlation with checks storage and results. Populated automatically when applying checks or saving results.                                                                                                                                                                                              |
 | `metric_name`         | `STRING`              | Name of the metric (e.g., 'input_row_count').                                                                                                                                                                                                                                                                                                                            |
-| `metric_value`        | `STRING`              | Value of the metric. All values are stored as strings — numeric metrics (e.g. `input_row_count`) are string-encoded integers, and `check_metrics` is a JSON string: `[{"check_name": "...", "error_count": N, "warning_count": N}, ...]`. Cast to the appropriate type when querying (e.g. `CAST(metric_value AS INT)` for counts, or parse JSON for `check_metrics`).    |
+| `metric_value`        | `STRING`              | Value of the metric. All values are stored as strings — numeric metrics (e.g. `input_row_count`) are string-encoded integers, and `check_metrics` is a JSON string: `[{"check_name": "...", "error_count": N, "warning_count": N, "status": "..."}, ...]`. Cast to the appropriate type when querying (e.g. `CAST(metric_value AS INT)` for counts, or parse JSON for `check_metrics`).    |
 | `run_time`            | `TIMESTAMP`           | Run timestamp when the summary metrics were calculated.                                                                                                                                                                                                                                                                                                                  |
 | `error_column_name`   | `STRING`              | Name of the error column in the output or quarantine table containing per row quality checking results (default: '_errors').                                                                                                                                                                                                                                             |
 | `warning_column_name` | `STRING`              | Name of the warning column in the output or quarantine table containing per row quality checking results (default: '_warnings').                                                                                                                                                                                                                                         |
--- a/src/databricks/labs/dqx/metrics_observer.py
+++ b/src/databricks/labs/dqx/metrics_observer.py
@@ -136,13 +136,25 @@ class DQMetricsObserver:
             json_check_name_sql_esc = json.dumps(check_name).replace("'", "''")
             err = self._error_column_name
             warn = self._warning_column_name
+            error_count = f"count(case when exists({err}, x -> x.name = '{check_name_escaped}') then 1 end)"
+            warning_count = f"count(case when exists({warn}, x -> x.name = '{check_name_escaped}') then 1 end)"
+            # *status* is derived from the same two aggregates rather than emitted as a separate
+            # metric: consumers (dashboard, Studio, Genie) need a pass/fail signal per check, and
+            # counts alone do not provide one. Errors take precedence over warnings.
+            status = (
+                f"case when {error_count} > 0 then 'error' "
+                f"when {warning_count} > 0 then 'warn' "
+                f"else 'passed' end"
+            )
             fragments.append(
                 f"concat("
                 f"'{{\"check_name\":{json_check_name_sql_esc},\"error_count\":',"
-                f"cast(count(case when exists({err}, x -> x.name = '{check_name_escaped}') then 1 end) as string),"
+                f"cast({error_count} as string),"
                 f"',\"warning_count\":',"
-                f"cast(count(case when exists({warn}, x -> x.name = '{check_name_escaped}') then 1 end) as string),"
-                f"'}}')"
+                f"cast({warning_count} as string),"
+                f"',\"status\":\"',"
+                f"{status},"
+                f"'\"}}')"
             )
         return f"concat('[', concat_ws(',', {', '.join(fragments)}), ']') as check_metrics"
 
--- a/src/databricks/labs/dqx/metrics_observer.py
+++ b/src/databricks/labs/dqx/metrics_observer.py
@@ -16,6 +16,25 @@ OBSERVATION_TABLE_SCHEMA = (
 )
 
 
+def _sql_literal_escape(value: str) -> str:
+    """Escape a string for inclusion in a single-quoted Spark SQL literal.
+
+    Spark's parser honours backslash escapes inside string literals by default
+    (*spark.sql.parser.escapedStringLiterals* is false), which drives both replacements:
+
+      * The backslash must be doubled, or the parser consumes it — collapsing a JSON-encoded
+        ``\\"`` back to a bare ``"`` and yielding malformed JSON for a check name containing a
+        double quote.
+      * The single quote must be escaped as ``\\'``, not doubled as ``''``. ANSI doubling is *not*
+        honoured in this mode: Spark drops the pair outright, so a check named ``it's_valid`` is
+        reported as ``its_valid``.
+
+    Both behaviours are verified by the round-trip integration test in
+    *tests/integration/test_summary_metrics.py*.
+    """
+    return value.replace("\\", "\\\\").replace("'", "\\'")
+
+
 @dataclass(frozen=True)
 class DQMetricsObservation:
     """
@@ -130,10 +149,10 @@ class DQMetricsObserver:
         """
         fragments: list[str] = []
         for check_name in check_names:
-            check_name_escaped = check_name.replace("'", "''")
+            check_name_escaped = _sql_literal_escape(check_name)
             # JSON-encode the check name (handles embedded quotes, backslashes, control chars),
-            # then SQL-escape any single quotes for safe inclusion in a literal.
-            json_check_name_sql_esc = json.dumps(check_name).replace("'", "''")
+            # then escape it again for safe inclusion in a SQL string literal.
+            json_check_name_sql_esc = _sql_literal_escape(json.dumps(check_name))
             err = self._error_column_name
             warn = self._warning_column_name
             error_count = f"count(case when exists({err}, x -> x.name = '{check_name_escaped}') then 1 end)"
--- a/src/databricks/labs/dqx/metrics_observer.py
+++ b/src/databricks/labs/dqx/metrics_observer.py
@@ -16,6 +16,25 @@ OBSERVATION_TABLE_SCHEMA = (
 )
 
 
+def _sql_literal_escape(value: str) -> str:
+    """Escape a string for inclusion in a single-quoted Spark SQL literal.
+
+    Spark's parser honours backslash escapes inside string literals by default
+    (*spark.sql.parser.escapedStringLiterals* is false), which drives both replacements:
+
+      * The backslash must be doubled, or the parser consumes it — collapsing a JSON-encoded
+        ``\\"`` back to a bare ``"`` and yielding malformed JSON for a check name containing a
+        double quote.
+      * The single quote must be escaped as ``\\'``, not doubled as ``''``. ANSI doubling is *not*
+        honoured in this mode: Spark drops the pair outright, so a check named ``it's_valid`` is
+        reported as ``its_valid``.
+
+    Both behaviours are verified by the round-trip integration test in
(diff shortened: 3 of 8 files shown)


docs/dqx/docs/guide/summary_metrics.mdx
```diff
@@ -46,16 +46,24 @@ The `check_metrics` value is a JSON-serialized array of structs, with one entry
 
 ```json
 [
-  {"check_name": "id_is_not_null", "error_count": 5, "warning_count": 0},
-  {"check_name": "name_is_not_null_and_not_empty", "error_count": 0, "warning_count": 3},
-  {"check_name": "passenger_incorrect_count", "error_count": 0, "warning_count": 0}
+  {"check_name": "id_is_not_null", "error_count": 5, "warning_count": 0, "status": "error"},
+  {"check_name": "name_is_not_null_and_not_empty", "error_count": 0, "warning_count": 3, "status": "warn"},
+  {"check_name": "passenger_incorrect_count", "error_count": 0, "warning_count": 0, "status": "passed"}
```
mwojtyczka: add an example where we have both error and warning count > 0

docs/dqx/docs/guide/summary_metrics.mdx
```diff
@@ -46,16 +46,24 @@ The `check_metrics` value is a JSON-serialized array of structs, with one entry
...
 ]
 ```
 
 Each entry contains:
 - `check_name` — the name of the check (either explicitly set via `name` in the rule definition, or auto-derived from the check function and arguments)
 - `error_count` — number of rows where this check triggered an error (`0` if no rows failed the check)
 - `warning_count` — number of rows where this check triggered a warning (`0` if no rows failed the check)
+- `status` — outcome of the check for this run: `error` if `error_count > 0`, `warn` if only `warning_count > 0`, otherwise `passed`
```
mwojtyczka: what if `error_count > 0` and `warning_count > 0`? I assume this will produce `error` status but this is currently not documented

docs/dqx/docs/guide/summary_metrics.mdx
```diff
@@ -46,16 +46,24 @@ The `check_metrics` value is a JSON-serialized array of structs, with one entry
...
+
+<Admonition type="tip" title="Why status and not just counts">
```
mwojtyczka: I don't think this explanation is necessary. I would drop this tip. We also care about how many rows violated depending on dashboard so this is not always true.

app/tasks/src/dqx_task_runner/runner.py
```diff
@@ -580,9 +580,18 @@ def _run_dryrun_sql_check(
...
+        # what ``DQMetricsObserver`` emits so both producers write one shape into
+        # the shared metrics table.
         synth_check_metrics = _json_dumps(
-            [{"check_name": str(check_name), "error_count": invalid_rows, "warning_count": 0}]
+            [
+                {
+                    "check_name": str(check_name),
+                    "error_count": invalid_rows,
+                    "warning_count": 0,
+                    "status": "error" if invalid_rows > 0 else "passed",
```
mwojtyczka: Please use `pass`, not `passed`. If we use `passed` then we should also use `errored`, `warned` which reads oddly
```suggestion
                    "status": "error" if invalid_rows > 0 else "pass",
```
In the dashboard we also use pass, error, warn. It would be good to check the dashboard (DQX_Dashboard.lvdash.json) to see if we can use the new status there. We are calculating it today inside the dashboard

src/databricks/labs/dqx/metrics_observer.py
```diff
@@ -130,19 +149,31 @@ def _build_check_metrics_expr(self, check_names: list[str]) -> str:
...
-            check_name_escaped = check_name.replace("'", "''")
+            check_name_escaped = _sql_literal_escape(check_name)
             # JSON-encode the check name (handles embedded quotes, backslashes, control chars),
-            # then SQL-escape any single quotes for safe inclusion in a literal.
-            json_check_name_sql_esc = json.dumps(check_name).replace("'", "''")
+            # then escape it again for safe inclusion in a SQL string literal.
+            json_check_name_sql_esc = _sql_literal_escape(json.dumps(check_name))
             err = self._error_column_name
             warn = self._warning_column_name
+            error_count = f"count(case when exists({err}, x -> x.name = '{check_name_escaped}') then 1 end)"
```
mwojtyczka, line 158: The `error_count` and `warning_count` aggregate expressions are each embedded twice in the emitted SQL — once in the `cast(... as string)` fragment and again inside the `status` CASE — so the per-check aggregate text is duplicated. Spark's common-subexpression elimination collapses the identical aggregates at execution, so runtime cost is unchanged, but the generated `observe()` SQL string is larger and harder to read/debug. A single derived-column alias isn't expressible inside one `observe()` expression, so the duplication is inherent to this approach — worth a short comment noting that CSE makes it free at runtime, rather than leaving it implicit.

mwojtyczka, line 158: The `error_count` and `warning_count` aggregate expressions are each embedded twice in the emitted SQL — once in the `cast(... as string)` fragment and again inside the `status` CASE — so the per-check aggregate text is duplicated. Spark's common-subexpression elimination collapses the identical aggregates at execution, so runtime cost is unchanged, but the generated `observe()` SQL string is larger and harder to read/debug. A single derived-column alias isn't expressible inside one `observe()` expression, so the duplication is inherent to this approach — worth a short comment noting that CSE makes it free at runtime, rather than leaving it implicit.

docs/dqx/docs/guide/summary_metrics.mdx
mwojtyczka: The new `status` field is a user-facing feature but is documented inline without a `<FeatureTags>` / `AvailableSinceVersion` marker. AGENTS.md ("Authoring documentation") asks new features to be tagged with lifecycle stage and version so readers can tell which DQX release introduced them. Consider adding a version tag for the `status` field. (Minor: the guidance targets subsections/headings, and this is inline prose on an already-tagged page, so it's arguably not a strict violation.)

Tag version 0.17.0

mwojtyczka (changes requested):
Generally looking good, left some small comments that need to be addressed before we merge

docs/dqx/docs/guide/summary_metrics.mdx
moomindani: Added `trip_distance_is_positive` with `error_count: 2, warning_count: 7` to the example, so the precedence case is visible in the JSON itself.

docs/dqx/docs/guide/summary_metrics.mdx
moomindani: Correct, `error` wins. That was undocumented — now covered explicitly in a new "Check Status" subsection with a condition table (`error` when `error_count > 0` whatever `warning_count` is; `warn` when only `warning_count > 0`; `pass` when both are `0`), and the example above now includes a check that triggers both.

docs/dqx/docs/guide/summary_metrics.mdx
moomindani: Dropped — you're right that row counts matter for some dashboards, so the framing was too absolute.

app/tasks/src/dqx_task_runner/runner.py
moomindani: Renamed to `pass` everywhere: the observer SQL, the Studio task runner's synthesised entry, the docs, and the tests.

On the dashboard — I checked `DQX_Dashboard.lvdash.json`. Per-check status is computed in the `dqx_dataset_check_status_for_period` dataset, which reads the checked/quarantine table, explodes `_errors`/`_warnings`, and emits `Error`/`Warn`/`Pass` bucketed by a user-chosen date column. No dataset reads `check_metrics` today; the three summary-table datasets (`dqx_summary_top_5_failures`, `dqx_summary_quality_stats`, `dqx_summary_table_quality`) only read the row-count metrics.

Switching that widget to `check_metrics.status` is not a drop-in:

- granularity changes from data date (`:date_col_name` on the checked table) to run (`run_time`/`run_id` on the metrics table), so the widget's x-axis changes meaning;
- the dataset would need the `summary_*` parameters instead of `inv_*`/`q_table_name`, and would only work when a `DQMetricsObserver` and `metrics_config` are configured, which is optional;
- the widget's colour mappings would need lowercase `error`/`warn`/`pass` added (they currently map `Error`/`Warn`/`Pass` plus legacy `errored`/`warned`/`passed`).

There is real upside, though: the current SQL derives `all_checks` from the exploded `_errors`/`_warnings`, so a check that never fired in the window is invisible and never shows as `Pass` — exactly what `check_metrics` fixes. I'd rather land that as a follow-up dashboard change than widen this PR; I'll open an issue for it. Happy to pull it in here instead if you prefer.

src/databricks/labs/dqx/metrics_observer.py
moomindani: Added a comment above the `status` expression: `observe()` takes a single expression so the aggregate cannot be aliased for reuse, and Spark's common-subexpression elimination collapses the duplicates during execution, so the repetition costs nothing at runtime and only lengthens the generated expression string.

docs/dqx/docs/guide/summary_metrics.mdx
moomindani: Added. Moved the `status` explanation into a `#### Check Status` subsection tagged `<AvailableSinceVersion version="0.17.0" heading={false} />`, which follows the convention of putting `<FeatureTags>` directly under a heading rather than inline in prose.

src/databricks/labs/dqx/metrics_observer.py
moomindani: Added a comment above the `status` expression: `observe()` takes a single expression so the aggregate cannot be aliased for reuse, and Spark's common-subexpression elimination collapses the duplicates during execution, so the repetition costs nothing at runtime and only lengthens the generated expression string.

app/tasks/src/dqx_task_runner/runner.py
moomindani: Filed as #1507.

moomindani:
@mwojtyczka All review comments are addressed and pushed — the passing status is now `pass`, the precedence rule is documented in a tagged "Check Status" subsection with a condition table, the tip is dropped, and the CSE note is in. The dashboard question is filed separately as #1507 rather than widened into this PR. `test_summary_metrics.py` passes 68/68 against serverless compute with a SQL warehouse configured. Ready for another look whenever you have time — and let me know if you'd like the "Fork test" mirror PR to run full CI.

app/tasks/src/dqx_task_runner/runner.py
mwojtyczka: sounds good, doing this as a separate PR makes sense

mwojtyczka:
@moomindani 

**Overall:** The escaping refactor is correct and the round-trip test coverage is good. My concern is with the `status` field itself — it's derivable from data already in the struct, and this PR doesn't actually wire up any consumer to use it. **If we're not going to consume `status`, this PR isn't worth shipping**: it's denormalized, redundant state plus new escaping/CSE complexity, with no reader. Details below.

Additional review comments:

**1. Escaping depends on an undocumented session precondition (`metrics_observer.py`).**
The backslash escaping is correct only when `spark.sql.parser.escapedStringLiterals=false` (the default). The new docstring now states this precondition clearly — good. But it's still only a docstring, not a guard. If a session enables Hive-compat mode, a name like `it's_valid` produces `'it\'s_valid'` where `\'` no longer collapses, and the `observe()` expression fails to parse (or yields malformed JSON so `json.loads` raises). Worth deciding: is documenting the precondition enough, or do we want a defensive check? I'd accept the docstring for now, but flag it.

**2. `status` is fully derivable redundant state — the core question.**
`status` is a pure function of `error_count`/`warning_count`, both present in the same struct (`>0 → error`, else `>0 → warn`, else `pass`). The Studio score view (`score_view_service.py:370`) already computes exactly this from the counts. Materializing it is only justified if we commit to it as a **canonical contract** — one authoritative precedence rule and a stable vocabulary for NL/Genie consumers. As shipped, that justification isn't met (see #3 and #4), so right now it's redundant state with no reader.

**3. The named consumer wasn't wired up (`score_view_service.py:103`).**
The PR justifies `status` because "consumers (dashboard, Studio, Genie) need a pass/fail signal." But `_CHECK_METRICS_JSON_SCHEMA` still omits `status`, so `from_json` silently drops it and Studio never surfaces the field. Not a crash — but the stated goal isn't delivered. This is the crux: if we're adding `status` *for* Studio, Studio has to read it.

**4. The second producer mislabels warn as error (`runner.py:706`).**
The synth SQL path hardcodes `status = 'error'` on any violation, even for a check authored with `warn` criticality. This breaks the "one canonical precedence rule" benefit that would justify materializing `status` in the first place — the two producers already disagree. (The underlying count convention is pre-existing, but adding `status` makes the mislabel consumable rather than latent.)

**Recommendation:** Decide whether `status` is a contract we're committing to.
- **If yes:** complete it in this PR — update `_CHECK_METRICS_JSON_SCHEMA` (#3), fix the runner precedence (#4), and keep the good docstring/CSE comments. Then it's a coherent, defensible feature.
- **If no:** drop `status` and let consumers derive it from the counts. The escaping refactor is a genuine fix and could land on its own.

src/databricks/labs/dqx/metrics_observer.py
mwojtyczka: Agreed, Spark's common-subexpression elimination collapses the two identical aggregates at execution, so the duplication is free at runtime and the single-observe()-expression constraint makes it unavoidable. That fully answers the performance objection. It does not answer the value objection (finding https://github.com/databrickslabs/dqx/pull/2) — cheap-to-compute isn't the same as worth-storing.

moomindani:
Went with the "if yes" branch — `status` is a contract worth committing to, so I completed it rather than dropping it.

**#3 Studio now reads it.** `_CHECK_METRICS_JSON_SCHEMA` gained `status: STRING`; the shaping view explodes it and projects `check_status`. Runs recorded before the observer emitted `status` carry `NULL`, so it `COALESCE`s onto the same error-over-warning precedence rather than dropping those rows out of the column. The REST model (`CheckMetricBreakdown`) still omits it — that needs an orval regeneration, so it stays a separate app-scoped change.

**#4 The runner no longer mislabels warn as error.** Extracted `_split_violations_by_criticality` so the rule is named and unit-testable without Spark, and applied it to the metric payload, the run row and the failed-checks summary.

One judgement call to flag: you scoped this as "fix the runner precedence" and noted the count convention is pre-existing. I fixed the counts too, because a `status` derived from criticality while `error_count`/`warning_count` stayed error-only would contradict its own struct — and the documented rule is that `status` is a function of those two counts. The visible consequence is that a `warn`-criticality cross-table SQL check now reports `warning_rows` instead of `error_rows` in run history and the failed-checks summary. I think that's the bug rather than a behaviour change, but say the word and I'll narrow it back to `status` alone.

**#1** Left as the docstring, per your call.

Thanks for the branch update — I built on your merge commit rather than force-pushing over it, so the two new commits sit on top. Also picked up the `productName="DQX"` prop on the new version tag to match what landed with the independent Studio versioning.

Verified: `make fmt`/lint clean, `basedpyright` 0 errors, 2503 core unit tests, app backend tests pass, and `test_summary_metrics.py -k check_metrics` 14/14 against serverless compute with a SQL warehouse. `make app-check`'s TypeScript step needs `bun`, which I don't have locally — no UI files changed.

app/tasks/src/dqx_task_runner/runner.py
moomindani: Wired up in the same push: `_CHECK_METRICS_JSON_SCHEMA` now reads `status`, and the shaping view projects it as `check_status` with a COALESCE fallback onto the counts for pre-`status` runs.

src/databricks/labs/dqx/metrics_observer.py
moomindani: Right — the CSE note only answered the performance half. The value half is now answered by wiring up the reader rather than by argument: `_CHECK_METRICS_JSON_SCHEMA` reads `status` and the shaping view projects `check_status`, so Studio consumes the stored field instead of recomputing it, and the runner applies the same precedence rule so the two producers can't disagree. That's what makes it a contract rather than a cached derivation. See the summary comment for the full rundown.

app/tasks/src/dqx_task_runner/runner.py
```diff
@@ -642,8 +666,6 @@ def _run_dryrun_sql_check(
         err_payload = [{"name": str(check_name), "message": "SQL check violation"}]
         sample_invalid = [{**row.asDict(recursive=True), "_errors": err_payload} for row in sample_rows]
```
mwojtyczka, line 667: **[correctness] `warn`-criticality SQL checks are still rendered as errors at the row level.**

This PR makes the *counts* criticality-aware — the metrics row now writes `error_rows, warning_rows` (and the old "SQL checks treat every violation as an error" comment was removed just below) — but the row-level violation rendering was left error-only:

- Here, `sample_invalid` unconditionally stamps `_errors` (`{**row, "_errors": err_payload}`), so for a `warn` check the UI Errors-column fallback shows the violations as **errors** even though `check_metrics` now reports `status='warn'` / `error_count=0`.
- The same happens in `_write_sql_quarantine_records` (~L823): it writes the synthesised payload into the `errors` VARIANT for every violation row, so `dq_quarantine_records` also presents `warn`-check rows as errors.

Net: the two producers this PR set out to reconcile (counts/status vs. row-level rendering) disagree for `warn` checks. Consider stamping `_warnings` / `warnings` (and leaving `_errors` empty) when the check's criticality is `warn`, keyed off the same split that already feeds `error_rows` / `warning_rows`.

mwojtyczka (commented):
Previous comments are addressed.

Automated re-review. One inline finding below (the row-level rendering left error-only while this PR made counts/status criticality-aware). I'm omitting the `metrics_observer.py` double-aggregate item — it's the same point as the already-resolved thread on `_build_check_metrics_expr` (CSE makes it free at runtime; the reader was wired up), so no need to reopen it.

moomindani:
Rescoped as discussed: `status` is gone and this is now just the escaping fix (`Resolves #1474`). Title and description updated.

All the earlier review points are either addressed or moot — the Studio wiring, the criticality split and the `warn`-row-rendering comment all went away with `status`.

Could you drop `needs-changes` and take another look? Happy to have the mirror "Fork test" PR run full CI — the fork gate skips integration/e2e here.

Separately, the `warn`-criticality SQL check defect on `main` still has no home (a check authored as `warn` is reported as an error everywhere). Let me know if you want me to file it.

PR #1472, opened 2026-08-14
Add Customer 360 industry demo

amising6:
This PR adds a Customer 360 industry demo using fully synthetic data.

The demo shows how DQX can validate common Customer 360 data quality rules, including customer identity, email format, non-negative revenue, ticket count validation, date validation, and active customer flag consistency.

The contribution is limited to a new demo file under demos/dqx_demo_industry and does not modify core package code.

No real customer, client, employer, or proprietary data is included.


amising6:
Thank you. I have signed the CLA and updated the PR with a signed commit.

Please let me know if you would like me to adjust the demo structure, naming, or any DQX rule usage to better align with the existing industry demos.

ghanse (commented):
@amising6 `dqx_customer_360_demo.py` is empty. Do you need to push it?

amising6:
Thanks for catching that. I had created the file but missed saving/pushing the actual notebook content. I have updated the branch with the Customer 360 demo content now.

demos/dqx_demo_industry/dqx_customer_360_demo.py
```diff
@@ -0,0 +1,354 @@
...
+  user_metadata:
+    domain: customer_360
+    rule_type: identity
+
+# 2. Customer ID should be unique
+- criticality: error
+  check:
+    function: is_unique
+    arguments:
+      column: customer_id
```
ghanse: `is_unique` takes `columns` as a list, not `column` — this actually trips the `validate_checks` step below, so the notebook stops at `assert not status.has_errors`. Quick fix:

```suggestion
      columns:
        - customer_id
```

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

demos/dqx_demo_industry/dqx_customer_360_demo.py
```diff
@@ -0,0 +1,354 @@
...
+from databricks.labs.dqx.engine import DQEngine
+
+customer_360_checks_yaml = """
+# 1. Customer ID must be present
+- criticality: error
+  check:
+    function: is_not_null_and_not_empty
+    for_each_column:
+    - customer_id
+    name: customer_id_not_null
```
ghanse: `name` is indented under `check:` here, so DQX ignores it and auto-generates a name instead — your nice readable names never show up in the results. It needs to sit at the top level, next to `criticality`:

```yaml
- criticality: error
  name: customer_id_not_null
  check:
    function: is_not_null_and_not_empty
    for_each_column:
    - customer_id
```

Same fix on all seven checks.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

demos/dqx_demo_industry/dqx_customer_360_demo.py
```diff
@@ -0,0 +1,354 @@
...
+    rule_type: contact_quality
+
+# 4. Revenue should not be negative
+- criticality: error
+  check:
+    function: is_in_range
+    arguments:
+      column: total_revenue
+      min_limit: 0
+      max_limit: 999999999.99
```
ghanse: The `999999999.99` ceiling is a magic number that'd wrongly quarantine a legitimately large revenue. Since the intent is just "not negative," `is_not_less_than` reads cleaner:

```yaml
    function: is_not_less_than
    arguments:
      column: total_revenue
      limit: 0
```

Same deal for `open_ticket_count` on line 254 (`999999`).

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

demos/dqx_demo_industry/dqx_customer_360_demo.py
```diff
@@ -0,0 +1,354 @@
...
+# MAGIC
+# MAGIC The dataset below is fully synthetic. It includes both valid and intentionally invalid records to demonstrate DQX validation behavior.
+
+# COMMAND ----------
+
+from pyspark.sql import Row
+from datetime import date
+
+customer_360_data = [
+    # Valid active customer
```
ghanse, line 83: Small gotcha: CUST-001 gets reused as the duplicate row down at line 171, and the error-level uniqueness check quarantines *both* copies — so this "valid active customer" never actually lands in `valid_df`. Give the duplicate row its own id so this one stays clean and the demo's headline record shows up where you'd expect.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

demos/dqx_demo_industry/dqx_customer_360_demo.py
```diff
@@ -0,0 +1,354 @@
...
+        email="missing.id@example.com",
+        customer_status="ACTIVE",
+        total_revenue=500.00,
+        last_purchase_date=date(2026, 6, 5),
+        last_campaign_engagement_date=date(2026, 6, 8),
+        open_ticket_count=0,
+        active_customer_flag=True
+    ),
+
+    # Invalid: bad email format
```
ghanse: This row (and the flag-mismatch one at line 156) are `warn`-level, so DQX keeps them in `valid_df`, not the quarantine output. A new user reading "Invalid: ..." will hunt for them in the invalid table and come up empty — worth a one-line note that warn-level issues stay in the valid data.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

demos/dqx_demo_industry/dqx_customer_360_demo.py
```diff
@@ -0,0 +1,354 @@
...
+# MAGIC
+# MAGIC * `customer_id` not null
+# MAGIC * `customer_id` uniqueness
+# MAGIC * valid email format
+# MAGIC * non-negative `total_revenue`
+# MAGIC * non-negative `open_ticket_count`
+# MAGIC * `last_purchase_date` not in the future
+# MAGIC * `active_customer_flag` consistency
+# MAGIC
+# MAGIC ### Why this matters
```
ghanse, line 25: This "Why this matters" section and the "Practical Takeaway" at the bottom make nearly the same point ("technically successful but still unreliable") twice. I'd trim one so the writeup doesn't loop back on itself.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

ghanse:
Thanks for adding this — a Customer 360 accelerator is a nice fit next to the other industry demos, and using fully synthetic data is the right call. A few things to sort out before it's ready:

**Blockers (the demo won't run as-is):**
- `is_unique` needs `columns:` as a list, not `column:` — this fails at `validate_checks` and stops the notebook.
- `name:` is nested under `check:` on all seven checks, so DQX ignores it. It belongs at the top level next to `criticality`.

**Worth fixing:**
- The duplicate CUST-001 row means your "valid active customer" gets quarantined by the uniqueness check — give the dupe its own id.
- Swap the magic `is_in_range` ceilings for `is_not_less_than` on the non-negative checks.
- A couple of rows commented "Invalid" are `warn`-level and stay in `valid_df`; a one-line note would save new users some confusion.

**Small stuff:**
- Add a trailing newline at the end of the file (trips some linters).
- Add this demo to `docs/dqx/docs/demos.mdx` alongside the other industry accelerators so people browsing the docs actually find it.
- The "Why this matters" and "Practical Takeaway" sections repeat the same point — trim one.

Good concept and structure — just needs the two YAML fixes to actually execute, plus the cleanups above.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

ghanse (dismissed):

demos/dqx_demo_industry/dqx_customer_360_demo.py
```diff
@@ -0,0 +1,354 @@
+# Databricks notebook source
```
mwojtyczka, line 1: Please link the demo here: https://github.com/databrickslabs/dqx/blob/main/docs/dqx/docs/demos.mdx

demos/dqx_demo_industry/dqx_customer_360_demo.py
```diff
@@ -0,0 +1,354 @@
...
+      max_limit: 999999
+    name: open_ticket_count_non_negative
+  user_metadata:
+    domain: customer_360
+    rule_type: operational_quality
+
+# 6. Last purchase date should not be in the future
+- criticality: error
+  check:
+    function: sql_expression
```
mwojtyczka: use `is_not_in_future` function instead

demos/dqx_demo_industry/dqx_customer_360_demo.py
```diff
@@ -0,0 +1,354 @@
...
+  user_metadata:
+    domain: customer_360
+    rule_type: identity
+
+# 3. Email should follow a basic email pattern
+- criticality: warn
+  check:
+    function: sql_expression
+    arguments:
+      expression: "email RLIKE '^[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\\\\.[A-Za-z]{2,}$'"
```
mwojtyczka: use `is_valid_email` function instead

mwojtyczka (changes requested):
Left some comments. In general, I would expect this demo to show example of some more comprehensive checks.

amising6:
Thank you for the detailed review. I have pushed an expanded version of the Customer 360 demo and addressed the requested changes.

The demo now includes 12 checks covering identity completeness and uniqueness, customer name and email completeness, email validity, customer status completeness and allowed values, non-negative revenue and ticket counts, purchase and campaign date validation, and cross-field active-status consistency.

I also corrected the DQX check syntax, removed the artificial range ceilings, separated the duplicate test records, clarified warning-level behavior, removed redundant documentation, and verified that validate_checks completes without errors.

Please take another look when convenient.

demos/dqx_demo_industry/dqx_customer_360_demo.py
```diff
@@ -0,0 +1,479 @@
...
+# MAGIC %md
+# MAGIC ### Persist Invalid Records for Review
+# MAGIC
+# MAGIC In a real pipeline, invalid records can be written to a quarantine table for review and remediation.
+
+# COMMAND ----------
+
+quarantine_table = f"{catalog}.{schema}.customer_360_quarantine"
+
+invalid_df.write.mode("overwrite").saveAsTable(quarantine_table)
```
mwojtyczka, line 482: **Warn-only records _are_ quarantined — this contradicts the demo's stated behavior.** `apply_checks_by_metadata_and_split` sets `invalid_df = get_invalid()`, which returns rows where `_errors` **OR** `_warnings` is non-null (`engine.py:372-385`). So the three warn-only customers (CUST-003, CUST-006, CUST-011) land in `invalid_df` and are written to `customer_360_quarantine` here. The comments at lines 121 / 161 / 227 and the markdown at line 444 — "Warning-level violations remain in valid_df and are not quarantined" — are therefore false. Either rewrite the narrative to match split semantics, or filter `invalid_df` to error-only rows before quarantining.

demos/dqx_demo_industry/dqx_customer_360_demo.py
```diff
@@ -0,0 +1,479 @@
...
+status = DQEngine.validate_checks(checks)
+print(status)
+assert not status.has_errors
+
+# COMMAND ----------
+
+# MAGIC %md
+# MAGIC ### Apply Checks and Split Valid vs Invalid Records
+# MAGIC
+# MAGIC DQX can split records into valid and invalid DataFrames. Error-level violations are quarantined in `invalid_df`. Warning-level violations remain in `valid_df` with warning metadata so they can be monitored without quarantining the record.
```
mwojtyczka: **`get_valid()` drops the warning metadata, so this claim is inaccurate.** `get_valid()` filters `_errors IS NULL` and then `.drop(_errors, _warnings)` (`engine.py:387-399`). Warn-only rows do stay in `valid_df`, but with **no** `_warnings` column — so a reader can't "monitor" warnings from `valid_df` as the text promises. The warning metadata exists only in `invalid_df`.

demos/dqx_demo_industry/dqx_customer_360_demo.py
```diff
@@ -0,0 +1,479 @@
...
+customer_360_data = [
+    # Valid active customer
+    Row(
+        customer_id="CUST-001",
+        customer_name="Aarav Mehta",
+        email="aarav.mehta@example.com",
+        customer_status="ACTIVE",
+        total_revenue=1250.75,
+        last_purchase_date=date(2026, 7, 10),
+        last_campaign_engagement_date=date(2026, 7, 15),
```
mwojtyczka: **Hard-coded near-future 2026 dates make "valid" rows fail an error-level check.** `last_purchase_date_not_in_future` is `criticality: error`, but the valid customers use 2026 dates. Run before those dates, `current_date` makes them "in the future" → error → these supposedly-valid customers move to `invalid_df`/quarantine, breaking the valid/invalid narrative. Prefer `F.current_date()`-relative dates (as sibling demos do) so the demo stays correct over time.

demos/dqx_demo_industry/dqx_customer_360_demo.py
```diff
@@ -0,0 +1,479 @@
...
+        email="future.date@example.com",
+        customer_status="ACTIVE",
+        total_revenue=300.00,
+        last_purchase_date=date(2099, 1, 1),
+        last_campaign_engagement_date=date(2026, 3, 10),
+        open_ticket_count=0,
+        active_customer_flag=True
+    ),
+
+    # Warning-level issue: active flag is false even though recent activity exists.
```
mwojtyczka: **This comment misdescribes the rule.** `active_customer_flag_consistency` (sql_expression) only compares `customer_status` vs `active_customer_flag` — it never inspects engagement/purchase dates. The warning fires on a status/flag mismatch (e.g. status='ACTIVE' with flag=false), not because "recent activity exists". Reword to describe the actual trigger.

docs/dqx/docs/demos.mdx
```diff
@@ -69,6 +69,7 @@ Import the following notebooks in the Databricks workspace to try DQX out:
...
 * [DQX for Fashion & Retail Accelerator Demo](https://github.com/databrickslabs/dqx/blob/v0.16.0/demos/dqx_demo_industry/dqx_fashion_demo.py) - demonstrates DQX rules tailored for the Fashion and Retail industry, including SKU validation, hex color formats, and pricing compliance.
 * [DQX for Manufacturing Industry Accelerator Demo](https://github.com/databrickslabs/dqx/blob/v0.16.0/demos/dqx_demo_industry/dqx_manufacturing_demo.py) - demonstrates how to use DQX to check data quality for Manufacturing Industry datasets, specifically for machine maintenance and sensor data.
+* [DQX for Customer 360 Industry Accelerator Demo](https://github.com/databrickslabs/dqx/blob/main/demos/dqx_demo_industry/dqx_customer_360_demo.py) - demonstrates DQX data quality rules for Customer 360 datasets, including customer identity, email format, revenue, ticket counts, purchase dates, and active customer flag consistency.
```
mwojtyczka: **Link uses `blob/main/` while every sibling demo link is pinned to `blob/v0.16.0/`.** `main` is a moving target — if the file is renamed/moved after release the published-doc link breaks, and it's inconsistent with the version-pinning convention used throughout this file. Pin to the release tag.

mwojtyczka (commented):
Inline review comments from a correctness pass. The main issues (first two) are that the demo teaches the opposite of DQX's actual split semantics; the rest are smaller.

demos/dqx_demo_industry/dqx_customer_360_demo.py
mwojtyczka: This is addressed, but the note that was added is wrong. The three warn rows are now labeled "Warning-level issue" (not "Invalid:") and carry the note "Warning-level violations remain in valid_df and are not quarantined" (lines 121/161/227, and the markdown at 444). So the cosmetic request was satisfied — but that note misstates DQX behavior, and so did the original comment's premise. Per get_invalid() (errors OR warnings), warn-only rows are in invalid_df and do get written to the quarantine table; and get_valid() drops both result columns, so the copies in valid_df have no warning metadata. So: the old concern is resolved, but the "fix" introduced a factual inaccuracy that still needs correcting.

mwojtyczka (changes requested):
Thanks for the fixes. 1 old comment is still not fully addressed and I added few more.

docs/dqx/docs/demos.mdx
```diff
@@ -69,6 +69,7 @@ Import the following notebooks in the Databricks workspace to try DQX out:
...
 * [DQX for Fashion & Retail Accelerator Demo](https://github.com/databrickslabs/dqx/blob/v0.16.0/demos/dqx_demo_industry/dqx_fashion_demo.py) - demonstrates DQX rules tailored for the Fashion and Retail industry, including SKU validation, hex color formats, and pricing compliance.
 * [DQX for Manufacturing Industry Accelerator Demo](https://github.com/databrickslabs/dqx/blob/v0.16.0/demos/dqx_demo_industry/dqx_manufacturing_demo.py) - demonstrates how to use DQX to check data quality for Manufacturing Industry datasets, specifically for machine maintenance and sensor data.
+* [DQX for Customer 360 Industry Accelerator Demo](https://github.com/databrickslabs/dqx/blob/v0.16.0/demos/dqx_demo_industry/dqx_customer_360_demo.py) - demonstrates DQX data quality rules for Customer 360 datasets, including customer identity, email format, revenue, ticket counts, purchase dates, and active customer flag consistency.
```
amising6, line 72: updated tag

amising6:
Hi, I have pushed the requested updates and addressed the latest review comments.

The changes include correcting the warning and quarantine explanation, using runtime-relative dates, updating the active-status flag description, and pinning the demo documentation link to `v0.16.0`.

I have also re-requested the review. Thanks again for the detailed feedback.

demos/dqx_demo_industry/dqx_customer_360_demo.py
```diff
@@ -0,0 +1,484 @@
...
+from databricks.sdk import WorkspaceClient
+from databricks.labs.dqx.engine import DQEngine
+
+customer_360_checks_yaml = """
+# 1. Customer ID must be present
+- criticality: error
+  name: customer_id_not_null
+  check:
+    function: is_not_null_and_not_empty
+    for_each_column:
```
mwojtyczka: Any reason this is using `for_each_column`? I would remove and use the standard syntax for a single column. Also if using `for_each_column` we don't define name as it is auto-generated.

mwojtyczka: resolved

mwojtyczka (approved):
LGTM - i added the demo to the e2e tests

PR #1486, opened 2026-08-25
Fix escaping during check execution

### LINKED ISSUE #1481: Column names requiring SQL identifier escaping pass validation but fail during check execution (`INVALID_IDENTIFIER`)

### Current Behavior

Checks defined against a column whose name requires SQL identifier escaping (non-ASCII letters, spaces, etc.) are accepted by validation but produce an invalid logical plan, and the run fails with `ParseException: [INVALID_IDENTIFIER]` at the first schema resolution.

The quoting introduced in [PR:1342](https://github.com/databrickslabs/dqx/pull/1342) `utils.quote_column_name()` is applied on the **validation** path only. The **execution** path builds the condition from the raw, unquoted string:

| Path | Code | Quoted? |
| --- | --- | --- |
| Validation | `manager.DQRuleManager._is_invalid_column()` → `_is_invalid_quoted_column()` → `F.expr(quote_column_name(column))` | yes |
| Skip / error messages | `manager.DQRuleManager._display_column_name()` | yes |
| Execution | `check_funcs.get_normalized_column_and_expr()` → `check_funcs._get_column_expr()` → `F.expr(column)` | **no** |

`_get_column_expr()` is reached from `get_normalized_column_and_expr()`, which has ~47 call sites in `check_funcs.py`, so this affects essentially every row-level check function.

Because `ParseException` subclasses `AnalysisException` in PySpark, `_is_invalid_column()` catches the parse failure of the bare name, retries successfully with backticks, and concludes the column is valid. The check is therefore **not** skipped, and the unquoted identifier is baked into the plan. Under Spark Connect `F.expr()` does not parse eagerly, so nothing fails until the first analyze RPC — in our case `df.columns` inside `reporting_columns.merge_info_columns()`, which makes the stacktrace point at a line unrelated to the offending check.

This looks like a behavioural regression from the partial fix: without the backtick fallback in `_is_invalid_column()`, such a column was reported as invalid and the check was silently skipped, so the pipeline completed. With the fallback, the same input now aborts the run.

Note that `datacontract/contract_rules_generator.py` emits `column_path` unquoted at all of its check-generation sites, so DQX's own ODCS contract generator produces failing checks for any contract whose schema uses such column names.

### Expected Behavior

A check defined as `{"function": "is_not_null", "arguments": {"column": "Päivämäärä"}}` should execute correctly (preferred — the column resolves fine once escaped, and validation already proves it).

### Minimal reproduction of the asymmetry

```python
from pyspark.sql import SparkSession, functions as F
from databricks.labs.dqx.check_funcs import get_normalized_column_and_expr
from databricks.labs.dqx.utils import quote_column_name

column = "Päivämäärä"

df = spark.range(1).select(F.col("id").alias(column))

# Validation path (manager._is_invalid_column -> _is_invalid_quoted_column): resolves fine,
# so the column is considered valid and the check is NOT skipped.
df.select(F.expr(quote_column_name(column))).schema
# -> StructType([StructField('Päivämäärä', LongType(), False)])

# Execution path (used by ~47 check functions via get_normalized_column_and_expr):
_, _, col_expr = get_normalized_column_and_expr(column)
df.select(col_expr).schema
# -> ParseException: [INVALID_IDENTIFIER] The unquoted identifier Päivämäärä is invalid
#    and must be back quoted as: `Päivämäärä`. ... SQLSTATE: 42602 (line 1, pos 0)
```

### End-to-end reproduction

```python
from pyspark.sql import SparkSession, functions as F

df = spark.range(3).select(F.col("id").alias("Päivämäärä"))

checks = [
    {
        "criticality": "error",
        "check": {"function": "is_not_null", "arguments": {"column": "Päivämäärä"}},
    }
]

engine = DQEngine(WorkspaceClient())
result_df, _ = engine.apply_checks_by_metadata(df, checks)
# raises ParseException [INVALID_IDENTIFIER] inside merge_info_columns -> df.columns

### Cloud

Azure

### Operating System

Linux

```shell
File ".../src/common/dqx/dqx_pipeline.py", line 140, in _apply_checks_with_metrics
    result_df, observation = dq_engine.apply_checks_by_metadata(source_df, dq_rules)

File ".../site-packages/databricks/labs/dqx/telemetry.py", line 119, in wrapper
    return func(self, *args, **kwargs)

File ".../site-packages/databricks/labs/dqx/engine.py", line 828, in apply_checks_by_metadata
    return self._engine.apply_checks_by_metadata(
        df=df, checks=checks, custom_check_functions=custom_check_functions, ref_dfs=ref_dfs
    )

File ".../site-packages/databricks/labs/dqx/engine.py", line 287, in apply_checks_by_metadata
    return self.apply_checks(df, dq_rule_checks, ref_dfs)

File ".../site-packages/databricks/labs/dqx/engine.py", line 185, in apply_checks
    result_df = self._create_results_array(...)

File ".../site-packages/databricks/labs/dqx/engine.py", line 631, in _create_results_array
    result_df = merge_info_columns(dest_info_col, result_df, info_col_names=info_col_names)

File ".../site-packages/databricks/labs/dqx/reporting_columns.py", line 43, in merge_info_columns
    if not info_cols and dest_name not in df.columns:

File "/databricks/spark/python/pyspark/sql/connect/dataframe.py", line 289, in columns
    return self.schema.names

File "/databricks/spark/python/pyspark/sql/connect/dataframe.py", line 2000, in schema
    self._cached_schema = self._session.client.schema(query)

File "/databricks/spark/python/pyspark/sql/connect/client/core.py", line 1271, in schema
    schema = self._analyze(method="schema", plan=plan).schema

ParseException: [INVALID_IDENTIFIER] The unquoted identifier Päivämäärä is invalid and must be
back quoted as: `Päivämäärä`. Unquoted identifiers can only contain ASCII letters ('a' - 'z',
'A' - 'Z'), digits ('0' - '9'), and underbar ('_'). Unquoted identifiers must also not start
with a digit. Different data sources and meta stores may impose additional restrictions on
valid identifiers. SQLSTATE: 42602 (line 1, pos 0)
```

ghanse:
## Changes

This PR adds escaping for column names during check execution.

***Note:** Auto-generated check names (e.g. when checks are not named by the user) are normalized to replace reserved characters with `_`.*

### Linked issues

Resolves #1481 

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

### Documentation and Demos

- [ ] added/updated demos
- [ ] added/updated docs


src/databricks/labs/dqx/check_funcs.py
```diff
@@ -4635,8 +4636,12 @@ def get_normalized_column_and_expr(column: str | Column) -> tuple[str, str, Colu
             - Spark Column expression corresponding to the input.
     """
     col_expr = _get_column_expr(column)
-    column_str = get_column_name_or_alias(col_expr)
-    col_str_norm = get_column_name_or_alias(col_expr, normalize=True)
+    if isinstance(column, str):
```
mwojtyczka: shouldn't this logic be implemented inside `get_column_name_or_alias`?
We already handle the case there:
```
    if isinstance(column, str):
        col_str = column
```
we just need to extend it with an option to normalize:
```
    if isinstance(column, str):
        if normalize:
            col_str = normalize_col_str(column)
        else:
             col_str = column
```

Then we don't need to change check_funcs.py
```
column_str = get_column_name_or_alias(col_expr)
col_str_norm = get_column_name_or_alias(col_expr, normalize=True)
```

src/databricks/labs/dqx/utils.py
```diff
@@ -308,6 +310,43 @@ def quote_column_name(name: str) -> str:
...
+    This keeps nested-field access (e.g. "struct_col.field1") working while escaping the segments that
+    require it.
+
+    Args:
+        column: Column reference provided as a string (plain name, nested path, or SQL expression).
+
+    Returns:
+        A string safe to pass to ``F.expr``.
+    """
+    if SQL_EXPRESSION_CHAR_PATTERN.search(column):
```
mwojtyczka: **Finding: word-only SQL expressions are back-quoted into a single identifier**

`normalize_column_expr` back-quotes word-only SQL expressions (containing only letters/digits/spaces, no operator characters) as a single identifier, breaking expressions like `col IS NOT NULL` or `a AND b` when passed as a check column.

**Failure scenario:** A check func receives `column="col IS NOT NULL"` (a supported string expression). `SQL_EXPRESSION_CHAR_PATTERN` finds no operator char (space is not in the class), so the whole string is treated as a name and quoted to `` `col IS NOT NULL` ``. `F.expr("`col IS NOT NULL`")` then resolves it as one literal column that does not exist, whereas the pre-PR `F.expr("col IS NOT NULL")` parsed a boolean expression. Regression for operator-free expressions.

src/databricks/labs/dqx/utils.py
```diff
@@ -308,6 +310,43 @@ def quote_column_name(name: str) -> str:
...
+
+    Strings containing expression characters (operators, grouping, ...) are treated as expressions and
+    returned unchanged. Otherwise the string is treated as a dotted column path: each segment is left
+    alone if it is a valid bare identifier or already back-quoted, and back-quoted if it needs escaping.
+    This keeps nested-field access (e.g. "struct_col.field1") working while escaping the segments that
+    require it.
+
+    Args:
+        column: Column reference provided as a string (plain name, nested path, or SQL expression).
+
```
mwojtyczka, line 361: **Finding: column names containing expression characters are not escaped (incomplete fix)**

`normalize_column_expr` does not escape column names that contain any expression character (hyphen, comma, parentheses, quote, etc.), so such names are passed through unquoted and misparsed — the escaping fix is incomplete.

**Failure scenario:** A DataFrame has a column literally named `gross-margin` or `amount (usd)`. `normalize_column_expr` sees the `-` / `(` as an expression character and returns the string unchanged, so `F.expr("gross-margin")` parses as subtraction (gross MINUS margin) and `F.expr("amount (usd)")` is a syntax error. These names require back-quoting, but the heuristic only handles spaces and non-ASCII letters, leaving a large class of names-requiring-escaping still broken.

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: **Finding: display/normalized names now derived from raw input, changing messages and check names**

For string column inputs, the display name (`col_expr_str`) and normalized name (`col_str_norm`) now come from the raw input rather than the parsed Column, changing error messages and auto-generated check names for pre-quoted or aliased/expression strings.

**Failure scenario:** User passes `column="`Customer Name`"` (already back-quoted). Pre-PR `column_str` was parsed to `Customer Name`; now `column_str = column`, so the error message reads ``Column '`Customer Name`' value is null`` with literal backticks. Likewise `column="a AS b"` previously produced display/name from the alias `b`; now it is the raw `a AS b` (`col_str_norm` becomes `a_as_b`), changing user-visible messages and check names.

src/databricks/labs/dqx/utils.py
```diff
@@ -308,6 +310,43 @@ def quote_column_name(name: str) -> str:
...
+    if SQL_EXPRESSION_CHAR_PATTERN.search(column):
+        return column
+
+    segments = column.split(".")
+    normalized_segments = [
+        (
+            segment
+            if VALID_UNQUOTED_IDENTIFIER_PATTERN.match(segment) or (segment.startswith("`") and segment.endswith("`"))
+            else quote_column_name(segment)
+        )
```
mwojtyczka: **Finding: unreachable dead-code branch**

The `` segment.startswith("`") and segment.endswith("`") `` guard inside this comprehension is unreachable dead code.

**Failure scenario:** `SQL_EXPRESSION_CHAR_PATTERN` (line 82) includes the backtick character, so any string containing a backtick returns early at the top of the function. A segment can only contain a backtick if the whole string did, meaning the already-back-quoted check can never evaluate. It is dead code that suggests a mis-scoped guard and adds maintenance noise.

mwojtyczka (changes requested):

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: This is still not resolved. The current implementation branches on `isinstance(column, str)` inside `get_normalized_column_and_expr` and calls `unquote_column_name` + `normalize_col_str` directly, rather than adding a `normalize` option to `get_column_name_or_alias` as suggested — so `check_funcs.py` was still changed. Leaving open to confirm whether we want to centralize this in `get_column_name_or_alias` instead.

src/databricks/labs/dqx/utils.py
```diff
@@ -308,6 +311,62 @@ def quote_column_name(name: str) -> str:
...
+        column: Column reference provided as a string (plain name, nested path, or SQL expression).
+
+    Returns:
+        A string safe to pass to ``F.expr``.
+    """
+    if SQL_EXPRESSION_STRUCTURE_PATTERN.search(column) or SQL_EXPRESSION_OPERATOR_PATTERN.search(column):
+        return column
+
+    return ".".join(
+        segment if VALID_UNQUOTED_IDENTIFIER_PATTERN.match(segment) else quote_column_name(segment)
```
mwojtyczka, line 369: **Finding: a column whose name equals a zero-arg SQL function resolves to the function, not the column (silent wrong result)**

A bare identifier that matches `VALID_UNQUOTED_IDENTIFIER_PATTERN` is returned unchanged, so a real column literally named `current_date`, `current_timestamp`, or `current_user` is passed to `F.expr` unquoted and evaluates the SQL function instead of the column.

**Failure scenario:** a DataFrame has a column named `current_date`. `normalize_column_expr("current_date")` returns `"current_date"` (valid identifier, left alone), so `F.expr("current_date")` yields today's date rather than the column's value. `is_not_null` etc. then check the wrong value with no error surfaced. Only closable by always back-quoting a name-classified segment, or schema-aware resolution. Lower priority (needs an unusually-named column), flagging for awareness.

src/databricks/labs/dqx/utils.py
```diff
@@ -308,6 +311,62 @@ def quote_column_name(name: str) -> str:
...
+    back-quoted identifier (plain names, dotted paths, SQL expressions) are returned unchanged.
+
+    Args:
+        name: Column reference provided as a string.
+
+    Returns:
+        The column name without surrounding back-quotes.
+    """
+    if len(name) >= 2 and name.startswith("`") and name.endswith("`"):
+        return name[1:-1].replace("``", "`")
```
mwojtyczka, line 333: **Finding: dotted path of separately back-quoted segments is mangled (cosmetic)**

`unquote_column_name` only strips the outer back-quotes, so a path whose segments are individually quoted comes back malformed.

**Failure scenario:** `column="`Odd Name`.`field`"` → `unquote_column_name(...)` returns `"Odd Name`.`field"`, from which `normalize_col_str` derives a garbled check name and the `_errors`/`_warnings` message shows a corrupted column name. Resolution itself is fine (`normalize_column_expr` leaves the string unchanged and `F.expr` resolves it), so this is display-only — but user-visible. Low severity.

src/databricks/labs/dqx/utils.py
```diff
@@ -308,6 +311,62 @@ def quote_column_name(name: str) -> str:
...
+    return name
+
+
+def normalize_column_expr(column: str) -> str:
+    """
+    Prepares a column reference string for use with ``F.expr``, back-quoting names that require SQL
```
mwojtyczka, line 339: **Nit: docstring uses double-backtick object references, against the repo convention**

AGENTS.md (Docstrings) says: *No backticks around object names — use italics instead (e.g., *arg1*, *column*). Backticks cause rendering issues in API docs.* This docstring uses ``F.expr`` (and `quote_column_name` / `unquote_column_name` use ``F.expr`` and `:func:` cross-refs) — should be italics, e.g. *F.expr*. Note `_get_column_expr` in check_funcs.py already does this correctly (*normalize_column_expr*).

mwojtyczka (changes requested):
Previous comments all addressed except 1, added a few new comments

src/databricks/labs/dqx/utils.py
ghanse: This is also DBSQL's convention. Users must escape zero-arg identifiers in any DBSQL query if they use them as column names. Should we change this or align with DBSQL?

mwojtyczka: sounds good, I think it's better to stay aligned with DBSQL

ghanse:
@mwojtyczka ready for another review. `test_apply_checks.py` is quite large and failing pylint rules now, should we decompose this into a few separate test modules?

PR #1488, opened 2026-08-27
Fix duplicate-key row multiplication in dataset comparisons

### LINKED ISSUE #1487: compare_datasets multiplies rows for duplicate matching keys
## Is there an existing issue for this?

- [x] I searched open and closed issues and pull requests for `compare_datasets`, duplicate/unique matching keys, row multiplication, cardinality, and duplicate primary keys.

## Current Behavior

`compare_datasets` silently performs a many-to-many join when the configured matching columns are not unique. This can multiply the output row count and report differences between two identical datasets.

For example, comparing these identical datasets using `id` as the matching column:

| id | value |
| --- | --- |
| 1 | A |
| 1 | B |

produces four joined pairs (`A/A`, `A/B`, `B/A`, and `B/B`). The two cross-pairs are then reported as value changes even though the datasets are identical.

The public parameter documentation describes `columns` and `ref_columns` as columns used for row matching, but it does not state that uniqueness is required. `_match_rows` performs the equality join without validating that either side has at most one row per matching-key combination.

This can produce false data-quality violations and inflated output counts in reconciliation or migration checks.

## Expected Behavior

`compare_datasets` must not silently return incorrect, grain-multiplied results.

Because duplicate matching keys make one-to-one row comparison ambiguous, the smallest safe behavior is to fail before joining when either the input or reference dataset contains a duplicate matching-key combination. The error should identify which dataset and matching columns violate the uniqueness requirement.

## Steps To Reproduce

```python
from databricks.labs.dqx.check_funcs import compare_datasets

df = spark.createDataFrame([(1, "A"), (1, "B")], "id int, value string")
ref_df = spark.createDataFrame([(1, "A"), (1, "B")], "id int, value string")

condition, apply = compare_datasets(
    columns=["id"],
    ref_columns=["id"],
    ref_df_name="ref_df",
)

result = apply(df, spark, {"ref_df": ref_df})
result.select("id", "value", condition).show(truncate=False)
```

The equality join in `_match_rows` creates four output rows. `_add_column_diffs` marks the `A/B` and `B/A` cross-pairs as changed.

## Proposed acceptance criteria

- Validate uniqueness of the configured matching-column combinations on both input datasets before `_match_rows`.
- Raise `InvalidParameterError` instead of returning multiplied or false-difference rows.
- Identify the offending dataset and matching columns in the error.
- Treat repeated null matching-key combinations as duplicates, consistent with null-safe row matching.
- Add public-API integration coverage for duplicate keys on both the input and reference sides.
- Document the matching-key uniqueness requirement and failure behavior.

## Environment

- DQX `main` at `191938dc677422c2f291d3202f0636e2e332a475`
- Latest affected release: `v0.16.0`
- Cloud: provider-independent Spark join behavior
- Operating system used for source inspection: macOS

## Relevant log output

No exception is currently raised; the failure is incorrect output cardinality and false comparison results.

mattfaltyn:
## Changes

- Prevent duplicate-key Cartesian fan-out in `compare_datasets` with lazy one-to-one pairing by default. Rows within each matching-key group receive a `row_number()` ordered by the string representation of compared values, with nulls first.
- Add `raise_on_duplicate_keys=True` to opt into eager uniqueness validation on both datasets. With standard null matching, rows containing null matching keys are excluded from uniqueness validation because they cannot match.
- Use the generated row number to distinguish present null-key rows from missing rows during default pairing.
- Document lexical numeric ordering and that pairing does not minimize reported changes.
- Explicitly reject streaming DataFrames on either side and document the batch-only restriction in the changelog.
- Restore repeated benchmark measurements for both lazy pairing and strict validation, including check construction in the timed section.

This prevents many-to-many joins from multiplying output rows and reporting false changes between identical datasets, while keeping the default comparison lazy.

BREAKING CHANGES!

* `compare_datasets` now explicitly rejects streaming DataFrames on either side with `InvalidParameterError`; its duplicate-key handling requires bounded batch inputs. Streaming callers must compare bounded snapshots or batch DataFrames within `foreachBatch`. ([#1488](https://github.com/databrickslabs/dqx/pull/1488))

### Linked issues

Resolves #1487

### Tests

- [x] manually verified duplicate-key join cardinality and existing unique-key comparisons
- [x] added unit tests for streaming rejection on either side in both modes
- [x] added integration coverage for strict validation, lazy construction, null keys, duplicate pairing, and unmatched rows
- [ ] added end-to-end tests — no end-to-end surface changed
- [x] updated performance benchmarks for both modes — workspace-scale benchmarks have not been run locally

Local validation:

- `make fmt`
- `make lint`
- `make test` — 2,484 passed
- Standalone PySpark 4.0.0 comparison checks — 27 passed using unchanged repository test bodies with local Spark/UTC fixtures; the Unity Catalog reference-table case was excluded
- `UV_FROZEN=1 UV_BUILD_CONSTRAINT=.build-constraints.txt uv run --exact --all-extras --group docs pydoc-markdown` — succeeded with three cross-reference warnings in unchanged aggregate helpers
- `corepack yarn --cwd docs/dqx build` — 150 documents indexed, production build succeeded

The repository's workspace-backed integration fixture could not initialize because `DATABRICKS_HOST` is unset. The local Spark checks do not replace Databricks Connect, Unity Catalog, or workspace-scale benchmark validation.

### Documentation and Demos

- [ ] added/updated demos — no demo flow changed
- [x] updated docstrings, reference documentation, and changelog
- [ ] added/updated agent skills — not applicable


ghanse (commented):
Left a comment here regarding design: https://github.com/databrickslabs/dqx/issues/1487#issuecomment-5465427917

Let me know what you think.

mattfaltyn:
> Failing fast with an `InvalidParameterError` is a reasonable default. I think we should optionally support duplicate keys.

Thanks @ghanse. Implemented in 41d9bc90. Strict uniqueness remains the default; `allow_duplicate_keys=True` opts into `row_number()` pairing ordered by compared values. Added docs, tests, and benchmark coverage.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -3017,15 +3032,42 @@ def apply(df: DataFrame, spark: SparkSession, ref_dfs: dict[str, DataFrame]) ->
...
+            ref_join_columns = [*ref_pk_column_names, row_number_col]
+        else:
+            match_count_col = f"__match_count_{unique_id}"
+            for dataset_name, dataset, matching_columns in (
+                ("source", df, pk_column_names),
+                ("reference", ref_df, ref_pk_column_names),
+            ):
+                matchable_rows = dataset if null_safe_row_matching else dataset.dropna(subset=matching_columns)
+                duplicate_keys = matchable_rows.groupBy(*matching_columns).agg(F.count("*").alias(match_count_col))
+                if not duplicate_keys.where(F.col(match_count_col) > 1).isEmpty():
```
ghanse, line 3140: The bigger thing: this default path fires two eager Spark jobs (`.isEmpty()` on each side) on every comparison, even when there are no dupes — but `compare_datasets` is otherwise lazy like the rest of our checks. The catch is you can't raise without an action, so any "validate then raise" is inherently eager.

Worth flipping this around: the `row_number()` pairing you added below is already fully lazy and already prevents the fan-out that #1487 is about. If that's the default, the bug is fixed with zero extra jobs, and unique keys are unchanged (rn=1 on both sides, so joining on `(pk, rn)` is the same as joining on `pk`). If we still want to surface non-unique keys, we can do it lazily instead of raising:

```python
duplicate_key_condition = F.count(F.lit(1)).over(Window.partitionBy(*pk_column_names)) > 1
```

and fold it into the status JSON like `row_missing`/`changed`. A hard raise-and-stop is still reasonable, but since it's inherently eager I'd make it the opt-in flag (`raise_on_duplicate_keys=True`) rather than the default. Curious what you think — happy to be talked out of it if the raise needs to be the default.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -2994,6 +3006,9 @@ def compare_datasets(
     def apply(df: DataFrame, spark: SparkSession, ref_dfs: dict[str, DataFrame]) -> DataFrame:
         ref_df = _get_ref_df(ref_df_name, ref_table, ref_dfs, spark)
 
+        if df.isStreaming or ref_df.isStreaming:
```
ghanse, line 3088: Good guard — the window/`groupBy` machinery wouldn't behave on streaming anyway, so a clear error beats a cryptic one. Just flagging it's technically a new restriction for any streaming caller, so worth a line in the changelog/release notes.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -3017,15 +3032,42 @@ def apply(df: DataFrame, spark: SparkSession, ref_dfs: dict[str, DataFrame]) ->
         # determine skipped columns: present in df, not compared, and not PK
         skipped_columns = [col for col in df.columns if col not in compare_columns and col not in pk_column_names]
 
+        join_columns = pk_column_names
+        ref_join_columns = ref_pk_column_names
+        if allow_duplicate_keys:
+            row_number_col = f"__match_row_number_{unique_id}"
+            order_columns = [F.col(col).cast("string").asc_nulls_first() for col in compare_columns] or [F.lit(1)]
```
ghanse, line 3123: Minor: casting everything to string means numerics order lexically (`"10" < "2"`). Doesn't affect correctness — the diff runs on the real values and identical rows still cluster together — just noting the pairing order is a bit arbitrary for numbers. Fine to leave as-is.

tests/perf/test_apply_checks.py
```diff
@@ -1306,12 +1307,14 @@ def test_benchmark_compare_datasets(benchmark, ws, generated_df, make_ref_df):
...
                 "ref_columns": ["ref_col1", "ref_col2"],
                 "ref_df_name": "ref_df",
+                "allow_duplicate_keys": allow_duplicate_keys,
             },
         ),
     ]
     refs_df = {"ref_df": make_ref_df}
-    checked = dq_engine.apply_checks(generated_df, checks, refs_df)
-    actual_count = benchmark(lambda: checked.count())
+    actual_count = benchmark.pedantic(
```
ghanse: `pedantic(rounds=1, iterations=1)` is a single measurement with no repeats, so this won't catch perf regressions anymore — it's basically a smoke test now. Was that intentional? If you want a real signal, keeping the default multi-round `benchmark()` on at least the `strict` case would help.

ghanse (commented):
Solid PR overall — it matches the design we landed on in #1487 (deterministic pairing for duplicate keys), with good test and doc coverage, and I didn't spot any correctness bugs.

My one real design question is whether the default should raise at all: raising needs an eager action, whereas the `row_number()` pairing you added is already lazy and already fixes the fan-out — so it might make a better default, with the hard-fail as an opt-in. Not a hard blocker, but worth a discussion. The rest are minor.

src/databricks/labs/dqx/check_funcs.py
mattfaltyn: > If that's the default, the bug is fixed with zero extra jobs

Agreed, thanks! Pairing is now lazy by default; `raise_on_duplicate_keys=True` opts into eager validation. Added regressions for laziness and null-key pairing.

src/databricks/labs/dqx/check_funcs.py
mattfaltyn: > worth a line in the changelog/release notes.

Added the batch-only restriction to the breaking-change notes, with guidance for streaming callers. Thanks for flagging it!

src/databricks/labs/dqx/check_funcs.py
mattfaltyn: > Fine to leave as-is.

Thanks! Kept the ordering and documented lexical numeric sorting; comparisons still use the original values.

tests/perf/test_apply_checks.py
mattfaltyn: > keeping the default multi-round `benchmark()` on at least the `strict` case would help.

Good catch! Restored multi-round `benchmark()` for both modes, including check construction so strict validation is measured.

CHANGELOG.md
```diff
@@ -48,6 +50,8 @@
 
 BREAKING CHANGES!
 
+* `compare_datasets` now explicitly rejects streaming DataFrames on either side with `InvalidParameterError`; its duplicate-key handling requires bounded batch inputs. Streaming callers must compare bounded snapshots or batch DataFrames within `foreachBatch`. ([#1488](https://github.com/databrickslabs/dqx/pull/1488))
```
mwojtyczka: Please don't update changelog. We are updating it when making new release. Please move the breaking change to the PR description.

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: we are generatic changelog during release, pls move it to PR description

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## 0.16.0
 
+* Fixed duplicate-key row multiplication in `compare_datasets` ([#1487](https://github.com/databrickslabs/dqx/issues/1487)). Duplicate-key groups now use lazy ordered one-to-one pairing by default. Set `raise_on_duplicate_keys=True` to opt into eager uniqueness validation.
```
mwojtyczka:
```suggestion
```

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
...
+
```
mwojtyczka:
```suggestion
```

CHANGELOG.md
mwojtyczka:
```suggestion
```

CHANGELOG.md
```diff
@@ -48,6 +50,8 @@
 
 BREAKING CHANGES!
 
+* `compare_datasets` now explicitly rejects streaming DataFrames on either side with `InvalidParameterError`; its duplicate-key handling requires bounded batch inputs. Streaming callers must compare bounded snapshots or batch DataFrames within `foreachBatch`. ([#1488](https://github.com/databrickslabs/dqx/pull/1488))
+
```
mwojtyczka:
```suggestion
```

mwojtyczka (changes requested):
I left a few comments, there are 2 correctness issues that needs addressing before we merge

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -3017,16 +3034,44 @@ def apply(df: DataFrame, spark: SparkSession, ref_dfs: dict[str, DataFrame]) ->
...
+                ("source", df, pk_column_names),
+                ("reference", ref_df, ref_pk_column_names),
+            ):
+                matchable_rows = dataset if null_safe_row_matching else dataset.dropna(subset=matching_columns)
+                duplicate_keys = matchable_rows.groupBy(*matching_columns).agg(F.count("*").alias(match_count_col))
+                if not duplicate_keys.where(F.col(match_count_col) > 1).isEmpty():
+                    raise InvalidParameterError(
+                        f"The {dataset_name} dataset contains duplicate matching keys for columns: "
+                        f"{', '.join(matching_columns)}."
+                    )
```
mwojtyczka, line 3144: **Correctness: eager path skips the row-number disambiguation.**

When `raise_on_duplicate_keys=True`, this branch validates uniqueness but leaves `join_columns`/`ref_join_columns` at their PK-only defaults (lines 3037-3038) — it never adds the `__match_row_number` column that the lazy branch above uses to distinguish a present null-key row from a missing side of the join.

With the default `null_safe_row_matching=True`, a single null-value matching key passes the uniqueness check (count = 1 per side), then matches itself via the null-safe join. `_add_row_diffs` (line 3074) then computes `row_missing = df.<pk>.isNull()` and `row_extra = ref.<pk>.isNull()`, so a genuinely matched row is flagged as **both** `row_missing:true` and `row_extra:true`.

Repro — `df=[(None,"X")]`, `ref=[(None,"Y")]`, `check_missing_records=True`:
- lazy (default): `{row_missing:false, row_extra:false, changed:{value: X->Y}}` ✅
- eager (`raise_on_duplicate_keys=True`): `{row_missing:true, row_extra:true, changed:{value: X->Y}}` ❌

The two modes disagree on identical input and the eager result is self-contradictory. Suggest extending the row-number disambiguation to this branch as well, and adding a both-null-keys-match test under `raise_on_duplicate_keys=True` — that scenario is currently uncovered (the existing null-key tests all run the lazy path).

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -3017,16 +3034,44 @@ def apply(df: DataFrame, spark: SparkSession, ref_dfs: dict[str, DataFrame]) ->
...
+            row_number_col = f"__match_row_number_{unique_id}"
+            order_columns = [F.col(col).cast("string").asc_nulls_first() for col in compare_columns] or [F.lit(1)]
+            df = df.withColumn(
+                row_number_col,
+                F.row_number().over(Window.partitionBy(*pk_column_names).orderBy(*order_columns)),
+            )
+            ref_df = ref_df.withColumn(
+                row_number_col,
+                F.row_number().over(Window.partitionBy(*ref_pk_column_names).orderBy(*order_columns)),
+            )
```
mwojtyczka, line 3131: **Perf trade-off: two window computations on every default comparison.**

The default (lazy) branch unconditionally adds two `row_number()` windows — one over `df`, one over `ref_df`, each `partitionBy(pk).orderBy(<all compare_columns>)`. That's two extra shuffles plus a full sort of both datasets on every compared column, even when the matching keys are already unique (the common case).

The only alternative today is `raise_on_duplicate_keys=True`, which replaces this with two eager `groupBy().count()` validation jobs — so there's no lightweight path back to the old single-join behavior when the caller already knows the keys are unique. Worth acknowledging as a trade-off; a fast opt-out for the known-unique case may be worth considering.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -2994,6 +3008,9 @@ def compare_datasets(
     def apply(df: DataFrame, spark: SparkSession, ref_dfs: dict[str, DataFrame]) -> DataFrame:
         ref_df = _get_ref_df(ref_df_name, ref_table, ref_dfs, spark)
 
+        if df.isStreaming or ref_df.isStreaming:
+            raise InvalidParameterError("compare_datasets only supports batch DataFrames")
```
mwojtyczka: **Nit: streaming guard message is unactionable.**

The guard itself is reasonable — dataset comparison needs bounded inputs, and `isStreaming` can only be checked after `_get_ref_df` resolves `ref_df`. But the message states the restriction without the remedy, which the docstring already documents. Consider pointing users at the workaround:

```python
raise InvalidParameterError(
    "compare_datasets requires bounded batch DataFrames and does not support streaming inputs. "
    "Compare bounded snapshots, or run the comparison inside foreachBatch on each micro-batch."
)
```

mwojtyczka (commented):
Automated review pass on the duplicate-key fix. The core lazy `row_number` pairing looks correct and prevents the Cartesian fan-out. Three inline notes below: one correctness issue on the eager path (worth fixing), plus a perf trade-off and an error-message nit to consider.

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: I would leave as is and improve docs

mwojtyczka: fixed

src/databricks/labs/dqx/check_funcs.py

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: fixed

mwojtyczka (approved):
LGTM

PR #1489, opened 2026-08-28
Row anomaly detection: a second detector, and the end of `segment_by`

### LINKED ISSUE #1484: Anomaly detection: grouped data has no scalable path to conditioning
## Problem

For data that is grouped by categorical attributes (events by market and type, metrics per
customer, readings per device), row anomaly detection has two paths and neither works:

- **Pooled** (the default, and what you get whenever `columns` is passed) cannot detect a
  *contextual* anomaly — a value that is ordinary globally but wrong for its own group.
- **`segment_by`** can, by training one model per group, but the model count is the product of
  the dimension cardinalities, and the per-group models turn out to be calibration-fragile.

This is currently invisible to users: the pooled path returns a confident, low severity rather
than an error or a warning.

## Evidence

Aggregate table, 90 groups (country x event_type x product) x 540 days. One group's volume was
dropped 80% on the scoring day, with the lost volume redistributed so the **daily total was exactly
preserved** -- the total moved only -5.9%, entirely from seasonality, with the collapse contributing
nothing to it. A second group was taken to zero. Threshold 95.

| config | the -80% group | group -> 0 | models | wall clock |
|---|---|---|---|---|
| pooled, group cols as categorical features | **45.1 — missed** | 82.0 — missed | 3 | ~5 min |
| `segment_by`, 8-group subset | **97.0 — flagged** | 96.8 — flagged | 24 | 7m55s |
| pooled + group-relative features | **95.5 — flagged** | 96.8 — flagged | **3** | **~4 min** |

The same group scored **42.0** at its normal volume the week before, so the collapse moved it 55
percentile points. 0 false positives on the control day in every config.

Two things this establishes:

1. **45.1 is not a near-miss.** It is the 45th percentile. No threshold recovers it — you would have
   to flag over half the table. This is a representation problem, not a tuning problem.
2. **Conditioning is the entire missing ingredient.** Algorithm, features, seasonality handling and
   threshold were identical across rows 1 and 2. Only the comparison basis changed.

### Segmentation does not scale

Segmented training passes `allow_ensemble=False` (`training_service.py:383`), and
`training_strategies.py:95` resolves the count as `params.ensemble_size if allow_ensemble and
params.ensemble_size else 1` — so **segmentation always trains exactly one model per segment,
regardless of `ensemble_size`**. Measured ~60s per model: 8 groups (8 models) completed in 7m55s
⇒ ~59s/model, which projects 90 groups to ~88 minutes. **90 groups (90 models) ran 70 minutes
without completing** and had to be cancelled — consistent with that projection. `_get_and_validate_segments` warns above 100
segments and proceeds; auto-discovery accepts any categorical with 2-50 distinct values and can pick
several, so ~125,000 segments is reachable behind a single log line.

### Per-group models are also calibration-fragile

On SMD (Server Machine Dataset, 8 entities, 199,362 test rows, 4.9% anomaly rate), per-entity models
were the **worst** of the three trained configurations, with catastrophic false positives on specific
entities — machine-3-1 produced 15,963 false positives out of 28,392 normal rows, a 56% false-alarm
rate at precision 0.012. Its aggregate recall came from overcalling, not discrimination.

Mechanism: severity is calibrated against each model's own training score distribution, so a 95th
percentile threshold flags ~5% of training-like rows by construction. 56% means that entity's test
scores sit far above its training scores. Each per-group model is calibrated on a small sample, so
each is independently fragile; pooling gives more stable calibration.

*That run used `ensemble_size=1`, but this is not a handicap the benchmark imposed: segmented
training never ensembles (`allow_ensemble=False`, `training_service.py:383`), so one model per group is
what this code path always does. The comparison is therefore the honest one, and the instability is a
property of per-group calibration rather than an artefact of the run.*

## Root cause

Isolation Forest is a global isolation method with axis-aligned splits. To catch the case above it
would need a tree path conjoining three one-hot features **and** the metric
(`country=DE AND event_type=spin AND product=casino AND event_count < 1000`), discovered by random
splits across ~23 features, then surviving averaging over the ensemble. The information is present;
the model class cannot exploit it.

Compounding it, `RobustScaler` is fit **globally** (`core.py:128`), so a value five deviations low
*within its group* is centred and scaled into an ordinary number before the model sees it.

## Proposal

Items 5 and 6 are self-contained defects and could land independently of any decision about the API;
items 1-4 are the part that needs design agreement.

### 1. `group_by` as the semantic parameter, with a strategy

`segment_by` currently conflates *"these columns identify a peer group"* with *"train one model per
group"*. Users must choose an implementation when all they know is their data's shape.

```python
train(df, ..., group_by=["country", "event_type"],
      params=AnomalyParams(group_strategy="auto"))   # auto | per_group | relative
```

- `per_group` — today's behaviour, one model per group.
- `relative` — one pooled model, group-relative features (below).
- `auto` — chosen from the data.

**Both mechanisms survive; only the parameter is replaced.**

| you write | what you get | models |
|---|---|---|
| `group_strategy="per_group"` | one model per group — today's `segment_by` behaviour, unchanged | N x ensemble |
| `group_strategy="relative"` | one pooled model, group-relative features | 1 x ensemble |
| `group_strategy="auto"` (default) | chosen from the data | either |
| `segment_by=[...]` (legacy) | maps to `group_by` + `per_group` — identical semantics | N x ensemble |

`segment_by` the *parameter* is deprecated; the *capability* it selects is not removed. Passing both errors.

Per-group models remain first-class because they are genuinely more expressive — they learn a different
distribution *shape* per group, not just a different scale — and because the evidence does not favour one
mechanism universally (item 3). Had relative features won everywhere, the right move would be to switch
unilaterally and drop the strategy parameter entirely. They did not, which is what makes the strategy
explicit rather than decorative.

Under `auto`, low cardinality with enough rows per group selects `per_group` — so **nothing changes for
the modest-cardinality case**. The new path only engages above the bounds, where today's behaviour is a
job that runs for over an hour without finishing.

**Introduce first, deprecate one release later.** Asking users to migrate to something with no track
record is the wrong order.

### 2. Group-relative features

For each numeric metric, add the deviation from that group's own baseline alongside the raw value:

```
signed_log(x) = signum(x) * log1p(abs(x))
rel_<metric>  = signed_log(value) - signed_log(group_training_median(metric))
```

log-ratio rather than raw ratio: stable when a baseline is near zero, and symmetric for halving vs
doubling. **Signed** log rather than plain `log1p`: `log1p(x)` is NaN for `x <= -1`, so a plain `log1p`
difference is correct for counts and shares but breaks on any signed metric of magnitude >= 1 (profit,
balance, delta). `signum(x) * log1p(abs(x))` is monotone over all reals and identical to `log1p` for
`x >= 0`. Keep the raw metric too — one extra column, and globally-absurd values stay detectable.

This fits existing machinery with **no registry schema change and no migration**:

- `SparkFeatureMetadata` already persists `categorical_frequency_maps: dict[str, dict[str, float]]`.
  A group-baseline map is the same shape, and `features.feature_metadata` is a JSON string column, so
  it needs two lines in `to_json`/`from_json`.
- `_apply_frequency_encoding` is the template: compute when `is_training`, persist the map,
  broadcast-join at scoring, `coalesce` the miss. `_apply_group_relative_encoding` sits beside it.
- Models without the key skip relative features, so no forced retrain.

Division of labour that makes it learnable: the ratio removes **scale**, the existing cyclical date
features handle **seasonality**, and what remains is the anomaly. A flat per-group baseline is
therefore sufficient — no need for per-group-per-weekday baselines.

SHAP confirms the mechanism rather than a coincidence: `rel_event_count` was the **#1** driver for the
zeroed group (14.6%) and #3 for the -80% group (14.1%).

### 3. Choose the strategy on heterogeneity, not group count

On SMD, relative features were **statistically indistinguishable** from plain pooling: aggregate
PR-AUC 0.1536 vs 0.1499, a 2.5% relative difference on a single seed, winning 6 of 8 entities (p~0.14).

I checked whether SMD simply lacked the heterogeneity that relative features correct for. **It does
not.** Features are globally scaled to [0,1], but cross-entity heterogeneity remains: the median ratio
of between-entity mean spread to within-entity spread is **1.11**, 19 of 35 features exceed 1.0, and
the worst reach **5.3**.

So the honest position is: relative features win decisively on synthetic data (45.1 -> 95.5) and show
**no measurable gain on real heterogeneous data**. I do not know why. Candidate explanations, none
tested:

- The synthetic anomaly was *constructed* to be purely group-relative, so relative features were
  tailor-made for it. SMD anomalies are real and multi-causal, so much of the signal is detectable
  without conditioning and the ceiling on what conditioning can add is far lower.
- Heterogeneity in *means* is not the same as heterogeneity in *scale*. The measurement above is a
  mean-spread ratio; the log-ratio feature corrects scale.
- SMD anomalies are contiguous temporal segments while DQX scores rows independently, which handicaps
  every configuration equally and compresses the differences between them.

What survives -- and is better grounded for having been measured -- is that the strategy should be
chosen from a **heterogeneity statistic**, not cardinality. Between-group variance over within-group
variance is cheap to compute during profiling and demonstrably discriminating: 1.11 on SMD against the
extreme spread deliberately built into the synthetic set. Cardinality remains a hard bound on
`per_group`, not the strategy signal.

Suggested bounds for `per_group`, reusing numbers already in the codebase: `group_count <= 20` (my
timings put 50 groups at ~37 min, too slow for a default) **and** `avg_rows_per_group >= 100` (already
the profiler's own viability warning). Today there are two inconsistent thresholds -- the profiler warns
above 50, training above 100 -- which this would unify.

### 4. Per-group score calibration

Per-group models get per-group `score_quantiles`; a pooled model calibrates against the global
distribution. The literature is explicit that raw anomaly scores are not comparable across entities and
should be normalised per entity before thresholding.

This likely explains the one weakness in the prototype: **95.5 vs 97.0**. Both clear the default
threshold, but 98 is documented as "strict", and at 98 the relative config misses while per-group
catches. Per-group quantiles therefore belong in the `features.feature_metadata` JSON string, following
the same pattern as `group_baselines`, with `training.score_quantiles` kept as the global fallback. Note
that `training.score_quantiles` is itself a typed `map<string,double>` column (`model_registry.py:30`), so
making *it* per-group would require a registry migration — which is why the JSON blob is the right home.

### 5. Cap the segment count

`training_service.py:116` should error above a ceiling with an explicit override, not warn and proceed.
A log line is the wrong response to a job that will attempt tens of thousands of model registrations.

### 6. Surface unseen groups deterministically

**Current behaviour is not "we skip it" — it is "we score it 0.0, the most normal value in the table."**
One-hot encoding emits all zeros for an unseen category, and because each one-hot column is 0 for the
majority of training rows, an all-zeros row looks like the majority on *every* axis; axis-aligned splits
cannot see the joint violation that no category is set at all. Frequency encoding is broken too, with the
opposite sign: `coalesce(freq, 0.0)` (`transformers.py:601`) gives an unseen category a frequency of zero,
which sits *below every value seen in training*, pushing the row toward a fabricated extreme rather than
toward the mean. So one-hot yields false negatives and frequency encoding false positives — and since
frequency encoding is the branch taken above the cardinality threshold, it is the one that fires on the
90-group case. Neither gives the caller a deterministic signal. Separately, `_estimate_feature_count:298`
already budgets a `+1 for MISSING category` that the encoder never creates.

**A model cannot be taught this.** An `is_new_group` indicator feature was built and tested: it
contributed **0.0% in every row**. No training row carries the value 1, so the feature is constant during
training, has zero variance, and the forest never learns a split on that axis. An unsupervised model on a
fixed feature space cannot flag novelty in a dimension it never observed varying.

Proposed, in three parts:

1. **Return NULL, not 0.0.** DQX already has a convention for "not scored": `create_null_anomaly_result`
   sets `severity_percentile` to NULL, and the `row_filter` path documents that unscored rows are "left in
   the output with null anomaly result". An unseen group belongs in exactly that state — *cannot judge*,
   rather than *fine*. This reuses machinery that exists rather than inventing anything.
2. **Detect by lookup, not inference.** The known category set is already persisted in
   `onehot_categories` (and would be in `group_baselines`). At scoring time this is a set-membership test.
   No model involved — which is the point, given the paragraph above.
3. **Surface it** in `_dq_info` as an explicit field naming which columns held unseen values.

**Whether that constitutes a violation should be the user's call, and the default should be no.** New
categories are routine for many users — a market launches, a SKU is added, a device is provisioned — so
auto-flagging every one would be noisy and wrong. But it must be *visible*, so a user for whom novelty is
the signal can act on it with a cheap presence check.

That case is common: anyone whose schema evolves through service migrations. A renamed event id arrives as
a novel category and is today scored as more reassuring than any real row in the table. For such users this
is likely a higher-value fix than the conditioning work above, because a missed volume drop is at least
*detectable* with the right configuration, whereas this is silent.

## API impact

| | change |
|---|---|
| `train()` | add `group_by`; `segment_by` deprecated later |
| `AnomalyParams` | add `group_by`, `group_strategy`; it also duplicates `segment_by` today |
| profiler | `recommended_segments` feeds `group_by`; add the heterogeneity statistic |
| registry | strategy + baselines into `feature_metadata` JSON — **no schema change** |
| **`has_no_row_anomalies`** | **none** — strategy is read from the model record |
| **checks YAML** | **none** |

Scoring and every user's check definitions are untouched, because segmentation is already inferred
from the trained model. The whole change is confined to training.

## Alternatives considered

- **Make `segment_by` auto-discover even when `columns` is passed.** Fixes detection, but makes the
  non-scaling path the default. Right answer, wrong architecture.
- **Silently reinterpret `segment_by` with auto strategy.** Rejected: severity semantics differ between
  the two strategies, so an existing user could see severities shift on retrain with no code change.
- **An unseen-category indicator feature.** Tested; contributes 0.0%. See item 6.
- **Role-typed API (`metrics=`, `groups=`, `time_column=`).** Reads well for aggregate tables but is
  wrong for DQX's original raw row-level case, where everything is a feature and there is no
  metric/dimension distinction. `columns` should stay a flat list with `group_by` as an optional refinement.

## What this proposal does not claim

- Relative features are **not** demonstrated to be universally better. They win decisively on the
  synthetic case and show no measurable gain on SMD, **and that divergence is unexplained** (see item 3
  -- SMD was verified to be heterogeneous, so "no heterogeneity to correct" is not the reason). The
  claim this proposal rests on is narrower: the *choice* currently cannot be expressed, and the only
  mechanism offered is unscalable and calibration-fragile.
- Absolute detection quality on SMD (PR-AUC ~0.15) is **not** comparable to published SOTA (>0.80).
  DQX scores rows independently; SOTA uses windowed sequence models. Only the A/B/C comparison is valid.
- Evidence is one synthetic harness plus one public benchmark at 8 entities, single seed.

## Reproducing

Harness, notebooks and analysis scripts are available, including the injected-incident generator
(one group -80% with the daily total held flat), the three training configurations, and the SMD
benchmark with PR-AUC and per-entity breakdowns. Evaluation deliberately avoids point-adjusted F1
(Kim et al., AAAI 2022 — random scores beat SOTA under it); PR-AUC is primary, with unadjusted
point-wise metrics and per-entity reporting.

## Not part of this proposal

Four independent documentation/small fixes, better as a separate PR:

- `drift_threshold` docstring says "default 3.0"; the signature is `None` (`check_funcs.py:167`).
- Reference docs describe `drift_threshold` as detecting drift in the *score* distribution; the code
  compares *input feature* distributions (`quality_checks.mdx:1960, 3619`).
- Docs imply sparse segments are commonly skipped; the real threshold is `row_count < 10`.
- `TemporalAnomalyConfig` is exported from `config.py` but referenced nowhere in `anomaly/`, and
  `temporal_config` on the model record is never assigned.

### LINKED ISSUE #1490: Anomaly detection: no detector for metrics that move together
## Problem

Row anomaly detection has one algorithm: Isolation Forest. It works by splitting on one feature at a time,
so it is good at finding rows where some value is unusual on its own, or where a combination of values is
rare in a way that repeated single feature splits can isolate.

That leaves a whole class of anomaly undetected. In machine telemetry, fleet metrics, service metrics and
sensor data, the metrics move together. Coolant flow tracks load, bearing temperature tracks speed, request
latency tracks queue depth. When something breaks, the individual readings often stay inside their normal
ranges and the *relationship* between them is what changes. There is no single feature to split on, because
no single feature is out of range.

Today a user in that situation gets a confident low severity, which reads as "this row is fine". The only
workaround is to hand engineer ratio columns for every pair of metrics they suspect, which needs them to
already know what to look for.

### Evidence

Server Machine Dataset (28 machines, 38 metrics per machine, real telemetry with labelled incidents),
scored inside an alert budget of 1% of rows and measured as the share of labelled incidents surfaced:

| detector | incidents surfaced |
|---|---|
| Isolation Forest | 33% |
| Mahalanobis distance | 79% |

Same features, same budget, same pipeline. Only the estimator changed. Isolation Forest was the weakest
estimator of everything measured on this data.

The figures above are measured with anomalies still present in the training sample, which is what DQX
actually does when it fits a sample of a user's table. On a curated clean training split the same comparison
is 36% vs 82%, so the gap is not an artifact of contamination.

## Proposal

Add a `profile` argument to `train()` that selects the detector, defaulting to today's behaviour:

```python
anomaly_engine.train(df, ..., profile="timeseries")   # "tabular" is the default
```

| `profile` | detector | finds |
|---|---|---|
| `"tabular"` (default) | Isolation Forest | rows whose values, or combination of values, are unusual |
| `"timeseries"` | Mahalanobis distance | metrics that normally move together and stopped |

What this adds:

1. **Correlation aware detection with no feature engineering.** The user passes their metrics. The detector
   learns how they normally relate and flags rows that break the pattern, including rows where every single
   reading is inside its healthy range.
2. **Attribution that says which relationship broke**, not just which column looked odd. The leave one out
   contribution has a closed form for this detector, so feature contributions come for free and are always
   non negative.
3. **Deterministic scoring.** One model instead of an ensemble, which also makes it cheaper to train.
4. **No new dependency.** The implementation is numpy only.

The default stays `"tabular"`, so existing users see no change in behaviour or score.

### Why the choice should be explicit and not automatic

An `"auto"` mode is tempting and should not be added. The right detector cannot be verified without labels,
which users do not have, and silently changing the estimator would move every score a user has already
calibrated thresholds against. Automatic detection was also tested: lag 1 autocorrelation does not separate
temporal from tabular data, because the statistic is confounded by any row ordering that correlates with the
values, which sorted storage produces routinely. Three of ten classical tabular benchmarks scored higher on
it than the weakest genuine time series entity.

### Two extension seams

The work should land as a pluggable training strategy and a pluggable scoring strategy, with the resolved
algorithm recorded in the model registry. That keeps the default path untouched and makes a third detector an
addition rather than a rewrite.

## Out of scope

This detector judges each row against the learned joint distribution. It is contextual, not sequential, so
it does not cover:

- trend, meaning slow drift where the metrics keep their relationships
- forecasting, meaning predicting the next value
- cycles that are not calendar based
- a single metric with no timestamp and no grouping column, which reduces to a plain range check

Trend handling is a separate piece of work and should get its own issue.

vb-dbrks:
# Row anomaly detection: a second detector, and the end of `segment_by`

Closes #1484 and #1490.

## TL;DR

Row anomaly detection gains **three orthogonal choices**, and loses 2 parameters:

| | what it answers | default |
|---|---|---|
| `profile` | *how a row is judged unusual* .. `"tabular"` (Isolation Forest) or `"correlation"` (Mahalanobis) | `"tabular"`, unchanged |
| `baseline_by` | *normal compared to what group* .. each row against its own group's baseline | off |
| `baseline_over_time` | *normal compared to when* .. each metric against its own history at that time | off |

- All three are opt-in. With none set, the default path is unchanged, and the isolation-forest inertness suite pins that.
- **`segment_by` and `AnomalyParams.max_segment_models` are removed.** `baseline_by` replaces them but is not a rename: `segment_by` trained one model per group, `baseline_by` conditions one pooled model. Scores differ.
- **Every existing model must be retrained.** The config hash now includes the comparison bases, so a stale model raises at scoring time rather than scoring against a different feature list.
- Registry `segmentation` → `grouping`, and a pre-release registry table is migrated in place on the next retrain.
- **`expected_anomaly_rate` is removed.** It set the estimator's `contamination`, which never changed what DQX flags, so its name promised the opposite of what it did. Nothing replaces it: `threshold` on the check is the alert budget. Default behaviour is unchanged.
- Severity between the persisted quantile knots was silently stricter than requested; fixed, with 90/95/99 byte-identical.
- **Contributions are keyed by the columns you passed**, under either profile, and are averaged across ensemble members. Previously one column's evidence was split across the features derived from it, and an ensemble's explanation came from whichever member trained first.

The feature is **Alpha**, so formats were allowed to change without a migration path. Migration steps are in the user guide.

## `profile="correlation"`

```python
anomaly_engine.train(df, ..., profile="correlation")   # "tabular" is the default
```

Isolation Forest splits on one feature at a time, which makes it strong on tabular data and close to blind when the anomaly *is* a broken relationship between two readings that are both in range .. there is no single feature to split on.

Measured on the full Server Machine Dataset (28 entities, 38 metrics, 708,420 rows, 327 labelled incidents), trained on an earlier period and scored on a later one .. which is what a user does when they train once and check new data.

| | incidents surfaced `@1%` budget | avg precision |
|---|---|---|
| `"tabular"` | 57.5% | 0.277 |
| `"correlation"` | **65.3%** | **0.423** |

Average precision is the larger relative gain, at +53% against +7.8 points of coverage. An incident counts as surfaced if the detector flags at least one of its rows, so coverage answers "would I have been paged".

<details>
<summary>Which protocol, and why the headline changed</summary>

Earlier versions of this description led with **65.4% against 96.8%**, measured by fitting a 30% sample of the very rows being scored. That protocol is legitimate for one question .. can a model re-find anomalies it has already seen, which is what scanning a table you already have amounts to .. but under it **272 of the 327 incidents had rows in training**, 271 shared rows with the held-out set, and 7,027 of the 29,444 anomalous rows were fitted. So it is not evidence for scoring data that arrives later, which is what the quickstart tells you to do. Same code, same data, **31 points of difference from the split alone.**

The table above therefore reports the chronological split, and this is the second correction to the same section. Round 2 flipped the headline from average precision to coverage because average precision was a wash (+0.006, 14W/14L) .. under the leakier protocol. Chronologically it is the stronger signal again, which is where it started. The per-entity statistics that justified the coverage framing (+31.4 points, sd 29.5, 22 wins to 1) belong to the retired protocol and are not carried over, because there are no per-entity chronological figures to replace them with.

Two protocols differing also does not isolate drift as the cause, and an earlier draft said it did.

Both protocols come from one archived run. The comparison is between **estimators on raw metric matrices** .. DQX's own feature engineering is not part of it. A benchmark through the real pipeline needs a workspace and 28 tables, and is tracked separately.

Two earlier sets of figures were withdrawn rather than adjusted: 79%/33%, from a harness truncating each entity to 4,000 rows, and 57.3%/67.6%, which mixed three runs one of which saved no output. Both were found by review.
</details>

No `"auto"`: the choice cannot be verified without labels, and silently swapping estimators would move every score people have set thresholds against.

## `baseline_by`

```python
anomaly_engine.train(df, ..., baseline_by=["region", "product"])
```

The same value can be fine in one group and wrong in another. One group's volume dropping 80% behind a flat daily total scored **45.1** .. the 45th percentile, which no threshold recovers. `baseline_by` adds each metric's deviation from its own group's median, on **one pooled model**, so cost does not grow with group count. The same collapse now scores above 95, and PR-AUC on a contextual anomaly goes from **0.0376 to 0.5703** while an already-globally-extreme anomaly stays at 1.0000.

Rows whose group was absent from training return a **null score with `is_new_baseline = true`**, rather than 0.0 .. which was the most normal-looking value in the table, and the wrong thing to say when the truth is "could not judge this".

## `baseline_over_time`

```python
anomaly_engine.train(df, ..., columns=["revenue"], baseline_over_time="event_ts")
```

A metric that has grown for a year sits outside the range it trained on, so ordinary rows start being flagged; and a value that is ordinary against the whole year can be badly wrong for where the trend had reached. On a fixture where every metric is rolled back together, correlations intact and every value inside the training range, `"tabular"` surfaces **0.0%** of incidents and `"correlation"` 0.8%. With the temporal baseline, **100%**.

DQX fits each metric's expected level as a *function* of time and persists the coefficients, so it extends to timestamps the training window never contained. That is why a time-bucket `baseline_by` cannot do this job.

- **Composes with `baseline_by`.** With both set, the fit runs on the group-relative value. On groups trending at different rates, one pooled fit on the relative value reaches **101% of per-group fits**; fitting raw values collapses from 80% to 29% as slopes diverge.
- **Independent of `profile`**: worth +39% mean event coverage on the tree detector and +40% on the correlation-aware one, harming neither. A third axis, not a third detector.
- **Not a forecaster.** It models the level expected *at* a time, never reads the previous row .. which is what keeps scoring valid on a streaming DataFrame, and why lag features were rejected rather than deferred.
- **Opt-in, and DQX warns rather than guessing.** On stationary data the transform measures *worse* than leaving it off, so a warning fires below 10% of variance explained. The statistic separates regimes sharply: 0.016 on real telemetry against 0.999 on trending data.

Four internal rules are measured rather than chosen, each pinned by a test carrying its number: a seasonal term needs **six complete cycles** (a daily period over 2.8 days cost 15 points of coverage; apparent seasonal strength *rises* as cycle count falls, so variance explained is the wrong gate and is inverted); the fit is **Huber, not least squares** (with 5% of rows at 6x normal, ridge recovered slope 0.0694 against a true 0.0500, Huber 0.0502); **changepoints are chosen on a held-out tail** (in-sample R² was identical from 0 to 25 changepoints while false flags one window out ranged 1.2% to 100%); and **the response is standardised before fitting**, because the design columns were already scaled while the metric was in its own units and Huber solves for coefficients and a scale parameter jointly. That last one is the largest model-quality fix here: across 20 seeds, a metric measured in billions had its slope **99.89% wrong (sd 0.00%) against 0.13%**, winning 20 of 20, and a contaminated sample went from 1.32% to 0.14%, also 20 of 20. Small magnitudes, rates near 1e-6, negatives, integer counts and zero-crossing metrics are a wash, and heavy-tailed noise is a 10/20 coin flip. It surfaced from a `ConvergenceWarning` in a demo's cell output, which is the sort of thing an exit code hides.

## Severity between the quantile knots

**`threshold=98` did not flag 2% of rows, and `threshold=99.9` flagged nothing at all.** Severity was interpolated linearly in *score* space, and a score quantile function is convex in the tail, so a straight line from p95 to p99 sits above the true quantile the whole way.

| `threshold` | promised | flagged, four score distributions |
|---|---|---|
| 98 | 2.0% | 1.41% to 1.56% |
| 99.5 | 0.5% | 0.00% to 0.03% |
| 99.9 | 0.1% | 0.00% |

Now interpolated in tail probability, which is what a percentile is: `severity = 100 - 5 * 5**-u`, `u = (score - q95) / (q99 - q95)`. Same distributions give 1.85% to 2.11% at 98 and 0.43% to 0.59% at 99.5.

Chosen over adding knots for two reasons: it **persists nothing new**, so models already in a registry are corrected by being scored (new keys would have broken every pre-existing model, since `extract_quantile_points` raises on a missing key); and it is **exact at p95 and p99**, so 90/95/99 are fixed points and the default configuration is byte-identical. It also breaks a tie .. the old expression clamped everything past the training maximum to exactly 100, which was 20% of a scored batch on average.

Above roughly p99 the tail assumes exponential decay, which held on four synthetic distributions and real telemetry but **overshoots on a heavy tail** (0.21% against 0.10% nominal at 99.9), so the guide says to read 99.5+ as "much stricter than 99" rather than as a budget.

What this does *not* fix: scoring a later period with a model fitted on an earlier one flagged 9.8% to 21.7% of rows at `threshold=95`, against 4.9% to 5.0% on same-period held-out data. Calibration is sound; the input moved. That is a retraining signal, and no training-relative scheme can fix it without a global pass the streaming contract forbids.

## Why Mahalanobis, and what was rejected

Chosen by measurement. The metric is **event coverage at a fixed alert budget**, not PR-AUC: SMD's anomalous rows sit in incidents of median length 6 and max 1041, so point-wise PR-AUC mostly measures whether the one huge incident was found. No point-adjusted F1 either .. Kim et al. (AAAI 2022) showed random scores reach state of the art under it.

It also gives **exact, never-negative attribution**: leave-one-out contribution has a closed form for this detector, so contributions come free with no SHAP dependency. The naive signed decomposition was rejected because its terms can go negative, and a feature that *reduced* the distance is not a driver of the anomaly .. it would now be dropped rather than misreported, which silently removes evidence from an explanation whose remaining shares no longer describe the distance they came from. Non-negativity has to come from the formula, not from a clip. Being deterministic, it trains one model rather than an ensemble (`ensemble_size` is ignored and `confidence_std` unavailable, both logged).

<details>
<summary>Rejected alternatives, with the numbers so nobody re-argues them</summary>

- **PCA, in any position .. read this one sceptically, because truncated PCA measured *better* than what ships.** At full rank it is a provable no-op (residual 8.6e-13 at ridge zero). Truncated it is neither: `PCA 95% → Mahalanobis` reaches 74.5% coverage against the shipping 65.3%, 15W/5L. Rejected anyway on three grounds independent of that: *k* is unchoosable (same fixture swings ROC-AUC 0.48 to 1.00 on *k* alone, and retained variance is *anti-correlated* with what the data wants); it collapses on the very anomaly class the second detector exists for (recall 81.6% → 0.8% on a correlation break); and it destroys per-feature attribution.
- **Robust (MAD) covariance.** Implemented, then reverted: it excluded 4.6% of legitimate tail rows and took coverage 91.9% → 86.4%, ROC-AUC 0.784 → 0.763. The distortion statistic meant to gate it is *inverted*. The contamination exposure is real and documented at the top of the detector.
- **Lag / rolling-window features.** Raw beat windowed on coverage; an earlier spike measured +9% PR-AUC for 190 extra features with ROC-AUC going backwards.
- **ECOD / COPOD as a third profile.** Best ADBench average precision measured (0.605) and 26 points of coverage behind the *default* on SMD, 1W/9L. Both sum per-dimension tail probability, so neither can see a correlation break.
- **Robust scaling.** Provably inert: Isolation Forest's partition is identical under any monotone affine map, and Mahalanobis is affine-invariant.
- **Automatic profile detection.** Lag-1 autocorrelation does not separate temporal from tabular data .. it is confounded by any row ordering that correlates with values, which sorted storage produces routinely.
- **Per-group models** (the removed `segment_by`). Worst of three configs on SMD (PR-AUC 0.1416 against 0.1536 conditioned), and failing in a way the average hides: one entity produced a 56% false alarm rate because each per-group model calibrates on its own rows. Cost is linear in groups .. 90 groups took ~88 minutes against 0.17s for one conditioned model.
</details>

## Explaining a flagged row

Detection and explanation were reviewed separately, and the explanation is where the defects were. Each
of these could name the wrong column, confidently, in business language, on a row the detector had
flagged correctly.

**One column, one entry.** Feature engineering can derive three features from one numeric column .. the
metric, its deviation from its group's baseline, its deviation from its expected level at that time ..
and contributions were reported per feature, which answers a question nobody asked. The correlation-aware
detector's per-view drops are each almost nothing, because dropping one view leaves another copy of the
same information: adding one affine duplicate of a metric moved the reported cause from **99.8% that
metric to 99.9% an unrelated one** while the score did not move. Blocking by source restores 99.7%/0.3%.
For a tree model the failure is different in shape .. the evidence is *split* rather than misdirected,
49.3% and 50.7% across two views of a column whose true share is 100% .. and still changes the answer,
because with three views a column whose true share was 74.5% left each view near 25%, handing the top
spot to a derived view of itself with an unrelated single-view column tied alongside. Both detectors now
group by source column, by an exact joint marginalisation for the one and a plain sum for the other,
which is exact because SHAP is additive.

**Provenance is read from the recorded transform, not from spelling.** A caller's own column named
`amount_rel_baseline` was decomposed into `amount` and labelled "amount vs its group baseline" on a model
with no grouping at all .. crediting one column's contribution to another, and making redaction of
`amount` sweep up an unrelated column. Suffixes now resolve only when the transform that generates them
actually ran, and an exact source-name match wins over decomposition.

**The SHAP orientation was wrong in kind, and the review's reading of it was backwards.** TreeSHAP on an
isolation forest explains the ensemble's average *path length*, not a score, so a **negative** value
shortens the path and drives the anomaly. Review reported this as `abs()` turning normalising evidence
into drivers, which implies the positive side is the anomaly-driving one; it is the opposite. Measured on
200 rows each carrying exactly one deliberately anomalous feature, the largest value named the true
culprit **199/200** under the orientation that ships, 199/200 using magnitude alone, and **0/200** using
the positive side .. so implementing the finding as described would have destroyed attribution accuracy.
Magnitude alone was very nearly right, since on rows anomalous enough to be shown the evidence is 97.8%
anomaly-driving at a severity of 95 and 99.5% at 99. It was still wrong in kind, adding evidence of
normality to evidence of anomaly, and correcting it left the top driver unchanged on every row measured
while reordering the ranks below it on roughly a quarter of rows. Not a schema change: the map stays
non-negative and sums to 100.

**A row with no anomaly-driving evidence said every feature contributed equally.** When the attribution
totalled zero the map was filled with `1 / n` .. an explanation with no input behind it, and the more
misleading of the two failures because the numbers look unremarkable. Such rows now carry the all-null
map that a row with a null feature already produced, which every consumer already renders as `unknown`.
Zero-valued entries are also dropped from messages, prompts and the pattern key, rather than being
listed as `quantity (0%)`.

**An ensemble scored with every member and explained with one.** The reported score is the mean across
members and so is `confidence_std`, but contributions came from `models[0]` .. arbitrary, since members
differ only by random seed. They disagree more than that framing suggests: on 200 rows with the default
three members, the least-agreeing pair named a different top driver on **130** of them and the closest
pair still differed on 92. Contributions are now the mean across members, which decomposes the mean path
length exactly, and the entry point takes the models behind the score rather than a model, so the mistake
cannot be remade by omission. It costs about **7% more** at three members, because attribution runs only
on rows above the threshold. Averaging makes the explanation unbiased between members, not stable .. with
three members it stays noisy, and the fix for that is more members.

**The prompt asserted a direction the inputs do not contain.** Contributions are unsigned magnitudes, so
a row displaced one way and a row displaced equally the other produce the identical score and the
identical map .. `(8, 0.5)` and `(-8, -0.5)` both score 65.387. The few-shot exemplars nonetheless said
"sits far above the norm" and "Inflated amount fields overstate revenue", and a smaller serving model
copies the shape of its examples. Both are rewritten so every clause is checkable against that
exemplar's own inputs; the instructions name the absence of direction explicitly; and "be direct, avoid
hedging" is reconciled rather than left to interpretation. `confidence` is relabelled as agreement
between seeds rather than confidence in the finding, and the header says the model is explaining what it
measured rather than diagnosing a root cause.

**The model was told a metric mattered but not what it was compared against.** Keying contributions by
source column closed a channel nobody had noticed was load-bearing: the map used to be keyed by engineered
feature, and the prompt labelled each key, so `revenue_rel_time` arrived as "revenue vs its expected level
at that time". With the key now `revenue`, that comparison left the prompt entirely. Grouping was never
affected, because `baseline_grouping` has always been its own field .. temporal conditioning had no such
field and only ever arrived through the label, so the two axes were asymmetric well before this and source
keying is what made it bite. It matters because a metric judged against its expected level at a moment can
sit inside every range the table has ever held and still be wrong for when it arrived, and a model given a
large share and no comparison reaches for "unusually high", which is the one claim the evidence cannot
support. `temporal_baseline` is now a first-class input beside `baseline_grouping`, both fields describe
what they *license* rather than only what they are, the instructions tie the claim to the comparison, and
both exemplars carry the field in different states.

<details>
<summary>What it actually generates, run against a workspace</summary>

A metric on a steady upward trend, scored on a later window where the value is ordinary for the table
(inside the training range of roughly 100 to 1,090) and wrong for its point in time:

```
orders=420.0  at 2026-05-06 01:00:00  severity=99.5
  contributions : {'orders': 52.1, 'revenue': 47.9}
  narrative     : 5 rows show orders (52%) and revenue (48%) both departing from the levels
                  expected at that point in time.
  impact        : Orders and revenue that depart from expected temporal patterns may distort
                  forecasts and period-over-period comparisons.
  action        : Compare orders and revenue against their expected levels for that time period.
```

Keys are the caller's columns, not engineered features. The shares total 100. The temporal reading is used,
and it is only derivable from the new field .. the map holds `orders` and `revenue`, and neither
`attribution_basis` nor `drift_summary` mentions time. The value is a *drop* (420 against an expected
1,090) and the narrative does not say so, because direction is not in the inputs. The action names the
right comparison rather than the metric's overall range. One explanation covers the group, and its "5 rows"
matches `group_size`.

</details>

**Novelty is not guaranteed, on either profile, and the bound is the encoding.** Retaining every
one-hot category puts an unseen value's distinction *in* the features, but acting on it needs a model that
reads them jointly: the correlation-aware detector scores an unseen category at 2,000,000 against 0.9 for a
known one, while Isolation Forest scores it 0.458 against 0.442 and 0.467, i.e. ordinary. Round 4 then
showed the remaining claim was still too broad, because it only holds while the column *is* one-hot .. at or
below the cardinality threshold of 20. Above it the column becomes one frequency number, and with
near-uniform training frequencies that coordinate is constant, the detector excludes it as such, and a known
and an unseen value both score **0.1837**. So neither profile substitutes for a membership check, which is
what the guide now says.

The same probe strengthens the case against pruning correlated features: dropping the anticorrelated dummy
collapses the one-hot separation to 1.1 against 0.9 while leaving ordinary rows untouched, so the redundancy
*is* the off-support constraint and no in-sample metric would notice its removal.

**Two explanation claims were corrected in round 4.** The prompt told the model that contributions spread
across several metrics meant those metrics had stopped agreeing with each other; two metrics with training
correlation 6e-18, each marginally extreme, produce 50%/50%, so the shape does not carry the mechanism. And
both the prompt and the guide treated the configured comparisons as the reason a metric contributed, when
`baseline_grouping` and `temporal_baseline` establish only which comparisons were *available* .. a metric
whose temporal fit was rejected carries a constant zero for its time-relative feature, so a time column does
not even establish the comparison ran for it. The exemplar mattered more than the prose here: its input line
demonstrated a label form the pipeline can no longer produce, and demonstrated exactly the pinning the
instruction was being corrected for.

## Limitations documented rather than fixed

- **`baseline_by` conditions on level, not on relationship.** One pooled model learns one set of
  relationships. On a counterexample where group A has `y` rising with `x` and group B has it falling,
  both centred identically, the pooled model scores a row that is impossible for B at **1.3** where a
  model fitted on B alone scores it **260** .. and since the medians match, group-relative features are
  identical, so the transform changes nothing. Groups that behave differently rather than sitting at
  different levels want one model each. `baseline_over_time` has the same limit: it removes one shared
  trend.
- **The temporal changepoint criterion is blind to constant forecast bias, and the blind spot is
  bounded rather than harmless.** An earlier version of this description claimed such a bias "cannot
  matter", on the strength of a test that shifted training and scoring data together and refitted. That
  measures translation invariance *of refitting* .. true, and relevant, because selection compares
  candidates all refitted on the same data and a shared level error cancels from the ratio they are judged
  on: 1.7e-14 relative on the correlation-aware detector, bit-identical on Isolation Forest. It says
  nothing about a bias that appears only *after* the fit window. With the detector fixed and +5 added to one
  held-out residual dimension, the residual MAD is unchanged at 1.0026 .. invisible to the criterion .. while
  rows above the training p99 go from **1.3% to 98.4%**. Drift detection and the staleness horizon are the
  mechanisms for that; this criterion is not one of them. Round 4 caught the overclaim.

## Breaking changes

- **`profile="timeseries"` is now `profile="correlation"`**, renamed before release with no alias. The old spelling named a data shape and a time dependence the detector never had: it needs no timestamp column, does not require time-ordered rows, and does not forecast .. it models how a row's metrics relate to each other. Nothing depends on the old string, so this costs nothing to change now and would cost a deprecation cycle later: the profile is resolved at training time and never persisted, and scoring resolves the detector from the algorithm a model records. Passing the old value raises and the message names the replacement, rather than silently falling back to the default and handing the caller a different detector than they asked for. Only relevant to anyone who tried the value on this branch.

- **`segment_by` removed**, along with `AnomalyParams.max_segment_models`.
- **Registry `segmentation` → `grouping`.** A pre-release table is migrated in place on the next retrain and the old column is *gone* afterwards, rather than both being carried forever by a `mergeSchema` append. A permission failure is translated into an error naming the grant needed.
- **Pre-release models cannot be scored at all**, because the config hash formula changed .. retraining is the only path, and the error says so.
- `_dq_info[].anomaly.segment` removed; `is_new_baseline`, `new_baseline_key`, `is_stale_baseline` and `stale_baseline_horizon` added. The struct is wider, so appending to an existing `_dq_info` table needs `mergeSchema`.
- Feature engineering now **refuses a schema where a generated feature would overwrite a real column** (`amount` alongside `amount_rel_baseline`), instead of silently replacing it and fitting on two copies of the derived value.
- **A model that cannot attribute by source column is now refused rather than explained wrongly.**
  Attribution keys are source columns, and an estimator that supplies only per-engineered-feature values
  would have had them reported under those keys .. width and names disagreeing, normalisation running over
  every column while only the leading ones are emitted. Observed output was `{'x': 0.0, 'y': 0.0}` on a row
  whose score matched a correctly explained one. Scoring such a model now raises, naming retraining or
  `enable_contributions=False`. Reachable without exotic setup: the config hash covers only columns and the
  comparison bases, so it is not version-sensitive, and a custom estimator arrives through a documented
  extension point.
- **`_dq_info[].anomaly.contributions` is keyed by source column** under both profiles, where `"tabular"` previously keyed by engineered feature. A query reading `signup_hour_sin` now reads `signup`. The map stays `map<string,double>`, non-negative and summing to 100, so only the key vocabulary changes .. and a column that argued the row was normal now shows `0` rather than a small positive share. A flagged row with no anomaly-driving evidence carries an all-null map instead of an invented even split.
- **`expected_anomaly_rate` removed from `train()`.** It supplied the estimator's `contamination`, which places only scikit-learn's own `predict`/`offset_` boundary; DQX ranks `score_samples` against training quantiles, so the parameter never changed which rows were flagged. Documenting around a name that promises a detection rate would have preserved the trap, so it is gone. `contamination` now defaults to 0.02 directly, which is exactly what the parameter filled in, so a call that never passed it behaves identically. To change how much is flagged, set `threshold` on the check; to set `contamination` because you load the registered model and call `predict` yourself, pass it through `params.algorithm_config`.
- **The `mlflow` floor moves to `>=3.13,<4`**, to pick up a Databricks unified-auth fix that exists in no 2.x release. One consequence needed handling: MLflow 3 validates saved sklearn models against skops' trusted types and refuses `MahalanobisDetector`, DQX's own class, so the correlation profile failed at registration while every unit test passed. `serialization_format="cloudpickle"` is now named explicitly, which is what MLflow 2 did by default, so this restores existing behaviour rather than introducing new. The floor also makes two pieces of MLflow 2.x scaffolding dead .. the `artifact_path=` fallback in `log_sklearn_model_compatible`, unreachable now that every admitted version accepts `name=`, and a dummy-credentials test fixture that existed only to work around the bug the floor picks up. Both are left in place and tracked in #1508: deleting the first changes no behaviour, and the second cannot be verified from CI at all, since the branch only runs when `DATABRICKS_CONFIG_FILE` is unset and CI always supplies real credentials.

## Docs, demos, tests

- **The feature is tagged Alpha, and `alpha` is a new lifecycle stage.** DQX's taxonomy went `experimental` → `beta` → GA, with no stage for "works, actively developed, not yet feature-complete" .. which is what this is. So `alpha` is added to `FeatureLifecycleStage`, given a badge colour that sits between Experimental's amber and Beta's brand red, and documented on the [Feature lifecycle](docs/dqx/docs/reference/feature_lifecycle.mdx) reference between the two. The authoring guide also advertised a `stage="ga"` that never existed .. GA features carry no badge at all .. so that is corrected while the list is being touched.
- **Version props are settled against the released tag, not by preference.** `v0.16.0`'s `anomaly_engine.py` contains no `profile`, `baseline_by` or `baseline_over_time`, so all three sections carry `0.17.0`; the page-level tag stays `0.14.0`, because row anomaly detection itself did ship then. The `baseline_by` section had no tag at all and now has one.
- **A documentation review pass** on top of the rename, from a black-box audit of the guide. The corrections a reader could have acted on and been wrong: the PII answer claimed there are no external API calls while AI explanations are on by default and call a serving endpoint; the reference told users to *increase* `categorical_cardinality_threshold` to get frequency encoding, where the code is `onehot if cardinality <= threshold`, so raising it does the opposite; and two passages disagreed about combining grouping with a temporal baseline, where one coefficient vector per metric is fitted, so groups at different *levels* share a model but different *slopes* do not. Also removed: segmented-model semantics left behind by `segment_by`, including a warning that segment values reach the prompt verbatim, which no longer happens. `AI explanations` is promoted out of "How it works under the hood" to a top-level section, since it is on by default and its cost and privacy are user decisions rather than internals.
- Guide gains "Choosing a profile", framed as *how does DQX decide a row is unusual* .. a question answerable without knowing ML, and one that does not imply either detector needs a particular shape of table. Threshold is documented as an **alert budget**, not a confidence score. Limitations become one FAQ entry, fixing two claims that were wrong (DQX *does* model seasonality; `drift_threshold` does *not* warn about trend).
- Quality is published in the **existing benchmarks report** next to core timings, fed by two new perf tests whose fixtures are generated in-repo from a fixed seed .. nothing downloaded, no third-party data redistributed.
- `dqx_row_anomaly_detection_demo.py` is replaced by two problem-led notebooks (card transactions that pass every rule but are jointly implausible; machine telemetry where every metric stays in band while the relationship breaks), both covered by a parametrised e2e test. Both print a **threshold sweep** with precision beside the ceiling that alert count allows, rather than a single detection rate, and both train on the whole table because `DataFrame.sample` draws per partition and is not reproducible across partition counts.
- New: isolation-forest inertness (16 tests, sabotage-checked), the Mahalanobis contract and its leave-one-out identity, feature-name collisions, temporal permutation invariance, one-hot determinism, registry migration from a **verbatim** pre-release schema, and an MLflow round-trip for a DQX-defined estimator class.
- New for attribution, each pinning a property rather than a wording: ground-truth orientation on rows with one known culprit (including the rejected alternative, asserted as wrong so nobody reaches for it again); the invented-uniform and all-null paths; the formatter's row cursor, whose off-by-one would misattribute every row after the first unexplained one; ensemble order-invariance, asserted at a tolerance because float addition is not associative; misaligned members raising rather than averaging; and source blocks recovering the undisturbed model's answer on both detectors.

Unit 2777 passing, mypy clean, pylint 10.00/10, `make fmt` and `make build` clean, `make docs-build` clean. AI explanations were additionally read off a live workspace, not only asserted .. see the collapsed output above.

## Still open

- A **supported calibration helper**. Tuning on one half of a batch and applying to the other cut median error against the requested budget from 3.1x to 1.7x, but only 20 to 30% of machines landed within double their budget: a large improvement, not a guarantee, so it wants its own PR. Both demos print the sweep by hand meanwhile.
- **Training is not reproducible run to run** at the default `sample_fraction`, because the sample is drawn per partition. A deterministic sample is a small change to `sample_df` but moves every existing user's model.
- **Detrending as a feature transform** (measured flat 2.8% false flags against raw's 19.6% to 99.2%) and a **trend-aware drift statistic**. Both are new capabilities rather than fixes.
- **Signed attributions and an explained-target field** in the public map, so a reader can tell a feature that drove the anomaly from one that argued against it. Deliberately not done here: it would reintroduce exactly the directional reading the prompt work just removed, and it wants its own acceptance tests.
- **A vocabulary check outside the learned ranking**, so an unrecognised category is caught under the default profile. `is_new_baseline` is the precedent to follow.
- **Removing the MLflow 2.x scaffolding the new floor made dead** (#1508, `good first issue`). Worth pairing with the workspace run below, since its second item needs one and provisioning twice would be wasteful.
- The anomaly gate (`make anomaly`, against a live workspace) and both demo notebooks need a re-run, because the contributions map's keys changed for the default profile. Unit tests cannot reach the registry or MLflow paths, so that gate is the signal that matters and this stays a draft until it is green.


vb-dbrks:
[code-walkthrough.html](https://github.com/user-attachments/files/31967447/code-walkthrough.html)
@mwojtyczka - might be useful for your review

mwojtyczka (commented):
## Automated code review

Reviewed the row anomaly detection changes (correlation detector, baseline grouping, temporal baselines, `segment_by` replacement). The code is carefully written and self-documented; the following four issues are the ones I can concretely justify, ranked by severity. Line numbers are anchored to head commit `dd6e50d`.

---

### 1. Correctness — severity gate can diverge from the published Spark severity
**`src/databricks/labs/dqx/anomaly/explainability.py:415`**

`severity_from_scores` (the numpy SHAP gate) lacks the equal-knot (`span == 0`) handling the Spark side now has. When the training score distribution has equal adjacent quantile knots (e.g. `approxQuantile` returns `q90 == q95` on a concentrated / low-cardinality distribution), Spark's `_linear_severity_chain` explicitly handles `span == 0` and uses the upper percentile, while the numpy path runs `np.interp` over non-increasing `xp`. A row that Spark flags (`is_anomaly = True`, `severity >= threshold`) can then get a different severity from the numpy gate and be classified below threshold — so `anomaly_contributions` comes back null for a flagged row, exactly the gate/flag disagreement the code comments say must not happen. **This is the only finding with runtime impact.**

### 2. Fragility — hidden mutation of `feature_metadata`
**`src/databricks/labs/dqx/anomaly/core.py:362`**

`compute_score_quantiles` / `compute_score_quantiles_ensemble` (via `_quantiles_from_scored`) mutate their `feature_metadata` argument (`baseline_score_quantiles = …`) as a hidden side effect that per-group severity calibration silently depends on. It works today only because the same object is threaded from training through persistence; any future refactor that copies `feature_metadata` between quantile computation and `to_json()` would silently drop per-group calibration and fall back to global severity with no error. Preferred: return the group quantiles and have the caller attach them explicitly (mirroring how `score_quantiles` is already returned).

### 3. Stale / trap dead code
**`src/databricks/labs/dqx/anomaly/explainability.py:550`**

`compute_feature_contributions` / `compute_contributions_for_matrix` have no production caller and remain TreeSHAP-only — they build a `TreeExplainer` directly and were never routed through the new duck-typed `_attribute` dispatch the live scoring path uses. Touched by this PR (the `named_steps.get('scaler')` fix) but left inconsistent: if a future caller wires them into the correlation (Mahalanobis) path, `TreeExplainer(MahalanobisDetector)` fails at runtime because the estimator is not a tree model.

### 4. Unreachable duplicate branch
**`src/databricks/labs/dqx/anomaly/feature_naming.py:168`**

`human_label` has a duplicate `if engineered_name in source_names: return engineered_name` after the same condition already returned earlier in the function — the second check can never execute. Harmless at runtime, but it implies a third resolution path that doesn't exist.

---

_Verified as correct (no action needed): the `stale_col` / `horizon_col` columns survive all four scoring paths; `permissive_quantile_points` is provably monotonic; the legacy `segment_by → baseline_by` registry migration is harmless (such models fail the config-hash check and retrain); and the Python-vs-Spark baseline-key machinery only guards a test-exercised contract._

mwojtyczka (changes requested):

demos/dqx_demo_anomaly_correlation_fleet.py
```diff
@@ -0,0 +1,775 @@
+# Databricks notebook source
```
mwojtyczka, line 1: rename the file to `dqx_row_anomaly_correlation_fleet.py`

demos/dqx_demo_anomaly_tabular_transactions.py
```diff
@@ -0,0 +1,765 @@
+# Databricks notebook source
```
mwojtyczka, line 1: rename the file to "`dqx_row_anomaly_tabular_transactions.py`"

docs/dqx/docs/guide/row_anomaly_detection/index.mdx
```diff
@@ -6,56 +6,23 @@ sidebar_position: 308
...
 # Row Anomaly Detection
 
 <FeatureTags>
+  <FeatureLifecycleStage stage="alpha" heading={false} />
   <AvailableSinceVersion productName="DQX" version="0.14.0" heading={false} />
 </FeatureTags>
 
-Use row anomaly detection to automatically find unusual rows in your data using ML (per‑record anomalies with explanations) without manually specifying thresholds so you can catch issues that rule-based checks miss. You provide recent good data; DQX trains a model and flags rows that don't fit typical patterns. No ML expertise required. Each flagged row includes an explainable breakdown of which columns drove the score, so you can see why it was flagged.
-
-<Deck>
```
mwojtyczka, line 20: so no more bananas?

docs/dqx/docs/guide/row_anomaly_detection/index.mdx
```diff
@@ -126,10 +108,10 @@ DQM and DQX each provide distinct capabilities. Together, they complement one an
 
 Each row is scored and enriched with:
 - **Severity percentile (0–100)**: how unusual the row is compared to training data.
-- **Anomaly flag**: whether it crosses your chosen score threshold (default 95). You can tune this to control how many alerts you get.
-- **Top contributors (explainability)**: which fields most influenced the anomaly score, so you can see *why* a row was flagged. Powered by SHAP, this turns a black-box score into an actionable insight.
+- **Anomaly flag**: whether it crosses your chosen score threshold (default 95). The threshold is a percentile of the severity seen during *training*, so a batch with real problems in it produces more alerts than the threshold's share of rows. See [How to choose a threshold](#how-to-choose-a-threshold).
```
mwojtyczka, line 111: I think this is hard to understand: `so a batch with real problems in it produces more alerts than the threshold's share of rows`

docs/dqx/docs/guide/row_anomaly_detection/index.mdx
```diff
@@ -109,6 +76,21 @@ Because results are **explainable**, you don't just get a list of flagged rows.
 **Use DQX for**: "Is this data unusual?"  
 **Use domain models for**: "Is this data fraudulent, faulty, or malicious?"
 
+### How well does it detect, and how to judge it
+
+Row anomaly detection scores each row **on its own**. It is not a time-series or sequence model and does not read a window of history to make a prediction. The high numbers you see on published anomaly-detection leaderboards are usually won by sequence models on time-series data, which is a different problem, so they are not a yardstick for DQX. The right yardstick is your own data.
```
mwojtyczka, line 81: what is a published leaderboard? are you refering to the benchmark? The whole explanation here is a bit hard to understand starting from `The high numbers you see on published ...`

docs/dqx/docs/guide/row_anomaly_detection/index.mdx
```diff
@@ -283,81 +272,405 @@ display(
...
-4. **Scoring and explanation**: Raw scores are normalized to a 0–100 severity percentile. SHAP contributions (which features drove each score) are on by default and computed only for anomalous rows; set `enable_contributions=False` to skip them for the fastest scoring.
-5. **Auto-discovery of columns and segments**: When you call `train()` without `columns` or `segment_by`, DQX automatically discovers both. It selects numeric columns with enough variance as features and may **auto-segment** when it finds suitable segment columns: categorical or low-cardinality columns with 2–50 distinct values, low null rate, and enough rows per segment (for example region, product category). If your auto-trained model is segmented, that is expected. To force a single global model, pass `segment_by=[]` (or omit segment columns from the data used for discovery) or set `columns` and `segment_by` explicitly.
+Each row above, with the measurement behind it:
 
-### Why Isolation Forest?
+- **Real problems add alerts on top of the budget.** On correlated machine telemetry where 4.0% of rows were
```
mwojtyczka, line 297: i don't get what we mean here by budget. I don't think any avg dqx user would be able to understand these analysis here

docs/dqx/docs/guide/row_anomaly_detection/index.mdx
```diff
@@ -283,81 +272,405 @@ display(
...
+  genuinely faulty, `threshold=95` flagged 8.4% of the batch — 5.2% of the *normal* rows, which is the budget
+  delivered almost exactly, plus most of the faults on top. A batch with nothing wrong with it gives you
+  roughly your budget and no more, provided it resembles the training data.
 
-Isolation Forest measures how "easy" it is to isolate a data point; anomalies are isolated in few splits, normal points need many. It is fast, handles mixed types, is robust to noise, and is explainable via SHAP. DQX uses it by default because it fits data quality use cases without tuning.
+- **A moved distribution can multiply alerts several times over.** Scoring a later time period with a model
+  fitted on an earlier one flagged 9.8%–21.7% of rows depending on the detector, where the same model on
```
mwojtyczka, line 303: where are all these crypting % numbers coming from?

docs/dqx/docs/guide/row_anomaly_detection/index.mdx
```diff
@@ -283,81 +272,405 @@ display(
...
+Pick the cutoff whose alert count matches how many rows someone can actually investigate. Both anomaly demos
+under `demos/` print this table, and one of them chooses its threshold from it rather than inheriting the
+default.
+
+A tuned threshold transfers only partly. Measured across machines in the same dataset, tuning on one half of
+a batch and applying to the other cut the median error against the requested budget from 3.1x to 1.7x and
+the spread across machines from 28 to 7.5 percentage points, but only 20-30% of machines landed within
+double their budget. Treat a tuned number as a much better starting point than a guess, not as a guarantee.
+
+:::note[Very high thresholds are directional, not calibrated]
```
mwojtyczka, line 349: adjsut to use dqx style notes

docs/dqx/docs/demos.mdx
```diff
@@ -49,7 +49,8 @@ Import the following notebooks in the Databricks workspace to try DQX out:
...
 * [DQX Demo Notebook for Profiling and Applying Checks at Scale on Multiple Tables](https://github.com/databrickslabs/dqx/blob/v0.16.0/demos/dqx_multi_table_demo.py) - demonstrates how to use DQX as a library at scale to apply checks on multiple tables.
-* [DQX Demo Notebook for Row Anomaly Detection](https://github.com/databrickslabs/dqx/blob/v0.16.0/demos/dqx_row_anomaly_detection_demo.py) - comprehensive demo showing how to use DQX Row Anomaly Detection to detect unusual patterns in your data.
+* [DQX Demo Notebook for Row Anomaly Detection on Transactions](https://github.com/databrickslabs/dqx/blob/v0.16.0/demos/dqx_demo_anomaly_tabular_transactions.py) - starts from a domain problem: card transactions that pass every rule but are jointly implausible. Shows why no threshold catches them, how `baseline_by` makes "normal" depend on the merchant category, and what the contributions tell you. Uses the default `tabular` profile.
```
mwojtyczka, line 52: I think these are more industry demos. We could move to the industry seciton and maybe have the existing one that we had as a generic one

docs/dqx/docs/guide/row_anomaly_detection/index.mdx
vb-dbrks: I removed a page from the docs, which was a separate one for Anomaly Benchmarks... I'll simplify these statements :) - I think I'll have to make a follow-up PR to add that in later.

tests/perf/generate_md_report.py
```diff
@@ -66,15 +72,36 @@
...
     )
 
-# Optional anomaly benchmark section (extra_info like roc_auc).
-anomaly_benchmarks = [b for b in data["benchmarks"] if b.get("group") == "anomaly_synthetic"]
+# Anomaly benchmark section: timings plus indicative detection quality carried in extra_info.
 if anomaly_benchmarks:
     lines.append("\n## Anomaly Benchmarks\n")
     lines.append(
-        "| Test | Mean (s) | Median (s) | Min (s) | Max (s) | Stddev (s) | Rounds | Ops/s | ROC-AUC | Precision | Recall | F1 | Precision@N |"
+        "* Every fixture is generated in-repo by `tests/integration_anomaly/synthetic_generators.py` "
```
mwojtyczka, line 79: Please don't leak internal implementation details in the docs

tests/perf/generate_md_report.py
```diff
@@ -66,15 +72,36 @@
...
+    )
+    lines.append(
+        "* Each row states the fixture it was measured on. They are deliberately different problems: "
+        "one blends overlapping and heavy-tailed distributions, one breaks the correlation between "
+        "metrics that normally move together (`profile=\"correlation\"`), and one collapses a single "
+        "group's volume while the overall total stays flat (`baseline_by`). Comparing quality numbers "
+        "*across* rows is meaningless."
+    )
+    lines.append(
+        "* Quality columns are **indicative and first-observed only**: the nightly baseline merge keeps "
```
mwojtyczka, line 91: again, remove internal details, users don't care if the benchmark was generated in the nightly

docs/dqx/docs/reference/quality_checks.mdx
```diff
@@ -3474,7 +3474,7 @@ Using non-curated aggregate functions is supported with the following limitation
 
 Detects unusual rows by learning what "normal" looks like across multiple columns, then flags rows that deviate. Train on recent "good" data so the anomaly model learns typical ranges and combinations (e.g., amount and quantity together), not just single-column thresholds.
 
-**Current algorithm**: IsolationForest (scikit-learn with Spark scoring, requires Spark >= 3.4). Future releases may add additional algorithms behind the same interface.
+**Detectors**: two, selected by `profile` (see the parameter table below). `"tabular"` (the default) is IsolationForest (scikit-learn with Spark scoring, requires Spark >= 3.4), which judges a row by its own values. `"correlation"` is a covariance-based detector that judges a row by whether its metrics still relate to each other as they normally do; it needs no timestamp column and trains a single model rather than an ensemble, because it is deterministic. Both share the same feature engineering, registry, check function, contributions and AI explanations.
```
mwojtyczka, line 3477: maybe think about the names: `tabular` does not really convey what it does

docs/dqx/docs/reference/quality_checks.mdx
```diff
@@ -3509,11 +3509,18 @@ run_configs:
   input_config:
     location: catalog.schema.orders
   anomaly_config:
-    columns: [amount, quantity]   # optional; omit to use all supported columns
+    columns: [amount, quantity]        # optional; omit to auto-discover
     model_name: catalog.schema.orders_monitor
     registry_table: catalog.schema.dqx_anomaly_models
+    baseline_by: [region, product]     # optional; judge each metric against its own group's baseline
+    profile: tabular                   # optional; "tabular" (default) or "correlation"
+    baseline_over_time: event_ts       # optional; time column each metric's level is fitted along
```
mwojtyczka, line 3517:
```suggestion
    baseline_over_time: event_ts       # optional, required only for "correlation" profile; time column each metric's level is fitted along
```
we only need it for "correlation" profile, but is it then still optional or required? what if the column is not provided?

docs/dqx/docs/reference/quality_checks.mdx
```diff
@@ -3525,12 +3532,13 @@ Required arguments: `df`, `model_name`, `registry_table`. Optional parameters:
...
+| `baseline_by` | list[str] | None | Columns identifying the group a row belongs to, so each metric is judged against its own group's baseline rather than against the whole table. One model whatever the group count. Auto-discovered when both `columns` and `baseline_by` are omitted; pass `baseline_by=[]` to suppress discovery and compare against the whole table. When you name `columns` but not `baseline_by`, the comparison stays pooled and a warning names the grouping to pass if the data looks grouped. |
+| `baseline_over_time` | str | None | A timestamp or date column each metric is judged *along*, so a value is compared with what its own history says to expect at that point in time. Composes with `baseline_by`: with both set the expectation is fitted on the group-relative value, so one model still covers every group. Independent of `profile`. Never auto-discovered, and DQX warns rather than acting when the training window shows little structure over time. Not a forecaster, and it never reads the previous row. The named column must not also appear in `columns`. See [Comparing against time](/docs/guide/row_anomaly_detection#comparing-against-time). |
```
mwojtyczka, line 3536: simple info missing, when it is required and when not

docs/dqx/docs/reference/quality_checks.mdx
```diff
@@ -3525,12 +3532,13 @@ Required arguments: `df`, `model_name`, `registry_table`. Optional parameters:
...
+| `profile` | str | `"tabular"` | Which detector to train. `"tabular"` is Isolation Forest — the behaviour that predates this option. `"correlation"` selects a correlation-aware detector for multivariate metrics whose anomalies are broken *correlations* rather than extreme single values; it needs no timestamp column and trains a single model rather than an ensemble. There is no automatic option: DQX never changes the algorithm on your behalf, because the choice cannot be verified without labels. See [Choosing a profile](/docs/guide/row_anomaly_detection#choosing-a-profile). |
```
mwojtyczka, line 3537: I think a note somewhere else better explains this: `"tabular"` (the default) is IsolationForest (scikit-learn with Spark scoring, requires Spark >= 3.4), which judges a row by its own values. `"correlation"` is a covariance-based detector that judges a row by whether its metrics still relate to each other as they normally do; it needs no timestamp column and trains a single model rather than an ensemble, because it is deterministic. Both share the same feature engineering, registry, check function, contributions and AI explanations.

PR #1491, opened 2026-08-29
Add opt-in semantic-aware profiling to DQProfiler

### LINKED ISSUE #1343: [FEATURE]: Profile classification support

IvannKurchenko:
## Changes

Introduce **opt-in semantic-aware profiling** in `DQProfiler`. A lightweight classification stage runs between metric collection and check generation, so each column receives *one* consistent family of checks instead of contradictory overlaps (e.g. `vehicle_type` no longer gets both `is_in([...])` and `min_max(...)`).

The feature is fully backward compatible: when a `DQProfiler` is constructed without a `semantic_registry`, no detection runs and the generated profiles are byte-identical to today.

**New public surface** (all in `src/databricks/labs/dqx/profiler/semantic.py`):

- `DQSemanticType` — Pydantic v2 model naming a column's semantic meaning (e.g. `enum`, `key`, `measurement`, `text`) plus optional properties.
- `DQSemanticTypeDetector` — named callable `(DQProfileContext) -> DQSemanticType | None`.
- `DQProfileContext` — frozen context passed to detectors and contextual builders. Carries `df`, `column_name`, `column_type`, `metrics`, `options`, and `semantic_type`. It exposes `with_metrics(...)` so the profiler loop can refresh the frozen context after earlier builders write back resolved metrics (e.g. min/max).
- `SemanticRegistry` — immutable, name-unique, ordered chain of detectors. Composition surface:
  - construction: `SemanticRegistry.default()`, `SemanticRegistry.of(*detectors)`, `SemanticRegistry(detectors=(...))`
  - composition: `prepend(detector)`, `append(detector)`, `insert(name, detector)` (after the named entry), `replace(name, detector)` (position-preserving swap), `remove(name)`
  - all methods return a new registry; uniqueness is enforced via a `model_validator`.
- `default_semantic_detectors()` and four built-in detectors: `DEFAULT_ENUM_DETECTOR`, `DEFAULT_KEY_DETECTOR`, `DEFAULT_MEASUREMENT_DETECTOR`, `DEFAULT_TEXT_DETECTOR` plus threshold constants (`ENUM_MAX_CARDINALITY_RATIO`, `KEY_MIN_DENSITY_RATIO`, `KEY_MIN_LENGTH_STABILITY_RATIO`).

**Profiler wiring** (`profiler/profiler.py`, `profiler/profile_builder.py`, `profiler/profile.py`):

- `DQProfiler.__init__` gains a keyword-only `semantic_registry: SemanticRegistry | None = None`. Presence of the argument opts the run into semantic detection — there is no per-call override.
- `DQProfile` gains an optional `semantic_type: str | None = None` field so generated profiles record *why* a check was emitted (survives YAML/JSON round-trip; defaults to `None`).
- `DQProfileBuilder` supports two mutually-exclusive callback shapes: the legacy 5-argument `builder` (unchanged) and the new `contextual_builder(ctx: DQProfileContext)`. `@register_profile_builder` gains a `kind=\"context\"` opt-in for the new shape; existing legacy registrations keep working without changes.
- The four library-native builders (`null_or_empty`, `is_in`, `min_max`, `has_no_outliers`) are migrated to the contextual form. `min_max` and `has_no_outliers` now skip emission unless `ctx.semantic_type` is `None` or `\"measurement\"`; `is_in` reuses the enum detector's already-collected distinct values so no second Spark `.distinct().collect()` runs.
- The `_detect_enum` gate and the legacy `is_in` gate both compute `distinct_count / count_non_null` and honour the profiler-wide `distinct_ratio` option, so semantic classification and legacy emission agree on low-repetition columns even in null-heavy datasets. `ShortType` is a first-class enum candidate (added to `_supports_distinct`).

### Linked issues

Resolves #1343

### Tests

- [x] added unit tests — `tests/unit/profiler/test_semantic.py` covers each detector (positive + negative), `SemanticRegistry` immutability and uniqueness invariants (constructor, `of`, `prepend`, `append`, `insert`, `replace`, `remove`), chain semantics (first-match-wins), the tightened enum cardinality guard, numeric-density and string-length-stability key guards including the empty-string edge case, and `DQProfileContext.with_metrics` snapshot semantics. `tests/unit/test_profile_builder.py` covers `kind=\"context\"` vs legacy registration, mutually-exclusive callback validation, and per-builder positive/negative behaviour (including `ShortType` for `is_in`) through the new contextual path.
- [x] added integration tests — `tests/integration/test_profile_semantic.py` covers the default-registry classification of the grounded design columns (`vehicle_type`, `cargo_weight`, `deal_value`, `user_id`, `order_id`, `user_name`, `work_description`), the no-registry byte-identical default, custom-chain composition via `prepend` / `SemanticRegistry.of(...)`, the `ShortType` enum → `is_in` regression guard, the enum-value reuse optimisation (single `.distinct()` invocation per enum column), and a public-API behavioural test that a contextual builder ordered after `min_max` observes the resolved `min`/`max` via `ctx.metrics`.
- [ ] added end-to-end tests
- [x] added performance tests

### Documentation and Demos

- [ ] added/updated demos
- [x] added/updated docs — new \"Semantic-aware profiling\" section in `docs/dqx/docs/reference/profiler.mdx` tagged `<FeatureLifecycleStage stage=\"beta\">` / `<AvailableSinceVersion version=\"0.17.0\">`. Documents the opt-in model, the four built-in detectors and their applicability rules and thresholds, the immutable `SemanticRegistry` composition patterns (`prepend`, `append`, `insert`, `replace`, `remove`, `of`), the `@register_profile_builder(kind=\"context\")` opt-in, and includes a copy-paste example specialised (UUID) detector.


mwojtyczka (commented):
Automated code review (5 findings). Comments posted inline below.

src/databricks/labs/dqx/profiler/semantic.py
```diff
@@ -0,0 +1,407 @@
...
+
+def _detect_enum(ctx: DQProfileContext) -> DQSemanticType | None:
+    """Detect *enum*-like columns: low cardinality and highly repeated values.
+
+    The detector is the sole owner of the enum value collection: when it
+    fires it runs a single distinct-collect on *ctx.df* and hands the result
+    to downstream builders through *ctx.semantic_type.properties.values*.
+    """
+    column_type = ctx.column_type
+    if not (_is_text(column_type) or isinstance(column_type, (T.IntegerType, T.LongType, T.ShortType))):
```
mwojtyczka, line 216: **ShortType columns lose all range/enum checks.** `_detect_enum` accepts `ShortType` here, so a low-cardinality `ShortType` column (e.g. a status code 1..5) is classified as `enum`. But `make_is_in_profile` returns `None` because `_supports_distinct` (profile_builder.py:325) only accepts `IntegerType`/`LongType`/text — not `ShortType` — so no `is_in` is emitted. `make_min_max_profile` is then also suppressed because the semantic type is `enum`, not `measurement`. Net result: the column gets **neither `is_in` nor `min_max`**. In the legacy/no-registry path the same column still gets a `min_max` profile (it is a `NumericType`), so enabling the default registry silently drops checks. Either add `ShortType` to `_supports_distinct`, or don't classify `ShortType` as enum.

src/databricks/labs/dqx/profiler/profiler.py
```diff
@@ -493,10 +600,31 @@ def _build_profiles_for_column(
         *metrics* so that downstream consumers (e.g. LLM primary-key detection) can read them
         without triggering a second Spark action.
         """
+        metadata_map = dict(column_metadata) if column_metadata else {}
+        semantic_type = self._detect_semantic_type(column_df, field_name, field_type, metrics, opts, metadata_map)
+
+        builder_ctx = DQProfileContext(
```
mwojtyczka, line 524: **min/max write-back never reaches contextual builders.** `builder_ctx` is created once here with `metrics=metrics`, but pydantic materializes `ctx.metrics` as a **new dict** (`ctx.metrics is metrics` is `False`). The min/max write-back below (~lines 633/635) mutates the outer `metrics` dict, not `ctx.metrics`. A contextual builder registered after `min_max` that reads `ctx.metrics['min']/['max']` — the exact scenario the docstring promises support for — receives the pre-resolution summary values, not the outlier-adjusted resolved ones. No built-in triggers it today, so it's a silent latent contract break for custom contextual builders.

src/databricks/labs/dqx/profiler/semantic.py
```diff
@@ -0,0 +1,407 @@
...
+
+    cardinality = ctx.metrics.get("count_distinct", 0)
+    if cardinality == 0:
+        return None
+
+    max_in_count = ctx.options.get(PROFILE_OPTION_MAX_IN_COUNT, 0)
+    if cardinality >= max_in_count:
+        return None
+
+    if (cardinality / count_non_null) > ENUM_MAX_CARDINALITY_RATIO:
```
mwojtyczka: **Enum cardinality gate is far more permissive than the `is_in` builder.** This ratio ceiling (`ENUM_MAX_CARDINALITY_RATIO = 0.95`) is much looser than the legacy `is_in` builder's `distinct_ratio` (default `0.05`). On a small table (e.g. 12 rows, 9 distinct strings): legacy `is_in` is suppressed because `9/12 = 0.75` is not `< 0.05`; but semantic profiling passes here (`cardinality 9 < max_in_count 10` and `9/12 = 0.75 <= 0.95`) and emits `is_in` with 9 allowed values. The enum branch never consults `distinct_ratio`, so semantic profiling generates `is_in` rules that ordinary profiling deliberately suppresses on low-repetition columns. Consider aligning the gate with the builder's `distinct_ratio`.

src/databricks/labs/dqx/profiler/profiler.py
```diff
@@ -415,6 +446,41 @@ def _stratified_sample(
...
+        (a dict keyed by column name, each carrying *column_comment*). Tags
+        are intentionally excluded — the SDK exposes them via separate
+        calls and current use cases only need the comment fields. Any SDK
+        failure (missing table, permission denied, non-UC location) is
+        logged as a warning and returns an empty dict, so metadata errors
+        never block profiling.
+        """
+        try:
+            table = self.ws.tables.get(location)
+        except DatabricksError as exc:
```
mwojtyczka: **Non-`DatabricksError` aborts profiling.** This only catches `DatabricksError`. If `ws.tables.get(location)` raises a client-side `ValueError` (e.g. a storage path or otherwise malformed/non-UC three-part name fails SDK name validation), it propagates and crashes `profile_table` — contradicting the docstring's "never block profiling" guarantee. Broaden the `except` (e.g. also catch `ValueError`, or `Exception`) so best-effort metadata fetch stays non-blocking.

docs/dqx/docs/reference/profiler.mdx
```diff
@@ -127,4 +129,158 @@ The `DQDltGenerator` class creates Delta Live Tables expectation statements from
...
+`type="context"` argument to opt into the preferred contextual callback shape,
+which receives a `DQProfileContext` (giving access to *ctx.semantic_type*,
+*ctx.metadata*, *ctx.metrics*, and *ctx.options*):
+
+```python
+from databricks.labs.dqx.profiler.profile import DQProfile
+from databricks.labs.dqx.profiler.profile_builder import register_profile_builder
+from databricks.labs.dqx.profiler.semantic import DQProfileContext
+
+@register_profile_builder("my_custom", type="context")
```
mwojtyczka: **Copy-paste example raises `TypeError`.** The decorator keyword is `kind`, not `type` — `register_profile_builder(profile_type, *, kind=...)`. As written, this example (and the prose mention on line 271) raises `TypeError: register_profile_builder() got an unexpected keyword argument 'type'`. Should be `@register_profile_builder("my_custom", kind="context")`.

docs/dqx/docs/reference/profiler.mdx
```diff
@@ -127,4 +129,158 @@ The `DQDltGenerator` class creates Delta Live Tables expectation statements from
...
+    <AvailableSinceVersion version="0.17.0" heading={false} />
+</FeatureTags>
+
+Semantic-aware profiling introduces a lightweight classification stage between metric
+collection and check generation so that a column receives *one* consistent family of
+checks. For example, a *vehicle_type* column classified as an *enum* now receives an
+*is_in* candidate only — not both *is_in* and *min_max* — while a continuous
+*cargo_weight* column is classified as a *measurement* and receives *min_max* only.
+
+The feature is fully opt-in: it runs iff a `semantic_registry` is supplied to the
```
mwojtyczka:
```suggestion
The feature is fully opt-in: it runs if a `semantic_registry` is supplied to the
```

docs/dqx/docs/reference/profiler.mdx
```diff
@@ -127,4 +129,158 @@ The `DQDltGenerator` class creates Delta Live Tables expectation statements from
...
+# 1. Default chain (enum → key → measurement → text)
+default_registry = SemanticRegistry.default()
+
+# 2. Prepend a custom detector so it wins first
+def _detect_uuid(ctx):
+    # ... your logic ...
+    return DQSemanticType(name="uuid") if ... else None
+
+uuid_detector = DQSemanticTypeDetector(name="uuid", detect=_detect_uuid)
+prepended = SemanticRegistry.default().prepend(uuid_detector)
```
mwojtyczka, line 184: a comment would be good here: 
* prepend(detector) → puts one detector at the front (highest priority)
* replace(detectors) → swaps the entire chain

docs/dqx/docs/reference/profiler.mdx
```diff
@@ -127,4 +129,158 @@ The `DQDltGenerator` class creates Delta Live Tables expectation statements from
...
+
+# 3. Replace the chain entirely
+custom = SemanticRegistry.default().replace([uuid_detector, *default_semantic_detectors()])
```
mwojtyczka: replace doesn't mean "replace a detector" — it means "set the whole chain. I would use something like with_detectors(...)

how about if i want to append or remove detector?

docs/dqx/docs/reference/profiler.mdx
```diff
@@ -127,4 +129,158 @@ The `DQDltGenerator` class creates Delta Live Tables expectation statements from
...
+
+# 4. Empty registry — semantic types are always None
+empty = SemanticRegistry()
+
+profiler = DQProfiler(ws, semantic_registry=default_registry)
```
mwojtyczka, line 205: They way this is constructed is different than registry approach like  @register_profile_builder(...) . Profile builders are customized with a decorator against a global registry; semantic detectors are customized by constructing an immutable value object and threading it through a constructor. Two different mental models for "add my thing to the chain."

src/databricks/labs/dqx/profiler/semantic.py
```diff
@@ -0,0 +1,407 @@
...
+                raise ValueError(f"Detector {detector.name!r} is registered more than once")
+            seen.add(detector.name)
+        return self
+
+    @classmethod
+    def default(cls) -> "SemanticRegistry":
+        """Return a registry populated with *default_semantic_detectors()*."""
+        return cls(detectors=default_semantic_detectors())
+
+    def prepend(self, detector: DQSemanticTypeDetector) -> "SemanticRegistry":
```
mwojtyczka, line 429: **The customization surface (`prepend` + whole-chain `replace`) is thin, asymmetric, and slightly mis-named.** The underlying model is sound — immutable, name-unique, first-match-wins, functional (returns new instances). The concern is scope, not correctness:

1. **`replace(detectors)` doesn't replace *a* detector — it sets the *entire* chain.** The name is misleading: a reader expects `replace(name, detector)` to swap one entry, but this discards everything. `with_detectors(...)` / `of(...)` would signal "this is the whole chain."

2. **Common operations don't exist, so they all degrade into a manual `replace([...])`.** The docs' own example proves it — to *add* one detector the user must re-spread the defaults by hand:
   ```python
   SemanticRegistry.default().replace([uuid_detector, *default_semantic_detectors()])
   ```
   There's no `append` (lower-priority fallback), no `insert_after(name, ...)` (e.g. after `enum` but before `key` — impossible without hand-rebuilding), no `remove(name)`, and no replace-one-by-name. `prepend` only covers the single "make mine win first" case.

3. **Inconsistent with the sibling extension point.** Profile *builders* use a `@register_profile_builder(...)` decorator against a global registry; semantic *detectors* use an immutable value object threaded through a constructor — two mental models for "add my thing to the chain."

**Suggested fix:** rename whole-chain `replace` → `with_detectors`/`of`, and add `append`, `remove(name)`, and either `insert_after(name, detector)` or a real `replace(name, detector)`, so realistic customizations don't require re-spreading `default_semantic_detectors()`.

mwojtyczka (commented):
Follow-up: ergonomics of the SemanticRegistry customization surface.

mwojtyczka (changes requested):

src/databricks/labs/dqx/profiler/semantic.py
IvannKurchenko: Hello, @mwojtyczka Thank you so much for review. I've tried to address all your comments from the review except the following:

> Inconsistent with the sibling extension point. Profile builders use a @register_profile_builder(...) decorator against a global registry; semantic detectors use an immutable value object threaded through a constructor — two mental models for "add my thing to the chain."

I understand and share the concern you raised. Although, the major difference between semantic detectors and profiler builders is order of execution for the first one. Unfortunately, with decorator style registry it won't be possible to guarantee or ensure order of added detectors or manage it in convenient way. 

In defence, recently released [actions api](https://databrickslabs.github.io/dqx/docs/guide/actions_and_alerts/) also follows explicit creation approach and actions are not registered in import time, like profile builders or check functions. 
Additionally, in other PR you pointed correctly to one of the flaws of the singleton registry is that user have to keep in mind operation side effects and needs to de-register it - https://github.com/databrickslabs/dqx/pull/1384#discussion_r3913770969.


Having said that, shall I consider filing ticket to start refactoring towards leveraging more explicit API for profiling module in terms of extension? Off-course preserving backward compatibility.

IvannKurchenko:
@mwojtyczka May I ask for another review round, please? Your previous comments has been addressed. Thank you

mwojtyczka:
## Code review findings — semantic-aware profiling (#1491)

Verified against head `a19ae55`. Ranked by severity.

### 1. `is_in` distinct-ratio denominator changed `total_count` → `count_non_null` — breaks the "byte-identical" no-registry guarantee
**`src/databricks/labs/dqx/profiler/profile_builder.py`** — `make_is_in_profile` (~L150)

```diff
- distinct_ratio = (1.0 * distinct_count) / total_count
+ distinct_ratio = (1.0 * distinct_count) / count_non_null
```

This runs on the **default (no `semantic_registry`) path** — when `semantic_type is None`, control falls through to this calc. For a null-heavy column (1000 rows, 200 nulls, 40 distinct): old ratio `40/1000 = 0.04 < 0.05` → `is_in` emitted; new ratio `40/800 = 0.05` → **not** emitted. This contradicts the PR's stated guarantee ("byte-identical to the pre-feature behaviour", and the `make_is_in_profile` docstring "applicability follows today's byte-identical rules") and the test asserting *"No registry → profiler output is identical to the pre-feature behaviour"* (which only passes because its fixtures contain no nulls).

This change came from aligning the `is_in` builder with `_detect_enum` (a good goal), but because the builder is shared with the legacy path, it shifted legacy output. Either accept the change and reword/drop the "byte-identical" guarantee, or align the two paths without altering the no-registry denominator.

### 2. `ShortType` added to `_supports_distinct` — also breaks "byte-identical"
**`src/databricks/labs/dqx/profiler/profile_builder.py`** — `_supports_distinct` (~L332)

```diff
- return isinstance(column_type, (T.IntegerType, T.LongType) + TEXT_TYPES)
+ return isinstance(column_type, (T.IntegerType, T.LongType, T.ShortType) + TEXT_TYPES)
```

`_supports_distinct` is used by the legacy `is_in` builder too, so a low-cardinality `ShortType` column now emits an `is_in` rule where it previously produced none — a behavior change for users who never opt into the feature. Same trade-off as #1 (this addressed the earlier "ShortType loses all checks" review comment, but via the shared helper). Worth an explicit decision on whether the no-registry path is allowed to change.

### 3. `profile_table` no longer emits the nested `profile` telemetry event (low)
**`src/databricks/labs/dqx/profiler/profiler.py`** — `profile_table` (~L164)

`profile_table` was rerouted from `self.profile(...)` (decorated `@telemetry_logger("profiler","profile")`) to the undecorated `self._profile_dataframe(...)`. Previously table profiling fired both `profile_table` **and** `profile`; now only `profile_table`. This is plausibly intentional de-duplication (the old path double-counted), but it's an observable, undocumented change — dashboards/usage counts keying on the `profile` event will see table-profiling volume drop. Please confirm it's intentional and note it in release notes.

### 4. Stat parse failure mislabels distribution as `"constant"` instead of `"unknown"` (low)
**`src/databricks/labs/dqx/profiler/semantic.py`** — `_detect_measurement` (~L329)

```python
except (TypeError, ValueError):
    span = 0.0; stddev_f = 0.0; mean_f = 0.0
if span <= 0:
    distribution = "constant"
```

When `min`/`max`/`mean`/`stddev` fail to cast to float, the zero-fallback drives `span <= 0` → `distribution="constant"` even for a highly variable column. Since the field is a best-effort guess documented to be `"unknown"` when stats are unavailable, the except branch should yield `"unknown"`, not `"constant"` — otherwise `MeasurementProperties.distribution` is misleading.

### 5. Redundant `DQProfileContext` metrics snapshot (minor cleanup)
**`src/databricks/labs/dqx/profiler/profiler.py`** — `_build_profiles_for_column` (~L524)

`builder_ctx` is constructed with `metrics=metrics`, but the loop's first contextual builder immediately does `builder_ctx = builder_ctx.with_metrics(metrics)` (metrics unchanged at that point), discarding the first snapshot. Plain (non-contextual) builders never read `builder_ctx`. Net: one wasted context/dict copy per column. Minor — skip the first refresh when metrics are unchanged, or construct once and refresh only after the `min_max` write-back.

IvannKurchenko:
@mwojtyczka  I see the conflicts, will work on resolution. Thanks.

PR #1492, opened 2026-08-31
Add allow_nulls option to comparison checks

### LINKED ISSUE #1395: [FEATURE]: additional is_not_null_and_* checks

### Problem statement

There are a small number of checks that combine an is_not_null check with another check (e.g. is_not_null_and_is_in_list). It would be useful to have more of those, e.g. `is_not_null_and_is_equal_to`, `is_not_null_and_is_not_equal_to` - or alternatively to have an "allow_nulls" flag to be set as an arg on functions to control this.

### Proposed Solution

Either supplementing the existing suite of checks with additional `is_not_null_and_{check}` checks, or to add an `allow_nulls` param to existing checks so that it is easy to manage whether null values should pass or fail a check without having to add separate null checks.

### Additional Context

_No response_

alvaromoureupm:
## Changes
Adds an `allow_nulls` argument (default `True`) to the six comparison row-level checks — `is_equal_to`, `is_not_equal_to`, `is_not_less_than`, `is_not_greater_than`, `is_in_range`, `is_not_in_range` — implementing the approach suggested by @ghanse in the linked issue.

Because of Spark's null-comparison semantics, null values always pass these checks today; making nulls fail requires pairing each rule with a separate `is_not_null` rule, which splits the failure metadata across two checks. With `allow_nulls=False`:

- null values fail the check with a dedicated message (`Column 'x' value is null`), while non-null failures keep their existing messages
- the auto-generated condition alias gains an `is_null_or_` prefix (e.g. `x_is_null_or_not_equal_to_value`), mirroring the existing `is_not_null_and_is_in_list` naming convention

The default behavior is fully backwards compatible: with `allow_nulls=True` (default) conditions, messages, and aliases are unchanged. The logic is shared via a single `_make_condition_handling_nulls` helper.

If this direction looks good, I'm happy to follow up extending `allow_nulls` to other null-passing checks (e.g. list membership checks) in a separate PR.

### Linked issues

Resolves #1395

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

Ran locally: `make fmt`, `make lint` (mypy + pylint 10.00/10), full `tests/unit/test_row_checks.py` + `tests/unit/test_check_func_signatures.py` (178 passed), and the new integration test `test_col_comparison_checks_with_nulls_failing` against a Databricks serverless workspace (passed).

### Documentation and Demos

- [ ] added/updated demos
- [x] added/updated docs


docs/dqx/docs/reference/quality_checks.mdx
```diff
@@ -43,12 +43,12 @@ You can also define your own custom checks in Python (see [Creating custom check
...
+| `is_in_range`                      | Checks whether the values in the input column are in the provided range (inclusive of both boundaries).                                                                                                                                                                                                                                                                                                                           | `column`: column to check (can be a string column name or a column expression); `min_limit`: min limit as number, date, timestamp, column name or sql expression; `max_limit`: max limit as number, date, timestamp, column name or sql expression; `allow_nulls`: (optional) if True (default), null values pass the check; if False, null values fail the check |
```
mwojtyczka, line 46: Only the arguments column documents `allow_nulls`; the prose description of each of the six checks (`is_in_range`, `is_not_in_range`, `is_not_equal_to`, `is_equal_to`, `is_not_less_than`, `is_not_greater_than`) still says nothing about the new null-failing behavior.

A user skimming the description column (rather than the long arguments column) sees e.g. "Checks whether the values in the input column are in the provided range" with no hint that null handling is now configurable — so they miss the feature and keep pairing the check with a separate `is_not_null` rule, which is exactly the workflow this PR set out to replace. Consider adding a short clause like "(null values pass by default; set `allow_nulls=False` to fail them)" to each description.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -208,6 +208,44 @@ def make_condition(condition: Column, message: Column | str, alias: str) -> Colu
...
+
+    Returns:
+        Column object for condition
+    """
+    if allow_nulls:
+        return make_condition(condition, message, f"{col_str_norm}_{alias_suffix}")
+
+    return make_condition(
+        col_expr.isNull() | condition,
+        F.when(col_expr.isNull(), F.lit(f"Column '{col_expr_str}' value is null")).otherwise(message),
```
mwojtyczka, line 244: For column-expression inputs (e.g. `column='a + b'`), the `allow_nulls=False` path emits `Column 'a + b' value is null` whenever **any** operand is null, attributing nullness to the whole composite expression rather than the individual column that is actually null.

Example: `is_in_range('a + b', 0, 10, allow_nulls=False)` on a row where only `b` is null produces `_errors` text "Column 'a + b' value is null". This is cosmetic and consistent with the existing `is_not_null` behavior, so low severity — just flagging it as a potential source of confusion during triage.

tests/integration/test_row_checks.py
```diff
@@ -4613,6 +4613,44 @@ def test_col_is_not_equal_to_with_tolerance(spark, set_utc_timezone):
     assertDataFrameEqual(actual, expected)
 
 
+def test_col_comparison_checks_with_nulls_failing(spark):
+    schema = "a: int"
```
mwojtyczka, line 4617: Nice end-to-end coverage of the `allow_nulls=False` path here. Per the check-addition checklist in `AGENTS.md`, though, a few coverage points appear to be missing for this change:

- **Performance test** — `tests/perf/test_apply_checks.py` isn't extended for the new `allow_nulls=False` branch (extra `isNull()` OR-ing + `F.when` message override), so the at-scale behavior of six widely-used checks ships unbenchmarked.
- **All-checks metadata resources** (`tests/resources/all_row_checks.yaml`) and **`test_apply_checks_all_checks_using_classes`** in `tests/integration/test_apply_checks.py` aren't extended to exercise the `allow_nulls=False` variants.

Without these, a future refactor of `_make_condition_handling_nulls` could regress the null-failing path unnoticed. The default (`allow_nulls=True`) path is unchanged, so this is completeness rather than a correctness gap.

mwojtyczka (approved):
LGTM - yes we can apply the same logic to other checks

tests/integration/test_row_checks.py
```diff
@@ -4613,6 +4613,44 @@ def test_col_is_not_equal_to_with_tolerance(spark, set_utc_timezone):
...
+    test_df = spark.createDataFrame([[1], [5], [None]], schema)
+
+    actual = test_df.select(
```
mwojtyczka, line 4620: Nit: It would be good to also cover the case where we provide column expresssion (e.g. a + b).

mwojtyczka: added

PR #1494, opened 2026-09-01
Make MCP integration coverage deterministic

vb-dbrks:
## Changes

The `mcp` coverage number moves on its own. Across eight consecutive nightly runs of unchanged `main`, `runner.py` reported anywhere from **49% to 75%** and the total from **75% to 81%** — against a 2% threshold on the `dqx-mcp-server` component, so the status flips on PRs that never touch `mcp-server/`.

The tests are not the problem: every operation ran and returned its result. What was lost is the **coverage upload**. The runner's task runs inside a Databricks python shell, and the task arguments are not in `sys.argv` when `site.py` execs our `.pth` — so the results volume, and with it the upload destination, cannot be derived yet. A checkpoint tick in that window can only leave the data in `/tmp`. At the old 15s interval a short-lived interpreter could spend its whole life there and ship nothing: **7 of 21** runner runs produced a data file, and which ones did tracked process lifetime, not outcome.

- Checkpoint interval 15s → 5s, so every interpreter gets a tick once `argv` is populated.
- Key the data file to the **interpreter** rather than to a run id that does not exist at startup.
- The bootstrap now writes its diagnostics beside the data file and the suite prints them. It previously logged only to raw stderr, which a job run does not retain, so a lost upload was undiagnosable from CI.

Measured after the change: **20/20 interpreters uploading, `runner.py` at 76%** — above the whole historical range.

### Linked issues

None.

### Tests

- [x] manually tested
- [ ] added unit tests
- [ ] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

Test-only change (everything is under `tests/integration_mcp/`); validated by running `make mcp-integration` against a coverage-enabled deploy and comparing collected data files and the merged report before and after.

### Documentation and Demos

- [ ] added/updated demos
- [ ] added/updated docs


tests/integration_mcp/conftest.py
```diff
@@ -348,17 +355,33 @@ class McpClient:
...
+        job_id: str = "",
+        workspace_host: str = "",
+        run_id_sink: list[int] | None = None,
+    ):
         self._url = url
         self._get_token = get_token  # mint a fresh bearer per request (no expiry over a long run)
         # Used only to build clickable job/run URLs when something fails — never for auth.
         self._job_id = job_id
         self._workspace_host = (workspace_host or os.environ.get("DATABRICKS_HOST", "")).rstrip("/")
+        # Every run this client submits, in order. Coverage collection needs the exact set: each
```
ghanse: These three spots all say each runner run writes its own data file, and that's why we track the run IDs — so we can spot a run whose file went missing. But down in `_log_runner_data_completeness` (line 515) you say the real story: there's one file per interpreter, not per run, so matching files to runs "would always look short and mean nothing." That second version is right — the per-run-file idea is the exact thing this PR fixes. Right now the run-ID list gets threaded through three functions just to print a count. I'd either drop it and get the count some simpler way, or reword these comments so they don't repeat the old, wrong idea.

tests/integration_mcp/conftest.py
```diff
@@ -348,17 +355,33 @@ class McpClient:
...
+        # runner run ships its own data file, so this is the only way to tell a file that was never
+        # written from one that was never expected. See collect_remote_coverage.
+        self._run_id_sink = run_id_sink if run_id_sink is not None else []
 
     @property
     def url(self) -> str:
         return self._url
 
+    @property
+    def submitted_run_ids(self) -> list[int]:
```
ghanse: `submitted_run_ids` is never read anywhere — only the internal `_run_id_sink` list gets used. Looks like dead code; safe to drop unless something's about to use it.

tests/integration_mcp/coverage_bootstrap/src/dqx_mcp_coverage_bootstrap/__init__.py
```diff
@@ -120,19 +195,34 @@ def _flush(final: bool) -> None:
     """
     if _cov is None:
         return
+    try:
+        _flush_locked(final)
+    finally:
+        # In a finally so the breadcrumb is published on every path, including the guarded
+        # already-finalized return — otherwise the last message never reaches the volume.
```
ghanse: The comment mentions a "guarded already-finalized return," but there's no such guard in `_flush_locked` — just the `cov is None` and no-destination early returns. The `finally` is fine either way; just tweak the comment so it's not describing code that doesn't exist.

tests/integration_mcp/conftest.py
```diff
@@ -523,7 +591,7 @@ def _download_coverage_files(ws: WorkspaceClient, coverage_dir: str, since: floa
     for entry in ws.files.list_directory_contents(coverage_dir):
         path = entry.path or ""
         name = path.rsplit("/", 1)[-1]
-        if not path or not name.startswith(".coverage"):
+        if not path or not (name.startswith(".coverage") or name.startswith("dqxcov-log")):
```
mwojtyczka: **Breadcrumb files defeat the zero-data-files warning.** `_download_coverage_files` now returns both `.coverage.*` and `dqxcov-log.*` entries in one list. In `collect_remote_coverage`, a run that produced a breadcrumb but **no real coverage data** makes `downloaded` non-empty, so the loud `if not downloaded:` WARNING ("downloaded 0 data files — the mcp flag will not be uploaded") is skipped, and the `else` branch instead prints `coverage files downloaded: N` counting breadcrumbs as data files. The two messages then contradict each other, since `_log_runner_data_completeness` separately warns `if not data_files`. Have the `if not downloaded` guard check data files (`.coverage*`) rather than the mixed list.

tests/integration_mcp/coverage_bootstrap/src/dqx_mcp_coverage_bootstrap/__init__.py
```diff
@@ -111,6 +154,38 @@ def _persist(local_path: str, remote_path: str) -> None:
...
+def _write_breadcrumb() -> None:
+    """Publish the diagnostics collected so far, so they are readable even if no flush ever runs.
+
+    Called at startup and after each flush. The filename deliberately does NOT start with
+    ``.coverage.`` — ``coverage combine`` globs that prefix and would choke on a text file.
+    """
+    destination = _destination_dir()
+    if not destination:
+        return
+    _persist_text(f"{destination}/dqxcov-log.{_key}.txt", "\n".join(_messages) + "\n")
```
mwojtyczka: **Unsynchronized `"\n".join(_messages)` can kill the checkpoint thread.** At process exit the atexit `_flush(True)` (main thread) and an in-flight checkpoint `_flush(False)` (daemon thread) both reach `_write_breadcrumb`, and this `join` runs outside `_lock` while `_log` appends to `_messages` from the other thread. If it raises (list mutated mid-iteration), it escapes `_write_breadcrumb`, propagates out of `_flush`'s `finally`, and since `_checkpoint_loop` has no guard the checkpoint thread dies permanently — reintroducing the atexit-only lost-upload failure this PR fixes. Snapshot the list under `_lock` before joining, and/or guard the checkpoint loop body.

tests/integration_mcp/coverage_bootstrap/src/dqx_mcp_coverage_bootstrap/__init__.py
```diff
@@ -111,6 +154,38 @@ def _persist(local_path: str, remote_path: str) -> None:
...
+        from databricks.sdk import WorkspaceClient
+
+        ws = WorkspaceClient()
+        ws.files.create_directory(os.path.dirname(remote_path))
+        ws.files.upload(remote_path, io.BytesIO(payload), overwrite=True)
+    except Exception:  # noqa: BLE001 — a diagnostic must never break the app or fail a job
+        pass
+
+
+def _write_breadcrumb() -> None:
```
mwojtyczka: **`_messages` grows unbounded → O(n²) volume traffic.** `_persist` appends a message on every flush and the interval dropped 15s→5s, so `_messages` grows ~1–2 entries per 5s for the whole suite. `_write_breadcrumb` re-joins and re-uploads the entire growing list each tick, and on the app (no FUSE) `_persist_text` builds a fresh `WorkspaceClient()` + `create_directory` + `upload` every time — total breadcrumb bytes shipped is quadratic in run length. Consider capping `_messages` (ring buffer) or writing incrementally.

tests/integration_mcp/conftest.py
mwojtyczka: **Dead property + over-engineered plumbing.** `submitted_run_ids` is never read anywhere; `expected_run_ids` is used solely to compute `submitted = len(set(...))` for one log line. Meanwhile `_log_runner_data_completeness` states "There is deliberately no per-run file check" — so the file-vs-expected comparison the docstring of `collect_remote_coverage` justifies ("the only way to tell never-written from never-expected") is never actually performed. This count could be derived locally in the test without threading a `run_id_sink` through `deploy_mcp_app` and `McpClient`.

mwojtyczka (commented):
Automated code review (5 findings). These are the ones worth a look before merge — findings 1 and 2 undermine the PR's own goal of reliable coverage upload.

tests/integration_mcp/conftest.py
```diff
@@ -370,8 +393,11 @@ def call(self, name: str, arguments: dict | None = None, *, poll: bool = True, t
         payload = _tool_payload(
             _mcp_request(self._url, self._get_token(), "tools/call", {"name": name, "arguments": arguments or {}})
         )
-        if poll and payload.get("status") == "submitted" and payload.get("run_id"):
-            return self.wait(payload["run_id"], timeout=timeout)
+        run_id = payload.get("run_id")
+        if payload.get("status") == "submitted" and run_id:
+            self._run_id_sink.append(int(run_id))
```
mwojtyczka: **`int(run_id)` adds a new crash path.** Previously `run_id` flowed only into `self.wait(...)`, which tolerates a string. This `int(run_id)` conversion — added purely for coverage bookkeeping — will raise `ValueError` in the normal `call` path if the MCP server ever returns a non-numeric run_id, turning a diagnostic detail into a test failure. Wrap it defensively or store the raw value.

mwojtyczka (changes requested):
Generally looking good, left some comments worth checking before we merge

tests/integration_mcp/conftest.py
vb-dbrks: Fixed, and it was my bug. I introduced it when I made the download pull breadcrumbs back too. Reporting is in one place now and counts .coverage* files only, so "nothing arrived" and "no coverage arrived" stopped being the same check.

tests/integration_mcp/coverage_bootstrap/src/dqx_mcp_coverage_bootstrap/__init__.py
vb-dbrks: Good catch, this was the important one. It would have quietly brought back the atexit-only failure this branch exists to fix. Snapshots under _lock now, and the loop body is guarded so one bad tick cannot kill the thread. Checked it by forcing _flush to raise, thread survives and logs it.

tests/integration_mcp/coverage_bootstrap/src/dqx_mcp_coverage_bootstrap/__init__.py
vb-dbrks: Fixed at source rather than just capped. A copy is logged once per destination, an unchanged breadcrumb is not re-uploaded, and the buffer is bounded. 2 messages after three ticks instead of one per tick. One note, the 15s to 5s premise is out of date now, the default is 0 and only the app sets 15s, but the concern still applied to the app so I fixed it.

tests/integration_mcp/conftest.py
vb-dbrks: Dropped all of it, sink, property and count. @ghanse flagged the same thing from the comments side. (edited, I had the wrong name in here before.)

tests/integration_mcp/conftest.py
vb-dbrks: Gone with the plumbing.

tests/integration_mcp/conftest.py
vb-dbrks: Dropped it rather than reworded it. You were right that the per-run-file idea is the thing this PR disproves, so leaving it in the comments was worse than the plumbing.

tests/integration_mcp/conftest.py
vb-dbrks: Removed.

tests/integration_mcp/coverage_bootstrap/src/dqx_mcp_coverage_bootstrap/__init__.py
vb-dbrks: Fixed. My first rewrite was also wrong, the "_cov is None" return never reaches the finally, so it now only says what actually happens.

tests/integration_mcp/coverage_bootstrap/src/dqx_mcp_coverage_bootstrap/__init__.py
```diff
@@ -108,7 +176,53 @@ def _persist(local_path: str, remote_path: str) -> None:
...
+    # evaluating _persist_text's argument, i.e. OUTSIDE that function's own try — so an exception here
+    # would escape _write_breadcrumb, propagate out of _flush's finally, and (see _checkpoint_loop)
+    # take the checkpoint thread with it permanently.
+    global _last_breadcrumb  # noqa: PLW0603 — module-level singleton, mirrors the other hooks
+    with _lock:
+        snapshot = list(_messages)
+    text = "\n".join(snapshot) + "\n"
+    if text == _last_breadcrumb:
+        return
+    _last_breadcrumb = text
```
mwojtyczka: **Failed breadcrumb upload is recorded as done and never retried.** `_last_breadcrumb = text` is set here, *before* `_persist_text(...)` on the next line — and `_persist_text` swallows every error (line 200-201). If the upload hits a transient Files-API error, the dedup state has already advanced, so every later flush with the same text short-circuits at the `if text == _last_breadcrumb: return` guard above and the breadcrumb is never re-sent. For the runner's single task-end breadcrumb (no further `_log` messages to change the text) the diagnostics this PR exists to deliver are lost precisely when an upload was failing. Advance `_last_breadcrumb` only after a *confirmed* write (have `_persist_text` return success/failure).

tests/integration_mcp/coverage_bootstrap/src/dqx_mcp_coverage_bootstrap/__init__.py
```diff
@@ -108,7 +176,53 @@ def _persist(local_path: str, remote_path: str) -> None:
...
+    # Snapshot under the lock rather than joining the live buffer: _log appends from whichever thread
+    # is running, and building the text outside the lock would race it. The join happens while
+    # evaluating _persist_text's argument, i.e. OUTSIDE that function's own try — so an exception here
+    # would escape _write_breadcrumb, propagate out of _flush's finally, and (see _checkpoint_loop)
+    # take the checkpoint thread with it permanently.
+    global _last_breadcrumb  # noqa: PLW0603 — module-level singleton, mirrors the other hooks
+    with _lock:
+        snapshot = list(_messages)
+    text = "\n".join(snapshot) + "\n"
+    if text == _last_breadcrumb:
```
mwojtyczka: **Dedup state is read/compared/assigned outside `_lock`.** The `with _lock:` above guards only the `list(_messages)` snapshot; this compare and the assignment two lines down touch the module-level `_last_breadcrumb` without it. Near shutdown the daemon checkpoint thread and the main thread's atexit `_flush(True)` can both reach here, both observe `text != _last_breadcrumb`, and either double-upload or interleave read/compare/set so one write is dropped. Genuine data race on shared state (minor impact, but worth folding the dedup check into the locked region).

tests/integration_mcp/coverage_bootstrap/src/dqx_mcp_coverage_bootstrap/__init__.py
```diff
@@ -108,7 +176,53 @@ def _persist(local_path: str, remote_path: str) -> None:
...
+                handle.write(payload)
+            return
+        except OSError:
+            pass
+        from databricks.sdk import WorkspaceClient
+
+        ws = WorkspaceClient()
+        ws.files.create_directory(os.path.dirname(remote_path))
+        ws.files.upload(remote_path, io.BytesIO(payload), overwrite=True)
+    except Exception:  # noqa: BLE001 — a diagnostic must never break the app or fail a job
```
mwojtyczka: **`_persist_text` swallows every failure with no diagnostic.** `_persist` logs on each branch (`FUSE copy unavailable…` / `uploaded →`), but this `except Exception: pass` emits nothing — so a breadcrumb-write failure leaves neither the breadcrumb nor any note that it couldn't be written, ironic for the mechanism whose whole purpose is diagnosing lost uploads. This is also what enables finding on line 224 to go unnoticed. At minimum log the failure.

mwojtyczka (commented):
Automated code review (2nd pass, 5 findings). Findings 1–3 form one cluster: a swallowed write failure advances the dedup state and blocks retry, so the breadcrumb is lost with no trace exactly when uploads are failing.

tests/integration_mcp/coverage_bootstrap/src/dqx_mcp_coverage_bootstrap/__init__.py
```diff
@@ -108,7 +176,53 @@ def _persist(local_path: str, remote_path: str) -> None:
     ws.files.create_directory(os.path.dirname(remote_path))  # upload does not create parents
     with open(local_path, "rb") as handle:
         ws.files.upload(remote_path, handle, overwrite=True)
-    _log(f"uploaded -> {remote_path}")
+    if remote_path not in _announced_copies:
+        _announced_copies.add(remote_path)
+        _log(f"uploaded -> {remote_path}")
+
+
+def _persist_text(remote_path: str, text: str) -> None:
```
mwojtyczka: **Duplicates `_persist`'s FUSE→Files-API fallback.** This is a second near-identical copy of the `try os.makedirs+open, else WorkspaceClient().files.upload` structure in `_persist` (line 153), and the two have already drifted — `_persist` logs on each branch, this one doesn't. A single helper taking a bytes/stream source would carry the platform fallback once. Maintainability, not a crash.

tests/integration_mcp/coverage_bootstrap/src/dqx_mcp_coverage_bootstrap/__init__.py
```diff
@@ -24,23 +26,72 @@
...
+    bug stayed unexplained across three full suite runs. The app, whose stderr *is* captured,
+    keeps working either way, and stderr is retained so a pre-logging-config failure still surfaces.
     """
+    _messages.append(message)
     sys.stderr.write(f"[dqx-coverage] {message}\n")
     sys.stderr.flush()
+    try:
+        import logging
+
+        logging.getLogger("dqx-coverage").warning(message)
```
mwojtyczka: **Every diagnostic is logged at `WARNING`.** Routine, expected events (tracer start, each successful `copied →`, each task-end flush) all flow through `_log` and land at WARNING severity — semantically wrong for informational lines, and on a long-lived app it emits a steady stream of spurious warnings on the flush path. Consider `info` for routine events and reserving `warning` for actual problems.

mwojtyczka (changes requested):
Previous comments are resolved, added a few more

tests/integration_mcp/coverage_bootstrap/src/dqx_mcp_coverage_bootstrap/__init__.py
vb-dbrks: Fixed, and gone rather than hardened. Breadcrumbs are now written only at the four moments that say something (start, task-end flush, final flush, any flush that did not land), so the dedup that caused this is deleted and there is no state left to advance early. Checked with a stub, 3 calls give 3 real attempts and two logged failures.

tests/integration_mcp/coverage_bootstrap/src/dqx_mcp_coverage_bootstrap/__init__.py
vb-dbrks: Also gone with the dedup, there is no shared _last_breadcrumb any more. Related though, and worse than this: the snapshot lock was not protecting anything either, because _log appended without taking it. The buffer now has its own lock (_log cannot take the flush lock, it is called from inside it and would deadlock).

tests/integration_mcp/coverage_bootstrap/src/dqx_mcp_coverage_bootstrap/__init__.py
vb-dbrks: Fixed, a failed write always logs now. You were right that this is what let the finding above hide, it was the only blind spot in a module that exists to explain missing uploads.

tests/integration_mcp/coverage_bootstrap/src/dqx_mcp_coverage_bootstrap/__init__.py
vb-dbrks: Merged into one _write_to_volume that both payloads use. Went with merging over the 3-or-more rule since the drift you spotted had already happened and the fallback is the subtle bit.

tests/integration_mcp/coverage_bootstrap/src/dqx_mcp_coverage_bootstrap/__init__.py
vb-dbrks: Fixed, routine events at INFO and problems at WARNING. Three stay at WARNING on purpose: tracing started, NOT started, and a failed checkpoint tick all fire before the task calls basicConfig, where lastResort drops INFO, and those are the lines that explain a silent 0%. My first pass demoted all three and a review caught it.

tests/integration_mcp/coverage_bootstrap/src/dqx_mcp_coverage_bootstrap/__init__.py
```diff
@@ -166,10 +368,13 @@ def _start() -> None:
     atexit.register(_flush, True)
     if _CHECKPOINT_SECONDS > 0:
         threading.Thread(target=_checkpoint_loop, daemon=True, name="dqx-coverage").start()
-    _log(f"tracing started (role={role}, data_file={_data_file})")
+    # WARNING, not INFO: emitted from the .pth before the task calls basicConfig, where lastResort
+    # drops INFO — and this is the line whose absence explains a silent 0% report. Once per process.
+    _log(f"tracing started (key={_key}, data_file={_data_file}, argv0={sys.argv[0]!r})", warning=True)
+    _write_breadcrumb()
```
mwojtyczka: **Network I/O during interpreter/`.pth` init.** `_start()` now calls `_write_breadcrumb()` at import time. For the app, `DQX_CATALOG`/`DQX_TMP_SCHEMA` are present at interpreter start, so `_destination_dir()` resolves and the startup breadcrumb goes through `_write_to_volume`. With no FUSE mount, the fallback constructs a `WorkspaceClient` and performs a **synchronous Files API upload inside the `.pth` exec, before uvicorn boots**. On slow auth/network that delays (up to the SDK timeout) or perturbs app startup on every coverage-enabled deploy — work the old `_start()` did not do (previously the first network call happened only on flush). Consider deferring the breadcrumb write off the import path, or skipping it at startup.

tests/integration_mcp/coverage_bootstrap/src/dqx_mcp_coverage_bootstrap/__init__.py
```diff
@@ -77,80 +139,220 @@ def _destination_dir() -> str:
...
+                # sys.argv yet. Callers for whom it is NOT routine say so themselves — see
+                # flush_at_task_end.
                 _log("no destination could be resolved; data left in /tmp only")
-                return
-            role, key = _role_and_key()
+                return False
             # The name must start with '.coverage.' (and never be exactly '.coverage') or
             # `coverage combine` will not glob it.
-            _persist(_data_file, f"{destination}/.coverage.{role}.{key}")
+            return _persist(_data_file, f"{destination}/.coverage.{_key}")
```
mwojtyczka: **`_lock` held across the Files API upload.** `_flush_locked` runs `_persist` → `_write_to_volume` (a network round-trip) while holding `_lock`. If the checkpoint thread is mid-upload holding `_lock` when the process gets SIGTERM, the `atexit` `_flush(final=True)` blocks on `_lock` waiting for the in-flight upload. A slow/hung upload can then exceed the platform's ~15s shutdown budget → SIGKILL → the final (freshest) coverage is never persisted — the exact loss this mechanism aims to prevent. Consider snapshotting the payload under the lock and uploading outside it, or a bounded timeout on the final flush.

tests/integration_mcp/coverage_bootstrap/src/dqx_mcp_coverage_bootstrap/__init__.py
```diff
@@ -77,80 +139,220 @@ def _destination_dir() -> str:
...
+    directory = os.path.dirname(remote_path)
     try:
-        os.makedirs(os.path.dirname(remote_path), exist_ok=True)
-        shutil.copyfile(local_path, remote_path)
-        _log(f"copied -> {remote_path}")
+        try:
+            os.makedirs(directory, exist_ok=True)
+            with open(remote_path, "wb") as handle:
+                handle.write(payload)
+            _announce_once(remote_path, announce, "copied")
```
mwojtyczka: **FUSE availability inferred only from exceptions → silent local-disk write counted as delivered.** `_write_to_volume` treats any successful `makedirs`+`open` as success. On an app container where `/Volumes` (or the target dir) happens to be locally creatable/writable, the write succeeds, this announces `copied …`, and returns `True` — coverage lands on **ephemeral local disk, is never uploaded to the UC volume, yet is counted as delivered**, with no failure breadcrumb. Consider verifying the path is an actual FUSE/volume mount rather than inferring it from the absence of an exception.

tests/integration_mcp/coverage_bootstrap/src/dqx_mcp_coverage_bootstrap/__init__.py
```diff
@@ -77,80 +139,220 @@ def _destination_dir() -> str:
...
+    if not announce:
         return
+    _log_once(remote_path, f"{verb} {announce} -> {remote_path}")
+
+
+def _log_once(key: str, message: str) -> None:
+    """Log *message* the first time *key* is seen, and never again."""
+    if key in _announced_copies:
+        return
+    _announced_copies.add(key)
```
mwojtyczka: **`_announced_copies` mutated from multiple threads without a lock.** The check-then-add (`if key in _announced_copies: return` … then `_announced_copies.add(key)`) isn't atomic, so the checkpoint thread and the atexit/main thread can both pass the guard and each emit the “once” announcement. Benign (double log, no crash), but inconsistent with the `_messages_lock` deliberately added elsewhere in this same PR for exactly this class of race — worth guarding the same way for consistency.

.github/actions/merge-mcp-coverage/action.yml
```diff
@@ -27,6 +27,18 @@ runs:
...
 
+        # The bootstrap's own diagnostics, downloaded beside the data files. Print them: the suite's
+        # stderr is not retained in the CI log, so without this the breadcrumbs exist on the runner
+        # and are visible nowhere — which is the situation that left three separate investigations of
+        # lost coverage with nothing to read. Globbed with a guard because this step runs under
+        # `bash -e`, and an unmatched glob would otherwise fail the job over a diagnostic.
+        for crumb in dqxcov-log.*; do
+          [ -e "$crumb" ] || continue
+          echo "::group::$crumb"
+          cat "$crumb"
```
mwojtyczka: **Unescaped breadcrumb content piped into the runner's command channel (log/command injection).** `cat "$crumb"` prints bootstrap-authored breadcrumb text unescaped inside a `::group::` block. Breadcrumbs include exception reprs and `sys.argv[0]` (e.g. `NOT started (…): {exc!r}`); if any line starts with `::` (`::error::`, `::set-output`, …) the Actions runner interprets it as a workflow command. Low risk since the content is internally generated, but consider masking workflow-command prefixes (e.g. pipe through `sed 's/^::/⁣::/'` or wrap so GitHub treats it as literal) rather than raw `cat`.

tests/integration_mcp/coverage_bootstrap/src/dqx_mcp_coverage_bootstrap/__init__.py
```diff
@@ -77,80 +139,220 @@ def _destination_dir() -> str:
...
+    _log("task-end flush (deterministic; checkpoint retired)")
+    # _flush publishes the breadcrumb itself when the flush did not land, so only the success path
+    # needs one here — otherwise the same bytes would be uploaded twice on the failure path.
+    if _flush(final=False):
+        _write_breadcrumb()
+    else:
+        # Not benign here, unlike an early checkpoint tick: the work is finished, so if this did not
+        # reach the volume the run's coverage is lost. WARNING because the breadcrumb cannot help —
+        # an unresolved destination is exactly the case where it has nowhere to be written either.
+        _log("task-end flush did not reach the volume; this run's coverage is lost", warning=True)
```
mwojtyczka, line 361: **The most informative failure line never reaches the durable breadcrumb.** On the resolved-destination-but-persist-failed path, `_flush(final=False)` writes the breadcrumb (capturing messages up to “flush failed”) *before* `flush_at_task_end` logs this `this run's coverage is lost` line. That decisive line goes only to stderr/logging (which a job run does not retain), never into the breadcrumb file that was just uploaded — so the single most useful line about the failure is absent from the one record that survives. Consider appending it to `_messages` before the breadcrumb write, or re-writing the breadcrumb after logging it.

mwojtyczka (commented):
Automated code-review pass. This is a test-only PR (MCP integration coverage bootstrap) and the core logic reads as sound; the findings below are robustness/CI concerns. #1-#3 are the substantive ones (real behavior in the deployed app); #4-#6 are lower-severity robustness/observability.

tests/integration_mcp/coverage_bootstrap/src/dqx_mcp_coverage_bootstrap/__init__.py
vb-dbrks: Gone, along with the breadcrumbs themselves. I checked on a real runner job first: the bootstrap logging records do survive in the retained job output once the task is running, on both stderr and the logging channel, so the volume breadcrumbs were duplicating what the job log already keeps. Only the interpreter start line is missing, since the .pth runs before log capture begins. No writes at import now, and the SDK client is created lazily on first publish rather than warmed at startup.

tests/integration_mcp/coverage_bootstrap/src/dqx_mcp_coverage_bootstrap/__init__.py
vb-dbrks: Fixed. Snapshot saves and reads the file under the tracer lock, publish uploads with it released. Worth noting the read under that lock is load-bearing rather than incidental: coverage opens its SQLite with journal_mode=off, so a copy taken mid-save can be structurally invalid, not just stale. A separate upload lock serializes publishes and is acquired differently at the three call sites, so a periodic tick can never delay the final flush. If the final flush cannot get it within 2s it publishes to a sibling path rather than risk interleaving on the canonical one.

tests/integration_mcp/coverage_bootstrap/src/dqx_mcp_coverage_bootstrap/__init__.py
vb-dbrks: Fixed properly rather than papered over: the write path is now chosen by runtime, not by catching OSError. The app has no mount and sets DATABRICKS_APP_NAME, so it goes straight to the Files API. Verified with a stub that the app path writes nothing to local disk.

tests/integration_mcp/coverage_bootstrap/src/dqx_mcp_coverage_bootstrap/__init__.py
vb-dbrks: Gone with the announce-once bookkeeping, which only existed to keep breadcrumb traffic flat. I dropped once-only logging entirely rather than adding a third lock for it.

vb-dbrks:
Summary of this round, and a request on process.

All four findings are addressed, but three of them by deletion rather than by hardening. Looking at the review history, every finding so far has landed in the breadcrumb machinery and none in the actual fix:

```
coverage_bootstrap/__init__.py    <- 21 of 21 findings
entry point / bundle / guard tests <- 0
```

The breadcrumbs were scaffolding for the investigation and the investigation is done, so I removed them instead of hardening them again. I checked the assumption first on a real runner job rather than guessing: the bootstrap's logging records do survive in a job's retained output once the task is running, so the volume breadcrumbs were duplicating what the job log already keeps.

That takes out the import-time upload (P1) and the announce-once race (P4). The other two are about the coverage data path, so they are fixed on their merits: the tracer lock is no longer held across the network, and the write path is chosen by runtime rather than by catching OSError.

I also ran an automated review locally before pushing this time, which found five more, two of them real: _log could raise on a closed stderr and, through the entry point's finally, replace the runner's real exception and fail a healthy run; and the checkpoint interval was parsed at module scope, where a bad value raises inside site.py so coverage never starts and the diagnostic that exists to explain that is never emitted. Both fixed.

One addition beyond the findings: downloaded data files are now validated before the merge. A single corrupt member makes coverage combine raise, and CI runs it under `|| true`, so one truncated file would silently zero the whole report while every job stayed green.

Net effect: the PR is smaller than it was before this round.

On the wobble itself, which is what this PR is for. Four CI runs now agree, and two attempts on identical code are byte-identical in every file:

```
runner.py  241 stmts, 23 miss, 15 brpart, 87%
missing    54, 162, 178, 210, 229->232, 253, 316, 344, 367-370, 377,
           379-380, 383-384, 390-392, 401-402, 407-408, 433, 459
TOTAL      1033 / 132 / 304 / 70 -> 84%
```

Identical missing lines is the real test, not the percentage. Before the fix, runner.py ranged 49-75% and TOTAL 75-81% across eight nightlies of unchanged main.

The process request: could you do one pass on the shape rather than line by line on the next iteration? Four rounds of line-level findings on instrumentation has cost more than the fix did, and I would rather get your read on whether the design is right than keep hardening a diagnostic layer that I have now mostly deleted.

Two things still open and not blocking: the stale-view sweeper failure at test_mcp_server.py:461 is unrelated to coverage (three local failures, zero in CI) and wants its own investigation; and my last seven local suite runs all died on infrastructure rather than on this code, so I am leaning on CI for validation here.

tests/integration_mcp/conftest.py
```diff
@@ -460,18 +466,58 @@ def collect_remote_coverage(app_name: str, since: float, tmp_schema: str = "") -
...
+    return usable
+
+
+def _partition_readable(data_files: list[str]) -> tuple[list[str], list[str]]:
+    """Split *data_files* into those ``coverage`` can read and those it cannot."""
+    usable: list[str] = []
+    unusable: list[str] = []
+    for path in data_files:
+        try:
+            coverage.CoverageData(basename=path).read()
```
mwojtyczka, line 515: **Per-file readability guard misses the cross-file incompatibility it's meant to catch.** `_partition_readable` reads each file in isolation (`CoverageData(basename=path).read()`), which only detects structurally-torn SQLite. `coverage combine` *additionally* hard-fails with `DataError` when merging **branch-mode data with line-mode data** (per the shipped coveragerc comment). If a downloaded file was produced with a different branch setting, every per-file `read()` passes, `combine` raises, and the CI merge step runs it under `|| true` — silently zeroing the whole report while jobs stay green. That's exactly the failure the guard's docstring claims to close, only partially closed. Consider a trial `coverage combine` on a copy (or partitioning by branch/line mode) rather than per-file reads.

mcp-server/tests/test_bundle_targets.py
```diff
@@ -245,3 +245,77 @@ def test_no_literal_dqx_version_can_drift_from_dqx_version_var(self):
...
+    """The app has no wheel entry point to wrap, so it is the one runtime that still needs the thread.
+
+    The bootstrap defaults the interval to 0 for the runner's sake; if the app's explicit value were
+    dropped, the app would fall back to atexit alone and lose coverage whenever the platform's
+    shutdown budget is exceeded.
+    """
+    bootstrap = _BOOTSTRAP_INIT.read_text(encoding="utf-8")
+    assert (
+        'os.getenv("DQX_COVERAGE_CHECKPOINT_SECONDS", "")' in bootstrap
+        and "return float(raw) if raw else 0.0" in bootstrap
```
mwojtyczka, line 317: **Guard test asserts on an exact source substring rather than behavior.** `assert "return float(raw) if raw else 0.0" in bootstrap` matches the literal source of `_checkpoint_seconds`; a behavior-preserving rewrite (e.g. `0.0 if not raw else float(raw)`) breaks the test though the default-off behavior is unchanged. Per the repo's testing rule ("test behaviour, not implementation details"), exercise it by calling `_checkpoint_seconds()` with and without the env var set and asserting the returned value, instead of grepping the source.

tests/integration_mcp/coverage_bootstrap/src/dqx_mcp_coverage_bootstrap/__init__.py
```diff
@@ -77,99 +140,258 @@ def _destination_dir() -> str:
...
+            _flush(final=False)
+        except Exception as exc:  # noqa: BLE001 — the thread must outlive any single bad tick
+            # Without this, one escaped exception ends the thread for the life of the process and
+            # silently reinstates the atexit-only failure this whole mechanism exists to avoid.
+            _log(f"checkpoint tick failed (non-fatal): {exc!r}", warning=True)
+        finally:
+            _upload_lock.release()
+
+
+def _final_flush() -> None:
```
mwojtyczka, line 316: **The app still delivers coverage via the atexit/SIGTERM/checkpoint path this PR argues is non-deterministic.** The runner got the deep fix (a wrapping entry point) precisely because a timer/atexit can't deliver reliably — but the app keeps that mechanism: if uvicorn's ~15s SIGTERM budget is exceeded, `_final_flush` falls back to the sibling `.final` write and the report reflects at best the last checkpoint. So the guarantee is asymmetric across runtimes. A symmetric wrapper (or a flush endpoint the suite invokes before teardown) would generalize the deterministic path rather than leaving the app on the fragile one. Not blocking — worth a follow-up.

tests/integration_mcp/coverage_bootstrap/src/dqx_mcp_coverage_bootstrap/__init__.py
```diff
@@ -77,99 +140,258 @@ def _destination_dir() -> str:
...
-    if _cov is None:
-        return
+    directory = os.path.dirname(remote_path)
+    try:
+        if not _is_app():
+            try:
+                os.makedirs(directory, exist_ok=True)
+                with open(remote_path, "wb") as handle:
+                    handle.write(payload)
+                _log(f"copied {len(payload)} bytes -> {remote_path}")
```
mwojtyczka, line 202: **Per-checkpoint log spam over the app's whole lifetime.** With `DQX_COVERAGE_CHECKPOINT_SECONDS=15`, this emits `[dqx-coverage] copied … bytes` to stderr plus an INFO record every 15s for the entire process life — hundreds of lines over a test run, unbounded in a long-lived deploy, and on the flush hot path. An "announce once per destination" line (like the one used elsewhere in this file) would carry the same information without the steady volume.

mwojtyczka (commented):
Second automated pass on the updated PR (head a77439b0). No hard correctness bugs this round — the `_upload_lock`→`_lock` ordering, concurrent same-object write serialization, the `parallel=false` guarantee, and the DAB `task_key` merge all check out, and the earlier substantive items look addressed. Four low-severity quality/robustness gaps below.

mwojtyczka (approved):
LGTM - left 4 nice to have comments but not bugs

PR #1500, opened 2026-09-02
feat(app): add Marketplace installation readiness and guided setup

### LINKED ISSUE #1108: [FEATURE]: Publish DQX App in the Databricks Marketplace

### Problem statement

DQX app can be installed using DAB

### Proposed Solution

Put the app on the Marketplace to enable no-code, one click deployment

### Additional Context

_No response_

laurencewells:
## Changes

- Package DQX Studio for Databricks Marketplace distribution.
- Add a guided setup wizard with live readiness checks, remediation guidance, progress visibility, and accessible resource descriptions.
- Reconcile required Unity Catalog schemas, wheels volume, Lakebase, SQL warehouse, task-runner job, application wheels, and migrations.
- Support a separate task-runner service principal while preserving existing `IS_OWNER` or `CAN_MANAGE` access.
- Add an executable shell workflow for Marketplace tag builds.
- Generate and force-stage the self-contained Marketplace artifact only on signed release branches; keep generated backend, frontend, lockfile, and wheel output off `main`.
- Keep dependency builds frozen and compatible with configured package indexes without modifying the root or Studio lockfiles.
- Expand deployment, development, and Marketplace documentation.

### BREAKING CHANGES

- When `DQX_ADMIN_GROUP` is unset, DQX Studio now bootstraps members of the workspace `admins` group with the in-app `ADMIN` role so Marketplace first-run setup always has an authorized administrator.
- DQX Studio now requires Lakebase for OLTP/application state. Delta-backed OLTP state is no longer supported, and existing Delta OLTP state has no automated migration path.

### Linked issues

Resolves #1108

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

Validated with 2,485 root unit tests, 3,978 Studio backend tests, 825 UI tests, 24 Marketplace tests, and clean TypeScript/Python type checks. A clean DAB deployment was also tested manually.

Marketplace dependency locking was validated using the same configured-index environment as `release_marketplace.sh`. The workflow stages the verified locked LiteLLM wheel while keeping both repository lockfiles unchanged.

### Documentation and Demos

- [ ] added/updated demos
- [x] added/updated docs


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 0.16.0
 
+* DQX Studio is now packaged for Databricks Marketplace with a readiness wizard for bound SQL warehouse, Lakebase, and Unity Catalog volume resources. Main excludes compiled Marketplace frontend assets; maintainers generate the complete artifact from an annotated signed version tag on a locally signed `marketplace/vX.Y.Z` branch, then inspect and push it manually. Marketplace uses published DQX Core pins, while DAB remains the one-command deployment path and continues to consume `.build/`.
```
mwojtyczka: let's remove it, we will update this when we add new version

CHANGELOG.md
```diff
@@ -111,6 +112,7 @@ BREAKING CHANGES!
 
 BREAKING CHANGES!
 
+* DQX Studio no longer supports Delta-backed OLTP/application state. Lakebase is required and existing Delta OLTP state has no migration path.
```
mwojtyczka: same as above, add BREAKING CHANGES to the PR description

app/src/databricks_labs_dqx_app/backend/setup/orchestrator.py
```diff
@@ -0,0 +1,383 @@
...
+        try:
+            await asyncio.to_thread(self.jobs.grant_setup_admin, self.runtime.require_job_id(), setup_user)
+        except Exception:
+            task_step = _failed(
+                SetupStepId.TASK_RUNNER,
+                "setup_admin_grant_failed",
+                "Could not grant the setup administrator access to the Studio task-runner job.",
+            )
+            steps = [task_step if step.id == SetupStepId.TASK_RUNNER else step for step in report.steps]
+            return self._publish_stopped(steps, SetupStepId.TASK_RUNNER)
```
mwojtyczka: **Availability bug: a failed permission grant takes a healthy app offline.**

When a deployment is already `READY`, `_reconcile_ready` still runs `grant_setup_admin`. If that call throws (e.g. a transient Databricks API error), this except branch calls `_publish_stopped(..., TASK_RUNNER)`, which publishes `state=SETUP_REQUIRED`. `SetupGateMiddleware` then returns 503 for every DB-backed API for **all** users until a later reconcile succeeds.

The reconcile endpoint is always allowed by the gate and is triggered whenever a second admin visits the setup page, so a healthy READY deployment can be knocked offline by a failed admin grant on an already-complete setup. A permission-grant failure on an already-READY app should be surfaced as a non-fatal warning, not downgrade the whole app to `SETUP_REQUIRED`.

app/src/databricks_labs_dqx_app/backend/setup/orchestrator.py
```diff
@@ -0,0 +1,383 @@
...
+        code=code,
+        summary=summary,
+        actions=(SetupActionId.RECONCILE,),
+    )
+
+
+def _sanitize_identity(value: str | None) -> str | None:
+    if value is None:
+        return None
+    sanitized = "".join(" " if unicodedata.category(character) == "Cc" else character for character in value)
```
mwojtyczka: **Duplicated control-character sanitization idiom (4+ copies).**

This `"".join(" " if unicodedata.category(character) == "Cc" else character for character in value)` expression is copy-pasted across `orchestrator.py:382` (`_sanitize_identity`), `services/app_settings_service.py:1087` (`_sanitize_audit_identity`), `dependencies.py:1189` (`sanitize_setup_display`), and a near-variant in `setup/checks.py:305/315`; `profiling_rule_builder.py:203` adds yet another variant. Per CLAUDE.md ("extract shared utilities when the same logic appears in 3+ places"), this warrants a single shared helper — the variants are already drifting (some check `category=='Cc'`, others compare against `' '`).

app/marketplace/src/databricks_labs_dqx_app/backend/config.py
```diff
@@ -0,0 +1,193 @@
...
+    # Hard cap on tokens generated per LLM call. Bounds cost/latency and
+    # mitigates LLM denial-of-service from pathological prompts (OWASP LLM04);
+    # rule-generation responses are small JSON payloads, so this is generous.
+    llm_max_tokens: int = Field(
+        default=4096,
+        validation_alias="DQX_LLM_MAX_TOKENS",
+        gt=0,
+        description="Maximum output tokens per ChatDatabricks call (LLM budget cap).",
+    )
+    admin_group: str = Field(
```
mwojtyczka: **Privilege-default change to confirm: `admin_group` now defaults to `"admins"`.**

The default changed from `None` to `"admins"`. `role_service.resolve_role` grants `UserRole.ADMIN` whenever `admin_group` is truthy and present in the user's groups, so any deployment that doesn't set `DQX_ADMIN_GROUP` now silently grants full app ADMIN to every member of the workspace `admins` group — and there's no longer a way to express "no bootstrap admin group" (previously `None`). Please confirm this privilege expansion is intended; if so, worth calling out in the release notes.

(Applies to both the `app/src/...` and `app/marketplace/...` copies of `config.py`.)

mwojtyczka (commented):
Automated code-review pass on the setup module and refactored routes. Four findings below — one availability bug worth fixing before merge, one privilege-default change to confirm, and two cleanup items. Note: `config.py` and `dryrun.py` are duplicated across the `app/src/...` and `app/marketplace/...` trees; comments are anchored to whichever copy GitHub exposes a diff for, but each applies to both copies.

app/marketplace/src/databricks_labs_dqx_app/backend/routes/v1/dryrun.py
```diff
@@ -0,0 +1,769 @@
...
+
+router = APIRouter()
+
+_DRYRUN_TABLE = "dq_validation_runs"
+_SQL_CHECK_PREFIX = "__sql_check__/"
+_ALL_ROLES = [UserRole.ADMIN, UserRole.RULE_APPROVER, UserRole.RULE_AUTHOR, UserRole.VIEWER]
+_NON_VIEWERS = [UserRole.ADMIN, UserRole.RULE_APPROVER, UserRole.RULE_AUTHOR]
+
+
+def _run_table_fqn() -> str:
```
mwojtyczka: **Redundant FQN re-derivation.**

`_run_table_fqn()` re-derives `resources.volume.catalog/schema` (and re-fetches `rt.require_resources()`), but the injected `sql` executor is already constructed with `catalog=resources.volume.catalog` / `schema=resources.volume.schema` (`dependencies.py`), so `sql.fqn("dq_validation_runs")` is equivalent. It's also used inconsistently — `list_validation_runs` uses `sql.fqn` while `submit_dry_run`/`get_dry_run_results` use `_run_table_fqn` for the same table. Prefer the standardized `sql.fqn()` helper throughout and drop the private one.

(Applies to both the `app/src/...` and `app/marketplace/...` copies of `dryrun.py`.)

CHANGELOG.md
laurencewells: Removed the unreleased CHANGELOG entries; CHANGELOG.md is no longer part of the PR diff.

CHANGELOG.md
laurencewells: Removed the CHANGELOG entries and kept the breaking changes in the PR description.

app/src/databricks_labs_dqx_app/backend/setup/orchestrator.py
laurencewells: Fixed in 9bb30637. A failed setup-admin grant refresh now emits a non-fatal warning and preserves the existing READY report. Added a regression test covering the transient failure.

app/src/databricks_labs_dqx_app/backend/setup/orchestrator.py
laurencewells: Fixed in 9bb30637. Extracted replace_control_characters() and reused it across setup display values, audit identities, setup checks, orchestration, and profiler rule names, including C0/C1 coverage.

app/marketplace/src/databricks_labs_dqx_app/backend/config.py
laurencewells: Confirmed intentional. Marketplace first-run setup needs a deterministic bootstrap administrator, and workspace admins already hold workspace-wide authority. The privilege-default change is now called out explicitly under BREAKING CHANGES in the PR description.

app/marketplace/src/databricks_labs_dqx_app/backend/routes/v1/dryrun.py
laurencewells: Fixed in 9bb30637. Removed _run_table_fqn() and standardized all validation-run table resolution on the injected SqlExecutor.fqn() helper.

app/src/databricks_labs_dqx_app/backend/setup/job_manager.py
```diff
@@ -0,0 +1,179 @@
...
+            str(job_id),
+            access_control_list=[_setup_admin_access(user_name)],
+        )
+
+    def validate_run_as(self, job_id: int, app_sp_id: str) -> SetupStep:
+        """Report whether a task-runner job uses a distinct service principal."""
+        job = self.workspace.jobs.get(job_id)
+        run_as = getattr(getattr(job, "settings", None), "run_as", None)
+        service_principal = getattr(run_as, "service_principal_name", None)
+        if service_principal and service_principal != app_sp_id:
```
mwojtyczka, line 87: **`run_as` gate is over-strict — first-run setup can never reach READY on its own.**

This requires the task-runner `run_as` to be a service principal *distinct* from the app SP (`service_principal != app_sp_id` → PASSED, everything else → ACTION_REQUIRED below). But when the app creates the job itself (`resolve()` on a fresh Marketplace install), the job's `run_as` defaults to the creating app SP — so `validate_run_as` returns ACTION_REQUIRED (`task_runner_run_as_matches_app`) and setup halts at TASK_RUNNER forever unless an operator manually reassigns `run_as`. That contradicts the PR's stated intent of *supporting* (not *requiring*) a separate SP. Consider treating `run_as == app SP` as an acceptable default rather than an error.

mwojtyczka (commented):
Second automated code-review pass, focused on the new guided-setup flow (`backend/setup/*`, `SetupGate.tsx`). Seven findings below. The first three cluster into a first-run failure mode worth addressing before merge; the rest are lower-severity cleanup/robustness notes. (Comments anchored to the `app/src/...` copies; the same code is mirrored under `app/marketplace/...`.)

app/src/databricks_labs_dqx_app/backend/setup/job_manager.py
```diff
@@ -0,0 +1,179 @@
...
+            return SetupStep(
+                id=SetupStepId.TASK_RUNNER,
+                state=StepState.PASSED,
+                code="task_runner_run_as_ready",
+                summary="The task-runner job uses a separate service principal.",
+            )
+        return SetupStep(
```
mwojtyczka, line 94: **ACTION_REQUIRED step carries no actions/instructions — dead-end in the wizard.**

This `SetupStep` sets only `id`/`state`/`code`/`summary` — no `actions`. `SetupWizard`'s `StepActions` only renders backend-advertised actions, and polling is off at `setup_required`, so the admin sees a card telling them to reassign `run_as` but no VERIFY_AGAIN/RECONCILE button. After fixing `run_as` via the Jobs UI they must manually reload the page for setup to advance. Attach a re-verify action (or a global reconcile affordance) to this step.

app/src/databricks_labs_dqx_app/ui/components/SetupGate.tsx
```diff
@@ -0,0 +1,98 @@
...
+  children: React.ReactNode;
+};
+
+const SETUP_POLL_INTERVAL_MS = 2_000;
+
+export function setupPollingInterval(
+  state: "checking" | "initializing" | "setup_required" | "ready" | undefined,
+  isReconciling = false,
+): number | false {
+  return isReconciling || state === "checking" || state === "initializing"
```
mwojtyczka: **Polling stops for waiting viewers — they never auto-advance when setup completes elsewhere.**

`setupPollingInterval` returns `false` for any state that isn't `checking`/`initializing` (unless the local client is reconciling). So a waiting non-admin (`state='setup_required'`, `isReconciling=false`) has `refetchInterval=false` and never re-queries `getSetupStatus`. When an admin completes reconcile on another session and the app goes READY, the waiting viewer stays stuck on the waiting screen until a manual browser reload. Consider polling (perhaps at a slower cadence) while `state === 'setup_required'` too.

app/src/databricks_labs_dqx_app/backend/setup/orchestrator.py
```diff
@@ -0,0 +1,383 @@
...
+        self.checkers = checkers
+        self.jobs = jobs
+        self.pg_migrations = pg_migrations
+        self.delta_migrations = delta_migrations
+        self.app_settings = app_settings
+        self.publish_wheels = publish_wheels
+        self.activation = activation
+        self.app_sp_id = _sanitize_identity(app_sp_id) or ""
+
+    async def inspect(self) -> SetupReport:
```
mwojtyczka: **Dead code: `inspect()` / `_inspect_locked()` has no production caller.**

`start_studio` calls only `reconcile()`, the `/status` route reads `setup_runtime.report()`, and `/reconcile` calls `reconcile()`. Nothing in production invokes `inspect()` — only unit tests do. This ~55-line non-mutating capability-scan path is maintained solely for its own tests and will drift from the `reconcile()` ordering it mirrors. Either wire it to a real caller or drop it (and its tests).

app/src/databricks_labs_dqx_app/backend/setup/checks.py
```diff
@@ -0,0 +1,315 @@
...
+            privilege = effective_privilege.privilege
+            value = getattr(privilege, "value", privilege)
+            if isinstance(value, str):
+                privileges.add(value)
+    return frozenset(privileges)
+
+
+def _missing_privileges(response: EffectivePermissionsList, required: frozenset[str]) -> frozenset[str]:
+    privileges = _privileges(response)
+    if "ALL_PRIVILEGES" in privileges:
```
mwojtyczka, line 286: **Owner-implicit privileges not recognized — an owning app SP can be falsely reported as missing privileges.**

`_missing_privileges` only treats `ALL_PRIVILEGES` as blanket-sufficient. If a deployment provisions the wheels volume/schema/catalog with the app SP as **owner** (implicit full rights) but without explicit grants of the enumerated privileges, `grants.get_effective` may enumerate neither the owner-implicit privileges nor `ALL_PRIVILEGES`, so this returns a non-empty set and `check_volume`/`check_unity_catalog` spuriously report ACTION_REQUIRED — blocking setup with grant instructions that are redundant for an owner. Consider also treating ownership as sufficient.

app/src/databricks_labs_dqx_app/backend/services/app_settings_service.py
```diff
@@ -133,6 +137,34 @@ def save_setting(self, key: str, value: str, *, user_email: str | None = None) -
...
+            job_id: Setup-resolved task-runner job identifier.
+            completed_at: Setup completion timestamp.
+            user_name: Optional setup administrator identity.
+        """
+        if job_id <= 0:
+            raise ValueError("The setup task-runner job ID must be positive.")
+        timestamp = completed_at.astimezone(timezone.utc).isoformat()
+        actor = _sanitize_audit_identity(user_name)
+        for key, value in (
+            (_SETUP_JOB_ID_KEY, str(job_id)),
```
mwojtyczka, line 162: **Persisted `setup_task_runner_job_id` is written but never read back.**

`record_setup_completion` persists `_SETUP_JOB_ID_KEY`, but restart-time job resolution goes through `conf.job_id` (env) + the managed-tag search in `_reconcile_job` — this stored value is never consulted. It's effectively dead/informational state, and recovery silently depends on the job's managed tags staying intact. Either use the persisted id as the primary resolution hint on restart, or drop the write.

app/src/databricks_labs_dqx_app/backend/startup.py
```diff
@@ -0,0 +1,680 @@
...
+    FAILING_ROWS_VIEW_NAME,
+)
+
+
+def _grant_user_view_access(delta_sql: SqlExecutor, resources: ActiveResources) -> None:
+    catalog = delta_sql.q(resources.volume.catalog)
+    schema = delta_sql.q(resources.genie_schema)
+    statements = [
+        f"GRANT USE SCHEMA ON SCHEMA {catalog}.{schema} TO `account users`",
+        *(f"GRANT SELECT ON TABLE {catalog}.{schema}.{name} TO `account users`" for name in _USER_READABLE_VIEWS),
```
mwojtyczka, line 485: **`_grant_user_view_access` assumes every readable view lives in the genie schema.**

This builds `catalog.genie_schema.<name>` for every entry in `_USER_READABLE_VIEWS`, including `FAILING_ROWS_VIEW_NAME` (imported from `entitlement_service`). If that view is actually created in the main schema (`volume.schema`) rather than `genie_schema`, the `GRANT SELECT` targets a non-existent FQN and fails — and since the failure is only logged as a warning, account users silently never receive access to the failing-rows view. Resolve each view's actual schema rather than assuming genie for all.

app/src/databricks_labs_dqx_app/backend/setup/job_manager.py
laurencewells: A distinct task-runner service principal is a strict requirement, not an optional enhancement. The app-scoped SP cannot be used outside the Apps framework, while the task-runner job executes externally. Setup therefore intentionally pauses until an administrator assigns that external SP. This is documented in `app/DEPLOYMENT.md` under the task-runner service-principal prerequisite.

app/src/databricks_labs_dqx_app/backend/services/app_settings_service.py
laurencewells: `setup_task_runner_job_id` is intentionally persisted as setup-completion/audit evidence alongside `setup_completed_at` and `setup_completed_by`; it is not a runtime resolution hint. Runtime recovery deliberately uses an explicitly configured ID or the managed ownership tags, rather than trusting a potentially stale persisted job ID.

app/src/databricks_labs_dqx_app/backend/startup.py
laurencewells: `v_dq_failing_rows` is created in the Genie schema. `EntitlementService.failing_rows_view_fqn_quoted` uses `_genie_schema_q`, and startup constructs that service with `resources.genie_schema`. `test_grants_use_genie_schema_plus_select_on_the_five_views` also explicitly asserts that all five user-readable views, including `v_dq_failing_rows`, are granted from the Genie schema.

app/src/databricks_labs_dqx_app/backend/setup/job_manager.py
laurencewells: Fixed in `34984a01`. ACTION_REQUIRED task-runner steps now advertise `VERIFY_AGAIN`, so the administrator can re-run reconciliation after assigning the external `run_as` identity.

app/src/databricks_labs_dqx_app/ui/components/SetupGate.tsx
laurencewells: Fixed in `34984a01`. Clients now poll every 10 seconds while setup is required, while retaining the 2-second interval during active checking/reconciliation.

app/src/databricks_labs_dqx_app/backend/setup/orchestrator.py
laurencewells: Fixed in `34984a01`. Removed the unused `inspect()` / `_inspect_locked()` path and its tests.

app/src/databricks_labs_dqx_app/backend/setup/checks.py
laurencewells: Fixed in `34984a01`. Setup now checks the catalog, schema, or volume owner when explicit effective privileges are incomplete and accepts ownership by the app SP as sufficient.

app/src/databricks_labs_dqx_app/backend/routes/v1/compute.py
```diff
@@ -180,17 +184,31 @@ def get_compute_settings(
...
     """Update one or both compute settings (admin only)."""
     if body.sql_warehouse_id is None and body.jobs_compute is None:
         raise HTTPException(
             status_code=400, detail="At least one of sql_warehouse_id or jobs_compute must be provided."
         )
     if body.sql_warehouse_id is not None:
+        warehouse_id = body.sql_warehouse_id.strip()
+        if warehouse_id:
+            step = await asyncio.to_thread(orchestrator.checkers.check_warehouse, warehouse_id, reader_ws=obo_ws)
+            if step.state != StepState.PASSED:
```
mwojtyczka: **Valid warehouse rejected with 409 when its ACL is unreadable.**

When neither the OBO client nor the SP client can read a warehouse's access-control list, `warehouse_access_status` returns `'unknown'`. With a non-`None` `warehouse_id`, `check_warehouse` skips the `SELECT 1` connectivity fallback (that fallback only runs when `warehouse_id is None`) and returns `ACTION_REQUIRED` / `warehouse_permission_unknown`, so this block raises 409 and the admin cannot save a warehouse the app SP actually has `CAN_USE` on.

Consider treating `'unknown'` as non-fatal (or running the `SELECT 1` probe even when a `warehouse_id` is supplied) so a working-but-unreadable warehouse can still be saved.

app/src/databricks_labs_dqx_app/backend/setup/resources.py
```diff
@@ -0,0 +1,159 @@
...
+            schema=schema,
+        )
+
+    host = environ.get("PGHOST", "").strip()
+    if not host:
+        return None
+    _reject_control_characters(host, "Lakebase host")
+    port = _parse_port(environ.get("PGPORT", "5432"))
+    database = _required_value(
+        environ.get("PGDATABASE", config.lakebase_database_name),
```
mwojtyczka: **Empty `PGDATABASE` discards the configured fallback.**

`environ.get("PGDATABASE", config.lakebase_database_name)` returns the *present-but-empty* value `""` rather than the default when the platform binding sets `PGDATABASE=""`. `_required_value("")` then raises `InvalidParameterError`, `_resolve_resources` catches it, and setup publishes `SETUP_REQUIRED` / `lakebase_binding_missing` permanently — even though a valid `lakebase_database_name` is configured.

Coalesce empties before defaulting, e.g. `environ.get("PGDATABASE") or config.lakebase_database_name`.

app/src/databricks_labs_dqx_app/backend/dependencies.py
```diff
@@ -1102,6 +1123,73 @@ async def _check(role: Annotated[UserRole, Depends(get_user_role)]) -> UserRole:
...
+    Returns:
+        Sanitized user identity and whether they are a member of the configured group.
+    """
+    configured_group = sanitize_setup_display(admin_group)
+    groups = {
+        display for group in (user.groups or []) if (display := sanitize_setup_display(group.display)) is not None
+    }
+    return SetupAccess(
+        user_name=sanitize_setup_display(user.user_name) or "unknown",
+        can_manage=configured_group is not None and configured_group in groups,
```
mwojtyczka, line 1151: **Empty/whitespace admin group permanently locks everyone out of setup.**

If `DQX_ADMIN_GROUP` (→ `config.admin_group`) is set to `""`/whitespace, `sanitize_setup_display` returns `None`, so `configured_group` is `None` and `can_manage` is always `False`. The same happens if `me()` returns groups without `display` populated. Since `/api/v1/setup/reconcile` is the only path to `READY` and is guarded by `require_setup_admin` (`can_manage`), `SetupGate` then 503s all API traffic forever with no way to complete setup.

Worth a guard: if `configured_group` is `None`, surface a clear misconfiguration error rather than silently denying every user.

app/src/databricks_labs_dqx_app/backend/dependencies.py
```diff
@@ -1102,6 +1123,73 @@ async def _check(role: Annotated[UserRole, Depends(get_user_role)]) -> UserRole:
...
+    )
+
+
+async def get_setup_access(
+    obo_ws: Annotated[WorkspaceClient, Depends(get_obo_ws)],
+    config: Annotated[AppConfig, Depends(get_conf)],
+) -> SetupAccess:
+    """Resolve setup access from exactly one caller OBO SCIM lookup."""
+    user = await asyncio.to_thread(obo_ws.current_user.me)
```
mwojtyczka, line 1180: **Fresh SCIM `me()` on every setup-status poll.**

`SetupGate` polls `getSetupStatus` while checking/initializing/reconciling, and each poll flows through `get_setup_access` → `obo_ws.current_user.me()` over the network. Over a multi-minute reconcile that's dozens of redundant SCIM lookups per client (group membership doesn't change mid-setup) — avoidable workspace API load / rate-limit pressure.

Consider caching the resolved `SetupAccess` for the poll interval.

app/src/databricks_labs_dqx_app/ui/components/setup/SetupWizard.tsx
```diff
@@ -0,0 +1,306 @@
...
+            {isWaiting
+              ? t("setup.waitingTitle")
+              : view.kind === "checking"
+                ? t("setup.checkingTitle")
+                : t("setup.title")}
+          </h1>
+          <p className="max-w-2xl text-sm leading-6 text-muted-foreground">
+            {isWaiting
+              ? t("setup.waitingDescription", { adminGroup: view.adminGroup })
+              : t("setup.checkingDescription")}
```
mwojtyczka: **Wizard shows "checking" copy to admins mid-remediation.**

For the `setup_required` + `canManage` wizard, `isWaiting` is false and `view.kind` is not `"checking"`, so this description ternary falls through to `t("setup.checkingDescription")`. The heading correctly shows `t("setup.title")` and the remediation step cards render, but the sub-text tells the admin the system is "checking" instead of explaining the required actions.

app/src/databricks_labs_dqx_app/backend/setup/orchestrator.py
```diff
@@ -0,0 +1,323 @@
...
+                    "studio_activation_failed",
+                    "Could not activate Studio background services.",
+                )
+                return self._publish_stopped([*steps, activation_step], SetupStepId.ACTIVATION)
+
+            steps.append(_passed(SetupStepId.ACTIVATION, "DQX Studio is active."))
+            report = SetupReport(state=SetupState.READY, steps=tuple(steps))
+            self.runtime.publish(report)
+            try:
+                await self.activation.start_background()
```
mwojtyczka, line 207: **`start_background()` failure re-gates a fully-activated app.** (distinct from the already-resolved permission-grant availability bug — this is the post-activation background path.)

By this point `activate()` has registered the OLTP executor, activated the runtime, run foreground hooks, and published `READY`. If `start_background()` raises, `reconcile` returns `_publish_stopped([...], ACTIVATION)`, flipping state back to `SETUP_REQUIRED`; `SetupGateMiddleware` then 503s all non-allowlisted API traffic even though the app is actually functional.

Mitigated in practice because `_start_scheduler` / `_maybe_start_ai_bootstrap` wrap their bodies in try/except, but a raise here still shouldn't tear down an already-READY app — consider keeping `READY` and surfacing background-start failure as a non-fatal warning.

mwojtyczka (commented):
Follow-up code review (setup/startup rewiring + marketplace repackaging). 6 findings on the new setup logic; none are definitive crashes. All previous review threads either resolved or replied-to; these are new items.

mwojtyczka (approved):
LGTM - isaac came back with a few more comments, worth checking before we merge

app/src/databricks_labs_dqx_app/backend/routes/v1/compute.py
laurencewells: Fixed in 5ba5cf12. We now allow warehouse_permission_unknown when saving a replacement warehouse, while still rejecting actual validation failures. Added a regression test for this case.

app/src/databricks_labs_dqx_app/backend/setup/resources.py
laurencewells: Fixed in 5ba5cf12. An empty PGDATABASE now falls back to the configured lakebase_database_name. Added a regression test.

app/src/databricks_labs_dqx_app/backend/dependencies.py
laurencewells: Fixed in 5ba5cf12. AppConfig now rejects an empty or whitespace-only DQX_ADMIN_GROUP during validation, so this fails clearly instead of locking everyone out. Added a regression test.

app/src/databricks_labs_dqx_app/backend/dependencies.py
laurencewells: Fixed in 5ba5cf12. Setup status now uses a 10-second cache keyed by the token and admin group. Reconcile still does a fresh SCIM lookup, so removing someone from the admin group takes effect immediately. Tests cover both paths.

app/src/databricks_labs_dqx_app/ui/components/setup/SetupWizard.tsx
laurencewells: Fixed in 5ba5cf12. The admin remediation view now shows setup guidance. The checking copy is only used while setup is actually being checked. Added UI coverage.

app/src/databricks_labs_dqx_app/backend/setup/orchestrator.py
laurencewells: Fixed in 5ba5cf12. A start_background failure after activation is now logged without changing the app back from READY to SETUP_REQUIRED. Added a regression test.

PR #1503, opened 2026-09-03
Cap Python at <3.13 across core, app, and mcp-server; modernize litellm/ruff pins

mwojtyczka:
Cap the supported Python version at 3.12 across DQX core and both sub-projects, and refresh their lockfiles. Previously core declared `requires-python = ">=3.10"` with no upper bound, so untested interpreters (3.13+) were implicitly advertised as supported.

## Changes

**Core** (`pyproject.toml`, `uv.lock`, `.build-constraints.txt`)
- `requires-python`: `">=3.10"` → `">=3.10,<3.13"` (classifiers already stop at 3.12). `<3.13` not `<=3.12` so 3.12 patch releases stay in range.
- Modernized two pins that were no longer served by the package index, blocking `make lock-dependencies`:
  - `litellm`: `<=1.82.6` → `<2.0` (resolves 1.98.0; floor still satisfies `dspy~=3.1.3`; ceiling is a range, not an exact pin, since this is a distributed library — the lock pins 1.98.0 exactly).
  - `ruff`: `~=0.3.4` → `~=0.12.0` (resolves 0.12.12; lint-only, zero new findings. 0.13+ expands the default rule set (~1.6k findings) — deferred to a separate lint-modernization effort).

**App** (`app/pyproject.toml`, `app/uv.lock`)
- `requires-python`: `">=3.12"` → `">=3.12,<3.13"`. Lock regenerated; also picks up litellm 1.98.0 from core’s `[llm]` extra (the app depends on `databricks-labs-dqx[llm,datacontract]` from the checkout and was still pinning the unavailable 1.82.6).

**mcp-server** (`mcp-server/pyproject.toml`, `mcp-server/uv.lock`)
- `requires-python`: `">=3.11"` → `">=3.11,<3.13"` (3.11 and 3.12 both supported). Lock regenerated.

## Validation
- Core: `make lint` (black ✓, ruff ✓, mypy ✓, pylint 10.00/10), `make test` — 2495 passed
- App backend: `make app-test` — 3885 passed
- mcp-server: `make mcp-test` — 207 passed

## Notes
- `app/yarn.lock` is intentionally untouched. `make lock-app-dependencies` runs `yarn install` first, which currently fails on a pre-existing unresolvable npm dep (`use-sync-external-store@^1.6.0`) — unrelated to the Python version cap. Only `app/uv.lock` (Python) needed updating here; the JS lockfile refresh is a separate issue.


laurencewells (approved):
LGTM

PR #1505, opened 2026-09-04
Fix filtered duplicate dataset pairing

### LINKED ISSUE #1504: Filtered compare_datasets mispairs duplicate keys

- [x] I have searched the existing issues and pull requests, including closed items.

### Current Behavior

`compare_datasets` can report an in-scope source row as extra even when it exactly matches the reference. This happens when the source contains an out-of-scope row with the same matching key that sorts before the in-scope row.

The duplicate-key pairing assigns `row_number()` across all source rows before the check filter is compiled. The excluded row consumes the reference row's sequence number, leaving the matching in-scope row unpaired. With an error-criticality rule, `apply_checks_and_split` places that valid row in the invalid/quarantine DataFrame.

This is a regression in the lazy duplicate-key pairing added by #1488. It is distinct from the Cartesian fan-out fixed there: output cardinality remains stable, but filter scope and pairing scope disagree.

### Expected Behavior

Rows excluded by the check filter must not consume reference pairing positions. In the reproduction below, source row `(1, "B", true)` exactly matches reference row `(1, "B")` and must pass.

The fix should preserve:

- all source rows in the output, including rows outside the filter;
- lazy pairing by default;
- existing `check_missing_records` behavior for reference-only rows;
- null-safe matching and `raise_on_duplicate_keys` semantics;
- the public `compare_datasets` signature.

### Steps To Reproduce

Run on current `main` (`407ffb41eb5bfe0d09b71585e662175cd7bb0e0f`) with PySpark 4.0.0:

```python
import json

from databricks.labs.dqx.check_funcs import compare_datasets

spark = SparkSession.builder.master("local[1]").getOrCreate()
source = spark.createDataFrame(
    [(1, "A", False), (1, "B", True)],
    "id int, value string, in_scope boolean",
)
reference = spark.createDataFrame([(1, "B")], "id int, value string")

condition, apply = compare_datasets(
    columns=["id"],
    ref_columns=["id"],
    ref_df_name="ref",
    row_filter="in_scope",
)

actual = (
    apply(source, spark, {"ref": reference})
    .select("id", "value", "in_scope", condition.alias("violation"))
    .orderBy("value")
    .collect()
)
print(json.dumps([row.asDict(recursive=True) for row in actual], sort_keys=True))
```

The reproduction was run twice with identical output and exit status 0.

Control: removing the excluded `(1, "A", false)` row makes `(1, "B", true)` pass with `violation = null`.

### Cloud

Cloud-independent. Reproduced with local Spark; no workspace service was involved.

### Operating System

macOS

```shell
[{"id": 1, "in_scope": false, "value": "A", "violation": null}, {"id": 1, "in_scope": true, "value": "B", "violation": "{\"row_missing\":false,\"row_extra\":true,\"changed\":{\"value\":{\"df\":\"B\"}}}"}]
```

Engine-level confirmation:

```shell
GOOD=[{"id": 1, "in_scope": false, "value": "A"}]
BAD=[{"id": 1, "in_scope": true, "value": "B", "_errors": [{"function": "compare_datasets", "message": "{\"row_missing\":false,\"row_extra\":true,\"changed\":{\"value\":{\"df\":\"B\"}}}"}]}]
```

### Suggested regression coverage

Add an integration test next to the duplicate-key pairing tests in `tests/integration/test_dataset_checks.py`. At minimum it should prove that an excluded duplicate cannot consume a pairing position and that the matching in-scope row has no violation. Boundary coverage should preserve behavior for `check_missing_records=True` and a null-valued filter predicate.

mattfaltyn:
## Changes

- Preserve exact-value matches before positional pairing when `compare_datasets` uses a filter.
- Pair remaining duplicate-key rows with in-scope source rows first, while retaining excluded rows in the output.
- Limit strict duplicate-key validation to source rows that are in scope; all reference rows remain in scope.
- Treat a null filter predicate like Spark filtering for present source rows while continuing to report reference-only missing rows.
- Add focused Spark regressions for lazy and strict modes, false and null predicates, exact and tolerant matches, output cardinality, and `check_missing_records=True`.

### Linked issues

Resolves #1504

### Tests

- [x] manually tested
- [ ] added unit tests: the new behavior requires Spark and is covered at integration level; the existing unit suite passes
- [x] added integration tests
- [ ] added end-to-end tests: no workflow or persistence contract changed
- [ ] added performance tests: the additional exact-match phase is limited to filtered lazy comparisons

Local validation:

- `make lint`: passed, including Black, Ruff, Mypy, and Pylint 10.00/10
- `make test`: 2,559 passed with coverage
- `make docs-build`: passed
- Standalone PySpark 4.0.0 regressions: passed for exact and tolerant matching, lazy and strict pairing, false and null filter predicates, stable cardinality, and missing-reference reporting

`make integration` and `make e2e` require Databricks workspace credentials, which are not configured locally. The non-fork PR acceptance workflows can run those suites.

### Compatibility and performance

- The public `compare_datasets` signature and result schema are unchanged.
- Unfiltered comparisons retain the existing two-window lazy pairing behavior.
- Filtered lazy comparisons add value-group counts plus exact and residual ranking so excluded rows cannot disturb available in-scope matches.
- Strict mode keeps its eager duplicate validation but ignores source rows excluded by the check filter.

### Documentation and Demos

- [ ] added/updated demos: not applicable; no usage or API changed
- [x] added/updated docs: duplicate-key filter pairing and its performance scope are documented
- [ ] added/updated agent skills: not applicable; no public API or metadata schema changed


mwojtyczka (approved):
LGTM - fixed minor docstring issues

mwojtyczka:
## Fix: filtered `compare_datasets` was splitting matched pairs into phantom rows

### Root cause

The `pairing_scope_col` (source = filter result, reference = `true`) was wired into **both** the `row_number` window partition **and the join keys**. Putting it in the join keys was the defect: an out-of-filter source row (`pairing_scope = false`) can then never join to a reference row (`pairing_scope = true`). A genuinely matched pair gets **split into two phantom rows** — a phantom "extra" on the source side and a phantom "missing" on the reference side.

That phantom missing-reference row is what broke the existing integration tests:

- `test_compare_datasets_check_missing_records_with_filter` → produced a spurious `Row(id=1, name=None)` (3 rows instead of 2)
- `test_compare_datasets_check_missing_records_with_partial_filter` → same spurious `Row(id=1, name=None)` (4 rows instead of 3)

The phantom row's warning shows as `None` only because the engine's outer rule-filter (`manager.py`, `_wrap_result`) suppresses it — but the **row itself** remains, which is what the assertions caught.

### Design conflict this surfaced

The branch's new test `test_compare_datasets_filter_preserves_missing_reference_rows` asserted the *split* behavior, which directly contradicts those integration tests **and** issue #1504's acceptance criterion to *"preserve existing `check_missing_records` behavior for reference-only rows."* In these scenarios `id=1` is **present in the source** (just filtered) — it is not reference-only — so reporting it as missing is incorrect.

### The fix

Give in-scope rows **pairing priority** through ordering, instead of *isolating* them via the join key:

- `row_number` now partitions by PK only, ordering by `pairing_scope.desc()` first (in-scope rows before excluded ones), then the compared values. This is what actually fixes #1504's duplicate-key regression: the in-scope row claims the reference slot before any excluded duplicate.
- `pairing_scope_col` is **removed from the join keys**. `_add_row_diffs` already uses the non-null `row_number` sentinel to distinguish a present-but-null-key row from an absent join side, so the scope column was never needed there.
- The `raise_on_duplicate_keys` path still validates uniqueness over in-scope rows only.

I also rewrote `test_compare_datasets_filter_preserves_missing_reference_rows` so its premise is correct: a filtered source row that matches a reference now pairs (violation suppressed, no phantom missing), while a **genuinely absent** reference key (`id=2`) is still reported as missing.

mwojtyczka:
## Follow-up: an accepted edge case in filtered duplicate-key pairing

While validating the filtered `compare_datasets` fix, I found a residual edge case in the **lazy** (`raise_on_duplicate_keys=False`) duplicate-key path. It's not fixed but it may be acceptable. 

### The edge case

When a single matching key carries **both in- and out-of-scope source rows** *and* the reference has **multiple rows with interleaving compared values**, the scope-first positional ordering can pair an in-scope row with a differently-valued reference row and report a spurious `changed` — even though a value-matching reference row exists.

Minimal repro:

```python
condition, apply = compare_datasets(
    columns=["id"], ref_columns=["id"], ref_df_name="ref_df", row_filter="in_scope",
)
df     = spark.createDataFrame([(1, "B", True), (1, "A", False)], "id int, value string, in_scope boolean")
ref_df = spark.createDataFrame([(1, "A"), (1, "B")], "id int, value string")
# in-scope (1,"B") has a value-matching reference (1,"B") but is paired with (1,"A")
# and flagged: {"row_missing": false, "row_extra": false, "changed": {"value": {"df": "B", "ref": "A"}}}
```

A fully correct result would pair `(1,"B")`↔ref`(1,"B")` and `(1,"A")`↔ref`(1,"A")` and flag nothing.

### Why it's inherent to single-pass pairing

The fix in this PR gives in-scope rows priority for reference slots (scope-first ordering), which is exactly what #1504 needs so an excluded row can't starve an in-scope row. But that same priority breaks value alignment in the case above. The two demands are irreconcilable in a single ORDER BY, because the correct choice depends on **whether the excluded row has a value-matching reference** — value ordering fixes this case but reinstates #1504; scope ordering fixes #1504 but hits this case.

### Why I'm proposing to accept it rather than fully fix it

A correct fix requires **two-phase, value-aware matching** (pair exactly-equal rows first, then pair residual rows positionally). I prototyped and validated it end-to-end — it fixes this case while preserving #1504, change detection, null-safe matching, tolerances, and `check_missing_records`. But it roughly triples the pairing code and adds several shuffles on the hot path (per-side value-group ranking + cross-side group-count exchange + residual ranking, then the join). For a narrow edge that requires duplicate keys **and** a filter **and** interleaving reference values, that cost/risk seems disproportionate.

### What's in the PR now

- The scope-first fix that resolves #1504 and preserves `check_missing_records` for reference-only rows.
- A code comment documenting this limitation on the lazy row-number ordering in `compare_datasets`.
- `test_compare_datasets_filter_duplicate_keys_interleaving_refs_known_limitation`, which pins the current behavior so any future change to the pairing heuristic is deliberate.


 @mattfaltyn let me know what do you think?

mwojtyczka:
## Follow-up: I prototyped the full fix and measured its cost — @mattfaltyn

Following on from the edge-case note above, here is the candidate correct fix for the interleaving-reference false positive and, more importantly, its performance consequences. **Recommendation: don't take it** — the cost is disproportionate to the corner it fixes. Details so you can decide.

### The candidate fix: two-phase, value-aware pairing

Replace single-pass positional pairing (lazy path only; the `raise_on_duplicate_keys=True` path is untouched) with two phases, encoded into one non-null string `pair_id` so the existing equi-join still does the work:

- **Phase 1 (exact):** rows equal on `(matching key + all compared values)` pair directly — a reference row is consumed by an equal source row before any positional guesswork.
- **Phase 2 (residual):** leftover rows pair positionally within each key, in-scope source rows first (surfacing `changed`/`missing`/`extra`).

**Correctness property:** a source row that has an available exact-value counterpart is *always* matched, so it can never be falsely flagged `changed`/`extra`; positional guessing survives only for rows with no exact counterpart (where the reported change *count* is invariant regardless of pairing). This fixes both the filtered interleaving case and the equivalent no-filter case (e.g. source `[A,A,B]` vs ref `[A,B]`, where single-pass mislabels the second `A` as changed-to-`B` and flags `B` as extra though ref `B` exists).

It is fully validated: 24 `compare_datasets` tests in `test_dataset_checks.py` + 9 in `test_apply_checks.py` pass on a live workspace, plus null-safe / tolerance / multi-column / `check_missing_records` cases; ruff, pylint (10.00), mypy clean.

### Why not take it: it taxes every lazy call

The two-phase path runs on **every** default (`raise_on_duplicate_keys=False`) call — not just when duplicates or filters or the edge shape are present. There is no cheap gate, because the lazy path exists precisely because the caller has *not* asserted unique keys (detecting duplicates to skip the work would cost the pass we'd be trying to save).

Measured on serverless, 10M rows, `noop` sink, median of 5 timed iterations after warmup:

| Scenario | Single-pass (current PR) | Two-phase | Ratio |
|---|---|---|---|
| Unique keys (common case) | ~3.0–3.5s | ~6.0s | **~2.0×** |
| Duplicate keys (~4/key) | ~2.6–2.9s | ~4.6–4.9s | **~1.7×** |

So roughly a **2× always-on runtime cost** (the shuffle count ~triples, but broadcastable count-exchange joins and fused window passes bring wall-clock to ~2×). Small-sample serverless numbers with the usual variance, but the direction is consistent across both scenarios. A real pipeline that also writes output would dilute the ratio; more compared columns or larger data could move it the other way.

### Recommendation

Keep the single-pass fix that's in this PR (resolves #1504, preserves `check_missing_records` for reference-only rows, and is strictly better than `main` — which produced this same spurious `changed` *and* cartesian row-multiplication for the shape). Treat the residual interleaving case as a **documented, pinned limitation of lazy duplicate-key pairing**, and steer correctness-sensitive duplicate-key users to `raise_on_duplicate_keys=True` (or a unique key), which needs no positional pairing and has no such edge case.

mattfaltyn:
@mwojtyczka thanks for the thoughtful review. I implemented exact-value-first matching for filtered lazy comparisons, kept unfiltered calls on the existing fast path, added regression coverage including tolerances, and updated the docs. All local checks pass. Could you please take another look at f728f728?

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -4740,8 +4945,9 @@ def _add_compare_condition(
     return df.withColumn(
         condition_col,
         F.when(
-            # apply filter but skip it for missing rows (null filter col)
-            (F.col(f"df.{filter_col}").isNull() | F.col(f"df.{filter_col}")) & ~all_is_ok,
+            # Missing rows have no source-side filter value and must still be reported. A present source
+            # row whose filter evaluates to NULL follows Spark filter semantics and remains out of scope.
+            (F.col(row_missing_col) | F.col(f"df.{filter_col}")) & ~all_is_ok,
```
mwojtyczka, line 4947: **Behavior change: present rows with a NULL filter result are no longer reported.**

The guard changed from `filter_col.isNull() | filter_col` to `row_missing_col | filter_col`. Previously any row whose filter column was NULL was treated as in-scope and reported; that included *present* source rows whose `row_filter` expression evaluates to NULL (e.g. `amount > 100` where `amount` is NULL). Those rows are now out of scope.

This looks intentional and arguably more correct (it matches Spark `WHERE` semantics, and the rewritten comment states it), and it *is* covered by `test_compare_datasets_filter_excludes_rows_from_duplicate_pairing`. The only real gap: it's a user-visible semantic change that isn't called out in the user docs / changelog. Consider a note in `docs/.../quality_checks.mdx` and the release notes so anyone relying on the old "NULL filter ⇒ in scope" reporting isn't surprised.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -3556,28 +3564,46 @@ def apply(df: DataFrame, spark: SparkSession, ref_dfs: dict[str, DataFrame]) ->
...
+                pairing_scope_col,
                 row_number_col,
-                F.row_number().over(Window.partitionBy(*pk_column_names).orderBy(*order_columns)),
-            )
-            ref_df = ref_df.withColumn(
-                row_number_col,
-                F.row_number().over(Window.partitionBy(*ref_pk_column_names).orderBy(*order_columns)),
+                unique_id,
+                null_safe_row_matching,
+                exact_value_pairing=bool(row_filter and compare_columns),
```
mwojtyczka, line 3593: **Exact-value pairing machinery runs unconditionally on the default filtered path.**

`exact_value_pairing = bool(row_filter and compare_columns)` is True for *every* filtered comparison that has compare columns, regardless of whether duplicate matching keys actually exist. When it's True, `_add_exact_value_pairing_columns` adds two `groupBy` aggregations, two extra joins, and several `row_number` windows. On the common case — a filtered comparison over unique keys with the default `raise_on_duplicate_keys=False` — this is a large, unconditional cost increase; previously a filtered comparison added only two window functions. The extra work is only needed when duplicate matching keys are present. Worth gating on actual duplicate presence, or documenting the cost.

mwojtyczka (commented):
Automated code-review pass on `compare_datasets` (filtered duplicate-key pairing). Findings below are inline. Severity ranges from a confirmed edge-case correctness gap (#4) to maintainability nits. One reproduction was run against a real Spark session at this commit.

This pull request review was written by Isaac.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -4441,6 +4484,168 @@ def _generate_field_presence_checks(
...
+    check_missing_records: bool | None,
+    null_safe_row_matching: bool | None,
+) -> DataFrame:
+    join_columns = [*pk_column_names, row_number_col]
+    ref_join_columns = [*ref_pk_column_names, row_number_col]
+    if pairing_phase_col is None:
+        return _match_rows(df, ref_df, join_columns, ref_join_columns, check_missing_records, null_safe_row_matching)
+
+    join_condition = F.lit(True)
+    for column, ref_column in zip(pk_column_names, ref_pk_column_names):
```
mwojtyczka: **Duplicated PK join-condition loop.**

This null-safe / non-null-safe PK join-condition loop is identical to the one in `_match_rows` (around line 4677). A future change to null-safe join semantics would have to be applied in both places; drift would silently change matching in only one path. Extract a shared PK-condition builder used by both.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -4441,6 +4484,168 @@ def _generate_field_presence_checks(
...
+        F.row_number().over(Window.partitionBy(*pk_column_names).orderBy(*order_columns)),
+    )
+    ref_df = ref_df.withColumn(
+        row_number_col,
+        F.row_number().over(Window.partitionBy(*ref_pk_column_names).orderBy(*order_columns)),
+    )
+    return df, ref_df, None
+
+
+def _add_exact_value_pairing_columns(
```
mwojtyczka, line 4519: **Residual positional pairing can still mispair an in-scope row across interleaving values (the #1504 fix is incomplete here). Reproduced on a real Spark session at this commit.**

Input (single PK group `id=1`):
- source: `(1,"X",in_scope=True)`, `(1,"X",True)`, `(1,"Y",False)`
- ref: `(1,"X")`, `(1,"Y")`

Result (`check_missing_records=True`, `row_filter="in_scope"`):
```
id=1 value='Y' in_scope=False -> None
id=1 value='X' in_scope=True  -> None                      (exact pair with ref X)
id=1 value='X' in_scope=True  -> changed {value: X -> Y}   (paired with ref Y)
```
Because the out-of-scope `Y` source row is excluded from the source value-count, ref `Y` gets `opposite_count=0`, falls to the phase-0 residual pool, and is positionally paired with the leftover in-scope `X` row — reported as `changed X→Y`, while the out-of-scope row's real counterpart is consumed. This is deterministic here (residual order is `[scope desc, value asc]`).

Whether this is a defect or an acceptable heuristic depends on the intended `row_filter` semantics (compare in-scope source against the *full* reference multiset vs. reserve the out-of-scope row's counterpart), which the docs don't pin down. Either way there's no test asserting this boundary — worth a test that documents the chosen behavior.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -3556,28 +3564,46 @@ def apply(df: DataFrame, spark: SparkSession, ref_dfs: dict[str, DataFrame]) ->
...
+                exact_value_pairing=bool(row_filter and compare_columns),
             )
         else:
             match_count_col = f"__match_count_{unique_id}"
             for dataset_name, dataset, matching_columns in (
                 ("source", df, pk_column_names),
                 ("reference", ref_df, ref_pk_column_names),
             ):
-                matchable_rows = dataset if null_safe_row_matching else dataset.dropna(subset=matching_columns)
+                matchable_rows = dataset.where(F.col(pairing_scope_col))
```
mwojtyczka: **`raise_on_duplicate_keys=True` validates uniqueness only over in-scope rows for filtered inputs.**

`matchable_rows = dataset.where(pairing_scope_col)` means an out-of-scope source row that shares a matching key with an in-scope row escapes the duplicate check. No `InvalidParameterError` is raised, both rows get `row_number=1`, and both join the same reference row — the fan-out that `raise_on_duplicate_keys` is documented to prevent. It's only non-user-visible when `check_missing_records` is disabled. Worth either validating over the unfiltered input or documenting that the guarantee is scoped to filter-passing rows.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -4441,6 +4484,168 @@ def _generate_field_presence_checks(
...
+        F.count("*").alias(source_count_col)
+    )
+    ref_counts = ref_count_input.groupBy(*dict.fromkeys(ref_value_columns)).agg(F.count("*").alias(ref_count_col))
+    df = _match_rows(
+        df.alias("df"),
+        ref_counts.alias("ref_df"),
+        source_value_columns,
+        ref_value_columns,
+        check_missing_records=False,
+        null_safe_row_matching=True,
```
mwojtyczka: **Value-count join hardcodes `null_safe_row_matching=True` regardless of the check's config.**

When the check is configured with `null_safe_row_matching=False`, `source_count_input` / `ref_count_input` are `dropna`'d on the PK columns, but this count join is still called with `null_safe_row_matching=True`. Correctness currently relies on the dropna removing null-PK rows. If the dropna is ever changed, or a null-PK row survives (e.g. a null in a compare column rather than a PK), the null-safe count join would over-count and mark rows as exact pairs that the non-null-safe main join then refuses to match, producing phantom missing/extra rows. Prefer honoring the configured `null_safe_row_matching` here rather than depending on the upstream dropna.

mattfaltyn:
> Implemented all review recommendations: strict duplicate-key validation now covers the full input, join predicates share one builder, exact-count joins honor configured row-key null safety, residual filter semantics have regression coverage, and the NULL-filter behavior plus filtered-path cost are documented in the reference and changelog.

Thanks for the thorough review. The branch is rebased on current `main`; 2,586 unit tests, formatting and lint, and the docs build all pass.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -3556,28 +3561,46 @@ def apply(df: DataFrame, spark: SparkSession, ref_dfs: dict[str, DataFrame]) ->
...
+                exact_value_pairing=bool(row_filter and compare_columns),
             )
         else:
             match_count_col = f"__match_count_{unique_id}"
             for dataset_name, dataset, matching_columns in (
                 ("source", df, pk_column_names),
                 ("reference", ref_df, ref_pk_column_names),
             ):
-                matchable_rows = dataset if null_safe_row_matching else dataset.dropna(subset=matching_columns)
+                matchable_rows = dataset
```
mwojtyczka: **Strict mode ignores `row_filter` — toggling `raise_on_duplicate_keys` changes correctness (correctness).**

When `raise_on_duplicate_keys=True`, duplicate-key validation runs over **all** source rows rather than only the in-scope ones. Duplicates that exist solely among filtered-*out* rows still raise, even though the scope-aware lazy path handles the same data without error. So `raise_on_duplicate_keys` stops being a pure performance/strictness toggle and changes the *result* — and it contradicts the PR description's claim that strict validation is limited to in-scope rows.

**Repro:** `df=[(1,'A',in_scope=False),(1,'B',in_scope=True)]`, `ref=[(1,'B')]`, `row_filter='in_scope'`, `raise_on_duplicate_keys=True`. Only one row (`id=1,'B'`) is in scope, so there is no in-scope duplicate, yet `apply()` raises `InvalidParameterError('source dataset contains duplicate matching keys ... id')`. The same inputs in lazy mode are handled cleanly.

**Fix direction:** apply the `row_filter` before the duplicate-key validation aggregation so strict mode validates the same in-scope row set the lazy path pairs over.

src/databricks/labs/dqx/check_funcs.py
```diff
@@ -4441,6 +4475,179 @@ def _generate_field_presence_checks(
...
+) -> DataFrame:
+    join_columns = [*pk_column_names, row_number_col]
+    ref_join_columns = [*ref_pk_column_names, row_number_col]
+    if pairing_phase_col is None:
+        return _match_rows(df, ref_df, join_columns, ref_join_columns, check_missing_records, null_safe_row_matching)
+
+    join_condition = _build_join_condition(pk_column_names, ref_pk_column_names, null_safe_row_matching)
+    join_condition &= F.col(f"df.{pairing_phase_col}") == F.col(f"ref_df.{pairing_phase_col}")
+    join_condition &= F.col(f"df.{row_number_col}") == F.col(f"ref_df.{row_number_col}")
+    for column in compare_columns:
```
mwojtyczka, line 4641: **Out-of-scope source row can absorb a genuinely-unmatched reference row (correctness).**

In the residual (phase-0) join, the join imposes no compare-column constraint, so an out-of-scope source row can pair positionally with a surplus reference row inside a duplicate-key group — suppressing a missing-record report even when that reference row has no in-scope source counterpart. This violates the PR's "genuine reference-only rows remain reportable" invariant.

**Repro:** `df=[(1,'X',in_scope=True),(1,'Z',in_scope=False)]`, `ref=[(1,'X'),(1,'X')]`, `row_filter='in_scope'`, `check_missing_records=True`. The in-scope `'X'` exact-pairs one ref `'X'`; the second ref `'X'` pairs positionally with the out-of-scope `'Z'` row and is reported `None` because the source side is out of scope. The extra reference `'X'` — which has no in-scope source match — is silently not reported as missing.

**Fix direction:** exclude out-of-scope source rows from the residual pairing (or constrain the phase-0 join so a reference row can only be absorbed by an in-scope source row), so surplus reference rows in a duplicate-key group remain reportable as missing.

mwojtyczka (changes requested):
found 2 new bugs and proposed fixes, otherwise lgtm

mattfaltyn:
> found 2 new bugs and proposed fixes, otherwise lgtm

Thanks for catching both. Fixed strict filter scoping and surplus reference pairing, with regression coverage for false and NULL filters and key-only comparisons. Local checks pass in 7bd32ddf.

src/databricks/labs/dqx/check_funcs.py
mwojtyczka: Fixed the concrete defect and documented the residual heuristic.

**Fix (`4c073f52`):** residual pairing ranked rows by the *string cast* of their compared values, so numeric columns sorted lexically (`"10.0" < "2.0"`). That reordered the two sides differently and misaligned the row-number pairing, flagging in-scope rows that actually had a within-tolerance partner. Numeric compare columns are now ordered by their native value on both sides; other types keep the string cast (natural order for dates/timestamps/booleans/strings, a uniform key for complex types). Regression test `test_compare_datasets_filter_tolerant_match_survives_interleaved_values` — `[2.0, 10.0]` vs ref `[2.3, 9.7]` with `abs_tolerance=0.5` now yields 0 violations (previously 2). Verified on a real Spark session.

**The repro in this thread is intended semantics, not that defect.** With two in-scope `X` rows and a single ref `X`, the extra in-scope `X` is a genuine discrepancy: in-scope rows are compared against the *full* reference multiset, so it is correctly reported (as `changed X→Y` by pairing with the leftover ref `Y`; the out-of-scope `Y` needs no counterpart). That boundary is pinned by `test_compare_datasets_filter_compares_residual_in_scope_rows_to_full_reference`.

**Remaining gap (documented, not closed):** exact values are paired up front, but values matching *only within tolerance* still fall to positional pairing, which is a greedy heuristic — a non-transitive tolerance chain, or an unmatched row shifting the positional order, can still pair a row with a different-but-tolerant reference. Native-value ordering removes the gross string-sort misalignment but not this heuristic. It is now called out in the `compare_datasets` docstring; when exact one-to-one tolerance matching matters, deduplicate the matching keys first.

mwojtyczka (approved):
LGTM - pending performance testing. The path with filter and raise_on_duplicate_keys=False runs heavy pairing machinery: two value-group groupBy aggregations + two left-outer count joins + extra window rankings, on top of the base per-key row-number windows. It runs unconditionally for every filtered comparison, even when keys are unique. The unfiltered and strict paths are essentially unchanged. Note: the existing benchmark don't set row_filter, so that regression-prone path isn't actually measured by the current baseline. Will add new test for this.

mwojtyczka:
## Perf: benchmark for the filtered exact-value pairing path

Added `test_benchmark_compare_datasets_filtered` (`tests/perf/test_apply_checks.py`, commit `426d4a9b`).

**Why:** the existing `compare_datasets` benchmarks all avoid the path this PR made heavier. `[lazy_pairing]` and `[foreach]` set no `row_filter` (plain positional pairing) and `[strict]` uses the single-join path — so the **filtered exact-value pairing** path (`_add_exact_value_pairing_columns`: two value-group aggregations, two count joins, and exact/residual/in-scope ranking windows) had no coverage. The new benchmark sets a check `filter` (pushed down as `row_filter`) over the heavily-duplicated `col1`/`col2` keys so that machinery does real work.

**Results** (100M rows, Databricks serverless, 5 warm rounds):

| Benchmark | mean |
|---|---|
| `compare_datasets[lazy_pairing]` (unfiltered) | ~7.4s |
| `compare_datasets_filtered` (new, exact-value path) | ~15.2s |

The filtered path costs **~2× the unfiltered path** on identical hardware — expected and now measured.

**No code regression indicated.** Absolute times are *not* comparable to `baseline.json` because that baseline is recorded on Azure Databricks serverless (nightly) whereas this run was on AWS serverless. Two independent signals confirm the differences are environmental, not code: the `strict` path (which never calls the changed ordering helper) shows the same offset, and the `strict`/`lazy` ordering inverts vs. the baseline env — neither is explicable by the localized `ORDER BY` change in this PR.

**Baseline handling:** `baseline.json` is intentionally left unchanged. The nightly perf job (same Azure serverless env, >25% mean-degradation threshold) is the authoritative regression check and will create the baseline entry for the new benchmark and open a baseline PR.

PR #1509, opened 2026-09-04
feat(app): version Studio releases independently

laurencewells:
## Changes

- Reset the independently versioned DQX Studio package to `0.1.0` while retaining DQX Core `0.16.0`.
- Require signed `studio-vX.Y.Z` release tags and generate `dqx-studio/marketplace/vX.Y.Z` branches.
- Update Marketplace release tooling, tests, and documentation.
- Show Studio `v0.1.0` in the application and Docusaurus documentation, linking to the Studio release tag.

### Linked issues

Related to #1108

### Tests

- [x] manually tested
- [x] added unit tests
- [ ] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

Validated with 29 Marketplace tests, 826 UI tests, 3,991 Studio backend tests, 2,495 Core unit tests, clean type checks, and a successful Docusaurus production build.

### Documentation and Demos

- [ ] added/updated demos
- [x] added/updated docs


app/scripts/release_marketplace.py
```diff
@@ -120,7 +121,7 @@ def release_marketplace(tag: str, repo_root: Path, commands: CommandRunner) -> s
 def main() -> int:
     """Create a local Marketplace release branch and print its manual push command."""
     parser = argparse.ArgumentParser(description=__doc__)
-    parser.add_argument("--tag", required=True, help="annotated signed version tag, for example v0.16.1")
+    parser.add_argument("--tag", required=True, help="annotated signed version tag, for example studio-v0.1.0")
```
mwojtyczka: The tool still prints only `git push origin {branch}` in `main()` ("Inspect it, then push manually…"), but this PR now also requires pushing the signed `studio-vX.Y.Z` tag — DEPLOYMENT.md/README.md/DEVELOPMENT.md were updated to two pushes, this stdout guidance was not. A releaser who follows the printed instruction never pushes the tag, so no GitHub release/tag is created and the docs' `AvailableSinceVersion` link (`.../releases/tag/studio-v0.1.0`) 404s. Consider updating the print to also instruct pushing the tag.

app/scripts/release_marketplace.py
```diff
@@ -67,7 +68,7 @@ def release_marketplace(tag: str, repo_root: Path, commands: CommandRunner) -> s
         project_version = tomllib.loads(tagged_pyproject.stdout)["project"]["version"]
     except (KeyError, TypeError, tomllib.TOMLDecodeError) as error:
         raise RuntimeError("Tagged application version is missing") from error
-    if project_version != tag.removeprefix("v"):
+    if project_version != tag.removeprefix("studio-v"):
```
mwojtyczka: The tag version is now parsed by two independent mechanisms: `release_branch_name()` captures it via the regex group `match.group(1)`, while here it's re-derived with `tag.removeprefix("studio-v")`. The two must stay in lockstep — a future prefix change needs edits in both places, and a mismatch would pass validation while producing a wrong branch name. Consider having `release_branch_name()` return (or a shared helper expose) the parsed version so this check reuses the same captured group.

docs/dqx/docs/studio/index.mdx
```diff
@@ -51,7 +51,12 @@ data-quality engine, exposed as a Python package.
 
 <FeatureTags>
   <FeatureLifecycleStage stage="beta" heading={false} />
-  <AvailableSinceVersion version="0.16.1" heading={false} />
+  <AvailableSinceVersion
```
mwojtyczka, line 54: `productName="DQX Studio"` only reaches the link's `title` (hover tooltip) in `FeatureTags.tsx` — the visible badge span still renders just `Available since v0.1.0`. Now that Studio (0.1.0) and Core (0.16.x) version independently, a reader who sees "v0.1.0" with no product qualifier could conflate it with a Core version. Consider surfacing the product name in the rendered badge text, not only the tooltip.

mwojtyczka (changes requested):
Generally looking good, left a few small comments

app/scripts/release_marketplace.py
laurencewells: Yep, updated this so the script prints both commands, branch first and then the signed tag.

docs/dqx/docs/studio/index.mdx
laurencewells: Updated the visible text as well. The badge now reads “Available since DQX Studio v0.1.0”, with a focused test around the label.

app/scripts/release_marketplace.py
laurencewells: Refactored this into a shared _studio_release() parser, so the version check and branch name both come from the same regex match.

docs/dqx/src/components/FeatureTagLabels.test.ts
```diff
@@ -0,0 +1,11 @@
+import assert from 'node:assert/strict';
```
mwojtyczka, line 1: This test never runs: `docs/dqx/package.json` has no `test` script and the CI docs job runs only `make docs-build` (Docusaurus build), so nothing executes `node --test` on this file. It reads as regression protection for `availableSinceLabel()` but provides none — a future regression would ship undetected. Either wire up a `test` script + CI step that runs it, or drop the test (and inline the helper — see the FeatureTagLabels.ts comment).

docs/dqx/src/components/FeatureTags.tsx
```diff
@@ -89,33 +90,40 @@ export function FeatureLifecycleStage({
...
+ * @param releaseTag  the Git tag containing the release notes. Defaults to the
+ *                    DQX Core tag for *version*.
+ * @param productName  the product name shown in the badge and accessible link title.
  * @param heading  when true (default), uses heading-friendly sizing so the tag can sit
  *                 inline next to a page or section title; set false for body text.
  */
 export function AvailableSinceVersion({
   version,
+  releaseTag = `v${version}`,
+  productName = 'DQX',
```
mwojtyczka, line 111: With `productName` defaulting to `'DQX'`, every existing Core `AvailableSinceVersion` tag now renders **"Available since DQX vX.Y.Z"** instead of the previous "Available since vX.Y.Z" — a sitewide visible-text change across ~30 tags in `guide/*.mdx` (0.14.0, 0.16.0, …), on pages unrelated to Studio.

Rather than leave that riding on a changed default, suggest making it explicit: update the existing Core `AvailableSinceVersion` usages to pass `productName` themselves (mirroring how the new Studio tag passes `productName="DQX Studio"`). That way every badge names its product intentionally and the sitewide relabel is a deliberate, reviewed change rather than a side effect of the default.

docs/dqx/src/components/FeatureTagLabels.ts
```diff
@@ -0,0 +1,7 @@
+/** Return the visible label for a feature's first product release. */
```
mwojtyczka, line 1: This module exists only so the (currently non-running) test can import `availableSinceLabel`. It's a one-line template string; splitting it into its own file + import in FeatureTags.tsx adds indirection for no runtime benefit. If the test is dropped, inline this back into `AvailableSinceVersion` and keep the change to a single file.

docs/dqx/src/components/FeatureTags.tsx
```diff
@@ -144,7 +152,7 @@ export function DeprecatedInVersion({
     : `Deprecated in DQX v${version}`;
   return (
     <Link
-      to={`${RELEASE_NOTES_BASE}${version}`}
+      to={`${RELEASE_NOTES_BASE}v${version}`}
```
mwojtyczka: `AvailableSinceVersion` was generalized with `releaseTag`/`productName` for Studio, but `DeprecatedInVersion` still hardcodes the Core tag (`${RELEASE_NOTES_BASE}v${version}`) and "Deprecated in DQX v…". If a Studio feature is ever deprecated, its badge will link to a non-existent Core tag (`.../tag/v0.1.0`) and mislabel the product. Consider applying the same `releaseTag`/`productName` params here so the abstraction is consistent.

mwojtyczka (commented):
Follow-up review after the FeatureTags rework. Four points below — the badge-label change (2) and the dead test (1) are the substantive ones.

docs/dqx/src/components/FeatureTagLabels.test.ts
laurencewells: Implemented. The docs build now runs the label tests.

docs/dqx/src/components/FeatureTags.tsx
laurencewells: Implemented. Core badges now pass productName explicitly.

docs/dqx/src/components/FeatureTagLabels.ts
laurencewells: Kept this helper because its tests now run in the docs build.

docs/dqx/src/components/FeatureTags.tsx
laurencewells: Implemented. DeprecatedInVersion now supports productName and releaseTag.

app/scripts/release_marketplace.py
```diff
@@ -37,16 +37,37 @@ def run(self, command: tuple[str, ...], *, cwd: Path, check: bool = True) -> Com
...
+    version = match.group(1)
+    return _StudioRelease(version=version, branch=f"dqx-studio/marketplace/v{version}")
+
+
 def release_branch_name(tag: str) -> str:
     """Return the release branch name for a valid Marketplace version tag."""
-    if re.fullmatch(r"v[0-9]+\.[0-9]+\.[0-9]+(?:[A-Za-z0-9.-]*)?", tag) is None:
-        raise ValueError("TAG must use vX.Y.Z release syntax")
-    return f"marketplace/{tag}"
+    return _studio_release(tag).branch
```
mwojtyczka: After this refactor `release_branch_name()` is dead production code: `release_marketplace()` now calls `_studio_release(tag)` directly, and this function just delegates to `_studio_release(tag).branch`. Across the repo its only remaining references are this definition and two calls in `test_release_marketplace.py` — no production caller. It reads as live public API but exists solely for the test to exercise. Consider removing it and having the test assert on `_studio_release(tag).branch` (or on `release_marketplace`'s observable output) instead.

mwojtyczka (approved):
LGTM - left one comment about dead code

PR #1510, opened 2026-09-05
Add row-level geo distance check and fix documentation examples

simplegaurav:
## Changes

Adds `is_geo_within_distance`, a row-level geospatial check that reports point values farther than a
maximum geodesic distance, in meters, from a reference point.

```yaml
- criticality: error
  check:
    function: is_geo_within_distance
    arguments:
      column: location
      reference_geometry: "POINT(4.90 52.37)"
      distance: 1000
      convert_column: true
      convert_reference_geometry: true
```

**Design — every point below was verified against a live serverless workspace (DBR 19.6), not assumed**

- **Distance via `st_distancespheroid`**, i.e. meters on the WGS 84 ellipsoid. The planar `st_distance`
  returns coordinate units (degrees for longitude/latitude data — 0.15 for two points 10 km apart),
  and no Databricks distance function accepts `GEOGRAPHY`: the runtime's own `DESCRIBE FUNCTION`
  defines `st_distance`, `st_distancesphere` and `st_distancespheroid` over `GEOMETRY` only. The check
  therefore operates on `GEOMETRY` parsed with `try_to_geometry` (WKT, WKB, EWKT, EWKB, GeoJSON),
  exactly like the other `is_geo_*` checks.
- **Points only.** `st_distancespheroid` raises `ST_INVALID_ARGUMENT` at runtime for any non-point
  argument, so the call is gated behind a `when` on `st_geometrytype`; a polygon row is reported as
  `is not a point geometry` instead of crashing the job.
- **SRID handling.** `try_to_geometry` assigns SRID 0 to WKT/WKB and 4326 to EWKT/GeoJSON, and
  `st_distancespheroid` raises `ST_DIFFERENT_SRID_VALUES` on a mismatch. Both operands are stamped
  4326 before measuring (SRID 0 is treated as WGS 84, 4326 is used as is, so formats can be mixed);
  any other SRID is reported per row rather than silently misread as degrees — `st_setsrid` relabels
  without reprojecting, so stamping a projected geometry would produce silently wrong distances.
- **Distinct messages** for an unparseable column value (raw input, rendered as hex for binary WKB —
  a raw string cast of WKB is invalid UTF-8 and crashes result collection), an unparseable reference,
  a non-point column value or reference, a projected SRID on either, and an empty column point or
  reference (none of these has a location to measure from). Null column values, null reference
  columns and null distances are skipped, matching the null semantics of the sibling checks.
- **`distance`** accepts a number, a `Column`, or a SQL expression evaluated per row. Numeric
  literals and numeric strings are validated up front: negative, NaN, infinite, boolean, and
  float-overflowing values raise `InvalidParameterError`.
- `convert_column` / `convert_reference_geometry` default to `False`, matching every sibling check.
  Operand preparation is shared with `_has_topological_relationship_precise` through a new
  `_prepare_geo_operands` helper, and the per-operand point/SRID/empty diagnostics live in
  `_diagnose_point_operand`. Every function the check relies on is documented for DBR 17.1+, so the
  declared minimum runtime holds.

**Also in this PR**

- Fixes the pre-existing programmatic examples for seven `is_geo_*` checks in the reference docs,
  which used `DQDatasetRule` although all of them are registered as row rules and raised
  `InvalidCheckError` as written.
- Registers the check in the Studio built-in rule severity seed at `Low`, alongside its siblings.

All review rounds from @ghanse and @mwojtyczka are addressed; see the inline replies for per-comment
details.

### Linked issues

None.

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] added end-to-end tests
- [x] added performance tests

Run against a live serverless workspace:

- `tests/integration/test_row_checks_geo.py` — the full file, **75 passed**, including 20 tests for
  this check: inside/outside radius, per-row and null radius, SQL-expression and numeric-string
  radius, per-row and null reference, unparseable column and reference, polygon column and reference,
  valid and invalid binary WKB, WKT mixed with GeoJSON in both directions, projected-SRID column and
  reference, empty column point and empty reference, and native `GEOMETRY` input with
  `convert_column=False`.
- `test_apply_checks_all_geo_checks_using_classes` and `test_apply_checks_all_geo_checks_as_yaml`,
  with the check added to `all_row_geo_checks.yaml` — both pass.
- Studio: `make app-test K=builtin_rules_seed` — 23 passed.
- Unit: 54 tests in `tests/unit/test_geo_check_funcs.py` plus the parameter-order contract in
  `test_check_func_signatures.py`; full unit suite green.
- Perf: `test_benchmark_is_geo_within_distance` alongside the `is_geo_covers` benchmarks. Not runnable
  locally on Windows (the perf fixture's 1900-01-01 start date fails in `datetime.timestamp()` there,
  for every benchmark); CI runs it on Linux.

Local gates: `black`, `ruff`, `mypy .`, and `pylint src tests` (10.00/10) all clean.

### Documentation and Demos

- [ ] added/updated demos
- [x] added/updated docs

`docs/dqx/docs/reference/quality_checks.mdx`: row-level table entry, YAML example and `DQRowRule`
example, plus the `DQDatasetRule` → `DQRowRule` fix described above.


src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -1340,3 +1341,103 @@ def is_geo_within(
...
+) -> Column:
+    """Checks if the column geography is within a geodesic distance of the reference geography using `st_distance`.
+
+    The distance is measured in meters along the WGS 84 ellipsoid, so the check is meaningful for
+    global data where planar `GEOMETRY` distances are not. A value is reported when the shortest
+    distance between it and the reference geography is strictly greater than *distance*.
+
+    Both the target column and the reference geometry are always handled as `GEOGRAPHY`.
+    When conversion is requested (*convert_column* or *convert_reference_geometry* set to True),
+    *try_to_geography* is applied to parse the value from any supported format (WKT, WKB, EWKT, EWKB).
```
ghanse: We should include GeoJSON in the supported formats as well.

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -1340,3 +1341,103 @@ def is_geo_within(
...
+    distance_expr = get_limit_expr(distance)
+
+    # `try_to_geography` yields NULL for values that fail to parse. The column and the reference are
+    # reported separately so the error message points at the value the user has to fix. Null input
+    # values are skipped before these are evaluated, so a NULL here always means "unparseable".
+    col_invalid = col_geog.isNull()
+    ref_invalid = ref_geog.isNull()
+    is_too_far = F.call_function("st_distance", col_geog, ref_geog) > distance_expr
+
+    condition = F.when(col_expr.isNull(), F.lit(None)).otherwise(col_invalid | ref_invalid | is_too_far)
```
ghanse, line 1413: This is great for handling input geometries that do not parse correctly. Opened #1513 to extend this to other check functions.

tests/integration/test_row_checks_geo.py
ghanse, line 1: I think we're missing a test case for `convert_column=False`

ghanse (changes requested):
Very good contribution. Left a few minor comments. Need to add 1 test case.

simplegaurav:
Thanks for the review — all three addressed in `3ce2ddd6`.

**GeoJSON in supported formats**

Good catch. The docstring listed the `try_to_geometry` formats (WKT, WKB, EWKT, EWKB) that I'd
copied from the sibling relationship checks, but this check parses with `try_to_geography`, which
also accepts GeoJSON. Updated in three places: the docstring, the `reference_geometry` argument
description, and the row-level checks table in `quality_checks.mdx`.

**Missing `convert_column=False` test**

Added two integration tests covering the native `GEOGRAPHY` path — one pass case, one violation:

```python
def test_is_geo_within_distance_native_geography_violation(skip_if_runtime_not_geo_compatible, spark):
    """A native GEOGRAPHY value outside the radius is flagged, with the value rendered via st_astext."""
    point = "POINT(5.05 52.37)"
    test_df = spark.createDataFrame([[point]], _GEO_SCHEMA).select(
        F.call_function("try_to_geography", F.col("geom")).alias("geom")
    )
    condition = is_geo_within_distance("geom", F.call_function("try_to_geography", F.lit(_POINT_INSIDE)), 1000)
    ...
```

They build the column with `try_to_geography` in a `select` and then leave both convert flags at
their defaults, so the check receives a genuinely `GEOGRAPHY`-typed column. The violation case also
exercises the `st_astext` rendering branch, which is only reachable on that path — with
`convert_column=True` the message renders the raw input instead, since `st_astext` returns NULL for
exactly the unparseable values the invalid-geography message is about.

I assert on the condition column alone rather than also selecting `geom`, since a `GEOGRAPHY`-typed
column can't be compared against a string schema.

**One thing I noticed while writing those**

No test in the repo currently builds a native `GEOGRAPHY` column. Every `is_geo_*` check defaults to
`convert_column=False`, but all existing tests pass `True`, so the default path is untested across
the geo module rather than just here. Might be worth a companion issue to #1513 — happy to open one
if useful.


Thanks for opening #1513.

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -1340,3 +1341,104 @@ def is_geo_within(
...
+    )
+    invalid_reference_message = F.lit(f"reference geometry for column `{col_expr_str}` is not a valid geography")
+    too_far_message = F.concat_ws(
+        "",
+        F.lit("value `"),
+        text_value_col,
+        F.lit(f"` in column `{col_expr_str}` is farther than "),
+        distance_expr.cast("string"),
+        F.lit(" meters from the reference geometry"),
+    )
```
ghanse, line 1436: This may not leave the most useful messages if the user passes WKB/EWKB with `convert_column=True`. For other checks, we surround the parsed geometry with `st_astext`:

```suggestion
    text_value_col = col_expr.cast("string") if convert_column else F.call_function("st_astext", col_geog)
    invalid_column_message = F.concat_ws(
        "",
        F.lit("value `"),
        text_value_col,
        F.lit(f"` in column `{col_expr_str}` is not a valid geography"),
    )
    invalid_reference_message = F.lit(f"reference geometry for column `{col_expr_str}` is not a valid geography")
    
    too_far_value_col = F.call_function("st_astext", col_geog)
    too_far_message = F.concat_ws(
        "",
        F.lit("value `"),
        too_far_value_col,
        F.lit(f"` in column `{col_expr_str}` is farther than "),
        distance_expr.cast("string"),
        F.lit(" meters from the reference geometry"),
    )
```

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -1340,3 +1341,104 @@ def is_geo_within(
...
+    Raises:
+        InvalidParameterError: If *distance* is a boolean, or a numeric literal that is negative,
+            NaN or infinite.
+
+    Note:
+        This function requires Databricks serverless compute or runtime 17.1 or above.
+    """
+    # `bool` is a subclass of `int`, so it would otherwise slip through as a 0/1 metre radius.
+    if isinstance(distance, bool) or (
+        isinstance(distance, (int, float)) and not (math.isfinite(distance) and distance >= 0)
```
ghanse, line 1395: `math.isfinite` may overflow on very large input distances. We might want to catch this and raise `InvalidParameterError`?

ghanse (changes requested):
Left a few minor suggestions.

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -1340,3 +1341,104 @@ def is_geo_within(
...
+    is_too_far = F.call_function("st_distance", col_geog, ref_geog) > distance_expr
+
+    condition = F.when(col_expr.isNull(), F.lit(None)).otherwise(col_invalid | ref_invalid | is_too_far)
+
+    # How the offending value is rendered depends on the input contract. When the column is converted,
+    # it holds a WKT/WKB-style value that casts to string losslessly, and the raw text is what the user
+    # needs to see - `st_astext` would be NULL on exactly the unparseable values the message is about.
+    # When conversion is off the column is already a native GEOGRAPHY, which has no string cast, so the
+    # value has to be rendered with `st_astext`; a NULL there is already excluded by the null guard.
+    text_value_col = col_expr.cast("string") if convert_column else F.call_function("st_astext", col_geog)
```
mwojtyczka, line 1420: **[correctness] Binary WKB renders as mojibake in the error message.** For a binary WKB column with `convert_column=True`, an unparseable value is rendered here via `col_expr.cast("string")`, which decodes raw bytes as UTF-8 — the message becomes `value ��... is not a valid geography` rather than readable text. The comment above ("casts to string losslessly") holds for WKT strings but not binary. Other geo checks avoid this via `st_astext(try_to_geometry(col_expr))` (NULL on unparseable values, dropped by `concat_ws`).

src/databricks/labs/dqx/geo/check_funcs.py
mwojtyczka, line 1395: **[correctness] Negative/NaN/inf validation is bypassed for string distances.** This guard only fires for `int`/`float`. A string literal like `distance="-100"` skips it, and `get_limit_expr("-100")` returns `F.expr("-100")` = -100, so `st_distance >= 0 > -100` for every row → all non-null rows are flagged. The numeric `-100` raises `InvalidParameterError`, but the string `"-100"` silently misbehaves — same value, opposite behavior by type. Consider validating string/expression distances too, or documenting that only numeric literals are range-checked.

mwojtyczka (commented):
Automated code review for `is_geo_within_distance` — 5 findings inline, ranked most-severe first (WKB rendering and the string-distance validation gap are the two worth acting on; the rest are lower severity). Line numbers verified against the PR head.

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -1340,3 +1341,104 @@ def is_geo_within(
...
+        isinstance(distance, (int, float)) and not (math.isfinite(distance) and distance >= 0)
+    ):
+        raise InvalidParameterError(f"'distance' must be a finite, non-negative number of meters, got {distance!r}.")
+
+    col_str_norm, col_expr_str, col_expr = get_normalized_column_and_expr(column)
+
+    ref_col = reference_geometry if isinstance(reference_geometry, Column) else F.lit(reference_geometry)
+    col_geog = F.call_function("try_to_geography", col_expr) if convert_column else col_expr
+    ref_geog = F.call_function("try_to_geography", ref_col) if convert_reference_geometry else ref_col
+    distance_expr = get_limit_expr(distance)
```
mwojtyczka, line 1404: **[security — low confidence] String `distance` reaches `F.expr()` without `is_sql_query_safe()`.** A YAML/metadata `distance: "<arbitrary SQL>"` flows through `get_limit_expr` to `F.expr(distance)` unvalidated, contrary to the CLAUDE.md SQL-safety rule. Low confidence: this mirrors the existing codebase-wide `get_limit_expr` pattern (`is_not_greater_than`, etc.), so it isn't a new pattern — but it's a fresh instance in a new public check. Worth confirming the shared pattern's safety story covers metadata-supplied expressions.

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -1340,3 +1341,104 @@ def is_geo_within(
...
+    """
+    # `bool` is a subclass of `int`, so it would otherwise slip through as a 0/1 metre radius.
+    if isinstance(distance, bool) or (
+        isinstance(distance, (int, float)) and not (math.isfinite(distance) and distance >= 0)
+    ):
+        raise InvalidParameterError(f"'distance' must be a finite, non-negative number of meters, got {distance!r}.")
+
+    col_str_norm, col_expr_str, col_expr = get_normalized_column_and_expr(column)
+
+    ref_col = reference_geometry if isinstance(reference_geometry, Column) else F.lit(reference_geometry)
```
mwojtyczka, line 1401: **[reuse] Operand-prep boilerplate duplicated from `_has_topological_relationship_precise`.** The `ref_col = ... if isinstance(..., Column) else F.lit(...)` branch plus the `convert_column`/`convert_reference_geometry` → `try_to_geography` wrapping (1401-1403) duplicate logic already in `_has_topological_relationship_precise`. A shared helper (e.g. `_prepare_geo_operands`) would keep WKT/GEOGRAPHY conversion semantics consistent — otherwise the same fix must be made in two places.

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -1340,3 +1341,104 @@ def is_geo_within(
...
+        text_value_col,
+        F.lit(f"` in column `{col_expr_str}` is not a valid geography"),
+    )
+    invalid_reference_message = F.lit(f"reference geometry for column `{col_expr_str}` is not a valid geography")
+    too_far_message = F.concat_ws(
+        "",
+        F.lit("value `"),
+        text_value_col,
+        F.lit(f"` in column `{col_expr_str}` is farther than "),
+        distance_expr.cast("string"),
```
mwojtyczka, line 1434: **[efficiency — minor & optional] `distance_expr` and `text_value_col` are materialized twice.** `distance_expr` appears in `is_too_far` (L1411) and again here as `.cast("string")`; `text_value_col` — which itself calls `st_astext` — is embedded in both `invalid_column_message` (L1425) and `too_far_message` (L1432), so `st_astext` runs for every row that reaches the message. Minor: Spark may fold common subexpressions, but the expression tree is larger than necessary.

mwojtyczka (changes requested):
Thank you for the PR, I left some comments

simplegaurav:
Pushed `eeb4f887`, which addresses the second round of comments and, more importantly, reworks the core of the check after verifying it against a live serverless workspace.

- The original design measured with `st_distance` on `GEOGRAPHY`. On this runtime `st_distance`, `st_distancesphere` and `st_distancespheroid` all reject `GEOGRAPHY` (`DATATYPE_MISMATCH`), and `st_distance` on `GEOMETRY` is planar — it returned 0.15 for two lon/lat points 10 km apart, i.e. degrees. The check now parses with `try_to_geometry` like the other `is_geo_*` checks and measures with `st_distancespheroid`, which returns meters on the WGS 84 ellipsoid (10,216.6 m / 68.1 m for the test points).
- `st_distancespheroid` is points-only and raises at runtime on any other geometry type, and on mismatched SRIDs (`try_to_geometry` assigns SRID 0 to WKT/WKB but 4326 to EWKT/GeoJSON). Both are guarded: a non-point column value or reference is reported instead of raising, and both operands are stamped SRID 4326 before measuring — 0 is treated as WGS 84, 4326 used as is, and any other SRID is reported rather than silently misread as degrees.
- An empty reference is reported on every row (it silently disabled the check), and a null reference column now leaves the row unmeasurable and skipped, matching the null semantics of the sibling checks.

Every behaviour above has an integration test that passes on the live workspace: the full `test_row_checks_geo.py` (75 tests, 20 for this check), the class-based and YAML all-geo tests, and the Studio seed tests.

src/databricks/labs/dqx/geo/check_funcs.py
simplegaurav: Done. The docstring, the `reference_geometry` description and the reference table now list GeoJSON alongside WKT, WKB, EWKT and EWKB, since parsing goes through `try_to_geometry`, which accepts all of them.

Listing it wasn't enough on its own: on the live workspace `try_to_geometry` assigns SRID 0 to WKT/WKB but 4326 to GeoJSON/EWKT, and `st_distancespheroid` throws `ST_DIFFERENT_SRID_VALUES` when the operands differ, so a WKT column against a GeoJSON reference failed on every row. Both operands are now stamped with `st_setsrid(..., 4326)` before measuring — it only requires SRID equality and always interprets coordinates as WGS 84 degrees, so the result is unchanged for WKT input. Two integration tests cover the mixed cases in both directions and pass against serverless.

src/databricks/labs/dqx/geo/check_funcs.py
simplegaurav: Thanks. That handling survived the redesign to `st_distancespheroid` (column and reference reported separately), and it now also gates non-point geometries so the spheroid call never raises. The parse and raw-value rendering live in `_prepare_geo_operands` and `_raw_geo_value_as_text`, which should make #1513 mostly a matter of reusing those helpers in the other checks.

tests/integration/test_row_checks_geo.py
simplegaurav: Good catch — the original commit only exercised the conversion path. Added `test_is_geo_within_distance_native_geometry_no_violation` and `..._violation`, which build the column and the reference with `try_to_geometry` up front and call the check with the default `convert_column=False` / `convert_reference_geometry=False`, asserting the pass case (~68 m) and the flagged case (~10.2 km) including the `st_astext`-rendered message.

An earlier iteration used `try_to_geography` here, but probing the runtime showed every distance function rejects GEOGRAPHY, so the check is GEOMETRY-only and the tests reflect that. The no-conversion path also avoids any string cast of the raw value, because `cast(GEOMETRY AS STRING)` fails analysis even in an untaken branch. Both pass against a live serverless workspace.

src/databricks/labs/dqx/geo/check_funcs.py
simplegaurav: Done. The messages now follow the pattern of the other geo checks: the "farther than" and "is not a point geometry" messages render the value with `st_astext(col_geom)`, so WKB, GeoJSON and native GEOMETRY input all show as WKT (e.g. `POINT(5.05 52.37)`). Only the "is not a valid geometry" branch shows the raw input, since `st_astext` is NULL there by definition; for that branch `_raw_geo_value_as_text` renders binary input as hex instead of casting the bytes — a raw cast of WKB is not valid UTF-8 and actually crashed result collection on Spark Connect. `test_is_geo_within_distance_binary_column` covers both rows (WKB → WKT in the too-far message, `0000` in the invalid message) and passes against serverless.

src/databricks/labs/dqx/geo/check_funcs.py
simplegaurav: Good catch — `math.isfinite` on an int that can't be represented as a float (e.g. `10**400`) raises `OverflowError` rather than returning False. `_validate_distance` now wraps the call and treats an overflow as non-finite, so it falls into the same `InvalidParameterError` as NaN, ±inf and negatives. Numeric strings such as `"1e400"` reach the same check via `float()`, which yields `inf` rather than overflowing. The Raises section lists the case, and `test_is_geo_within_distance_rejects_invalid_distance` is parametrized with `10**400`.

src/databricks/labs/dqx/geo/check_funcs.py
simplegaurav: Fixed in `eeb4f887`. Parsed values now always render through `st_astext`, so a valid WKB point shows as `POINT(5.05 52.37)`.

For the unparseable branch I didn't adopt the NULL-and-drop approach: on the live workspace `cast(BINARY AS STRING)` of WKB bytes was worse than mojibake — it produced invalid UTF-8 that crashed Spark Connect result collection with `UnicodeDecodeError` — and dropping the value would leave the user with no clue which input failed. Instead `_raw_geo_value_as_text` gates on `typeof(x) = 'binary'` and renders bytes as `hex(x)`, keeping the string cast only for string input. The native-GEOMETRY path reuses the `st_astext` rendering because `cast(GEOMETRY AS STRING)` fails analysis even in an untaken branch. `test_is_geo_within_distance_binary_column` covers both the parsed and unparseable WKB rows and passes against serverless.

src/databricks/labs/dqx/geo/check_funcs.py
simplegaurav: Fixed. `_validate_distance` now runs `float()` on string distances first: anything that parses as a number is held to the same finite/non-negative rule as numeric literals, so `"-100"`, `"nan"`, `"inf"` and `"-1e3"` raise `InvalidParameterError` exactly like their numeric counterparts. Strings that don't parse (`"radius_m * 2"`) are SQL expressions and pass through to `get_limit_expr`, as do `Column` values; those are evaluated per row and documented as not range-checked. `bool` is rejected too (a subclass of `int` that would otherwise be a 0/1 m radius). Unit tests parametrize the rejected and accepted cases, and the expression and Column paths are covered by integration tests that pass against serverless.

src/databricks/labs/dqx/geo/check_funcs.py
simplegaurav: I left this as is, and I think the shared pattern's safety story does hold — I checked it empirically against the workspace rather than by reasoning:

- Every destructive statement I tried through `F.expr()` — `DROP TABLE …`, `DELETE FROM …`, a `;`-chained second statement — is a `ParseException`. `F.expr` parses a single expression, so statements can't execute through it regardless of any guard.
- The one thing `F.expr` *does* permit that is a real exposure — a scalar subquery reading another table, e.g. `(SELECT count(*) FROM information_schema.tables)` — is allowed by `is_sql_query_safe`, which explicitly permits `SELECT`.

So adding `is_sql_query_safe` here would block nothing `F.expr` can run and allow the only thing it can. If we want to tighten this it should be a subquery guard inside `get_limit_expr`, so every limit-taking check (`is_not_greater_than`, `has_dimension`, …) benefits consistently — happy to open an issue for that rather than special-case one check.

src/databricks/labs/dqx/geo/check_funcs.py
simplegaurav: Extracted into `_prepare_geo_operands`, which returns a `_GeoOperands` NamedTuple; both `_has_topological_relationship_precise` and `is_geo_within_distance` call it, so the literal-vs-Column branch and the `try_to_geometry` wrapping live in one place. The conversion semantics actually converged along the way: the distance check no longer uses `try_to_geography`, because every distance function on this runtime rejects GEOGRAPHY and `st_distance` on GEOMETRY is planar. It now takes the same `try_to_geometry` path as the relationship checks and measures with `st_distancespheroid`, with both operands passed through `st_setsrid(..., 4326)` and gated on `st_geometrytype = 'ST_Point'`. The full geo integration file passes against serverless, so the sibling checks are unaffected by the refactor.

src/databricks/labs/dqx/geo/check_funcs.py
simplegaurav: Left as is, deliberately. Each repeated subexpression — `distance_expr`, `st_astext(col_geom)`, the `st_geometrytype` checks — is referenced from a different branch of one `CASE`, and `make_condition` only evaluates the message for failing rows, so per row `st_astext` runs at most once and only on a violation; the cost is tree size, not runtime. Folding them into a single evaluation would need a `withColumn` or CTE, which the row-check contract (`Column` in, `Column` out) doesn't allow, and `is_geo_within` / `is_geo_contains` embed `st_astext` the same way. Spark's subexpression elimination also dedupes these within the projection at codegen time.

simplegaurav:
@ghanse  thanks for the two rounds of review — all five of your comments are addressed in `eeb4f887`, with a reply on each thread:

- GeoJSON added to the documented formats (and the SRID mismatch that mixing it with WKT would have caused is handled).
- `convert_column=False` now has integration coverage on the native `GEOMETRY` path.
- Too-far values render via `st_astext`, so WKB shows as WKT; unparseable binary input renders as hex.
- `math.isfinite` overflow is caught and raises `InvalidParameterError`.

One thing worth your attention beyond the comments: verifying against a live serverless workspace (DBR 19.6, so not a runtime-age issue) showed the original `st_distance`-on-`GEOGRAPHY` design couldn't work — the runtime's own `DESCRIBE FUNCTION` defines `st_distance` as 2D Cartesian over `GEOMETRY`, and no Databricks distance function accepts `GEOGRAPHY`. The check is now built on `st_distancespheroid` with point-type and SRID guards; every function it uses is documented for DBR 17.1+, so the declared minimum still holds. The PR description has the details, and the full geo integration file passes live (75 tests, 20 for this check).


Could you take another look when you have a moment?

simplegaurav:
@mwojtyczka thanks for the review — all five findings are addressed in `eeb4f887`, with a reply on each thread:

- **Binary WKB rendering:** unparseable binary now renders as hex; parsed values go through `st_astext`. On the live workspace the raw cast was worse than mojibake — it crashed result collection — so this was a real fix, not cosmetic.
- **String distance validation:** numeric strings are now held to the same rule as numeric literals, so `"-100"` and `-100` behave alike.
- **Operand-prep duplication:** extracted into `_prepare_geo_operands`, shared with `_has_topological_relationship_precise`.
- **`F.expr` / `is_sql_query_safe`:** left as is, with evidence on the thread — `F.expr` can't execute statements, and the one thing it can do (scalar subqueries) `is_sql_query_safe` permits anyway. I've proposed a subquery guard in `get_limit_expr` as a follow-up so all limit-taking checks benefit consistently.
- **Duplicated subexpressions:** left as is; reasoning on the thread.

The check itself was also reworked after live verification showed the `st_distance`/`GEOGRAPHY` design didn't run on serverless — it's now `st_distancespheroid` over `GEOMETRY` with point-type and SRID guards, all covered by integration tests that pass live. Details in the updated PR description.

Would you be able to re-review?

src/databricks/labs/dqx/geo/check_funcs.py
```diff
@@ -1340,3 +1440,128 @@ def is_geo_within(
...
+                ),
+            )
+        )
+    )
+
+    return make_condition(
+        F.when(col_expr.isNull(), F.lit(None)).otherwise(unmeasurable | (geodesic_distance > distance_expr)),
+        message,
+        alias=f"{col_str_norm}_is_not_within_distance_from_reference_geometry",
+    )
```
ghanse, line 1573: I think there's an edge case where an input of `"POINT EMPTY"` is skipped only because `st_distancespheroid` returns a null value. We respond to similar values (e.g. unparseable geometries) with a message.

It might be best for us to handle this more explicitly. I'm not sure it should be skipped outright. We should also document the behavior for the users in docstrings/documentation.

ghanse (approved):
Thanks for your work on this. LGTM pending the changes for `"POINT EMPTY"` values.

src/databricks/labs/dqx/geo/check_funcs.py
simplegaurav: @ghanse , Agreed — it was being skipped by accident, not by design. Fixed in `3d2d63c4`.

Emptiness is now one of the explicit reasons an operand can't be measured, next to unparseable, non-point and projected-SRID values. An empty column point is reported as ``value `POINT EMPTY` in column `geom` is an empty geometry`` — the same wording `is_non_empty_geometry` uses — and an empty reference is reported on every row with the matching message. The docstring and the reference table now describe this instead of deferring to `is_non_empty_geometry`; the null semantics (null column value, null reference column, null distance) are unchanged.

Verified against the live serverless workspace: the full geo integration file passes (75 tests, including the empty-point and empty-reference cases).

simplegaurav:
@mwojtyczka , can you please review?

ghanse (approved):
LGTM

Issue #137, opened 2025-01-27
[FEATURE]: Document public methods of the engine


mwojtyczka:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Problem statement

We only have usage documented but not specification

### Proposed Solution

Provide documentation of the public methods

Issue #134, opened 2025-01-24
[FEATURE]: Streamline synchronization of config changes with the dashboard


mwojtyczka:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Problem statement

Currently, the quarantine table name used in the dashboard is provided during the installation (`quarantine_table` field from the "default" run config). If the run config is changed later, this is not reflected in the dashboard. User must manually update the query. 

### Proposed Solution

Possible solutions:
* Add option to specify quarantime table in the dashboard as a parameter
* Provide option to re-deploy dashboards via cli command, using the current config

Issue #131, opened 2025-01-24
[FEATURE]: Better documentation of the quality rules


mwojtyczka:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Problem statement

The current documentation just list functions and arguments but it may still be hard to use this in different contexts (e.g. check sdefined code and yaml/json).

### Proposed Solution

Provide documentation for check functions so it is more clear how to use them when specified in the code or as a file.

Issue #130, opened 2025-01-23
[FEATURE]: Improve DQX dashboard


mwojtyczka:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Problem statement

The current dashboard just show basic count over errors and warnings.

### Proposed Solution

Improve the visualisations and create new useful queries.

PR #1515, opened 2026-09-07
fix(app): DQX Studio bug bash — scheduling, results, settings, reset, and scheduled-run access

OGordon100:
## DQX Studio bug bash (bug-bash-6)

A batch of DQX Studio fixes off `main`, executed via subagent-driven development (each change implemented, code-reviewed, and fix-looped; plus whole-branch reviews). Deployed and verified on a live workspace.

### Scheduling
- Human-readable cron now renders **any** day-of-week (e.g. `0 3 * * 0` → "Runs weekly on Sunday at 03:00"), not just Monday; removed the raw "five fields…" helper and the "Full table is the default…" hint; scan-scope is now the second option; added a **Frequency** heading; consistent dropdown styling.

### Results & thresholds
- Each run now **retains the pass threshold in effect when it executed** — changing a threshold no longer retroactively re-judges historical runs (#129).
- The **frozen per-run threshold is surfaced** ("threshold used: N%") in the By-rule drilldown (#130).

### Rules / tables / collections UX
- Bulk-action overlay bar: round only the top corners.
- Collections subtitle copy → "Group related tables into data products to consolidate owners, scheduling, and results."
- Table pickers show short schema/table names instead of the full `catalog.schema.table`.
- Removed the **"Describe a rule"** flow; the inline "create a new one" is now a bold link (#134).
- Removed the **run** and **save/approve** confirmation modals — actions fire directly (reject keeps a plain confirm) (#137).
- Removed the free-text **"or paste a table name"** FQN box from the table picker (#135).

### Home & settings
- Homepage tile relabeled **"Overall Score"** (was cropping) with the help "?" aligned inline.
- Settings → Compute: dropped the separate "use the app default" option; the default warehouse row is labelled **"<name> (default)"**.
- Added **descriptions to severity levels** (Critical/High/Medium/Low), with a startup backfill for already-seeded deployments.

### Reliability
- **Demo runs no longer fail:** demo `is_in_list` enums were unquoted (treated as columns); now SQL-quoted. Demo-deploy seed now **fails fast** on a broken run instead of grinding for hours (fixes "demo deployment never ends").
- **"Reset DQX Studio" no longer spins forever:** reset now runs asynchronously (background thread + status endpoint + UI polling), with a client timeout and a 409 guard so reset/demo-seed can't collide.

### Scheduled-run access (#131)
- When a schedule is set up on a table (or every table in a collection), the app grants `SELECT` on the **underlying Unity Catalog table** — via the user's OBO token — to the service principals a scheduled run reads as (the app SP, plus the task-runner SP), so scheduled runs can read the data (manual "Run now" already worked via OBO views).
- Gated on Unity Catalog **MANAGE/ownership** (including via group): a user without it is **hard-blocked** from saving the schedule and shown a warning naming the users/groups who can set it up.

### Deferred / follow-up (non-blocking)
- Adding a collection member and setting its schedule in the same save leaves the new table ungranted (normally separate saves).
- A few orphaned i18n keys; run-gating on unapproved threshold edits (the per-run freezing above covers the correctness concern).


app/src/databricks_labs_dqx_app/backend/services/reset_status.py
```diff
@@ -0,0 +1,154 @@
+"""Settings-backed store for the long-running "Reset database" job status.
+
+Mirrors :mod:`backend.demo.status`: the reset now runs on a background daemon
+thread (32 cross-backend DELETEs plus a full Genie space reprovision can outlive
+the Databricks Apps gateway idle timeout), so the UI polls this store to drive
+its spinner and terminal toast rather than blocking on the request. The status is
+persisted as a JSON blob under *RESET_STATUS_KEY* in the *dq_app_settings*
```
mwojtyczka, line 7: **Reset status lives in the table the reset wipes.** Reset status is stored in `dq_app_settings`, but the reset itself `DELETE`s that table (`dq_app_settings` is in `OLTP_TABLE_NAMES`) early in `reset_all_data()`. From that point until the thread writes `succeeded`, the running row is gone, which breaks three things:

1. `ResetStatusStore.get()` returns the idle default → `is_running()` is False → a concurrent second reset or demo-deploy is no longer blocked by the new 409 guard (the exact race it was added to prevent).
2. The Settings poll sees running→idle, and `useResetStatus`'s `refetchInterval` only continues while `state === 'running'`, so polling stops and never observes the final `succeeded` — spinner stops early, dialog becomes closable mid-reset.
3. The success/failed toast + global `queryClient.invalidateQueries()` never fire, leaving stale data.

The code documents both halves without reconciling them (this module says status lives in `dq_app_settings`; the reset service clears `dq_app_settings`). This undercuts the PR's headline reliability fix — consider storing reset status outside the wiped table.

app/src/databricks_labs_dqx_app/ui/routes/_sidebar/monitored-tables.$bindingId.tsx
```diff
@@ -1256,7 +1222,7 @@ function RunTableAction({
       <TooltipTrigger asChild>
         <span className={cn(draftDisabled && "cursor-not-allowed")}>
           <Button
-            onClick={() => promptScope({ source: "draft" })}
+            onClick={() => void handleRunDraft(0)}
```
mwojtyczka: **Silent full-table-scan regression from removing `RunSampleDialog`.** Previously 'Run draft' opened `RunSampleDialog` with `defaultKind='records'` (1000-row sample) and 'Run now' let the user pick a sample. Now the buttons call `handleRunDraft(0)` / `handleRun('approved', 0)` where `0` = full table. On a large monitored table a draft spot-check now scans the entire table every time, with no way to sample a manual run. The PR frames this as removing 'confirmation modals', but this modal also carried the sample-size choice. Same change in `ProductHeader.tsx`.

app/src/databricks_labs_dqx_app/backend/routes/v1/schedules.py
```diff
@@ -26,6 +32,53 @@
...
+    (idempotently) to the scheduler SPs otherwise. A no-op for dormant schedules
+    or scopes that resolve to no tables.
+    """
+    if not _schedule_is_enabled(config):
+        return
+    target_fqns = svc.resolve_scope_table_fqns(config)
+    if not target_fqns:
+        return
+    blocked: list[tuple[str, list[dict[str, str]]]] = []
+    for fqn in target_fqns:
```
mwojtyczka: **O(N) synchronous UC round-trips in one save for `scope_mode='all'`.** For a scope-config schedule with `scope_mode='all'` (or a broad catalog/schema), `resolve_scope_table_fqns` returns every approved-rule table in the workspace. `_enforce_scheduler_grants` then calls `user_can_manage(fqn)` (up to 3 owner GETs + paginated `grants.get_effective`) and `grant_select_to_schedulers(fqn)` for each. On a workspace with hundreds of approved tables this is O(N) blocking SDK calls in a single save — can exceed the new 300s axios timeout, re-runs on every save, and (unlike the table/collection pages) has no preflight, so a single unmanageable table hard-blocks saving an all-scope schedule with no pre-block.

app/src/databricks_labs_dqx_app/backend/routes/v1/admin.py
```diff
@@ -90,6 +125,16 @@ def reset_database(
             detail="Confirmation phrase does not match. Type the exact phrase to confirm the reset.",
         )
 
+    # Mutual exclusion: reset and demo deploy share the SP warehouse + Lakebase,
```
mwojtyczka: **Non-atomic check-then-set (TOCTOU) in the reset/demo mutual-exclusion 409.** Two near-simultaneous requests can both pass `is_running()` before either persists `running`: both call `is_running()`/`demo_status_store.is_running()` (both False), both proceed, both set `running`, and two destructive operations run concurrently against the shared SP warehouse + Lakebase — the exact collision the guard intends to prevent. Compounded by the `dq_app_settings`-wipe finding, which makes the guard read False for the whole reset duration.

app/src/databricks_labs_dqx_app/backend/services/schedule_grant_service.py
```diff
@@ -0,0 +1,393 @@
...
+
+        Prefers ``DATABRICKS_CLIENT_ID`` (injected into a deployed Databricks
+        App) and falls back to the SP's own SCIM ``me()`` for local dev.
+        """
+        env_id = (os.environ.get("DATABRICKS_CLIENT_ID") or "").strip()
+        if env_id:
+            return env_id
+        try:
+            me = self._sp_ws.current_user.me()
+            return (me.user_name or me.id or "").strip()
```
mwojtyczka, line 142: **`app_sp_id()` fallback may return the wrong principal.** When `DATABRICKS_CLIENT_ID` is unset (local dev, or a deployment where the env var is absent), the fallback returns `me.user_name` or `me.id`. For a service principal the SCIM `user_name` may not be the application (client) id UC expects as a grant principal, so `grant_select_to_schedulers` either grants to the wrong principal or the `grants.update` call fails — and since the app-SP grant is 'essential', the whole schedule save 502s.

app/src/databricks_labs_dqx_app/backend/services/dq_results_service.py
```diff
@@ -295,9 +295,19 @@ class _GroupAcc:
...
+    # can pool rows from several runs whose frozen thresholds differ; we surface
+    # the threshold stamped on the NEWEST run in the group (by run_date), which
+    # mirrors how the by-rule group takes its display label from the newest run.
+    # Stays None when no contributing row carried a frozen threshold (legacy runs).
+    pass_threshold: int | None = None
+    threshold_run_date: str = ""
 
     def add(self, row: CheckResultRow, resolve: ThresholdResolver | None = None) -> None:
         self.failed += row.failed
+        if row.pass_threshold is not None and (row.run_date or "") >= self.threshold_run_date:
```
mwojtyczka: **Group display-threshold latches on an empty `run_date` sentinel.** In `_GroupAcc.add`, `threshold_run_date` starts as `''`. A contributing row with `pass_threshold` set but `run_date` None → `(row.run_date or '') >= ''` is True → it sets `pass_threshold` with `threshold_run_date=''`. Real-dated rows correctly win, but if the newest real run carries `pass_threshold=None` while an older undated row carried a value, the group surfaces the stale older threshold as 'threshold used', mislabelling which run's threshold the drilldown shows.

mwojtyczka (commented):
Automated code review (Isaac / Claude Code) — 6 findings, most load-bearing first. These are for discussion; verify before acting.

mwojtyczka (changes requested):
Left some comments

app/src/databricks_labs_dqx_app/ui/routes/_sidebar/monitored-tables.$bindingId.tsx
OGordon100: Will double check the logic here

app/src/databricks_labs_dqx_app/backend/services/reset_status.py
OGordon100: We don't need to keep track of resets; once it's done we don't care about knowing that it was previously done. Will double check that this functionality works as intended, however!

app/src/databricks_labs_dqx_app/backend/services/schedule_grant_service.py
OGordon100: This must always be set anyway, so not a concern

app/src/databricks_labs_dqx_app/backend/services/reset_status.py
OGordon100: Confirmed that functionality works correctly - no change needed

app/src/databricks_labs_dqx_app/ui/routes/_sidebar/monitored-tables.$bindingId.tsx
mwojtyczka: Fixed at be476ac — draft runs now fetch `useGetDraftRunSampleLimit()` and pass a non-zero/undefined `sample_size` (buttons call `handleRunDraft(draftSampleSize)`, no longer `0`). Approved runs intentionally scan the full table. Resolving.

app/src/databricks_labs_dqx_app/backend/routes/v1/schedules.py
mwojtyczka: Fixed at be476ac — grant enforcement is now bounded-concurrency async (`asyncio.Semaphore(8)`), caller/scheduler identities are primed once up front, and the grant path uses `grant_select_precleared_async` so MANAGE is checked exactly once per table. Resolving.

app/src/databricks_labs_dqx_app/backend/routes/v1/admin.py
mwojtyczka: The non-atomic check-then-set is fixed at be476ac — a shared `threading.Lock` (`_job_lock`) now wraps check-and-set for both reset and demo endpoints. Resolving this specific finding.

Caveat: mutual exclusion is still defeated by a separate mechanism — the `running` status lives in `dq_app_settings`, which `reset_all_data()` wipes mid-run (`DELETE FROM dq_app_settings`), so a second request arriving after the DELETE reads `is_running()==False` and passes the gate under the lock. That's tracked in the still-open reset_status.py thread.

app/src/databricks_labs_dqx_app/backend/services/dq_results_service.py
mwojtyczka: Fixed at be476ac — `threshold_run_date` now starts as `None` with a strict `>` comparison and 'newest run wins' semantics (adopts `None` from the newest run rather than latching a stale older threshold). Resolving.

app/src/databricks_labs_dqx_app/backend/services/schedule_grant_service.py
```diff
@@ -0,0 +1,457 @@
...
+        return user_name, principals
+
+    def _owners(self, fqn: str) -> list[str]:
+        """Return the owners of the table and its parent schema/catalog (lowercased).
+
+        An owner of any containing securable can grant on the table, so we walk
+        up the hierarchy. Each read is best-effort — a missing read simply
+        contributes no owner rather than failing the whole check.
+        """
+        catalog, schema, _table = fqn.split(".", 2)
```
mwojtyczka, line 235: **Service principals are misclassified as `group` in the MANAGE-holder warning.** `_classify_principal` returns `"user" if "@" in principal else "group"`, so a service principal that holds MANAGE (application id / UUID — no `@`) is labelled `group`. The ScheduleGrantWarning card then tells a blocked user to ask that 'group' to set up the schedule — misleading guidance pointing at a non-human principal that can't act on it. Consider detecting SPs (application-id/UUID shape, or a servicePrincipals SCIM lookup) and labelling them accordingly.

app/src/databricks_labs_dqx_app/backend/routes/v1/data_products.py
```diff
@@ -234,8 +248,31 @@ def update_data_product(
...
     )
+    updates = body.model_dump(exclude_unset=True)
+
+    # Only gate/grant when a schedule is actually being set/enabled on the
+    # collection — a non-empty cron. Clearing or unrelated partial updates skip.
+    if (updates.get("schedule_cron") or "").strip():
+        member_fqns = svc.member_table_fqns(product_id)
+        blocked: list[tuple[str, list[dict[str, str]]]] = []
+        for fqn in member_fqns:
+            if not grant_svc.user_can_manage(fqn):
```
mwojtyczka, line 262: **`user_can_manage(fqn)` is computed twice per member table.** The precheck loop calls `user_can_manage(fqn)` here to build the blocked set, then `grant_select_to_schedulers(fqn)` (a few lines below) re-runs the same MANAGE check internally — 2×N Unity Catalog round-trips on the save path, each doing owner reads + paginated `grants.get_effective`. The schedule-save path in `schedules.py` was refactored to check MANAGE once via `grant_select_precleared_async`; this collection path still uses the non-precleared call. Consider a precleared grant / pre-verified flag here too.

app/src/databricks_labs_dqx_app/ui/routes/_sidebar/settings.tsx
```diff
@@ -1538,40 +1541,51 @@ function WarehouseSelect({
...
   onChange,
   warehouses,
+  defaultWarehouseId,
   disabled,
 }: {
   value: string;
   onChange: (value: string) => void;
   warehouses: WarehouseOut[];
+  defaultWarehouseId?: string;
   disabled?: boolean;
```
mwojtyczka, line 1555: **No way to clear an explicit warehouse override back to the effective default.** Unlike `ServingEndpointSelect` (which has a `NO_ENDPOINT_VALUE` sentinel), `WarehouseSelect` offers only concrete warehouse ids and `handleWarehouseChange` always writes a concrete id — there is no option that maps to empty / 'follow the effective default'. `selectedId = value || defaultWarehouseId || ""` falls back to the default for *display*, but once an admin picks a warehouse the `value` can never be cleared, so the selection is pinned and won't track future default changes. Consider a sentinel 'use default' option that writes an empty override.

app/src/databricks_labs_dqx_app/ui/routes/_sidebar/settings.tsx
```diff
@@ -2363,6 +2476,7 @@ function ComputeSettingsCard() {
           value={warehouseId}
           onChange={handleWarehouseChange}
           warehouses={warehouses}
+          defaultWarehouseId={settings?.effective_warehouse_id ?? undefined}
           disabled={!isAdmin || saveMutation.isPending}
         />
```
mwojtyczka, line 2586: **A fast reset can fire no toast and skip cache invalidation.** The terminal-state handler is gated by `if (prev !== "running") return;`, and polling only runs while `state === "running"` (`refetchInterval`). If the reset completes before the client ever observes `running` (idle→succeeded between 2s polls, or a near-instant reset), `prev` is never `"running"`, so `toast.success`/`toast.error` and this `queryClient.invalidateQueries()` never fire — the admin sees no outcome and stale caches aren't refreshed. Consider optimistically setting `running` on click, or handling a terminal state that was never preceded by an observed `running`.

app/src/databricks_labs_dqx_app/backend/services/schedule_grant_service.py
mwojtyczka: Fixed at the current head (`e07bdfd3`): `_classify_principal` now returns `service_principal` for UUID-shaped principals (`_UUID_RE.fullmatch`), so a service principal holding MANAGE is no longer mislabelled as a group in the warning card. Resolving.

app/src/databricks_labs_dqx_app/backend/routes/v1/data_products.py
mwojtyczka: Fixed at the current head (`e07bdfd3`): the save path now runs a single `user_can_manage(fqn)` precheck and grants via `grant_select_precleared(fqn)` (data_products.py:262/272), so MANAGE is verified once per member table instead of twice. Resolving.

app/src/databricks_labs_dqx_app/ui/routes/_sidebar/settings.tsx
mwojtyczka: Still open at the current head (`e07bdfd3`). `WarehouseSelect` (settings.tsx ~1540) writes only concrete warehouse ids (`onValueChange={(v) => onChange(v)}`), and `selectedId = value || defaultWarehouseId || ""` is display-only — there is still no "use default" sentinel like `ServingEndpointSelect`'s `NO_ENDPOINT_VALUE`. Once an admin picks a warehouse the override cannot be cleared back to tracking the effective default.

app/src/databricks_labs_dqx_app/ui/routes/_sidebar/settings.tsx
mwojtyczka: Still open at the current head (`e07bdfd3`). The terminal-state handler still bails with `if (prev !== "running") return;` (settings.tsx:2582) and polling only runs while `state === "running"` (`refetchInterval`, 2567). A reset that goes idle→succeeded between 2s polls (or when the `dq_app_settings` wipe makes `is_running` read False mid-run) is never observed as running, so `toast.success`/`toast.error` and `invalidateQueries` never fire.

mwojtyczka (changes requested):

mwojtyczka:
@OGordon100 all previous issues resolved, except 2 that are sitll open

mwojtyczka:
## Code review findings — profiler sampling (originally reviewed on #1519, now merged into this branch)

These findings come from reviewing the admin-configurable profiler sampling work (`dqx-studio/profiling`, merged via #1519), which is now part of this PR's diff. Most severe first.

### 1. `TABLESAMPLE` without `REPEATABLE` seed produces inconsistent samples
**`app/src/databricks_labs_dqx_app/backend/services/view_service.py`** (~L650)

`records`/`percent` samples are built as `TABLESAMPLE` inside a **non-materialized temp view** with no `REPEATABLE (seed)`. A view re-runs its query on every scan and `TABLESAMPLE` re-randomizes each time, so:
- `rows_profiled = df.count()` and the subsequent `profiler.profile(df)` see **different random rows**.
- This directly undermines the stated fix that "rows_profiled counts what the profiler actually reads."
- The profiler's multi-pass per-column statistics become mutually inconsistent across columns.

**Fix:** add `REPEATABLE (<seed>)` to the `TABLESAMPLE` (or materialize the sample) so every scan sees the same rows.

### 2. Profiler Run button not gated on `sampleHydrated`
**`app/src/databricks_labs_dqx_app/ui/routes/_sidebar/profiler.tsx`** (~L1006)

During the `useGetProfilerSample()` hydration window, the SampleSelector shows its placeholder but `sampleOverride` is `{}`. The Run button is only disabled on `selectedTables.length === 0 || !canRunRules`. A user who clicks Run in that window sees "Random sample / 10 percent" while the backend silently resolves the admin setting (possibly full-table or records/50000). The monitored-tables modal fixed exactly this by disabling confirm until hydrated; the Profiler page did not.

### 3. `sample_limit=0` cannot distinguish a full scan from a percent sample
**`app/src/databricks_labs_dqx_app/backend/routes/v1/profiler.py`** (~L286)

`recorded_sample_limit` returns 0 for both `full` and `percent` runs (and runner.py independently sets `sample_limit = sample_value if sample_kind == 'records' else 0`). A 10%-of-1M-rows run and a whole-table run both persist `sample_limit=0` in `dq_profiling_results`, so a reader (or ETA logic keying on prior runs) cannot tell a sampled run from a full scan. `rows_profiled` is the only differentiator — and per finding #1 it is unreliable for sampled runs.

### 4. records → percent switch clamps to 100% (silent full scan)
**`app/src/databricks_labs_dqx_app/ui/routes/_sidebar/settings.tsx`** (~L1655)

In `ProfilerSampleSettings.handleKindChange`, switching from records (e.g. 50000) to percent computes `Math.min(100, Math.max(1, 50000)) = 100` and immediately `handleSave('percent', 100)`. 100% is the whole table (the opposite of a cap), persisted without the admin re-confirming a sensible percentage — the very "cap silently widens to full scan" hazard the backend `model_validator` was added to prevent, reached here through the unit switch.

### 5. Percent selector still labeled "Sample Limit (rows)"
**`app/src/databricks_labs_dqx_app/ui/routes/_sidebar/profiler.tsx`** (~L973)

The label above the records/percent/full selector reads `profiler.sampleLimit` = "Sample Limit (rows)" in every locale. When the user picks the percent unit, the control is labeled as a row count, which is misleading. The label key should be neutral (e.g. "Sampling").

### 6. Demo duplicates sample-options pinning instead of the shared helper
**`app/src/databricks_labs_dqx_app/backend/demo/seed_service.py`** (~L97)

The demo hard-codes `profile_options={'sample_fraction': None, 'limit': 0}` inline instead of calling the new shared `routes/v1/profiler.sample_profile_options()` helper. If the pinning convention changes, the demo path will silently drift back to DQX core defaults (~300 rows). Call the shared helper instead of copying its output.

---

_Line numbers are approximate (the review was run against #1519's head via the GitHub API rather than a local checkout)._

mwojtyczka:
## Code review findings — DQX Studio bug bash (#1515)

Full-PR review, ranked by severity. (Finding #1 overlaps with the profiler-sampling comment above and is included here for completeness.)

### 1. `TABLESAMPLE` in a non-materialized view re-samples on every read
**`app/src/databricks_labs_dqx_app/backend/services/view_service.py`**

`records`/`percent` profiles build `CREATE OR REPLACE VIEW ... AS SELECT * FROM src TABLESAMPLE (p PERCENT) [LIMIT n]`. In `runner._run_profile`, `rows_profiled = df.count()` and the subsequent `profiler.profile(df)` each re-execute the view with a fresh Bernoulli seed, so `rows_profiled` disagrees with the rows actually profiled, per-column stats are computed over inconsistent samples, and a `records` run can return fewer than `n` rows. The old code used a plain `LIMIT` and did not reseed. **Fix:** materialize the sample (or fix a seed) so every scan sees the same rows.

### 2. MANAGE-via-group detection can falsely hard-block schedule saves
**`app/src/databricks_labs_dqx_app/backend/services/schedule_grant_service.py`**

`user_can_manage`'s `principal=user_name` loop only matches *direct* grants. Group-inherited MANAGE is caught only by the `principal=None` full enumeration — but a user who holds MANAGE only via group membership often cannot list all grants on the table. `_manage_assignments` swallows that error and returns `[]`, so the method returns `False` and the schedule save is hard-blocked (403) even though the user could actually grant. This defeats the stated Task-12 goal of supporting MANAGE-via-group.

### 3. Hard-block warning banner renders even when the user isn't scheduling
**`app/src/databricks_labs_dqx_app/ui/components/monitored-tables/MonitoredTableSchedulingTab.tsx`** (same pattern in `ProductSchedulingTab.tsx`)

The "You can't set up this schedule" banner is `banner={cannotManage ? ... : undefined}`, so it renders whenever any table is unmanageable — even when `cron === null`. The actual save gate (`blockForSchedule`) is correctly gated on cron, but the banner is not, so a view/modify user opening the tab on an unscheduled table sees an alarming hard-block for a schedule they aren't setting up.

### 4. Stale comment: draft-run default is now 0, not 1000
**`app/src/databricks_labs_dqx_app/backend/services/binding_run_service.py`** (same stale wording in `ProductHeader.tsx` and `monitored-tables.$bindingId.tsx` `RunTableAction`)

The comment reads "the admin-configured draft-run sample limit (compiled-in default of 1000 when unset)", but `DRAFT_RUN_SAMPLE_LIMIT_DEFAULT` was changed to `0` (whole table) in this PR. A maintainer reading it will believe unset draft runs sample 1000 rows when they actually scan the full table.

### 5. Reset completion toast can be missed on a fast reset
**`app/src/databricks_labs_dqx_app/ui/routes/_sidebar/settings.tsx`**

The reset success/failure toast only fires on a `running → terminal` edge (`if (prev !== "running") return`). `prevResetStateRef` starts undefined, and `refetchInterval` polls every 2s only while `state === 'running'`. If the background reset completes (or the first poll lands) after the state already flipped to `succeeded`/`failed`, `prevResetStateRef` never held `running`, so the toast is skipped and the global `invalidateQueries` refresh never runs — the admin gets no completion feedback.

### 6. Schedule-grant preflight fires on every scheduling-tab mount
**`app/src/databricks_labs_dqx_app/ui/components/monitored-tables/MonitoredTableSchedulingTab.tsx`**

`preflightQuery` is `enabled: canEdit` (and the collection preflight is enabled whenever `memberFqns.length > 0`), so merely opening the Scheduling tab issues per-table ownership reads (table/schema/catalog) plus fully-paginated `grants.get_effective` calls. For a large collection that is many blocking SDK round-trips per open, done purely for a pre-block that only matters if the user actually sets a cron. Gate the preflight on an intent signal (cron non-null / editing).

---

_Line numbers omitted; findings are anchored by file and symbol. Reviewed a ~11k-line PR — focus was backend service logic (schedule grants, DB reset + status, profiler sampling, threshold freezing) and the highest-risk UI logic._

OGordon100:
Thanks — went through all 12 findings against the code. Verdicts below; **11 confirmed, 1 I think is based on a wrong premise.** Most of the confirmed ones are in my own profiler-sampling work.

## Profiler sampling (comment above)

| # | Verdict |
|---|---|
| 1 · `TABLESAMPLE` without `REPEATABLE` | **Confirmed — most serious of the set.** No seed, non-materialized view, and `df.count()` then `profiler.profile(df)` are separate scans, so each re-draws. This directly falsifies the claim in my commit message that `rows_profiled` "counts what the profiler actually reads", and a `records` run can return fewer than `n` rows. Will fix with `REPEATABLE`. |
| 2 · Run button not gated on `sampleHydrated` | **Confirmed.** I made `sampleOverride` `{}` while unhydrated so the backend resolves the setting, but left the display/actual mismatch you describe: the selector shows "10 percent" while the run uses the admin policy. The button is only gated on `selectedTables.length === 0 \|\| !canRunRules`. |
| 3 · `sample_limit=0` can't distinguish full from percent | **Confirmed.** `recorded_sample_limit` returns 0 for both, and per #1 `rows_profiled` isn't a reliable differentiator either. |
| 4 · records → percent clamps to 100% | **Confirmed — a real bug, and my code.** `Math.min(100, Math.max(1, 50000))` is 100, saved immediately. My comment says "clamp into the new unit's range"; clamping lands on the maximum, which is the whole table — the exact hazard the backend validator exists to prevent, walked in through the unit switch. |
| 5 · Label still "Sample Limit (rows)" | **Confirmed.** `profiler.tsx:973` uses `t("profiler.sampleLimit")` = "Sample Limit (rows)" in all five locales. I swapped the input for the selector and left the label. |
| 6 · Demo duplicates the pinning dict | **Confirmed.** `seed_service.py:455` hard-codes `{"sample_fraction": None, "limit": 0}`. |

## Bug bash (comment above)

| # | Verdict |
|---|---|
| 1 · `TABLESAMPLE` re-samples | Same as above — confirmed. |
| 2 · MANAGE-via-group false hard-block | **I think the premise is wrong — please push back if I'm missing something.** The finding assumes `get_effective(..., principal=user_name)` returns only *direct* grants. It returns **effective** privileges for that principal, which include group- and parent-securable inheritance — so the group case is caught by that call, before the `principal=None` enumeration is reached. That first call also reads only the caller's *own* permissions, so it isn't subject to the "can't list all grants" problem; that limitation applies to the `principal=None` fallback, which is belt-and-braces. Two comments in the file assert this (`_manage_assignments` docstring, and the `user_can_manage` comment "folds in parent + group inheritance"). If effective-permissions genuinely excludes group inheritance then the finding stands and it's a real hard-block bug — but I can't reproduce the failure path as described. |
| 3 · Banner renders when not scheduling | **Confirmed.** `banner={cannotManage ? …}` vs `blockForSchedule = cannotManage && cron !== null`. |
| 4 · Stale "default 1000" comments | **Confirmed, partially fixed.** I corrected the two in `monitored-tables.$bindingId.tsx` when changing the default but missed `binding_run_service.py:255` and `ProductHeader.tsx:313`. |
| 5 · Reset toast missed on fast reset | **Confirmed** — same finding as the still-open `settings.tsx` thread. Left deliberately for now, not overlooked. |
| 6 · Preflight fires on tab mount | **Confirmed.** `enabled: canEdit`, so opening the Scheduling tab issues the ownership + paginated `grants.get_effective` reads regardless of intent. |

Nothing here was already fully fixed; #2 and #4 of the second list were partially addressed. I'll follow up with the fixes rather than resolve anything on the strength of this comment alone.

_Written by Isaac._

app/src/databricks_labs_dqx_app/backend/routes/v1/schedules.py
OGordon100: Addressed — the work was made faster rather than smaller (`schedules.py:60-95`).

Caller and SP identities are now resolved once per request (`prime_caller_identity`, memoized on the grant service) instead of each gate check re-issuing `current_user.me()`, and the per-table MANAGE gate plus grant round-trips run concurrently under a bounded semaphore (`_GRANT_CONCURRENCY`). MANAGE is checked exactly **once** per table — the grant path trusts the gate's decision rather than recomputing it.

Coverage is deliberately not trimmed to a delta: re-saving is how a revoked or previously-failed grant gets re-detected and re-applied, so narrowing the set would silently lose that self-healing property.

app/src/databricks_labs_dqx_app/backend/routes/v1/admin.py
OGordon100: Fixed in `1d44a394`. Both handlers now perform the check-and-set under a single module-level `_job_lock`, so `is_running()` and `set(running)` can no longer interleave.

Worth noting it is the **same** lock in the demo and reset handlers, which is what makes it enforce demo⇄reset exclusion and not merely demo⇄demo. The lock is released before the ~30 minute seed runs on its daemon thread, so it guards only the decision, not the job.

app/src/databricks_labs_dqx_app/backend/services/dq_results_service.py
OGordon100: Fixed in `d4a7b6b2`. `threshold_run_date` now starts as `None` and the comparison is `self.threshold_run_date is None or row_date > self.threshold_run_date`.

The two sentinels are now distinct, which was the actual bug: `None` means nothing seen yet, `""` means an undated contributing row (which sorts oldest). Strict `>` keeps first-seen winning ties, matching the by-rule label's tie behaviour, and the newest contributing run's threshold is adopted even when it is `None` — that case matters when the pass-threshold feature is switched off after an earlier run had it on.

app/src/databricks_labs_dqx_app/backend/services/schedule_grant_service.py
OGordon100: Fixed in `4327b987`. `_classify_principal` now tests the principal against `_UUID_RE` and returns `"service_principal"` for a UUID-shaped value, so an application id is no longer described as a group.

The warning copy renders it through a `holderTypeLabel` in `ScheduleGrantWarning.tsx`, so the three kinds (user / group / service principal) each read correctly instead of the previous `"@" in principal` two-way split.

app/src/databricks_labs_dqx_app/backend/routes/v1/data_products.py
OGordon100: Fixed in `4327b987` by adding a public `grant_select_precleared()`. The precheck loop's MANAGE decision is now passed through to the grant path, so `user_can_manage(fqn)` is computed once per member table rather than twice.

Kept as a separate public method rather than a flag on the existing one so the "already gated" precondition is explicit at the call site.

app/src/databricks_labs_dqx_app/ui/routes/_sidebar/settings.tsx
OGordon100: Valid, and **not** addressed in this PR — flagging that plainly rather than implying otherwise.

The analysis is right: `handleWarehouseChange` only ever writes a concrete id, so once an admin picks a warehouse the override cannot be cleared and stops tracking future default changes. `selectedId = value || defaultWarehouseId || ""` masks it at display time, which is what makes it easy to miss.

The fix is the sentinel you describe, mirroring `NO_ENDPOINT_VALUE` in `ServingEndpointSelect` — a "use default" option that writes an empty override. Happy to add it here if you would rather it not ship as a follow-up.

app/src/databricks_labs_dqx_app/ui/routes/_sidebar/settings.tsx
OGordon100: Acknowledged as real, and deferred deliberately by the repo owner rather than disputed — recording that here so the thread is not mistaken for an oversight.

The race is as described: `if (prev !== "running") return;` combined with polling that only runs while `state === "running"` means an idle→succeeded transition between two 2s polls fires no toast and skips `invalidateQueries()`. The fix would be to set `running` optimistically on click, or to handle a terminal state never preceded by an observed `running`.

Left for a follow-up because the reset is an admin-only, low-frequency action and the missed toast has no data-correctness consequence — the caches refresh on the next navigation.

OGordon100:
Review pass addressed — replies are on each thread. Summary of what landed since the `CHANGES_REQUESTED` review:

| Finding | Status | Where |
|---|---|---|
| Reset status stored in the table the reset wipes | fixed | `reset_status.py` |
| `RunSampleDialog` removal → silent full scan | fixed | `monitored-tables.$bindingId.tsx` |
| O(N) sync UC round-trips on scope save | fixed (parallelised, MANAGE checked once) | `schedules.py` |
| TOCTOU in reset/demo 409 guard | fixed | `1d44a394` |
| `app_sp_id()` fallback wrong principal | fixed | `schedule_grant_service.py` |
| Threshold latches on empty `run_date` sentinel | fixed | `d4a7b6b2` |
| SPs misclassified as `group` | fixed | `4327b987` |
| `user_can_manage` computed twice | fixed | `4327b987` |
| Warehouse override cannot be cleared | **open** — valid, deliberately not in this PR | thread |
| Fast reset fires no toast | **open** — deferred by owner, no data-correctness impact | thread |

Also on the branch since the review: `TABLESAMPLE` now carries `REPEATABLE (seed)` (a plain view re-executes per scan, so each scan was drawing a different Bernoulli sample and `rows_profiled` did not describe the rows profiled); the records→percent unit switch no longer saturates to 100%; and `sample_kind` is persisted so a percent run is distinguishable from a full scan.

The two open threads are the only outstanding items — both flagged rather than quietly closed. `reviewDecision` is still `CHANGES_REQUESTED`, which resolving threads doesn't clear, so this needs a re-review when you have a moment.

mwojtyczka (approved):
LGTM

PR #1519, opened 2026-09-08
feat(app): admin-configurable profiler sampling

OGordon100:
## What

Adds an admin setting under **Settings → Compute → Profiling** controlling how much of a source table the profiler reads:

- **Full table**
- **Random sample** → **Records** (a row cap) or **Percent**

The two sampled forms are mutually exclusive by construction — the kind decides which unit the value carries. Default is **10%**.

The setting is the default for **both** profiling entry points:

- **Profiler page** — pre-filled from the setting, overridable per run.
- **Monitored table → Profile tab** — one-off runs (`Refresh profile` and the empty-state `Run profile`) now ask for their sampling in a modal, instead of silently using a hard-coded 50,000.

## Two pre-existing defects fixed along the way

**1. Profiling only ever saw ~300 rows.** The app never sent `limit` / `sample_fraction`, so DQX core's `DEFAULT_PROFILE_OPTIONS` filled in `sample_fraction: 0.3` and `limit: 1000` (`profile_options.py:74,79`) — every profile run was capped at 1,000 rows then sampled to ~300, regardless of the 50,000 the UI asked for. Both keys are now always sent explicitly so core defaults cannot leak in.

**2. `rows_profiled` over-reported.** It counted the DataFrame *before* core's sampling, so a run reported "50,000 rows profiled" when the profiler had seen ~300. It now counts what the profiler actually reads.

## Random sampling is genuinely random

A row cap does **not** use `TABLESAMPLE (n ROWS)` — Spark implements that as a plain `LIMIT`, returning the *first* n rows. Instead the temp view over-samples by percentage (derived from a metadata-fast `COUNT(*)`, with a 1.5x margin) and caps with `LIMIT`:

```sql
-- 50k of 1M rows
SELECT * FROM `c`.`s`.`t` TABLESAMPLE (8 PERCENT) LIMIT 50000
```

If the count is unavailable it falls back to a bare `LIMIT` and logs a warning — a non-random sample beats a failed run. A percentage sample maps straight onto `TABLESAMPLE (p PERCENT)` (true Bernoulli) and never pays for a count.

## DQ runs are deliberately unaffected

Scheduled runs and dry runs still scan the full table, so their pass rates describe the table rather than a subset — capping them would quietly corrupt threshold/pass-rate logic. A regression test pins this.

## Implementation notes

- Two keys in the existing `dq_app_settings` key-value table (`profiler_sample_kind`, `profiler_sample_value`) — **no migration needed**.
- The value fallback is **per-kind**: a single global default is unsafe across units, since 50,000 clamped into percent units becomes 100% — the whole table, the opposite of a cap. Covered by a test.
- The sampling SQL builder is a **pure function** (`build_sample_select`), with the row-count I/O kept at the `ViewService` boundary.

## Testing

- 49 new tests in `app/tests/test_profiler_sample_setting.py` covering storage round-trips, corrupt/out-of-range fallbacks, the SQL shape for every kind, option pinning, route validation, and the DQ-runs-unsampled guard.
- Updated `tests/unit/test_app_backend.py` for the new `create_view` signature.
- `make app-check` clean (tsc, basedpyright 0 errors, 837 UI tests); `make app-test` 4102 passed; root `tests/unit` 2497 passed; `make fmt` / `make lint` 10.00/10.
- Deployed and exercised end-to-end against a live workspace: settings round-trip across all four states, both validation paths return 400, and the modal works from both buttons.

## Not covered

No profile run has been observed end-to-end against real data, so the honest `rows_profiled` figure is verified by unit test rather than in practice.


app/src/databricks_labs_dqx_app/backend/routes/v1/compute.py
```diff
@@ -165,6 +172,63 @@ def _settings_out(app_settings: AppSettingsService) -> ComputeSettingsOut:
...
+
+    sample_kind: Literal["full", "records", "percent"]
+    sample_value: int | None = Field(default=None, ge=1)
+
+
+@router.get(
+    "/profiler-sample",
+    response_model=ProfilerSampleOut,
+    operation_id="getProfilerSample",
+    dependencies=[require_role(UserRole.ADMIN)],
```
mwojtyczka: **[correctness] `GET /profiler-sample` is ADMIN-only, but non-admin runners need it.**

This GET is gated on `require_role(UserRole.ADMIN)`, yet `RULE_AUTHOR` is in `CAN_RUN_ROLES` and can run profiles. The profiler UI calls `useGetProfilerSample()` unconditionally, so for a `RULE_AUTHOR` the request 403s → `profilerSampleQuery.data` stays undefined, the hydration effect never fires, and `sampleKind`/`sampleValue` stay at the hardcoded `percent`/`10`. Net effect: every profile run by a non-admin silently ignores the admin-configured default and samples 10%.

The GET should be readable by `CAN_RUN_ROLES` (Admin + Author); only the PUT below needs to stay ADMIN-only.

app/src/databricks_labs_dqx_app/backend/routes/v1/profiler.py
```diff
@@ -45,6 +52,49 @@
...
+    """Resolve the sampling policy for a profile run.
+
+    A request that names a *sample_kind* overrides the admin setting for
+    that run; anything else falls back to the configured default. This is
+    the single place the precedence is decided, so the single-table and
+    batch routes cannot drift apart.
+    """
+    if body.sample_kind is None:
+        return app_settings.get_profiler_sample()
+    return ProfilerSample(kind=body.sample_kind, value=body.sample_value or 0)
```
mwojtyczka: **[correctness] Per-run sample overrides bypass the clamping the admin path enforces.**

`resolve_sample` returns `ProfilerSample(kind=body.sample_kind, value=body.sample_value or 0)` with no bounds, and `ProfilerSampleOverride.sample_value` only carries `ge=1` (no `le`), despite its description saying "1-100 when percent". So a client (stale UI or direct API caller) can POST `{sample_kind: "percent", sample_value: 5000}` → `build_sample_select` emits `TABLESAMPLE (5000 PERCENT)`. Bernoulli sampling with fraction > 1 includes every row, so this silently returns the *whole table* — the opposite of a cap.

The admin PUT rejects `percent > 100` and `get_profiler_sample` clamps stored values, but this per-run path enforces neither, so the "single place precedence is decided" comment does not hold the same invariants. Consider clamping percent to 1-100 (and bounding records) here, or adding `le` on the model.

app/src/databricks_labs_dqx_app/ui/routes/_sidebar/monitored-tables.$bindingId.tsx
```diff
@@ -1743,9 +1746,46 @@ function ProfileTab({
...
   const running = submitMutation.isPending || runId !== null;
 
-  const handleRunProfile = useCallback(() => {
+  // One-off profile runs ask for their sampling in a modal, pre-filled from the
+  // admin default (Settings → Compute → Profiling). Confirming overrides that
+  // default for this run only; it never writes the setting back.
+  const [sampleDialogOpen, setSampleDialogOpen] = useState(false);
+  const [sampleKind, setSampleKind] = useState<SampleKind>("percent");
+  const [sampleValue, setSampleValue] = useState(10);
+  const [sampleHydrated, setSampleHydrated] = useState(false);
```
mwojtyczka, line 1755: **[correctness] Sampling modal can submit `percent`/10 before the admin default hydrates.**

`sampleKind`/`sampleValue` initialise to `percent`/`10` and `sampleHydrated` starts `false`; the admin default only arrives once `profilerSampleQuery` resolves. A user who opens the modal and confirms before hydration (or, per the profiler-sample 403 for non-admins, when the query never resolves) submits `sample_kind="percent", sample_value=10` — overriding the admin-configured policy (e.g. `records`/50000) with 10%. Disabling the confirm action until `sampleHydrated` is `true` would prevent the wrong override.

app/src/databricks_labs_dqx_app/backend/services/view_service.py
```diff
@@ -54,12 +126,47 @@ def _ensure_schema(self) -> None:
...
         Returns the fully qualified view name.  The view is created using
         the caller's OBO token so that the user's table permissions are
         enforced.
+
+        Args:
+            source_table_fqn: Fully qualified source table.
+            sample: Optional sampling policy. ``None`` (the default) and the
+                ``full`` kind both select the whole table; ``percent`` and
+                ``records`` narrow it to a *random* subset. See
+                :meth:`_sample_body` for the SQL shape.
```
mwojtyczka: **[minor] Stale symbol reference in the docstring.**

The `create_view` docstring points at `:meth:`_sample_body``, but that method does not exist — the SQL shape is built by `_sample_select` (this class) / `build_sample_select` (module-level). `runner.py` carries the same stale reference (`ViewService._sample_body`). Update the cross-references to the real symbols so they resolve.

mwojtyczka (changes requested):
generally looking good, left 3 comments that need to be addressed

app/src/databricks_labs_dqx_app/backend/routes/v1/compute.py
OGordon100: Fixed at 62b2ec13 — the GET now uses `require_role(*CAN_RUN_ROLES)`; the PUT stays ADMIN-only. Confirmed the failure mode you described: an author 403s, the hydration effect never fires, and the run silently uses the UI placeholder.

Added tests asserting the GET admits `RULE_AUTHOR`, excludes `VIEWER`, and the PUT stays admin-only, using the route-introspection pattern from `test_data_products_routes.py`.

app/src/databricks_labs_dqx_app/backend/routes/v1/profiler.py
OGordon100: Fixed at 62b2ec13, though not quite as suggested — clamping percent to 1-100 does not actually remove the harm, because `TABLESAMPLE (100 PERCENT)` is still the whole table. Clamping 5000 to 100 produces exactly the full scan the finding is about, just via a legal value.

So an out-of-range percentage is now **rejected**: a `model_validator` on `ProfilerSampleOverride` returns 422, mirroring the 400 the admin PUT returns. A plain `le` cannot express it, since `sample_value` carries two different units and the bound depends on `sample_kind`.

`resolve_sample` also clamps with the same (now public) helper the stored setting uses, as defence in depth for internal callers that bypass the model. That surfaced an adjacent bug: a kind supplied without a value collapsed to 0, clamped to 1 — a 1-row/1% sample. It now falls back to that kind default.

Verified live: `POST /profiler/run` with `{"sample_kind":"percent","sample_value":5000}` returns 422.

app/src/databricks_labs_dqx_app/ui/routes/_sidebar/monitored-tables.$bindingId.tsx
OGordon100: Fixed at 62b2ec13 — the confirm action is disabled until `sampleHydrated`, as suggested.

Also applied the same guard to the Profiler page, which had the identical window: both entry points now omit the override entirely while unhydrated, so the backend resolves the configured setting instead of receiving a placeholder nobody chose.

app/src/databricks_labs_dqx_app/backend/services/view_service.py
OGordon100: Fixed at 62b2ec13 — stale reference from the pure-function refactor. The docstring now points at `build_sample_select`, and the `runner.py` comment was updated to `view_service.build_sample_select`. No `_sample_body` references remain.

PR #1522, opened 2026-09-09
Fix Studio warehouse errors and Genie provisioning

laurencewells:
## Changes

- render structured API failures as actionable messages when changing the SQL warehouse
- provision the managed DQX Genie space under `/Shared/dqx-studio`
- create the shared workspace folder idempotently and recover safely from stale or inaccessible stored space IDs
- restrict Genie-space reuse to candidates in the configured parent folder

### Linked issues

None.

### Tests

- [x] manually tested
- [x] added unit tests
- [ ] added integration tests
- [ ] added end-to-end tests
- [ ] added performance tests

Validated with 3,996 passing backend tests, UI regression tests, Ruff, and basedpyright. The deployed app completed startup and created the Genie space under `/Shared/dqx-studio`.

### Documentation and Demos

- [ ] added/updated demos
- [ ] added/updated docs

No documentation or demo changes are required for these bug fixes.


mwojtyczka:
## Code review findings — Genie space provisioning (#1522)

Ranked by severity. Findings #1–#3 share one underlying cause: the verify-before-trust change treats a *transient* API failure as an authoritative "not found"/"error" rather than distinguishing it from `NotFound`.

### 1. Transient verify failure creates a duplicate space and orphans the original
**`app/src/databricks_labs_dqx_app/backend/services/genie_space_service.py`**

In steady state (`SETTING_SPACE_ID` points at a live space, hash matches), a transient blip on the new verify `GET /genie/spaces/{id}` (timeout / eventual-consistency 5xx) sets `existing=None`. `_find_space_id_by_title` then lists and, due to paging/replication lag, finds no match in `/Shared/dqx-studio` → POST creates a **second** space. The original is orphaned and `SETTING_SPACE_ID` is overwritten with the new one. This is exactly the path exercised by `test_replaces_unverifiable_stored_id_after_clean_list`.

### 2. Two Studio deployments in one workspace collide on the shared space
**`app/src/databricks_labs_dqx_app/backend/services/genie_space_service.py`**

With a fixed `SPACE_PARENT_PATH` (`/Shared/dqx-studio`) and fixed `SPACE_TITLE`, App B (catalog Y) booting with no stored id matches App A's space by title+parent and adopts it — but the reuse path never PATCHes `serialized_space`. It stores that id + `desired_hash(Y)` and marks READY, so B's Genie space queries **catalog X's** tables. Moving from per-SP `/Users/<sp>` parents to a single shared folder makes this collision deterministic rather than incidental.

### 3. Resilience regression: a transient blip marks Genie unavailable
**`app/src/databricks_labs_dqx_app/backend/services/genie_space_service.py`**

The formerly free no-op path (stored id + matching hash → instant return) now depends on a verify GET. A transient error on both the verify GET and the subsequent list GET raises `_SpaceLookupError` → `STATUS_ERROR`, `return None` (`test_stored_space_verification_failure_does_not_create_duplicate`). Before this PR the same boot was an instant no-op immune to API blips; now a momentary hiccup marks the Genie feature unavailable until the next successful startup.

### 4. Structured-error helper wired into only one file → `[object Object]` toasts
**`app/src/databricks_labs_dqx_app/ui/lib/api-error.ts`**

The new canonical `extractApiError` (handles structured `{summary}` details) was wired only into `settings.tsx`. ~14 other UI files (`PermissionsTab.tsx`, `RegistryRuleFormDialog.tsx`, `ProductHeader.tsx`, `monitored-tables.*`, `collections.*`) keep a string-only local copy: `return err?.response?.data?.detail ?? fallback`. When one of their endpoints returns the new structured `detail` object, `detail` is truthy so `?? fallback` is skipped and the object is passed to `toast.error`, rendering `[object Object]`. The shared helper's purpose (render-safe structured errors) is duplicated-away and inconsistent.

### 5. Serial detail-GET per title match; a single flap aborts provisioning
**`app/src/databricks_labs_dqx_app/backend/services/genie_space_service.py`**

`_find_space_id_by_title` issues one sequential detail `GET /genie/spaces/{id}` per title-prefix match across all parents on every no-stored-id boot, to read `parent_path`. A workspace that accumulated many "DQX Studio — DQ Results …" spaces (repeated deploys/resets) yields a large `matches` list fetched serially; cost scales with match count, and if any single detail GET flaps, `raise _SpaceLookupError` discards all work and skips provisioning for that boot — even though a valid candidate may exist later in the list.

### 6. Redundant nested try/except (cleanup)
**`app/src/databricks_labs_dqx_app/backend/services/genie_space_service.py`**

The inner `except Exception: raise _SpaceLookupError from error` wrapping the detail GET is fully subsumed by the function's outer `except Exception: raise _SpaceLookupError from error`. Deleting the inner try/except produces identical behavior — the outer handler catches the raw error and wraps it the same way — so it is extra nesting for no behavioral gain.

---

_Line numbers omitted; findings are anchored by file and symbol._

mwojtyczka (changes requested):
Left some comments

laurencewells:
Addressed findings #1–#3 and #5–#6 in the latest push. Finding #4 requires a wider UI-wide error-handling cleanup, so I have split that into follow-up issue #1528.

mwojtyczka:
## Re-review — previous findings closed, three new items

Thanks — verified the latest push (`f4dfb97`). **Previous findings #1–#3 and #5–#6 are addressed, and #4 is correctly deferred to #1528.** Confirmed in code:

- **#1/#3** — `ensure_dq_genie_space` now classifies the verify-GET error and returns the existing id on a transient (non-`NotFound`) failure instead of recreating / marking unavailable.
- **#2** — `space_parent_path()` is now per-app-client (`/Shared/dqx-studio/{client_id}`) and reuse requires a matching `parent_path`, so two deployments no longer collide.
- **#5** — the title lookup now skips a flapped detail-GET and still returns a valid later match, rather than discarding all work.
- **#6** — the inner `except` now does skip-and-continue, so it's no longer redundant.

Three new items introduced by the fix, ranked by severity:

### 1. `>20 pages → raise _SpaceLookupError` is a permanent provisioning failure
`_find_space_id_by_title`: after the paging cap, `if page_token: raise _SpaceLookupError`. On a workspace with >2000 Genie spaces the loop exits with `page_token` still set → raise → `ensure_dq_genie_space` sets `STATUS_ERROR` and returns `None`. Since no id is ever stored, this repeats on every startup and the space is never created. Consider degrading to create instead of hard-failing.

### 2. `PERMISSION_DENIED` treated as "space gone" can orphan the space and create a duplicate
In the verify branch, `error_code in {"NOT_FOUND", "PERMISSION_DENIED"}` clears `existing`. A transient/ACL permission blip on the verification GET therefore drops the stored id, the title lookup can't see the space either (no permission), and a brand-new space is created — orphaning the original. `NOT_FOUND` deserves recreate; `PERMISSION_DENIED` is not evidence the space is gone — suggest treating it like the other transient errors (return `existing`).

### 3. `%`-style logging violates the repo convention
The two new `logger.info("… %s", type(error).__name__)` calls use `%`-style args; AGENTS.md requires f-strings for log messages, and every other logger call in this file already uses them.

app/src/databricks_labs_dqx_app/backend/services/genie_space_service.py
```diff
@@ -1664,14 +1678,35 @@ def _find_space_id_by_title(ws: WorkspaceClient, title: str) -> str | None:
...
+            try:
+                detail = ws.api_client.do("GET", f"/api/2.0/genie/spaces/{space_id}")
+            except Exception as error:
+                logger.info(f"Genie space detail lookup skipped: {type(error).__name__}")
+                unclassified_match = True
+                continue
+            if isinstance(detail, dict) and detail.get("parent_path") == parent_path:
+                return space_id
+        if unclassified_match:
+            raise _SpaceLookupError
```
mwojtyczka, line 1705: **Permanent self-repeating provisioning failure (correctness).**

When a title-matching space can't be classified (its detail `GET` fails) and none matches the parent, `_find_space_id_by_title` raises `_SpaceLookupError`, so `ensure_dq_genie_space` sets `STATUS_ERROR` and returns `None` **without creating a space** — and repeats this on every startup.

**Repro:** first-time provisioning (no stored id) in a workspace that also holds another deployment's space titled `DQX Studio — DQ Results …` that this app SP cannot `GET` (ACL/permission denial, or a persistent API error). No candidate matches this deployment's `parent_path`, so `unclassified_match=True` → raise → `STATUS_ERROR`, `return None`. The space is never created; every subsequent startup repeats the failure.

This re-introduces exactly the permanent failure the adjacent paging-cap branch was deliberately written to avoid — that branch degrades to *create* instead. Consider degrading to create here too rather than raising.

app/src/databricks_labs_dqx_app/ui/routes/_sidebar/settings.tsx
```diff
@@ -2513,8 +2513,7 @@ function ComputeSettingsCard() {
           toast.success(t("config.computeGrantSucceeded"));
         },
         onError: (err: unknown) => {
-          const axErr = err as AxiosError<{ detail?: string }>;
-          toast.error(axErr?.response?.data?.detail ?? t("config.computeGrantFailed"));
+          toast.error(extractApiError(err, t("config.computeGrantFailed")));
```
mwojtyczka, line 2516: **Incomplete `extractApiError` migration (correctness).**

Only the two `ComputeSettingsCard` `onError` handlers were migrated to `extractApiError`. The other ~11 handlers in this file still do `axErr?.response?.data?.detail ?? fallback`.

If any non-warehouse settings endpoint returns the new structured detail shape `{code, summary, instructions}` (an object, not a string), those handlers pass the object straight to `toast.error`, rendering `[object Object]` instead of a message. The new helper exists precisely to prevent this — apply it to the remaining handlers too.

app/src/databricks_labs_dqx_app/ui/lib/api-error.ts
```diff
@@ -0,0 +1,15 @@
+interface StructuredApiErrorDetail {
+  summary?: unknown;
+}
+
+/** Return a render-safe message from string or structured FastAPI error details. */
+export function extractApiError(error: unknown, fallback: string): string {
```
mwojtyczka, line 6: **Reuse / DRY.**

This new canonical helper duplicates ~9 identical local `extractApiError` definitions still scattered across the UI (RegistryRuleFormDialog, ProductHeader, useEditProductState, PermissionsTab, AddMonitoredTableModal, ApplyRuleModal, RegistryRuleJsonDialog, MonitoredTableSchedulingTab).

The copies that only read `detail?: string` silently mishandle the new structured detail, and future backend shape changes must be replicated in ~10 places. Consider replacing those copies with imports of this helper as part of the extraction.

app/src/databricks_labs_dqx_app/backend/services/genie_space_service.py
```diff
@@ -1731,7 +1766,24 @@ def ensure_dq_genie_space(
         stored_hash = settings.get_setting(SETTING_CONFIG_HASH)
```
mwojtyczka, line 1766: **Extra Genie API round-trip on every startup (efficiency — minor).**

The steady-state no-op path (id present, hash unchanged) now issues a `GET /api/2.0/genie/spaces/{id}` on every startup where it previously made zero calls (the test flipped from `assert_not_called` to `assert_called_once`). Combined with the per-candidate detail `GET` loop in `_find_space_id_by_title`, provisioning in a workspace with many `DQX Studio` spaces costs N detail GETs per attempt. Added blocking work on boot — worth confirming it's intended.

app/src/databricks_labs_dqx_app/backend/services/genie_space_service.py
```diff
@@ -81,13 +83,25 @@
...
 SETTING_SPACE_ID = "dq_genie_space_id"
 SETTING_CONFIG_HASH = "dq_genie_space_config_hash"
 SETTING_STATUS = "dq_genie_space_status"
 
+
+class _SpaceLookupError(RuntimeError):
+    """Raised when existing spaces cannot be safely classified."""
+
+
+def space_parent_path() -> str:
```
mwojtyczka, line 99: **Shared-root fallback collapses two deployments (correctness — minor).**

`space_parent_path()` falls back to the shared root `/Shared/dqx-studio` when `DATABRICKS_CLIENT_ID` is unset/empty. Two deployments both lacking a client id resolve to the same `parent_path`, and the new parent-scoped reuse guard then matches the *other* deployment's space by title+parent — so B reuses/attaches to A's space, defeating the client-id-subfolder isolation this change relies on. Consider failing closed (or using a per-deployment unique fallback) when the client id is absent.

mwojtyczka (commented):
Isaac Review — PR #1522 (`dqx-studio/smoke-bug-fixes`). 5 findings, most severe first. #1 (permanent provisioning failure) and #2 (`[object Object]` toasts) are the priorities.

app/src/databricks_labs_dqx_app/backend/services/genie_space_service.py
laurencewells: Fixed. The unclassified match branch now falls through to create instead of raising, so a title match we cannot classify no longer blocks provisioning on every boot. Covered by a new test.

app/src/databricks_labs_dqx_app/ui/routes/_sidebar/settings.tsx
laurencewells: Agreed. This is the wider UI error handling cleanup tracked in #1528, so deferring it there.

app/src/databricks_labs_dqx_app/ui/lib/api-error.ts
laurencewells: Agreed. Swapping the duplicated local copies for this shared helper is part of #1528.

app/src/databricks_labs_dqx_app/backend/services/genie_space_service.py
laurencewells: Intended. The verify GET is the cost of checking the stored id before trusting it, which fixed the earlier duplicate and resilience findings. It is one call on the steady state path.

app/src/databricks_labs_dqx_app/backend/services/genie_space_service.py
laurencewells: Fixed. space_parent_path now raises when DATABRICKS_CLIENT_ID is missing and ensure fails closed instead of falling back to the shared root. Covered by two new tests.

app/src/databricks_labs_dqx_app/backend/services/genie_space_service.py
```diff
@@ -1664,14 +1688,42 @@ def _find_space_id_by_title(ws: WorkspaceClient, title: str) -> str | None:
...
+            space_id = match.get("space_id")
+            if not space_id:
+                continue
+            try:
+                detail = ws.api_client.do("GET", f"/api/2.0/genie/spaces/{space_id}")
+            except Exception as error:
+                logger.info(f"Genie space detail lookup skipped: {type(error).__name__}")
+                unclassified_match = True
+                continue
+            if isinstance(detail, dict) and detail.get("parent_path") == parent_path:
```
mwojtyczka, line 1712: **Reuse hinges on an unverified `parent_path` round-trip (correctness).**

Reuse fires only on exact string equality between the value the detail `GET` returns and the path the app built itself:

```python
detail = ws.api_client.do("GET", f"/api/2.0/genie/spaces/{space_id}")
if isinstance(detail, dict) and detail.get("parent_path") == parent_path:
    return space_id
```

`parent_path` is constructed locally as `/Shared/dqx-studio/{client_id}` and sent verbatim on create (`build_create_payload` → POST body). Nothing verifies the **read** endpoint echoes it back in the same form the **write** endpoint received. If the detail response uses a different key (`parentPath`, nested, an object), or the API normalizes the value (resolves `/Shared` to a folder id, adds/strips a trailing slash, returns an absolute/canonical form), the `==` is silently false.

**Why it degrades to "always create":** when the equality never matches, the detail GETs still *succeed*, so `unclassified_match` stays `False`, no error is raised, and the function returns `None` cleanly → `ensure_dq_genie_space` sees `space_id is None` and creates a brand-new space. The reuse feature becomes a silent no-op that leaks a duplicate space on every boot that lost its stored id (fresh deploy, reset, or a `STATUS_ERROR` retry) — the exact duplicate-proliferation this rework was meant to prevent.

This is now the *only* guard against unconditional creation on the reuse path: the paging-cap and unclassified-match branches were both intentionally relaxed to "degrade to create," so if this assumption is wrong nothing else catches it. A fake `api_client` that echoes the posted `parent_path` back would make this comparison the identity function and pass in tests while failing against the real service.

**Suggestions (any one):**
- Capture one real `GET /api/2.0/genie/spaces/{id}` response and confirm `parent_path` equals what `space_parent_path()` produced — closes this or confirms the leak.
- Compare on a normalized form (`rstrip("/")` both sides, or match only the `{client_id}` segment you control) rather than raw equality.
- Drop `parent_path` as the identity key entirely — store the space id in a durable setting keyed by `client_id`, so reuse doesn't depend on how the server echoes a path.

laurencewells: The documented Genie API contract defines `parent_path` as a top-level string on both Create Space and Get Space, representing the parent folder path. That matches the current implementation; the suggested `parentPath`, nested-object, and folder-ID response shapes would violate the published contract. Neither side emits a trailing slash, so normalization would not materially strengthen the identity check. I’m therefore keeping the exact comparison.

mwojtyczka (approved):
LGTM

PR #1530, opened 2026-09-11
Let Lakebase scale to zero when the scheduler is idle

laurencewells:
## Problem

The in-process `SchedulerService` woke every 60 s and queried Lakebase (`dq_schedule_configs`) on **every** tick. That interval is shorter than any Lakebase suspend timeout, so the scale-to-zero endpoint never went idle and never suspended. The scheduler pinned it awake 24/7 even with **no** schedules configured. Logs showed the tick succeeding every minute with `Scheduler tick: no schedule configs found`.

## Changes

**Scheduler idle back-off** (`services/scheduler_service.py`)
- `_tick` now returns whether any **active** schedule exists: a non-manual/non-paused scope config, a cron-scheduled data product, or a cron-scheduled monitored table. `_tick_products` / `_tick_monitored_tables` return their found-count instead of `None` (no extra queries).
- `_loop` polls every `_ACTIVE_POLL_SECONDS` (60) while work is active and backs off to `_IDLE_POLL_SECONDS` (3600) when idle, so the tick query stops holding the endpoint awake. `reload()` still interrupts the sleep instantly on a config save, so the idle interval only ever bounds first pickup of a schedule created while idle.
- The interval is 1 h to match the current idle floor: the hourly metadata-dim refresh reads the Lakebase Rules Registry every hour regardless, so polling less often wouldn't extend suspension until that is made skip-when-unchanged.

**Pool drains to zero** (`config.py`, `pg_executor.py`)
- Default Lakebase pool `min_size` `1 → 0` so the pool drops idle connections and lets the endpoint suspend. A held-open connection is re-established after suspension kills it, nudging the endpoint awake; the existing pre-ping absorbs the cold-connect on the next real request.

## Tests

New unit tests (TDD): idle-vs-active tick reporting, product/table found-counts, poll-interval selection, and the pool `min_size = 0` default (executor + config). Full app backend suite passes (4168); basedpyright / black / ruff clean.

## Deferred (for deeper multi-hour suspension)

- Make the hourly metadata-dim refresh skip-when-unchanged (it's today's ~1 h floor).
- Wire `notify_scheduler()` into the product / monitored-table schedule + approval routes so those schedules wake the loop instantly rather than waiting up to the idle interval for first pickup.

### DIFF
--- a/app/src/databricks_labs_dqx_app/backend/pg_executor.py
+++ b/app/src/databricks_labs_dqx_app/backend/pg_executor.py
@@ -245,7 +245,7 @@ class PgExecutor:
         token_refresh_retry_seconds: int = 10,
         token_refresh_retry_jitter: float = 0.3,
         token_refresh_max_failures: int = 12,
-        pool_min_size: int = 1,
+        pool_min_size: int = 0,
         pool_max_size: int = 10,
     ) -> None:
         self._ws = ws
@@ -314,6 +314,13 @@ class PgExecutor:
         # dead connection and the fresh connect re-auths with the current
         # OAuth token (read from ``_connect_kwargs['password']``, which
         # the refresh loop keeps current) and resumes the endpoint.
+        #
+        # ``min_size`` defaults to 0 (see AppConfig.lakebase_pool_min_size) so
+        # the pool drains to zero idle connections and lets the endpoint
+        # actually suspend. A held-open connection would be re-established by
+        # the pool after suspension killed it, nudging the endpoint awake and
+        # defeating scale-to-zero; the pre-ping above absorbs the cold-connect
+        # on the next real request instead.
         self._pool: ConnectionPool = ConnectionPool(
             conninfo="",
             min_size=pool_min_size,
@@ -945,7 +952,7 @@ def build_pg_executor(
     token_refresh_retry_seconds: int = 10,
     token_refresh_retry_jitter: float = 0.3,
     token_refresh_max_failures: int = 12,
-    pool_min_size: int = 1,
+    pool_min_size: int = 0,
     pool_max_size: int = 10,
 ) -> PgExecutor:
     """Construct a :class:`PgExecutor` from a Databricks workspace client.
@@ -1009,7 +1016,7 @@ def build_pg_executor_from_connection(
     token_refresh_retry_seconds: int = 10,
     token_refresh_retry_jitter: float = 0.3,
     token_refresh_max_failures: int = 12,
-    pool_min_size: int = 1,
+    pool_min_size: int = 0,
     pool_max_size: int = 10,
 ) -> PgExecutor:
     """Construct an executor from endpoint or platform-bound connection values."""
--- a/app/src/databricks_labs_dqx_app/backend/services/scheduler_service.py
+++ b/app/src/databricks_labs_dqx_app/backend/services/scheduler_service.py
@@ -207,6 +207,26 @@ _METADATA_DIM_REFRESH_INTERVAL_HOURS = 1
 # ``tag_reconcile_service`` was wired or when tag-auto-apply is off.
 _TAG_RECONCILE_INTERVAL_HOURS = 6
 
+# Scheduler loop poll cadence. When any schedule is active (a non-manual,
+# non-paused scope config, a cron-scheduled data product, or a cron-scheduled
+# monitored table) the loop wakes every ``_ACTIVE_POLL_SECONDS`` so due runs
+# fire promptly. When nothing is scheduled it backs off to
+# ``_IDLE_POLL_SECONDS`` so the ``_tick`` query against the OLTP store stops
+# pinning a scale-to-zero Lakebase endpoint awake every minute — the endpoint
+# can then suspend in the gaps. ``reload()`` still wakes the loop immediately
+# when a schedule config is saved (see the config/schedule routes), so the
+# longer idle interval only ever bounds first pickup of a schedule created
+# while idle.
+#
+# The idle interval is set to one hour to match the current idle floor: the
+# hourly metadata-dim refresh (``_METADATA_DIM_REFRESH_INTERVAL_HOURS``) reads
+# the Lakebase Rules Registry every hour regardless, so polling less often than
+# that would not let Lakebase suspend any longer. Lowering that floor (and thus
+# raising this interval to multi-hour) requires making the metadata refresh
+# skip-when-unchanged — deferred.
+_ACTIVE_POLL_SECONDS = 60
+_IDLE_POLL_SECONDS = 3600
+
 # System attribution for scheduler-initiated writes (mirrors the
 # ``user_email="scheduler"`` the product/table run ticks already use).
 _SCHEDULER_SYSTEM_USER = "scheduler"
@@ -448,12 +468,25 @@ class SchedulerService:
     # ------------------------------------------------------------------
 
     async def _loop(self) -> None:
-        """Check schedules every 60 seconds and trigger due runs."""
+        """Trigger due runs, then sleep until the next tick or a ``reload()``.
+
+        The sleep cadence adapts to whether anything is scheduled (see
+        :meth:`_poll_interval_seconds`): the tight ``_ACTIVE_POLL_SECONDS``
+        while schedules exist, and the longer ``_IDLE_POLL_SECONDS`` when the
+        app is idle so the ``_tick`` query stops pinning a scale-to-zero
+        Lakebase endpoint awake every minute. ``reload()`` interrupts the sleep
+        immediately when a schedule is saved, so the longer idle interval never
+        delays a schedule the user just created.
+        """
         while True:
+            # Default to the active cadence: if a tick raises before it can
+            # report, keep polling tightly rather than risk backing off while
+            # real work is pending.
+            has_active_work = True
             try:
                 recalc = self._force_recalc
                 self._force_recalc = False
-                await self._tick(recalc=recalc)
+                has_active_work = await self._tick(recalc=recalc)
                 await self._maybe_gc_orphan_views(datetime.now(timezone.utc))
                 await self._maybe_sweep_stale_tmp_views(datetime.now(timezone.utc))
                 await self._maybe_run_retention(datetime.now(timezone.utc))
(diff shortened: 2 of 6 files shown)


app/src/databricks_labs_dqx_app/backend/pg_executor.py
```diff
@@ -314,6 +314,13 @@ def __init__(
         # dead connection and the fresh connect re-auths with the current
         # OAuth token (read from ``_connect_kwargs['password']``, which
         # the refresh loop keeps current) and resumes the endpoint.
+        #
+        # ``min_size`` defaults to 0 (see AppConfig.lakebase_pool_min_size) so
```
mwojtyczka, line 318: **Eager fail-fast is lost with the new `min_size=0` default.** The pool is still opened eagerly at line 335 (`self._pool.open(wait=True, timeout=30.0)`), and the comment at line 330 still reads `open=False,  # opened explicitly below so failures surface eagerly`. But with `min_size=0`, `open(wait=True)` only waits for `min_size` (0) connections, so it returns without establishing any physical connection — a bad host, wrong schema, or unauthorized credential that used to raise at startup now surfaces only on the first real request. Either keep a one-shot startup probe connection (independent of the steady-state pool size) or update the now-stale comment to reflect that failures are deferred to first use.

app/src/databricks_labs_dqx_app/backend/services/scheduler_service.py
```diff
@@ -599,6 +664,11 @@ async def _tick(self, *, recalc: bool = False) -> None:
         except Exception:
             logger.exception("Scheduler failed refreshing the score cache for completed runs")
 
+        # Active when any due-ness source has scheduled work (or errored, via
```
mwojtyczka, line 667: **Idle back-off can delay completed-run score refresh by up to `_IDLE_POLL_SECONDS`.** This active-work signal deliberately excludes the completion-refresh path (`_pending_score_runs` / `_refresh_scores_for_completed_runs`). When a run is pending completion refresh and there are no active schedules, the loop backs off to `_IDLE_POLL_SECONDS` (3600s), so the score cache and last-run timestamps for that completed run stay stale for up to an hour instead of ~60s. If manual / out-of-band runs can populate `_pending_score_runs` with no schedule configured, consider keeping the signal active while `_pending_score_runs` is non-empty.

app/src/databricks_labs_dqx_app/backend/services/scheduler_service.py
```diff
@@ -465,14 +498,34 @@ async def _loop(self) -> None:
                 logger.exception("Scheduler tick failed")
```
mwojtyczka, line 499: **First run of a newly created cron-scheduled data product / monitored table can be delayed up to `_IDLE_POLL_SECONDS`.** The idle back-off relies on `notify_scheduler()` to wake the loop the instant a schedule is saved, but `notify_scheduler` is only wired into `routes/v1/config.py` and `routes/v1/schedules.py`. `routes/v1/data_products.py` and `routes/v1/monitored_tables.py` write `schedule_cron`/`schedule_tz` without notifying the scheduler. So if the loop is idle (sleeping 3600s) when the first cron-scheduled product or monitored table is created/approved, it won't be picked up until the idle timer elapses — up to an hour, vs ~60s under the old fixed poll. Worth wiring `notify_scheduler()` into those two routes rather than deferring.

app/tests/test_pg_executor.py
```diff
@@ -1507,6 +1507,29 @@ def test_platform_binding_uses_supplied_credentials_without_control_plane(self)
...
+            password=SecretStr("secret"),
+            schema="dqx_studio",
+        )
+
+        with (
+            patch("databricks_labs_dqx_app.backend.pg_executor.ConnectionPool") as pool,
+            patch("databricks_labs_dqx_app.backend.pg_executor.threading.Thread"),
+        ):
+            pool.check_connection = MagicMock()
+            build_pg_executor_from_connection(ws, connection)  # no pool_min_size override
```
mwojtyczka, line 1530: **This insertion silently weakens `test_platform_binding_uses_supplied_credentials_without_control_plane`.** The new `test_pool_min_size_defaults_to_zero_for_scale_to_zero` is inserted right after that test's last statement (`thread_factory.assert_not_called()`), so its trailing `ws.postgres.get_endpoint.assert_not_called()`, `generate_database_credential.assert_not_called()`, and `ws.current_user.me.assert_not_called()` now belong to the new method (asserting against the new test's `ws`). The platform-binding test loses its verification that the control-plane APIs are never called. Move the new test method below those three assertions so the original test keeps them.

app/src/databricks_labs_dqx_app/backend/services/scheduler_service.py
```diff
@@ -574,20 +632,27 @@ async def _tick(self, *, recalc: bool = False) -> None:
         # product-tick failure is fully isolated inside
         # :meth:`_tick_products` and cannot roll back or skip anything
         # the config loop already did.
+        # A sentinel of -1 marks "source errored, count unknown" so the
+        # active-work check below treats it as active (keep polling tightly)
```
mwojtyczka, line 636: **The `-1` sentinel doesn't buy anything over a boolean.** `product_count`/`table_count` are set to `-1` on error to mark 'errored, count unknown', but the active-work return only tests `!= 0`, so `-1` and any positive count are treated identically. A boolean (e.g. `product_active = True` in the except block) would express the intent more directly and avoid a future reader mistaking `-1` for a real count. Minor.

mwojtyczka (commented):
Review pass on the scale-to-zero changes. One confirmed test-coverage regression (#4 below), the eager-startup fail-fast interaction with the new `min_size=0` default, and a couple of idle-back-off edge cases where work can wait up to an hour. Details inline; all verified against the diff/code.

_AI-assisted review (Isaac); findings checked against the code before posting._

app/src/databricks_labs_dqx_app/backend/pg_executor.py
laurencewells: Added a one-shot startup probe that opens and validates a single connection right after open(), then returns it to the pool so it drains back to min_size. Eager fail-fast is restored and scale-to-zero is preserved. Also updated the stale comment on the open= line.

app/src/databricks_labs_dqx_app/backend/services/scheduler_service.py
laurencewells: Fixed. _tick now stays active while _pending_score_runs is non-empty, so a pending completion refresh keeps the tight cadence instead of waiting out the idle interval. Out-of-band manual runs do populate the set via _sweep_recent_run_sets with no schedule configured, so this path was reachable.

app/src/databricks_labs_dqx_app/backend/services/scheduler_service.py
laurencewells: Wired it in rather than deferring. notify_scheduler now fires on data product approve and on auto-approve-during-submit, and on monitored table schedule set (not on clear, since clearing activates nothing). Both route files get a small _notify_scheduler helper matching the one in config.py and schedules.py.

app/tests/test_pg_executor.py
laurencewells: Fixed. Moved the three control-plane assert_not_called checks back into test_platform_binding_uses_supplied_credentials_without_control_plane so it keeps its verification. The new min_size test now holds only its own assertion.

app/src/databricks_labs_dqx_app/backend/services/scheduler_service.py
laurencewells: Replaced the -1 sentinel with product_active and table_active booleans. The errored branch sets the boolean to True directly, and the return expression reads more clearly now.

app/src/databricks_labs_dqx_app/backend/services/scheduler_service.py
```diff
@@ -207,6 +207,26 @@ def _normalize_schedule_kind(value: object) -> str:
...
+# while idle.
+#
+# The idle interval is set to one hour to match the current idle floor: the
+# hourly metadata-dim refresh (``_METADATA_DIM_REFRESH_INTERVAL_HOURS``) reads
+# the Lakebase Rules Registry every hour regardless, so polling less often than
+# that would not let Lakebase suspend any longer. Lowering that floor (and thus
+# raising this interval to multi-hour) requires making the metadata refresh
+# skip-when-unchanged — deferred.
+_ACTIVE_POLL_SECONDS = 60
+_IDLE_POLL_SECONDS = 3600
```
mwojtyczka, line 228: **Scale-to-zero isn't actually delivered by this change.** Even when fully idle, `_loop` still runs `_maybe_refresh_metadata_dims` every iteration (reads the Lakebase Rules Registry hourly), and each `_tick` still issues `_load_schedule_configs` / `_load_scheduled_products` / `_load_scheduled_tables`. That's exactly why `_IDLE_POLL_SECONDS` is capped at 3600s here "to match the current idle floor". Net effect: the endpoint receives Lakebase traffic at least once per hour and can never suspend longer than that — the actual root cause (metadata refresh always reading Lakebase) is explicitly deferred, so this PR by itself does not achieve scale-to-zero. The real fix belongs in making the metadata refresh skip-when-unchanged, not in the poll cadence. Worth confirming whether this PR is intended as a first step vs. delivering the stated goal.

app/src/databricks_labs_dqx_app/backend/routes/v1/data_products.py
```diff
@@ -72,6 +72,18 @@ def _current_user_email(obo_ws: WorkspaceClient) -> str:
...
+def _notify_scheduler() -> None:
+    """Wake the in-process scheduler so a newly approved cron-scheduled product
+    is picked up immediately rather than after the idle poll interval (which can
+    be an hour once the loop has backed off to let Lakebase suspend)."""
+    try:
+        from databricks_labs_dqx_app.backend._scheduler_registry import notify_scheduler
+
+        notify_scheduler()
+    except Exception:
+        pass
```
mwojtyczka: `except Exception: pass` swallows every failure of the wake with no diagnostic. If the lazy `from ..._scheduler_registry import notify_scheduler` path ever breaks (rename/refactor) or `notify_scheduler()` raises, the wake is silently dropped — a just-approved cron-scheduled product then waits up to `_IDLE_POLL_SECONDS` (1h) for first pickup while operators see no error explaining the latency. Log at debug/warning before `pass`. Same pattern in `monitored_tables.py`.

app/src/databricks_labs_dqx_app/backend/routes/v1/data_products.py
```diff
@@ -72,6 +72,18 @@ def _current_user_email(obo_ws: WorkspaceClient) -> str:
     return user.user_name or "unknown"
 
 
+def _notify_scheduler() -> None:
```
mwojtyczka: **DRY:** this identical `_notify_scheduler` try/except + lazy-import wrapper now exists in 4 route modules (config.py, schedules.py, data_products.py, monitored_tables.py). AGENTS.md says extract shared utilities once the same logic appears in 3+ places — this PR adds copies 3 and 4. The try/except belongs inside `_scheduler_registry.notify_scheduler()` (or a single shared helper) so route modules call it directly; four copies mean any future change to the wake semantics must be edited in four places and can drift.

app/src/databricks_labs_dqx_app/backend/pg_executor.py
```diff
@@ -314,18 +314,34 @@ def __init__(
...
         )
         self._pool.open(wait=True, timeout=30.0)
+        # ``open(wait=True)`` only waits for ``min_size`` connections, so with
+        # the scale-to-zero default of 0 it returns without ever contacting the
+        # database — a bad host, wrong schema, or unauthorized credential would
+        # otherwise surface only on the first real request. A one-shot probe
+        # restores eager fail-fast at startup: it opens and validates a single
+        # connection, then returns it to the pool, which drains back to
+        # ``min_size`` and leaves scale-to-zero intact.
+        with self._pool.connection() as probe:
```
mwojtyczka: The startup probe opens a connection via `with self._pool.connection()` and checks it back into the pool on exit. With `min_size=0` the pool keeps that idle connection until `psycopg_pool`'s `max_idle` elapses (default ~10 min), so after every app start/restart Lakebase is held awake ~10 minutes before it can first suspend — partially undercutting the scale-to-zero goal on a warm-restart-heavy deployment. Consider closing the probe connection outside the pool, or setting `max_idle` explicitly.

mwojtyczka (commented):
Automated code-review pass (via /code-review). 4 findings, quality/altitude rather than crashes. Most important is the first: the PR's headline goal may not actually be delivered by these changes alone.

app/src/databricks_labs_dqx_app/backend/services/scheduler_service.py
laurencewells: This PR still enables scale-to-zero: with a 300s suspend timeout and hourly idle polling, Lakebase can suspend for about 55 minutes each hour. Multi-hour suspension is explicitly deferred.

app/src/databricks_labs_dqx_app/backend/routes/v1/data_products.py
laurencewells: Fixed in `9f7911e0`. Scheduler notification failures are now logged centrally without failing the route, with regression coverage.

app/src/databricks_labs_dqx_app/backend/routes/v1/data_products.py
laurencewells: Fixed in `9f7911e0`. Removed the four route wrappers; all callers now use the shared registry helper.

app/src/databricks_labs_dqx_app/backend/pg_executor.py
laurencewells: Fixed in `9f7911e0`. The startup probe now uses a short-lived direct connection and never enters the pool. Regression coverage added.

mwojtyczka (approved):
LGTM

PR #1531, opened 2026-09-11
feat(app): dqx-core compatible rule loading + backend clean up

mwojtyczka:
## What & why

Lets external pipelines load DQX Studio-authored rules with the standard `DQEngine.load_checks` API, and makes the Studio storage table names clearer.

### 1. `dq_rules_core` compatibility view
A dqx-core-compatible read view over **approved** rules in `dq_resolved_rules`, so pipelines can load Studio rules with unmodified core:

```python
engine.load_checks(LakebaseChecksStorageConfig(
    location="databricks_postgres.dqx_studio.dq_rules_core",
    instance_name="<lakebase-instance>",
    run_config_name="<table_fqn>"))
```

`run_config_name ← table_fqn`; the inner `{function, arguments}` is extracted from the stored JSONB; only `status = 'approved'` rows appear.

### 2. Fingerprints
- `rule_fingerprint` (dqx-core `compute_rule_fingerprint`) is stored on **every** write path into the table (`RulesCatalogService.save` / `update_rule`, `Materializer`).
- `rule_set_fingerprint` is computed **in the view** per `table_fqn` (a GROUP BY over the approved set) so `load_checks` selects the current set and returns all of it. A parity test guards that the SQL reproduction matches core's Python.

### 3. Rename `dq_quality_rules` → `dq_resolved_rules`
Includes `dq_quality_rules_history` → `dq_resolved_rules_history`, indexes, and constraints — edited into the v1 Postgres baseline in place (per the single-baseline convention; existing dev deployments re-provision). All backend, UI, test, and regenerated `api.ts` references updated.

### 4. Docs
- New **"Use outside the app"** Studio section: load rules via `LakebaseChecksStorageConfig`; read `dq_quarantine_records` (schema + example queries).
- Front page: hero → "Battle-tested data quality framework", expanded Studio capability list, removed the placeholder Marketplace/Cost Calculator banners, **DQX Studio** section renamed with a proper `<FeatureTags>` Beta tag.
- Installation guide: Marketplace made the **recommended** install path (Studio index links to it).
- Wording fixes (Collections/Tables labels, "human readable name").

## Testing
- 4157 backend tests pass · basedpyright 0 errors · tsc clean · 837 UI tests · docs build + tests pass.

### DIFF
--- a/docs/dqx/docs/studio/integrations/index.mdx
+++ b/docs/dqx/docs/studio/integrations/index.mdx
@@ -1,0 +1,38 @@
+---
+sidebar_position: 0
+title: Use outside the app
+---
+
+import { FeatureLifecycleStage, FeatureTags } from '@site/src/components/FeatureTags';
+
+# Use DQX Studio data outside the app
+
+<FeatureTags>
+  <FeatureLifecycleStage stage="beta" heading={false} />
+</FeatureTags>
+
+DQX Studio is where you *author, review, and govern* rules — but the rules and
+results it stores are meant to be used by your pipelines too. Studio persists its
+state in your own Unity Catalog catalog, so anything the app knows, your jobs and
+notebooks can read directly.
+
+Two things you'll most often want from outside the app:
+
+- **[Load approved rules](/docs/studio/integrations/load-rules)** — read the rules
+  you approved in Studio into a pipeline with the standard
+  `DQEngine.load_checks` API, and run them wherever your data lives (production
+  included). Studio exposes a dqx-core-compatible view (`dq_rules_core`) for exactly
+  this.
+- **[Read quarantined records](/docs/studio/integrations/read-quarantine)** — query
+  the failing rows Studio's runs captured (`dq_quarantine_records`), for dashboards,
+  alerting, or your own triage.
+
+<div className="admonition admonition-note">
+
+DQX Studio's own runs target dev/UAT data and write results to the app's catalog —
+they are **not** intended to run rules as part of your production pipelines. To run
+Studio-authored rules against production data, load them with
+[`DQEngine.load_checks`](/docs/studio/integrations/load-rules) from your own
+pipeline.
+
+</div>
--- a/app/AGENTS.md
+++ b/app/AGENTS.md
@@ -74,8 +74,9 @@ but is protected transitively by the instance-level guard.
  │   ├── dq_quarantine_records        (Delta) invalid rows captured by runs
  │   ├── dq_metrics                   (Delta) per-run quality metrics for trend tracking
  │   ├── dq_app_settings              (OLTP*) key/value app configuration
- │   ├── dq_quality_rules             (OLTP*) active/approved rules
- │   ├── dq_quality_rules_history     (OLTP*) rule change audit log
+ │   ├── dq_resolved_rules             (OLTP*) active/approved rules
+ │   ├── dq_rules_core                 (VIEW, Lakebase) dqx-core-compatible read view over approved dq_resolved_rules — external pipelines load it via DQEngine.load_checks(LakebaseChecksStorageConfig)
+ │   ├── dq_resolved_rules_history     (OLTP*) rule change audit log
  │   ├── dq_role_mappings             (OLTP*) role → workspace group mappings (RBAC)
  │   ├── dq_comments                  (OLTP*) comment threads on rules/runs
  │   ├── dq_schedule_configs          (OLTP*) per-schedule config (cron/interval, target rules)
@@ -88,8 +89,8 @@ but is protected transitively by the instance-level guard.
 Lakebase project (when enabled, default `lakebase_project_id` = `dqx-studio-db`):
  └── databricks_postgres              (database — always-present admin DB; no per-app DB provisioned)
      └── dqx_studio                   (schema — created by PgMigrationRunner on first start; configurable via DQX_LAKEBASE_SCHEMA)
-         ├── dq_app_settings, dq_role_mappings, dq_quality_rules,
-         │   dq_quality_rules_history, dq_comments, dq_schedule_configs,
+         ├── dq_app_settings, dq_role_mappings, dq_resolved_rules,
+         │   dq_resolved_rules_history, dq_comments, dq_schedule_configs,
          │   dq_schedule_configs_history, dq_schedule_runs
          └── dq_migrations             (Postgres migration version tracker)
 ```
@@ -511,7 +512,7 @@ choice is driven entirely by `databricks.yml`:
 | Backend | Tables | Why |
 |---------|--------|-----|
 | **Delta Lake** (always) | `dq_validation_runs`, `dq_profiling_results`, `dq_quarantine_records`, `dq_metrics` | Spark task runner writes these; high-volume append-mostly; columnar reads. |
-| **Lakebase Postgres** *(default — opt-out via `lakebase_endpoint="-"`)* | `dq_app_settings`, `dq_role_mappings`, `dq_quality_rules`, `dq_quality_rules_history`, `dq_comments`, `dq_schedule_configs`, `dq_schedule_configs_history`, `dq_schedule_runs` | Low-latency point reads/writes from FastAPI request handlers; row-level upserts; primary-key/foreign-key semantics. |
+| **Lakebase Postgres** *(default — opt-out via `lakebase_endpoint="-"`)* | `dq_app_settings`, `dq_role_mappings`, `dq_resolved_rules`, `dq_resolved_rules_history`, `dq_comments`, `dq_schedule_configs`, `dq_schedule_configs_history`, `dq_schedule_runs` | Low-latency point reads/writes from FastAPI request handlers; row-level upserts; primary-key/foreign-key semantics. |
 
 When Lakebase is **disabled** (no `lakebase_endpoint` set), the OLTP
 tables fall back to Delta — `MigrationRunner` runs both
@@ -549,7 +550,7 @@ per `_RETENTION_INTERVAL_HOURS` (24h). Two knobs, both stored in
 
 | Setting key                  | Default | Tables affected |
 |------------------------------|--------:|-----------------|
-| `retention_days`             | 90      | `dq_validation_runs`, `dq_profiling_results`, `dq_metrics`, plus the OLTP history tables (`dq_quality_rules_history`, `dq_schedule_configs_history`). Picked to match what trend dashboards expect. |
+| `retention_days`             | 90      | `dq_validation_runs`, `dq_profiling_results`, `dq_metrics`, plus the OLTP history tables (`dq_resolved_rules_history`, `dq_schedule_configs_history`). Picked to match what trend dashboards expect. |
 | `quarantine_retention_days`  | 30      | `dq_quarantine_records` only. Tighter because that table holds the full source row payload (PII surface). |
 
 Both resolvers share a `_RETENTION_DAYS_MIN = 7` floor so a
--- a/app/src/databricks_labs_dqx_app/backend/migrations/__init__.py
+++ b/app/src/databricks_labs_dqx_app/backend/migrations/__init__.py
@@ -85,7 +85,7 @@ Two status families intentionally use different casing:
   (``RUNNING``/``SUCCESS``/``FAILED``/``CANCELED``). These mirror the
   Databricks Jobs SDK ``life_cycle_state`` / ``result_state`` values
   that are passed straight through ``RunStatusOut`` to the frontend.
-- **App-domain workflow** (``dq_quality_rules.status``,
+- **App-domain workflow** (``dq_resolved_rules.status``,
   ``dq_schedule_runs.status``) — lowercase. These are pure DQX
   vocabulary (``draft``/``approved``, ``pending``/``partial_failure``)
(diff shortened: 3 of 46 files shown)


docs/dqx/docs/studio/integrations/index.mdx
ghanse, line 1: #### Suggestion:
Update [this section](https://databrickslabs.github.io/dqx/docs/guide/quality_checks_storage/#loading-quality-checks-from-a-storage) to mention that `load_checks` can be used to load rules written in DQX Studio.

ghanse (approved):
LGTM

OGordon100 (changes requested):
## Docs feedback: the two "Use outside the app" pages should be one page, in a different place, and much shorter

The backend work looks fine — this is entirely about `docs/dqx/docs/studio/integrations/`. As written, the three new pages (305 lines) read as generated prose: verbose, train-of-thought, and heavy on internal plumbing a reader doesn't need. My concrete asks below.

### 1. One page, not three — and not a top-level section

Splitting this across `index.mdx` + `load-rules.mdx` + `read-quarantine.mdx` gives each page too little to say, and "Use outside the app" as a new top-level section at `position: 6` overweights it relative to `Authoring rules` / `Monitor your tables` / `Review results`.

**Suggested:** delete `docs/studio/integrations/` entirely and add a single page under the existing **Monitor your tables** section:

- **Path:** `docs/dqx/docs/studio/monitoring/use-as-code.mdx`
- **Title / sidebar label:** `Use as Code in Lakehouse`
- **`sidebar_position: 6`** (directly after `Run checks & schedule`, which is `5`)
- No `_category_.json` changes needed beyond deleting `integrations/_category_.json`.

I checked for inbound links: every reference to `/docs/studio/integrations/*` is from within the section being deleted, so nothing else in the repo breaks.

### 2. The Lakebase access paragraph is not actionable

This is the part that made me stop:

> Access to the Lakebase database instance. To open a connection, `load_checks` mints a short-lived Postgres credential from Databricks for Studio's Lakebase instance. The reader's Databricks identity must be allowed to use that database instance (grant it on the instance in Databricks). Without this, the connection can't be established.

It explains the mechanism but never names the permission, so an admin reading it cannot act on it. "Grant it on the instance in Databricks" — grant *what*?

The underlying call is `ws.database.generate_database_credential(request_id=..., instance_names=[instance_name])` (`src/databricks/labs/dqx/checks_storage.py:555`), which needs **`CAN_CONNECT` on the Lakebase database instance**. Please name that permission explicitly and link to the Databricks database-instance permissions doc, and drop the credential-minting explanation — a reader doesn't need to know a token is minted per connection, only what to ask their admin for.

### 3. Cut the internal plumbing

- **`## The dq_rules_core view` section (load-rules.mdx:20–43)** — the reader does not need to know rules are stored in an internal `dq_resolved_rules` table that "carries Studio-specific columns (lifecycle status, versioning, provenance) that dqx-core doesn't understand" and that a view "reshapes" them. One sentence covers it: Studio exposes approved rules through a view that `load_checks` reads.
- **The column table** — mostly restates its own column names ("`name`, `criticality`, `filter` | The check's identity and severity"). Also incomplete: the view selects **10** columns including `message_expr` and `created_at` (`app/src/databricks_labs_dqx_app/backend/migrations/postgres.py:779-788`), which the table omits. Either make it complete or drop it; I'd drop it.
- **`## Notes and limits` (load-rules.mdx:130–146)** — the `for_each_column` compact-vs-expanded fingerprint paragraph is accurate but is a footnote to a footnote, and it ends by reassuring the reader it "never affects *which* rules load, only the informational fingerprint values" — i.e. it does not matter. Cut the whole section; keep only "approved rules only", which belongs in the body.

### 4. The SQL queries are mostly filler

`read-quarantine.mdx` has **five** SQL snippets. Queries 3–5 (explode `errors`, count failures by check, join to `dq_validation_runs`) are a SQL cookbook, not documentation for this feature, and the "Which tables have approved rules?" tip in `load-rules.mdx` is another. Keep one query that shows how to read the table; let people write their own `GROUP BY`. The "Reading VARIANT columns" tip can go too — it links out to Databricks' own VARIANT docs, which is enough on its own.

### 5. Smaller wording issues

- "DQX Studio is where you *author, review, and govern* rules — but the rules and results it stores are meant to be used by your pipelines too" — the "but" framing plus italics-for-emphasis recurs on all three pages.
- `index.mdx` opens by explaining what the section will contain, then the same admonition about production runs appears here and in the body copy.
- One thing to double-check rather than change: both `index.mdx` and `read-quarantine.mdx` state quarantine holds a **sample** of failing rows. I could only find `LIMIT` on the read/paging path (`routes/v1/dq_results.py:1316`), not on the write path — if the writer stores all failing rows, "sample" is wrong.

### Suggested replacement page

Target is ~3–4 paragraphs, one Python example, one SQL example. Facts kept deliberately vague where I could not verify them:

````mdx
---
sidebar_position: 6
title: Use as Code in Lakehouse
---

import Admonition from '@theme/Admonition';
import { FeatureLifecycleStage, FeatureTags } from '@site/src/components/FeatureTags';

# Use as Code in Lakehouse

<FeatureTags>
  <FeatureLifecycleStage stage="beta" heading={false} />
</FeatureTags>

You author and approve rules in Studio, but you don't have to run them there. Studio
keeps its rules and results in your own catalog, so a pipeline can read them directly —
there's no export step. Studio's own runs are for dev and UAT data; to check production
data, load the rules from your own job.

## Load approved rules

Point `LakebaseChecksStorageConfig` at Studio's rules view and pass the table you want
rules for as `run_config_name`:

```python
from databricks.labs.dqx.config import LakebaseChecksStorageConfig

engine = DQEngine(WorkspaceClient())

checks = engine.load_checks(
    LakebaseChecksStorageConfig(
        location="databricks_postgres.dqx_studio.dq_rules_core",
        instance_name="<your-lakebase-instance>",
        run_config_name="main.sales.orders",  # the table the rules apply to
    )
)

valid_df, quarantined_df = engine.apply_checks_by_metadata_and_split(df, checks)
```

Only **approved** rules load — drafts and rejected rules are never returned. You get the
table's current approved set, so no version pinning is needed. Cross-table SQL checks have
no single home table and appear under `__sql_check__/<name>` instead of a table name.

<Admonition type="note" title="Access you'll need">
The rules view lives in Studio's Lakebase Postgres database, so an administrator grants a
reader two things: **`CAN_CONNECT`** on the Lakebase database instance, and `SELECT` on the
view itself:

```sql
GRANT USAGE  ON SCHEMA dqx_studio        TO "<principal>";
GRANT SELECT ON dqx_studio.dq_rules_core TO "<principal>";
```

For a service principal, `<principal>` is its application id — pass the same id as
`client_id` in the config above.
</Admonition>

## Read quarantined records

Failing rows captured by Studio's runs land in a Delta table in the Studio catalog, which
you can query from SQL, a notebook, or an alerting job:

```sql
SELECT quarantine_id, run_id, created_at, row_data, errors, warnings
FROM <catalog>.dqx_studio.dq_quarantine_records
WHERE source_table_fqn = 'main.sales.orders'
ORDER BY created_at DESC
LIMIT 100;
```

`row_data` holds the full source row and `errors` / `warnings` the checks it failed, all as
[`VARIANT`](https://docs.databricks.com/aws/en/semi-structured/variant). Because the rows
are real source data, treat the table as a possible PII surface and grant access
accordingly. Records are deleted after `quarantine_retention_days` (default 30).
````

Happy to be talked out of the placement, but I do think this should be one page rather than a section.

**Comment by OGordon100** [comment]
Two follow-ups to my review above — one gap in what's documented, one placement change that supersedes part of my earlier suggestion.

### 1. Loading a *single* rule, and pinning a *version*, are both undocumented

Neither page tells a reader how to do either, and both are things you'd immediately reach for.

**Single rule.** There's no config field for it — `load_checks` returns the whole approved set for a `run_config_name` as a `list[dict]`. That's fine, but say so and show the one-liner, otherwise readers will go looking for an argument that doesn't exist:

```python
checks = engine.load_checks(LakebaseChecksStorageConfig(...))
order_id_checks = [c for c in checks if c["name"] == "order_id_is_not_null"]
```

**Version pinning.** `LakebaseChecksStorageConfig` already takes `rule_set_fingerprint` (`src/databricks/labs/dqx/config.py:560-562`) and the loader filters on it (`src/databricks/labs/dqx/checks_storage.py:808-813`), so this is a real, supported feature. The page only mentions it obliquely — "With no `rule_set_fingerprint` given, the loader returns the **current** approved rule set" — which implies you could pass one to get a *different* version, and never shows how.

**But here's the catch, and I think it needs stating explicitly:** against Studio's `dq_rules_core` view, the set fingerprint is computed *on the fly* from whatever is approved right now, grouped by table (`app/src/databricks_labs_dqx_app/backend/migrations/postgres.py:770-777`). So exactly one `rule_set_fingerprint` exists per table at any moment, and passing an older one returns an **empty list**, not an older rule set. Version pinning works when dqx-core owns the Lakebase *table*; via Studio's *view* it can only ever confirm the current set.

That distinction is the sort of thing someone will lose an afternoon to. Please either document `rule_set_fingerprint` as "use it to assert you got the set you expected, not to retrieve history", or — if pinning historical Studio rule sets is meant to work — treat that as a gap in the view rather than a docs fix.

### 2. Move the quarantine content into "Review results" as "Analyse in Lakehouse"

Rather than the quarantine half of the page I proposed above, reading quarantined records belongs on the existing **Review results** page (`docs/dqx/docs/studio/running/index.mdx`) — that page is already where a reader goes to look at failing rows, and its `Investigate and act on a failure` section already walks through the in-app version of exactly this.

**Suggested:** add a `## Analyse in Lakehouse {#analyse-in-lakehouse}` section at the end of that page, after `## Ask Genie`. The progression then reads: inspect in the app → ask Genie → drop to SQL when you want to go further. One short paragraph, the single `SELECT` from my draft, and the PII/retention warning is plenty.

That leaves the new page under **Monitor your tables** to be purely about loading rules as code, which makes "Use as Code in Lakehouse" a much more honest title than the two-topic page I sketched. Net effect: one new page plus one new section on an existing page, and the `integrations/` section still goes away.

docs/dqx/docs/studio/integrations/index.mdx
mwojtyczka: addressed

mwojtyczka:
> Two follow-ups to my review above — one gap in what's documented, one placement change that supersedes part of my earlier suggestion.
> 
> ### 1. Loading a _single_ rule, and pinning a _version_, are both undocumented
> Neither page tells a reader how to do either, and both are things you'd immediately reach for.
> 
> **Single rule.** There's no config field for it — `load_checks` returns the whole approved set for a `run_config_name` as a `list[dict]`. That's fine, but say so and show the one-liner, otherwise readers will go looking for an argument that doesn't exist:
> 
> ```python
> checks = engine.load_checks(LakebaseChecksStorageConfig(...))
> order_id_checks = [c for c in checks if c["name"] == "order_id_is_not_null"]
> ```
> 
> **Version pinning.** `LakebaseChecksStorageConfig` already takes `rule_set_fingerprint` (`src/databricks/labs/dqx/config.py:560-562`) and the loader filters on it (`src/databricks/labs/dqx/checks_storage.py:808-813`), so this is a real, supported feature. The page only mentions it obliquely — "With no `rule_set_fingerprint` given, the loader returns the **current** approved rule set" — which implies you could pass one to get a _different_ version, and never shows how.
> 
> **But here's the catch, and I think it needs stating explicitly:** against Studio's `dq_rules_core` view, the set fingerprint is computed _on the fly_ from whatever is approved right now, grouped by table (`app/src/databricks_labs_dqx_app/backend/migrations/postgres.py:770-777`). So exactly one `rule_set_fingerprint` exists per table at any moment, and passing an older one returns an **empty list**, not an older rule set. Version pinning works when dqx-core owns the Lakebase _table_; via Studio's _view_ it can only ever confirm the current set.
> 
> That distinction is the sort of thing someone will lose an afternoon to. Please either document `rule_set_fingerprint` as "use it to assert you got the set you expected, not to retrieve history", or — if pinning historical Studio rule sets is meant to work — treat that as a gap in the view rather than a docs fix.
> 
> ### 2. Move the quarantine content into "Review results" as "Analyse in Lakehouse"
> Rather than the quarantine half of the page I proposed above, reading quarantined records belongs on the existing **Review results** page (`docs/dqx/docs/studio/running/index.mdx`) — that page is already where a reader goes to look at failing rows, and its `Investigate and act on a failure` section already walks through the in-app version of exactly this.
> 
> **Suggested:** add a `## Analyse in Lakehouse {#analyse-in-lakehouse}` section at the end of that page, after `## Ask Genie`. The progression then reads: inspect in the app → ask Genie → drop to SQL when you want to go further. One short paragraph, the single `SELECT` from my draft, and the PII/retention warning is plenty.
> 
> That leaves the new page under **Monitor your tables** to be purely about loading rules as code, which makes "Use as Code in Lakehouse" a much more honest title than the two-topic page I sketched. Net effect: one new page plus one new section on an existing page, and the `integrations/` section still goes away.

Feedback implemented. Mostly approved, with few exception. Kept the cookbook queries but moved to the "Query Results Cookbook" page. Move dloading checks under monitoring tables section and moved querying quarantined data under results section as suggested. The content is now also simplified to only required info, without exposing app internals.

